PR #92, opened 2014-08-21
Make C-t in bash/zsh respect FZF_DEFAULT_COMMAND.

eventualbuddha:
I had to move the configuration into the generated config files instead
of where one would normally put it (e.g. ~/.bashrc). This is to properly
support tmux, since when using tmux a new shell is spawned that may not
be configured the same way.

I was unable to figure out how to make this work for fish. For some
reason the new shell in the tmux pane does not get variables set with
`set -x` (I tried placing them as the first lines of `fzf_key_bindings`).

### DIFF
--- a/install
+++ b/install
@@ -90,6 +90,17 @@ fzf() {
 }
 export -f fzf > /dev/null
 
+# Custom configuration
+# --------------------
+
+# Use this section to customize your fzf settings.
+
+# Use fzf with ag to respect .gitignore, .hgignore, and svn:ignore.
+# export FZF_DEFAULT_COMMAND='ag -l -g ""'
+
+# Adjust the default options for fzf:
+# export FZF_DEFAULT_OPTS="--sort 20000"
+
 # Auto-completion
 # ---------------
 $fzf_completion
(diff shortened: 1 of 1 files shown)


yifanchen:
Nice plugin, I love it. 

And I am still looking for an answer for this question. Does this mean CTRL-T doesn't work? If it doesn't, how come it still shows in README.md?

junegunn:
You're looking at an ancient issue. FZF_CTRL_T_COMMAND works.

```sh
FZF_CTRL_T_COMMAND='seq 100'

# Press CTRL-T
```

yifanchen:
Doesn't work, instead, it threw me an error sound.

junegunn:
@yifanchen Okay, then please create a separate issue with the information of your environment and a minimal shell configuration that can be used to reproduce the problem. (No other shell plugins and stuff that might get in the way)

yifanchen:
@junegunn sorry, I was wrong. Actually nothing happens when I `ctrl - t`. 

For example: If I do `rm` + `ctrl - t `. Nothing happens.

junegunn:
@yifanchen Please create a new issue and provide information about your environment.

yifanchen:
Will do :).

yifanchen:
Quick question. Is there any way to set a default opening directory?

PR #114, opened 2014-12-29
Speed up git clone for installation

PeterDaveHello:


junegunn:
Thanks. But since the size of the repository is still relatively small, less than 500KB, the benefit is not significant. And a quick Google search shows people having various kinds of issues with this, so I'm not going to merge it at the moment. But I'll leave this PR open for future consideration.

References:
- https://github.com/bower/bower/issues/217
- https://github.com/gmarik/Vundle.vim/issues/28
- http://git.661346.n2.nabble.com/BUG-shallow-clones-over-http-td7581458.html

PeterDaveHello:
Okay, I think the issues you mentioned are in different situation, if we just want to setup from the latest commit, there is no problem, but you are right, the benefit is not significant, haha ...

junegunn:
> if we just want to setup from the latest commit, there is no problem

Probably. I honestly don't know. I haven't looked carefully into those issues, so maybe I'm just being too cautious.

justinmk:
Interesting, I wasn't aware of any problems with this. Reading the links, seems the problems occur if `git:` is not used as the protocol.

junegunn:
@justinmk Maybe it's a bug from older versions of git? Haven't had time to really look into. Homebrew and NeoBundle also stopped supporting shallow clones for some reasons.

- https://github.com/Homebrew/homebrew/commit/8a1c2ec
- https://github.com/Shougo/neobundle.vim/issues/81

justinmk:
Note that the neobundle thread also shows `https:` being used. The suggestion that `git:` is more efficient is a good reason to use it anyway.

junegunn:
GitHub recommends using `https`:
https://help.github.com/articles/which-remote-url-should-i-use/

I agree with the point that `https` are more universal. At my workplace, servers are behind the firewalls so `https` through proxy is the only option there.

PR #116, opened 2015-01-03
Rewritten in Go

junegunn:
Good to Go? :santa: 

```sh
git checkout go
./install
```

Binaries: https://github.com/junegunn/fzf-bin/releases


justinmk:
Very excited about this! Any plans for a windows build (http://www.appveyor.com seems popular)?

junegunn:
@justinmk There are some hurdles for supporting Windows. fzf depends on ncurses which as far as I know not available on Windows platform. Fortunately, there is a cross-platform alternative to ncurses called [termbox](https://github.com/nsf/termbox-go), but it lacks some features fzf needs, such as proper mouse support and escape key sequences (for alt-something). Those features are not really crucial to fzf, so given that we can live without them on Windows, it's doable. But I don't have a Windows PC for the development, so I could use some help.

justinmk:
> There are some hurdles for supporting Windows. fzf depends on ncurses

Would it be possible to decouple the UI from the fuzzy matching engine? Then something like unite.vim could use fzf as a backend, and it would just decorate fzf's output[1]. That's really what I'm interested in--using fzf as a cross-platform component shipped with Neovim.

junegunn:
It should be possible. The core components, Reader, Matcher and Terminal, are already decoupled from each other.

PR #122, opened 2015-01-24
Improve `find` command for cd widgets: exclude proc/dev

blueyed:
When using the widget in "/", it would descend into 'dev/'.
Using '*' for the starting path would do so also with the new '-fstype'
excludes.

The drawback / change now is that the './' prefix is being added, which
I don't mind.  This could be fixed by throwing in some `cut -b3-` into
the pipe.

### DIFF
--- a/install
+++ b/install
@@ -202,10 +202,10 @@ EOF
 # Key bindings
 # ------------
 __fsel() {
-  command find * -path '*/\.*' -prune \
+  command find -L . \( -path '*/\.*' -o -fstype 'dev' -o -fstype 'proc' \) -prune \
     -o -type f -print \
     -o -type d -print \
-    -o -type l -print 2> /dev/null | fzf -m | while read item; do
+    -o -type l -print 2> /dev/null | sed 1d | cut -b3- | fzf -m | while read item; do
     printf '%q ' "$item"
   done
   echo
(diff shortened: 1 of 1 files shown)


junegunn:
Please ignore Travis failure. I momentarily gave up making it work.

blueyed:
> Please ignore Travis failure

Heh.. :) - I was just wondering.

blueyed:
Feel free to squash-merge the commits, or tell me to rebase the PR.

junegunn:
Doh, one thing I noticed is that `find .` includes `.` itself, so fzf gives an empty entry.

```sh
jg@mbp:~> find . | cut -b3- | head

.agignore
.ansible
.ansible/cp
.ansible/tmp
.ansible/tmp/ansible-1355468543.5-175586869495702
.ansible/tmp/ansible-1355468608.74-15792789855106
.ansible/tmp/ansible-1355468672.32-65045913254503
.ansible/tmp/ansible-1355468717.3-263255040481250
.ansible/tmp/ansible-1355468798.05-74592742241722
```

We could append `grep -v ^$`, but I wonder if there's a better way. Except for that, the change looks good to me. Could you fix the issue above and squash the commits into one?

install
```diff
@@ -369,14 +369,15 @@ function fzf_key_bindings
   end
 
   function __fzf_list
-    command find * -path '*/\.*' -prune \
+    command find -L ${1:-.} \( -path '*/\.*' -o -fstype 'dev' -o -fstype 'proc' \) -prune \
```
junegunn: Fish complains about this:

```
fish: Did you mean {$VARIABLE}? The '$' character begins a variable name. A bracket, which directly followed a '$', is not allowed as a part of a variable name, and variable names may not be zero characters long. To learn more about variable expansion in fish, type 'help expand-variable'.
Standard input:     command find -L ${1:-.} \( -path '*/\.*' -o -fstype 'dev' -o -fstype 'proc' \) -prune \
                                    ^
```
Let's just change it to simple `.`

blueyed:
Fish should be fixed now, and I've added `sed 1d` to remove the first entry.

junegunn:
Merged. Thanks!

Issue #83, opened 2014-07-25
Implement test cases for shell extension


junegunn:
- Key bindings
    - [x] bash
    - [x] zsh
    - [x] fish
- Auto completion
    - [x] bash - file
    - [x] bash - dir
    - [x] bash - process
    - bash - host (TBD)

Is there any tool that can be used to test *interactive* shell scripts?

Assuming there isn't one already:
- Using tmux seems like a nice idea
    - Send keystrokes to the split pane
    - Monitor the content of the split pane and act accordingly

PR #133, opened 2015-02-27
fix bug with __fcd /path/to/dir

acornejo:
There is a bug in the way `__fcd` is currently implemented.

To observe the problem, simply run on your shell `$(__fcd $HOME)`.

*Expected behavior:* You are displayed a list of directories inside `$HOME` and can use `fzf` to select one of them.
*Observed behavior:* You get a truncated list of directories, and on selection `fzf` fails to change folder (since the paths are truncated).

This bug was caused due to a naive use of cut to remove the "./" prefix from folders. However, this prefix is absent if using absolute paths.

I replaced the use of cut by `sed`, this also has the side benefit of require less pipes, since `sed` was already being used to remove the first line (and `sed` allows you to chain several operations in the same call).

In this commit I also made allowed other `fzf` commands, such as `__fsel`, receive the starting path as a parameter (as `__fcd` already did) using the current directory as the default when nothing else is specified.

Another change which I was tempted to make, was to add a global variable `FZF_DEFAULT_STARTPATH` that allows the users to specify in which directory they want to start fcd or fsel by default, allowing them to specify something else other than the current directory (aka "."). For instance, in my own use case I have the default be $HOME/src, since I rarely want to use fzf outside that. I thought I should get some feedback before creating that pull-request.

### DIFF
--- a/install
+++ b/install
@@ -202,10 +202,10 @@ EOF
 # Key bindings
 # ------------
 __fsel() {
-  command find -L . \( -path '*/\.*' -o -fstype 'dev' -o -fstype 'proc' \) -prune \
+  command find -L ${1:-.} \( -path '*/\.*' -o -fstype 'dev' -o -fstype 'proc' \) -prune \
     -o -type f -print \
     -o -type d -print \
-    -o -type l -print 2> /dev/null | sed 1d | cut -b3- | fzf -m | while read item; do
+    -o -type l -print 2> /dev/null | sed -e 1d -e s#^\./## | fzf -m | while read item; do
     printf '%q ' "$item"
   done
   echo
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks for your suggestion. But those `__fxx`s are not supposed to be used directly but only by the supplied key bindings, hence the prefix `__` denoting that they are not the part of public interface. The README page also does not mention they are provided by default, see https://github.com/junegunn/fzf#installation. There's nothing wrong with your patch, and it's nice that it shortens the command pipeline, but I don't want to give the wrong impression that they can be used directly for other purposes (e.g. `${1:-.}`), because they are not meant to be.

junegunn:
By the way, I'm trying to refrain from updating those key binding codes until #83 is done. It has been really annoying to test all three shells after every update.

acornejo:
:(

It was worth a shot, I wanted to avoid having to maintain my own changes just for that micro "feature", but I understand your decision.

PR #136, opened 2015-03-06
Add fzf-tmux script

junegunn:
... and clean up the documentation


Tranquility:
Maybe it's better to use tmux terminology and talk about tmux panes instead of splits?

junegunn:
@Tranquility Good point. I'm too used to Vim terminology. I'll update the doc.

PR #138, opened 2015-03-08
[fzf-tmux] Allow opening fzf on any position (up/down/left/right)

junegunn:


marcmo:
is opening fzf on any position only possible with tmux?
I'd like to use it in my vim integration but can't figure out how to have the query prompt on top instead of at the bottom...is this possible?

junegunn:
Hmm, did you try `--reverse` option? (it's not related to this PR)

marcmo:
thanks @junegunn !! all I needed!
fzf is such a great idea! love it! slowly I migrate more and more functionality that I used unite for...it's insanely fast!

PR #148, opened 2015-03-18
Add support for ANSI color codes

junegunn:
Interpret ANSI color codes from the input.

### Why?

Because... colors.

#### Use cases

- https://github.com/D630/fzf-fs
- https://github.com/peco/peco/issues/188

#### `CLICOLOR_FORCE=1 ls -al | fzf`
fzf-ls

#### `CLICOLOR_FORCE=1 ls -al | fzf --ansi`
fzf-ansi-ls

#### `screenshotTable.sh | fzf --ansi`
screenshot-table

### Then, why hesitate?

- Code complexity
- Performance overhead in reader (*how much?*)
- Increased memory footprint for large input (> 1M items)
- A typical command-line program suppresses ANSI color codes when its output is redirected
    - Must explicitly enable colors.
        - OSX: `CLICOLOR_FORCE=1 ls -al`
        - Linux: `ls --color=always -al`
- :sleepy: 

### TBD

- Should fzf remove ANSI codes from its output?
    - Probably. It might make it less flexible, but more convenient/suitable for most cases.

### DIFF
--- a/src/algo/algo.go
+++ b/src/algo/algo.go
@@ -12,6 +12,10 @@ import "strings"
 
 // FuzzyMatch performs fuzzy-match
 func FuzzyMatch(caseSensitive bool, input *string, pattern []rune) (int, int) {
+	if len(pattern) == 0 {
+		return 0, 0
+	}
+
 	runes := []rune(*input)
 
 	// 0. (FIXME) How to find the shortest match?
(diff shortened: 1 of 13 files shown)


D630:
+1

junegunn:
I'll continue development on `devel` branch instead of feature-branches. I'll recreate another issue with the same content.

PR #165, opened 2015-03-27
zsh: If point is on a dir, use that instead of PWD

quodlibetor:
So:

    ~$ vim /etc<C-t>

Will do fzf search from within /etc instead of ~. But:

    ~$ vim /etc <C-t>

Note the space - will still do fzf from within ~.

### DIFF
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -11,6 +11,31 @@ __fsel() {
   echo
 }
 
+fzf-current-word-widget() {
+    local current
+    current="${LBUFFER/* /}${RBUFFER/ */}"
+    if [[ -n $current && ${current[0,1]} = '~' ]] ; then
+        printf %q ${HOME}${current[2,${#current}]}
+    else
+        printf %q $current
+    fi
+}
+zle -N fzf-current-word-widget
+
+fzf-dir-at-point() {
+    local current_word
+    local current_dir
+    current_word=$(fzf-current-word-widget)
+    if [[ -d $current_word ]] ; then
+        current_dir="$current_word"
+    elif [[ -d ${current_word:h} ]] ; then
+        current_dir=${current_word:h}
+    else
+        current_dir="$PWD"
+    fi
+    printf %q $current_dir
+}
+
 if [[ $- =~ i ]]; then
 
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks, makes sense. I don't like zsh version deviates from bash and fish versions, but what can we do? 

However, there are some issues in your patch.
- Non-tmux version is not updated accordingly
- Does not use 2-space indentation
- No test cases
- Does not work if the prefix does not have trailing slash
    - `ls ~<CTRL-T>`
    - `ls ~USERNAME<CTRL-T>`
    - `ls PATH<CTRL-T>`
- Your code looks at RBUFFER as well, which is nice, but does not yield the correct result.
    - `ls PA<CTRL-T>TH`

I can understand some of these issues are not easy to address at the moment as the current version uses `tmux send-key`. We probably should replace the use of `send-key` with the new `fzf-tmux` script, not just in zsh, but also in bash and fish.

quodlibetor:
Thanks for looking at this! I admit I only checked what I actually use, but I agree with all your critiques.

I don't know anything about bash and fish completion, although I'm willing to try.

I can add test cases for all the issues you brought up, making them work.

junegunn:
Another point we have to consider is that `--multi` option is enabled for CTRL-T, so we are allowed to choose multiple entries. The code should attach the prefix to every item selected except the first one which already has the prefix.

> I don't know anything about bash and fish completion, although I'm willing to try.

I don't think it's really possible with bash (there's no such thing as `$LBUFFER`), but it's okay since bash has fuzzy completion that can be used instead (`ls PATH/**<TAB>`). I don't know nor care much about fish. Honestly I'm a bit tired in dealing with non-standard behaviors of it. You can try if you want, but it's okay to skip it.

> I can add test cases for all the issues you brought up, making them work.

Thanks. Let's first try to fix the issues I mentioned here, then I'll tell you how to write test cases or I can just do it myself.

quodlibetor:
Sounds great!

junegunn:
Hi, this PR is superseded by the new zsh completion with empty trigger sequence support. See #232. Thanks.

```sh
export FZF_COMPLETION_TRIGGER=''
bindkey '^I' expand-or-complete
bindkey '^T' fzf-completion
```

Issue #168, opened 2015-03-29
git ls-tree snippet doesn't work


xcambar:
Hi,
I can't have the tip about using `git ls-tree` (https://github.com/junegunn/fzf#git-ls-tree-for-fast-traversal) to work. I'm afraid it's outdated since `fzf` has been rewritten in Go.

Honestly, my only clue here is the `declare -f fzf` statement, which, according to SO, outputs the contents of a function. But as `fzf` is a binary executable, it doesn't realy make sense to me...

Correct me if I'm wrong, and if I am, I would nonetheless appreciate some help on getting the snippet to work under zsh 5.0.7.

Thanks for the project and the others, they're great!

junegunn:
You're right, it was an outdated snippet, thanks for letting me know :)
I replaced the snippet with an equivalent `$FZF_DEFAULT_COMMAND`.

PR #170, opened 2015-03-30
Updates git ls-tree example

xcambar:
Following up on https://github.com/junegunn/fzf/issues/168,
this PR fixes the snippet using `git ls-tree`

`find` had an incorrect syntax.

### DIFF
--- a/README.md
+++ b/README.md
@@ -425,7 +425,7 @@ speed of the traversal.
 ```sh
 export FZF_DEFAULT_COMMAND='
   (git ls-tree -r --name-only HEAD ||
-   find * -name ".*" -prune -o -type f -print -o -type l -print) 2> /dev/null'
+   \find * -name ".*" -prune -o -type f -print -o -type l -print) 2> /dev/null'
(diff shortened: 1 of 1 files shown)


junegunn:
It works as expected on OS X and Linux, what problem did you have?

xcambar:
After more digging, it seems to be the same issue that appeared in https://github.com/junegunn/fzf/pull/97.

My `find` is aliased to `noglob find`, which leads to the following error:

```sh
$ find * -name ".*" -prune -o -type f -print -o -type l -print
find: *: No such file or directory
```

I've updated the PR accordingly

xcambar:
Still doesn't work. I'll come back with something more solid.

george-b:
@xcambar you way want to have `unsetop NOMATCH` in your zshrc somewhere. This results as the `*` getting past to find as is rather than zsh throwing an error about it not matching glob.

xcambar:
@george-b Works perfectly, thanks.

PR #171, opened 2015-03-30
Support for vi insert mode in upcoming fish 2.2.0

oschrenk:
Turns out fish key bindings support different modes, so that adding the functionality was a matter of binding the given shortcuts also to the insert mode.

### DIFF
--- a/shell/key-bindings.fish
+++ b/shell/key-bindings.fish
@@ -76,5 +76,11 @@ function fzf_key_bindings
   bind \ct '__fzf_ctrl_t'
   bind \cr '__fzf_ctrl_r'
   bind \ec '__fzf_alt_c'
+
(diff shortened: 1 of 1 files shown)


junegunn:
Hi, thanks. Looks like Fish 2.2.0 it not yet officially released, and the code causes errors on Fish 2.1.2, which I have. Can you make the code conditionally executed only on Fish 2.2.0 or above? Or maybe we can simply suppress the errors messages.

oschrenk:
That makes sense. Will do.

oschrenk:
Wrapped the mode bindings in an `if bind -M insert > /dev/null 2>&1` which should do the trick and push forced the changes.

oschrenk:
Removed comment.

junegunn:
Thank you!

PR #180, opened 2015-04-10
Check if :FZF is already executing

mhinz:
Hi, Junegunn!

Prior to this change, you'd get a longer error message if you did:

    :FZF
    <esc>
    :FZF

The main problem being that `:file [FZF]` can be used only once. This commit uses that to its advantage and just adds a simple `bufexists('[FZF]')` check (and a warning message, that you can remove if you want).

Probably easier than using an ugly `s:is_active` variable. :)


junegunn:
Merged, thanks!

PR #183, opened 2015-04-11
Check if fzf#run() is already executing

qiemem:
#180 only fixes invocations of `:FZF`. This simply moves the check to `fzf#run()` so that it works with all invocations of fzf.


junegunn:
Thank you! I commented on the code. Could you change them and force-push to the branch?

qiemem:
Ah woops, changed before I saw the force-push comment. Just a moment.

qiemem:
Done!

junegunn:
:+1:

PR #188, opened 2015-04-14
install: wait for LF in non-interactive shell

justinmk:
Problem: `yes n | ./install` should answer "n" to every prompt, but instead it answers "n" to the first question, and the next question is getting some non-printable input which `ask()` interprets as "y". This means that shell keybindings are always installed. 

Solution: omit `-n 1` for non-interactive shells.

### DIFF
--- a/install
+++ b/install
@@ -5,8 +5,15 @@ version=0.9.7
 cd $(dirname $BASH_SOURCE)
 fzf_base=$(pwd)
 
+# If stdin is a tty, we are "interactive".
+[ -t 0 ] && interactive=yes
+
 ask() {
-  read -p "$1 ([y]/n) " -n 1 -r
+  # non-interactive shell: wait for a linefeed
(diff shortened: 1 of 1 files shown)


justinmk:
Ack, this has a bug.

justinmk:
Ok, tested this in interactive and non-interactive mode, answering "y" and "n". No bugs now.

install
```diff
@@ -5,8 +5,15 @@ version=0.9.7
 cd $(dirname $BASH_SOURCE)
 fzf_base=$(pwd)
 
+# If stdin is a tty, we are "interactive".
+[ -t 0 ] && interactive=yes
```
justinmk: This is actually posix compatible in case anyone is curious. http://stackoverflow.com/a/911213/152142

junegunn:
Good catch, thanks!

PR #189, opened 2015-04-15
ctrl-t is my tmux character, so...

rlpowell:
Here are ways to use other characters in vim for tmux splitting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -282,6 +282,19 @@ Similarly to [ctrlp.vim](https://github.com/kien/ctrlp.vim), use enter key,
 `CTRL-T`, `CTRL-X` or `CTRL-V` to open selected files in the current window,
 in new tabs, in horizontal splits, or in vertical splits respectively.
 
+The type of splitting can be overriden using
+`g:fzf_default_split_style`.  The split characters can be overridden
+using `g:fzf_tab_char`, `g:fzf_split_char`, and `g:fzf_vsplit_char',
+
+For example:
+```vim
+let g:fzf_default_split_style = 'tabedit'
+let g:fzf_tab_char = 'alt-t'
+```
+
+See the `--extend` section of the fzf man page for a description of
+the syntax of these character options.
+
 Note that the environment variables `FZF_DEFAULT_COMMAND` and `FZF_DEFAULT_OPTS`
 also apply here.
 
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -26,6 +26,10 @@ let s:launcher = 'xterm -e bash -ic %s'
 let s:fzf_go = expand('<sfile>:h:h').'/bin/fzf'
 let s:fzf_rb = expand('<sfile>:h:h').'/fzf'
 let s:fzf_tmux = expand('<sfile>:h:h').'/bin/fzf-tmux'
+let s:default_split_style = 'e'
+let s:tab_char = 'ctrl-t'
+let s:split_char = 'ctrl-x'
+let s:vsplit_char = 'ctrl-v'
 
 let s:cpo_save = &cpo
 set cpo&vim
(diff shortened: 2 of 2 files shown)


rlpowell:
I am almost totally certain that my changed didn't cause that build failure, as the build failure is about Go or something and I only touched the vim plugin.

junegunn:
Hi, thanks for suggestion.

> my changed didn't cause that build failure

I know, don't worry about that.

I have to mention that I'm not sure if I want to put configuration variables all over the code for different customization needs, when the entire definition of `:FZF` spans only 20~30 lines. Users with different tastes can easily copy the code onto their .vimrc and change it to their liking with no restriction. The added bonus is that by doing so, they'll learn how to use `fzf#run()` function. (Actually I even want to remove the pre-existing config knobs and simplify the code and the documentation. For example, `g:fzf_tmux_height` has now become an invalid name since Neovim terminal integration.)

Having said that, there are some points I don't agree with your implementation. I think the code should be more generic. For example, what if a user wants to have two separate keys for opening file on the left and on the right? We don't want to add more configuration variables.

Other minor points that don't match my personal taste :)
- Commit message
- Excessive margins around `|`
- Lines well over 80 column-limit
- Invalid description on README (`--extend`?)

I'll write my own version of the code tonight and let you know.

rlpowell:
On Wed, Apr 15, 2015 at 03:12:47AM -0700, Junegunn Choi wrote:
> Hi, thanks for suggestion.
> 
> > my changed didn't cause that build failure
> 
> I know, don't worry about that.
> 
> I have to mention that I'm not sure if I want to put configuration
> variables all over the code for different customization needs,
> when the entire definition of `:FZF` spans only 20~30 lines. Users
> with different tastes can easily copy the code onto their .vimrc
> and change it to their liking with no restriction. The added bonus
> is that by doing so, they'll learn how to use `fzf#run()`
> function. (Actually I even want to remove the pre-existing config
> knobs and simplify the code and the documentation. For example,
> `g:fzf_tmux_height` has now become an invalid name since Neovim
> terminal integration.)

The flip side of that, though, is that users lose access to any
good changes you implement that they might have wanted, unless they
specifically make a point to compare their code to yours on
occassion.

> Having said that, there are some points I don't agree with your
> implementation.

It's my first vimscript.  :D

> I think the code should be more generic. For example, what if a
> user wants to have two separate keys for opening file on the left
> and on the right?

Fair point.

> Other minor points that don't match my personal taste :)
> - Commit message
> - Excessive margins around `|`
> - Lines well over 80 column-limit
> - Invalid description on README (`--extend`?)
> 
> I'll write my own version of the code tonight and let you know.

Sure.

-- 
http://intelligence.org/ :  Our last, best hope for a fantastic future.
.i ko na cpedu lo nu stidi vau loi jbopre .i dafsku lu na go'i li'u .e
lu go'i li'u .i ji'a go'i lu na'e go'i li'u .e lu go'i na'i li'u .e
lu no'e go'i li'u .e lu to'e go'i li'u .e lu lo mamta be do cu sofybakni li'u

junegunn:
Okay, now you can use `g:fzf_action` to set up key bindings. See: https://github.com/junegunn/fzf/wiki/Examples-(vim) Thanks!

junegunn:
By the way, binding ALT key bindings (which is essentially ESC + ?) doesn't work on Neovim, and it's a known issue.

https://github.com/neovim/neovim/issues/2269

justinmk:
Works for me (cb7bb70c4902a38165e3d1cb3158704dbac228c4):

```
:nn <m-/> :echo '/'<cr>
:nn <m-?> :echo '?'<cr>
```

rlpowell:
Oh that's *way* better.  Thank you!

On Wed, Apr 15, 2015 at 06:56:06AM -0700, Junegunn Choi wrote:
> Okay, now you can use `g:fzf_action` to set up key bindings. See: https://github.com/junegunn/fzf/wiki/Examples-(vim) Thanks!
> 
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/junegunn/fzf/pull/189#issuecomment-93410654

-- 
http://intelligence.org/ :  Our last, best hope for a fantastic future.
.i ko na cpedu lo nu stidi vau loi jbopre .i dafsku lu na go'i li'u .e
lu go'i li'u .i ji'a go'i lu na'e go'i li'u .e lu go'i na'i li'u .e
lu no'e go'i li'u .e lu to'e go'i li'u .e lu lo mamta be do cu sofybakni li'u

edi9999:
> The added bonus is that by doing so, they'll learn how to use fzf#run() function.

Completely agree with this, I would like to **have** to put the fzf#run in my .vimrc, so that I can see what exactly is run by FZF and hack on it if you want

Issue #214, opened 2015-04-23
[zsh] <Ctrl-R> not working


chrisbra:
Hi, I have `setopt no_bang_hist` in zsh and when I use the <Ctrl-R> command key, fzf tries to use the !expansion form, which obviously does not work.

junegunn:
Do you think it's possible to temporarily change "no_bang_hist"? Is there an easy solution? I'm a bash user and have no clue.

chrisbra:
That would be fine for me.

junegunn:
```sh
# CTRL-R - Paste the selected command from history into the command line
fzf-history-widget() {
  local selected restore_no_bang_hist
  if selected=$(fc -l 1 | $(__fzfcmd) +s --tac +m -n2..,.. --tiebreak=index --toggle-sort=ctrl-r -q "$LBUFFER"); then
    num=$(echo "$selected" | head -1 | awk '{print $1}' | sed 's/[^0-9]//g')
    LBUFFER=!$num
    if setopt | grep nobanghist > /dev/null; then
      restore_no_bang_hist=1
      unsetopt no_bang_hist
    fi
    zle expand-history
    [ -n "$restore_no_bang_hist" ] && setopt no_bang_hist
  fi
  zle redisplay
}
zle     -N   fzf-history-widget
bindkey '^R' fzf-history-widget
```

This seems to work.

chrisbra:
Great and so fast!

junegunn:
Okay, I just pushed a fix. Let me know if you see any issues. Thanks!

Issue #142, opened 2015-03-11
[ZSH] multiple line command histroy problem?


xuchunyang:
For example,
```
➜  ~  git \
> --version
git version 2.3.2
```
`fzf` (by using `C-r` in ZSH) gave me
```
➜  ~  git \\n--version
error: invalid key: pager.\n--version
error: invalid key: alias.\n--version
git: '\n--version' is not a git command. See 'git --help'.
```
which doesn't work like the normal `C-r` .

Any ideas?

junegunn:
Hmm, thanks for the report. bash version also seems to have problem with multi-line commands.

junegunn:
Should be fixed now. Please [upgrade fzf](https://github.com/junegunn/fzf#upgrading-fzf) and let me know if you see any issue. Thanks.

xuchunyang:
It works now. Thanks for your work. @junegunn

PR #219, opened 2015-04-28
don't use numbered history expansion for fzf-history-widget

DanielFGray:
fzf already returns the command itself, deleting it and using history expansion seems silly. this is much simpler.


junegunn:
Thanks, but please see #142, #214

PR #226, opened 2015-05-05
changed fish shell commands

tactualjota:
Fish was hanging up on ctrl_t binding for fish. Not sure why ,but it was the 2>/dev/null that was doing it. Removed it and it works perfectly

### DIFF
--- a/shell/key-bindings.fish
+++ b/shell/key-bindings.fish
@@ -17,7 +17,7 @@ function fzf_key_bindings
     command find -L . \( -path '*/\.*' -o -fstype 'dev' -o -fstype 'proc' \) -prune \
       -o -type f -print \
       -o -type d -print \
-      -o -type l -print 2> /dev/null | sed 1d | cut -b3- | eval (__fzfcmd) -m > $TMPDIR/fzf.result
+      -o -type l -print | sed 1d | cut -b3- | eval (__fzfcmd) -m > $TMPDIR/fzf.result
     and commandline -i (cat $TMPDIR/fzf.result | __fzf_escape)
     commandline -f repaint
     rm -f $TMPDIR/fzf.result


junegunn:
I can't reproduce. Which version of fish are you running? And not suppressing stderr messages from find command can cause visual glitches on fzf. (e.g. trying to traverse directories that you don't have permission for.)

tactualjota:
I’m using fish, version 2.0.0, it comes with ubuntu 14.04 LTS.

You’re absolutely right, we shouldn’t just keep stderr around, but 
on my machine it hangs the eval (__fzfcmd) -m. History | eval works 
perfectly.

How can I help you with this issue? It may very well be something 
outside fzf!

-Jota

junegunn:
Okay, it looks like a bug of fish 2.0. I don't see the problem on fish 2.1, but I can reproduce it on fish 2.0 and confirmed that your workaround helps. (Also noticed that alt-c has the same issue)

Normally I would just tell the users to upgrade fish and move on but as it's the default version on the latest Ubuntu LTS, I'll probably have to address the issue somehow :(

I'll see what I can do when I get some time.

junegunn:
I confirmed that upgrading fish to 2.1.2 fixes the issue.

https://launchpad.net/~fish-shell/+archive/ubuntu/release-2

```sh
# sudo apt-get install software-properties-common
sudo apt-add-repository ppa:fish-shell/release-2
sudo apt-get update
sudo apt-get install fish
```

If we can't come up with a simple workaround, I'll just mention the version requirement on the README page or on the wiki.

junegunn:
So there's no good workaround to the problem? I've also tried a few different methods to no avail.

tactualjota:
This would be a temporary change, that would have to be reset on fish 
2.1.1. I'd rather have good code (i.e: no hacks) and just let fish 
update itself eventually.

Also, when updating fish I ran into a lot of issues. Commands wouldn't 
load, etc. I eventually had to reset my fish installation (and 
configuration files) etc. I've since been running fzf on osx and ubuntu 
without any troubles.

So, given that there is no clear solution and I ended up running into 
more troubles, I've closed the pull request.

---

I'm not trying to replace the find with locate to speed control-T :) If 
I get something working you'll get another pull request one of these days.

Thanks for your efforts. I really like fzf. If you ever need any fish 
shell testing, please let me know.

-Jota

Junegunn Choi wrote:

junegunn:
Thanks, I completely agree what you said about hackless code.

sandric:
well, I'm not sure if this is the same issue, but I think I can reproduce on 2.1.2 - just input random characters as command and hit `ctrl-t`, it hangs for a little time mb 3 seconds. Found it through trying it with `ll` command, which behaves similary, which is even more strange..

sandric:
well, I need to correct - I don't really know what is the problem, and try to explain:
After typing ctrl-t to open fzf, it hangs up from time to time. It is even look like that its based on how often and how quick I try to open fzf. Really strange bug, and I don't know how to reproduce - I just tried to add `pkill -9 fzf` at beginning of `function __fzf_ctrl_t`, because as I get, when it hangs, process remain in memory, and after terminal 'hang out', `ps -aux | grep fzf` gives me
`sandric  31541  0.0  0.0  13556  2916 pts/9    T    16:18   0:00 bash /usr/bin/fzf-tmux -d40% -m
sandric  31542  0.0  0.0 326864  6848 pts/9    Tl   16:18   0:00 fzf -m`
And the number of those hanged out processes equal to times it hanged out.
But this doesn't help..

junegunn:
If you're running the latest version of fzf and fish, it's probably an issue of fish. The code for fish support here is [pretty minimal](https://github.com/junegunn/fzf/blob/master/shell/key-bindings.fish), so I don't think there's something terribly wrong with it. Consider reporting the issue to fish guys.

junegunn:
@sandric If you're running the latest version of fzf and fish, it's probably an issue of fish. The code for fish support here is [pretty minimal](https://github.com/junegunn/fzf/blob/master/shell/key-bindings.fish), so I don't think there's something terribly wrong with it. Consider reporting the issue to fish guys.

sandric:
ok, thx.

tactualjota:
Btw. I've tried on my machines. Random typing, then ctrl-t. Can't 
reproduce.

I would worry too much. Doesn't seem like a reproducible bug.
-Jota

sandric wrote:

PR #231, opened 2015-05-09
fix typo in argument of head (zsh completion)

robinro:
at least my version of head wants -n1 to only display the first line


junegunn:
Thanks!

junegunn:
Related: http://www.spinics.net/lists/crypto/msg03235.html

robinro:
I have to thank! fzf looks pretty awesome. Coming from command-t in vim I started to use fzf in the shell and it looks pretty good so far (as long as it doesn't break ctrl-r :smile:  )

Issue #230, opened 2015-05-09
[zsh-completion] Known issues


junegunn:
Bugs:
- [x] `~USERNAME**<TAB>`
- [x] `~INVALID_USERNAME**<TAB>`
    - nonomatch
- [x] Fall back to default completion
    - Need to see what is currently mapped to `^I`. it may not be `expand-or-complete`
    - `bindkey '^I' | awk '{print $2}'`
- [x] `foo\ bar**<TAB>`
- [x] `setopt vi` resets TAB binding from fzf
    - I can't think of any way to handle this. We'll just have to tell the users about it.
    - https://github.com/junegunn/fzf/wiki/Fuzzy-completion

Enhancements:
- [x] Empty trigger sequence #232
- [x] Custom bindkey #232
```sh
export FZF_COMPLETION_TRIGGER=''
bindkey '^T' fzf-completion
bindkey '^I' $fzf_default_completion
```

tiziano88:
Thanks for all this @junegunn ! Also I think we should add support for configuring a different bound key and having no trigger at all. Specifically, the use case I'd like to support is: start writing a path name, use tab completion as usual up to a certain point, and then use a different shortcut (e.g. <kbd>CTRL</kbd>+<kbd>P</kbd>) to trigger fzf from that point on (without ever typing a specific trigger).

PR #232, opened 2015-05-10
zsh completion script empty trigger and bound key personalisation

tiziano88:
This allows for the use case I mentioned in #230 by setting the following env variables:

```zsh
export FZF_COMPLETION_TRIGGER=''
export FZF_COMPLETION_KEY='^P'
```

I'm not super happy about the name of `FZF_COMPLETION_KEY`, please do let me know if you have any suggestions, or feel free to go ahead and rename it after merging,

### DIFF
--- a/shell/completion.zsh
+++ b/shell/completion.zsh
@@ -110,7 +110,7 @@ fzf-zsh-completion() {
   fi
 
   cmd=${tokens[1]}
-  trigger=${FZF_COMPLETION_TRIGGER:-**}
+  trigger=${FZF_COMPLETION_TRIGGER-'**'}
 
   # Trigger sequence given
   tail=${LBUFFER:$(( ${#LBUFFER} - ${#trigger} ))}
--- a/shell/completion.zsh
+++ b/shell/completion.zsh
@@ -110,6 +110,8 @@ fzf-zsh-completion() {
   fi
 
   cmd=${tokens[1]}
+
+  # Explicitly allow for empty trigger.
   trigger=${FZF_COMPLETION_TRIGGER-'**'}
 
   # Trigger sequence given
--- a/shell/completion.zsh
+++ b/shell/completion.zsh
@@ -9,6 +9,7 @@
 # - $FZF_TMUX_HEIGHT        (default: '40%')
 # - $FZF_COMPLETION_TRIGGER (default: '**')
 # - $FZF_COMPLETION_OPTS    (default: empty)
+# - $FZF_COMPLETION_KEY     (default: '^I')
 
 _fzf_path_completion() {
   local base lbuf find_opts fzf_opts suffix tail fzf dir leftover matches nnm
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks, I like the idea in general but I don't like is that `FZF_COMPLETION_KEY` is zsh only and there's no easy way to apply it on bash in a consistent way. (Note that bind key expression for tab on bash is `"\C-i"` and not `^I`)

So instead of introducing the variable, here's what I would suggest to the users:

```sh
export FZF_COMPLETION_TRIGGER=''
bindkey '^I' expand-or-complete
bindkey '^P' fzf-zsh-completion
```

which I think is clear and not unduly verbose.

By the way, it looks like your PR doesn't seem to handle the prefix-less cases correctly. e.g. `vim <CTRL-P>` / `vim foobar <CTRL-P>`. Can you look into it?

junegunn:
Hey, I cherry-picked your commits and fixed the issue with no prefix. So no `FZF_COMPLETION_KEY`, and you'll have to use `bindkey` as described above. What do you think about renaming `fzf-zsh-completion` to `fzf-completion`? `zsh` in the name sounds redundant. Hmm.

junegunn:
Okay, renamed it. So it becomes

```sh
export FZF_COMPLETION_TRIGGER=''
bindkey '^I' expand-or-complete
bindkey '^P' fzf-completion
```

tiziano88:
Sounds good, thanks for merging.

junegunn:
By the way, you don't use `CTRL-P` on shell? I tried the above for a while and I quickly realized that I was missing the old `CTRL-P`. Anyway, due to the change now we can close #165 which is yet an incomplete effort to extend `CTRL-T`.

tiziano88:
Did you mean if I use the default `CTRL-P` behaviour in zsh (which AFAICT does nothing useful for me out of the box), or whether I use the `CTRL-P` plugin (or something else)? Anyway, in short, I didn't use `CTRL-P` for anything so far, that's why I wanted to bind it to fzf.

junegunn:
Oh, I see. `CTRL-P` is `previous-history` on bash and I always use it. I didn't know it was unbound by default on zsh. I was just thinking that we should suggest the empty-trigger method above as a way to enhance `CTRL-T` binding fzf provides.

tiziano88:
TBH I must say in general I don't see the advantage of having to explicitly type a trigger and pressing <kbd>TAB</kbd> as opposed to using a dedicated keybinding. In my case, I think I basically get the `CTRL-T` behaviour for free.

junegunn:
Well, I agree with that. The thing is, writing context-aware key bindings in bash is almost impossible and I had to take advantage of its completion system to implement fuzzy completion. So I was forced to use the `TAB` key, and the trigger sequence to denote the intention of fuzzy completion was necessary. However in zsh, we are not limited to using `TAB` key and the trigger sequence is no longer necessary. I just translated the bash version to zsh for consistency's sake. But considering that no one really switches between zsh and bash, the point is a little moot. Maybe we can rethink the interface for zsh later, but since there's a way to customize the key binding, it's not urgent. By the way in the latest revision `$fzf_default_completion` is set when `completion.zsh` is loaded (it remembers the function that TAB was originally bound to), so I recommend that you use it like follows.

```sh
export FZF_COMPLETION_TRIGGER=''
bindkey '^T' fzf-completion
bindkey '^I' $fzf_default_completion
```

PR #233, opened 2015-05-11
Fix ssh completion's grep.

lazywei:
Without the quote `"`, the grep will output unwanted results.
For example, in `~/.fzf`:

```
$ cat ~/.ssh/config 2> /dev/null | \grep -i ^host
grep: bin: Is a directory
grep: man: Is a directory
grep: plugin: Is a directory
grep: shell: Is a directory
grep: src: Is a directory
grep: test: Is a directory
```

### DIFF
--- a/shell/completion.zsh
+++ b/shell/completion.zsh
@@ -82,7 +82,7 @@ EOF
 
 _fzf_ssh_completion() {
   _fzf_list_completion "$1" "$2" '+m' << "EOF"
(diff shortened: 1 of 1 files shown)


junegunn:
Hmm, it works fine on my zsh.

```
mbp% ls | grep ^hadoop
hadoop
hadoop-common-2.0.0-cdh4.2.0
hadoop-snappy-read-only
```

Do you know why `^` works differently on your machine? I agree that quoting the pattern is probably a safer choice, but I'd like to know what's really going on. And please update the bash version as well.

junegunn:
Fixed in c14aa99. I fixed it myself to update bash code as well. Thanks for the patch!

lazywei:
No problem. Thanks for doing that :+1:

Issue #203, opened 2015-04-20
Force fzf-tmux to be used for CTRL-R (command history) usage


megalithic:
Hi there, is there a way (other than editing/overriding the function) to force fzf-tmux to be used when executing command history (CTRL+R) behaviors?

Thanks.

junegunn:
Unfortunately, no. But it's one of the things I've been meaning to address. We currently have `$FZF_TMUX` and `$FZF_TMUX_HEIGHT`, but they only affect `CTRL-T`. I think we should instead make separate vars for each key binding. Also, `CTRL-T` was implemented before `fzf-tmux` was added so it does not leverage the script.

junegunn:
I applied fzf-tmux to CTRL-R and ALT-C, so that all three key bindings now use tmux pane when available. I didn't try to add FZF_XXX_TMUX / FZF_XXX_TMUX_HEIGHT pairs for each one of them as it introduces too many configuration knobs.

megalithic:
Thanks tons June, i'll give it a try here shortly.

Tranquility:
The change broke CTRL-R for me. After using it a few times it stops working. When I hit ENTER the prompt is still empty and it no longer gets the text. 

After restarting the terminal it works a few times but the problem always comes back.

I am using zsh. Do you need any other information to find the bug?

junegunn:
@Tranquility Can't reproduce. So you mean that tmux split correctly appears and you can select an item there, but the selected one does not update command line, right? And how many is "a few times"?  Did you try re-sourcing `~/.fzf.zsh` when that happens?

Tranquility:
> So you mean that tmux split correctly appears and you can select an item there, but the selected one does not update command line, right?

Exactly!

> And how many is "a few times"?

I can't reproduce it reliably. Sometime it's two sometimes it's twenty. Sometimes it starts to work again after a few times. Sometimes it doesn't (or maybe I did not try enough times).

> Did you try re-sourcing ~/.fzf.zsh when that happens?

Yes, I doesn't help.


I changed the fzfcmd to back from fzf-tmux to fzf for now.

junegunn:
@Tranquility I can't reproduce no matter how many times I try. I'm a bash user and there's nothing in my .zshrc. Maybe fzf conflicts with other parts of your zsh config.

joaociocca:
[...]

PR #242, opened 2015-05-26
fix for broken CTRL-R in zsh

### LINKED ISSUE #241: CTRL-R broken in zsh
For me ctrl-r stops working after a couple of uses. This sounds a lot like the bug mentioned after the patch in #203

robinro:
fixes #241 and the followup of #203 
When looking for the bug I found that whenever CTRL-R is broken, it never entered the ```if selected=...```
This change fixes that and covers all cases I could think of.

### DIFF
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -36,7 +36,8 @@ bindkey '\ec' fzf-cd-widget
 # CTRL-R - Paste the selected command from history into the command line
 fzf-history-widget() {
   local selected restore_no_bang_hist
-  if selected=$(fc -l 1 | $(__fzfcmd) +s --tac +m -n2..,.. --tiebreak=index --toggle-sort=ctrl-r -q "$LBUFFER"); then
+  selected=$(fc -l 1 | $(__fzfcmd) +s --tac +m -n2..,.. --tiebreak=index --toggle-sort=ctrl-r -q "$LBUFFER")
+  if [ -n $selected ]; then
     num=$(echo "$selected" | head -n1 | awk '{print $1}' | sed 's/[^0-9]//g')
     if [ -n "$num" ]; then
       LBUFFER=!$num


junegunn:
Thanks, but can you explain how this slight adjustment of the code fixes the issue? I don't see anything wrong with the current code (and I can't reproduce the problem in the first place). Is this a known issue of zsh? By the way, you have to change `[ -n $selected ]` to `[ -n "$selected" ] `or it will always evaluate to true even when it's empty.

/cc @Tranquility

junegunn:
Which version of zsh are you running? Do you have any zsh plugin that might affect the behavior of fzf?

robinro:
I have a non-trivial zsh setup and didn't have time to test this bug in more detail.
Sourcing the original keybinding didn't fix the bug. Sourcing the modified one did.

The patch corresponds to the change that came out of 5 mins playing around with the zsh-function bound to CTRL-R. I overlooked the missing ", so if your right, the fix is equivalent to removing the outer if. The if for ```num``` should be enough anyway.

I'll let you know if the bug shows up again when I have more time to spend investigating it.

junegunn:
The outer if allows you to early-exit the code when fzf is terminated with ctrl-c or esc, but as you mentioned the end result is the same with or without it. But anyway I don't think the code is flawed in itself, and if it somehow causes the problem, then it is likely an issue of zsh (or tmux) and that's why I asked you about the zsh/tmux version you have. I'm not against having workarounds in the code for external/environmental issues, but in this case we seem to have little knowledge of the exact situation.

lyeoh:
I had the same problem and found it was due to git_prompt_short_sha() in my prompt, defined in `oh-my-zsh/lib/git.zsh`. It somehow breaks fzf whenever the pwd is not in a git repo. Simply defining a new function

```
function git_prompt_short_sha() {
  echo $(command git rev-parse --short HEAD 2> /dev/null)
}
```

fixed it for me.

Further, using `set -A selected ...` in place of `selected=( ... )` also fixed it independently.

robinro:
@lyeoh thanks for the suggestion
I also have git_prompt_short_sha defined via oh-my-zsh, so this seems plausible. My issue is, that I'm not able to get the issue reproducibly, inside or outside of repositories. Nevertheless it appeared less often recently and if your suggestion fixes it for you it should be included in my opinion.

ajorgensen:
This fix seems to only select the second change in the list? No matter what item I select in the list I always get the same entry unless I change
```
    num=$selected[1]
```
to
```
    num=$selected[5]
```

or some other number.


I have a similar situation where I have a function that adds some git information to my prompt and when I remove it the tmux script works however I tried putting command in front as suggested by @lyeoh and it still seems to fail.

PR #252, opened 2015-06-06
Portable implementation of swapOutput

dominikh:
Assigning to stdin, stdout and stderr is not portable. They're specified
to be macros, and for example musl will not allow such code to be
compiled.

Use dup2 instead.

### DIFF
--- a/src/curses/curses.go
+++ b/src/curses/curses.go
@@ -4,11 +4,6 @@ package curses
 #include <ncurses.h>
 #include <locale.h>
 #cgo LDFLAGS: -lncurses
-void swapOutput() {
-  FILE* temp = stdout;
-  stdout = stderr;
-  stderr = temp;
-}
 */
 import "C"
 
@@ -142,6 +137,7 @@ var (
 	_colorMap     map[int]int
 	_prevDownTime time.Time
 	_clickY       []int
+	_screen       *C.SCREEN
 	Default16     *ColorTheme
 	Dark256       *ColorTheme
 	Light256      *ColorTheme
@@ -254,10 +250,9 @@ func Init(theme *ColorTheme, black bool, mouse bool) {
 		// syscall.Dup2(int(in.Fd()), int(os.Stdin.Fd()))
 	}
 
-	C.swapOutput()
-
 	C.setlocale(C.LC_ALL, C.CString(""))
-	C.initscr()
+	_screen = C.newterm(nil, C.stderr, C.stdin)
+	C.set_term(_screen)
 	if mouse {
 		C.mousemask(C.ALL_MOUSE_EVENTS, nil)
 	}
@@ -316,7 +311,7 @@ func initPairs(theme *ColorTheme, black bool) {
 
 func Close() {
 	C.endwin()
-	C.swapOutput()
+	C.delscreen(_screen)
 }
 
 func GetBytes() []byte {


junegunn:
Thanks, then can we use [syscall.Dup2](https://golang.org/pkg/syscall/#Dup2) instead?

dominikh:
You tell me. You authored b277f5ae6fe9b263410945796263aa52d95d4ab5 which switched from using syscall.Dup2 to writing a C function instead and I couldn't quite figure out why.

junegunn:
Ha, true. I vaguely remembered that I had issues with `syscall.Dup2` (which is probably equivalent to raw `dup2`) but I completely forgot that I actually reverted it in the commit history. I'm noticing that the 32-bit binary built with `dup2` (your patch) on OS X does not run as expected from time to time not showing the list from the default command correctly.

dominikh:
Unfortunately I don't have access to OS X and can't debug this issue, although I'm a bit surprised that it only happens from time to time.

How about using newterm instead of initscr? newterm allows one to choose the output and input FILE*, so instead of swapping stdout and stderr, one could simply pass stderr in.

junegunn:
I don't have experience with newterm so I can't really say, but I have no objection replacing initscr with newterm if it solves the issue with no side effect.

By the way, actually I don't worry too much about the portability of the source code since fzf is basically distributed as compiled binaries and I build linux binaries with the environment set up with [the attached Dockerfile](https://github.com/junegunn/fzf/blob/master/src/Dockerfile.arch).

dominikh:
> By the way, actually I don't worry too much about the portability of the source code since fzf is basically distributed as compiled binaries

No Linux distribution out there will use your binaries, they will compile from source. And some distributions don't use glibc. In fact, that's how I noticed the issue in the first place, becaue fzf wouldn't build for the musl variant of Void Linux.

Using newterm has one "drawback". It returns a screen object, which theoretically needs to be freed, to reclaim memory. If you intend the curses package to be used multiple times over the lifetime of the application, Init would have to return a value that tracks the screen, so we can free it later. If the package is only used in the context of "run it once, let the process exit", then we don't need to free the screen and can keep the current API. Would you prefer a more robust design or not changing the API?

junegunn:
Users can still follow the official instruction and download the binaries by running the install script. Packages from various linux distros are not my concern as I don't use or maintain them :) But the case of void linux sounds like a valid reason to address this.

What do you mean by changing the API, which one? Can you elaborate?

junegunn:
> What do you mean by changing the API, which one?

Maybe you meant adding a function to curses package to close the screen object? Fine with me. But it's true that it's unnecessary as fzf is just a standalone program.

By the way, I just realized that the issue with `dup2` I mentioned above is reproducible with 64-bit binary as well.

dominikh:
Actually, it probably doesn't matter.

The point was that newterm returns a screen object we need to remember and later free, that is `curses.Close` needs to call `delscreen(obj)`. But I see now that `curses` acts as a singleton already, with a lot of global state (such as `_in`), so I guess you're not concerned with concurrent use of the package and we can just track the screen that way? Then this would be the new patch:

```diff
diff --git c/src/curses/curses.go w/src/curses/curses.go
index 985dd87..fc0fef2 100644
--- c/src/curses/curses.go
+++ w/src/curses/curses.go
@@ -4,11 +4,6 @@ package curses
 #include <ncurses.h>
 #include <locale.h>
 #cgo LDFLAGS: -lncurses
-void swapOutput() {
-  FILE* temp = stdout;
-  stdout = stderr;
-  stderr = temp;
-}
 */
 import "C"
 
@@ -142,6 +137,7 @@ var (
        _colorMap     map[int]int
        _prevDownTime time.Time
        _clickY       []int
+       _screen       *C.SCREEN
        Default16     *ColorTheme
        Dark256       *ColorTheme
        Light256      *ColorTheme
@@ -254,10 +250,9 @@ func Init(theme *ColorTheme, black bool, mouse bool) {
                // syscall.Dup2(int(in.Fd()), int(os.Stdin.Fd()))
        }
 
-       C.swapOutput()
-
        C.setlocale(C.LC_ALL, C.CString(""))
-       C.initscr()
+       _screen = C.newterm(nil, C.stderr, C.stdin)
+       C.set_term(_screen)
        if mouse {
                C.mousemask(C.ALL_MOUSE_EVENTS, nil)
        }
@@ -316,7 +311,7 @@ func initPairs(theme *ColorTheme, black bool) {
 
 func Close() {
        C.endwin()
-       C.swapOutput()
+       C.delscreen(_screen)
 }
 
 func GetBytes() []byte {
```

If you're happy with that, I'll update the PR.

junegunn:
I see. The new patch looks good.

dominikh:
I pushed the new version of the patch.

junegunn:
Merged, thanks! I wish I'd knew about `newterm` before :)

dominikh:
Thanks for merging.

junegunn:
No problem. I've always wanted to get rid of the hackery.

PR #255, opened 2015-06-08
add support to nil-byte separated input strings; closes #121

### LINKED ISSUE #121: Accept input separated by NUL bytes
This would be useful with `find`'s `-print0`.

dullgiulio:
Simplified reader and added a configurable delimiter. Added "--null" option to interface.

Not sure I added the help line in the right place. Feel free to fix it yourself.


dullgiulio:
Committed with wrong identity, will retry.

PR #256, opened 2015-06-08
add support to nil-byte separated input strings, closes #121

### LINKED ISSUE #121: Accept input separated by NUL bytes

dullgiulio:
Simplified reader and added a configurable delimiter. Added "--null" option to interface.

Not sure I added the help line in the right place. Feel free to fix it yourself.

### DIFF
--- a/src/core.go
+++ b/src/core.go
@@ -113,7 +113,7 @@ func Run(opts *Options) {
 	// Reader
 	streamingFilter := opts.Filter != nil && !sort && !opts.Tac && !opts.Sync
 	if !streamingFilter {
-		reader := Reader{func(str string) { chunkList.Push(str) }, eventBox}
+		reader := Reader{func(str string) { chunkList.Push(str) }, eventBox, opts.ReadZero}
 		go reader.ReadSource()
 	}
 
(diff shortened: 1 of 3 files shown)


junegunn:
Aha, so `ReadString` doesn't suffer the problem of fixed buffer size of 64kb, right? That's nice. Thanks. I'll have to add a test case and update man page before merging it.

dullgiulio:
AFAIK the 64k problem is only in the Scanner. See: http://golang.org/src/bufio/bufio.go#L358 for ReadString implementation. It's similar to what you did.

I also fixed the spacing and period and that made the build fail?

junegunn:
Probably not. Travis CI has been flaky due to its reliance on timeouts and tmux send-keys.

junegunn:
By the way, can you squash the commits into one? I'll do the rest of the job later in the day or tomorrow.

dullgiulio:
Squashed. Let me know if you need anything more here.

junegunn:
Merged to master, thanks!
I'll update the man page when I release the next version.

PR #264, opened 2015-06-13
Avoid conflict with other neoterm plugins.

kassio:
To avoid conflict with other neoterm plugins that manage terminals,
prefer named terminals.

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -98,7 +98,7 @@ function! fzf#run(...) abort
 try
   let oshell = &shell
   set shell=sh
-  if has('nvim') && bufexists('[FZF]')
+  if has('nvim') && bufexists('term://*:FZF')
     echohl WarningMsg
     echomsg 'FZF is already running!'
     echohl None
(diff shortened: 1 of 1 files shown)


kassio:
Also remove unnecessary `bd!`.

Instead of use a non-documented behaviour of `:file` command, with named terminal buffers, the `bd!` used to delete the FZF buffer after it's closed will not be required anymore.

junegunn:
Thanks, but with the patch the name of the buffer is shown with `term://` prefix instead of simple `[FZF]` and I don't think I like it.

Is there a way to fix this?

> Instead of use a non-documented behaviour of :file command, with named terminal buffers, the bd! used to delete the FZF buffer after it's closed will not be required anymore.

I'm not sure if I follow you here. I'm using `:file` command to change the name of the buffer (`:help :file_f`) and I do not understand how giving `name` to the terminal buffer can be a fix to the problem of superfluous use of `bd!`. @justinmk mentioned that it's a known issue of neovim: see https://github.com/junegunn/fzf/commit/34e5e2dd8200fc2cf4b298cd2d5583f2875c7a70#comments

justinmk:
Terminal buffer that is not named `term://...` seems like a bad idea. At least, autocmds looking for `term://` won't work. Not sure of any other specific problems, yet. I advocated for making terminal buffer names readonly.

kassio:
My bad about `:file_f` I have not found the documentation. :blush:

@justinmk just explained for me about the `term://...`, thank you. :smile: 

About the `bd!`, the `name` thing fixes it because `:file` creates another buffer, from `:h :file_f`:

>If the buffer did have a name, that name becomes the |alternate-file| name.  An unlisted buffer is created to hold the old name

So, with `:file [FZF]`, actually you have 2 buffers (one named as `term://...` and named as `[FZF]`), this is why you were needing the `bd!`:

nvim 2015-06-13 11-54-10

kassio:
About the visual name, you could add to your plugin something like:
```viml
au TermOpen *FZF setlocal statusline=[FZF]
```

And you would have:
nvim 2015-06-13 12-03-49

junegunn:
@kassio Thanks for the explanation. So, `term://` prefix is the convention that we are going to have to follow, whether I like it or not. Overriding `&statusline` can be an option but people using fancy statusline plugins will not appreciate it. Anyway, I'll merge the PR, thanks!.

justinmk:
So far this seems to have fixed the problem I mentioned in another issue where the file is opened i a `[Program exited, press any key to close]` state. This would make sense based on my past experience with the `:file` command.

PR #268, opened 2015-06-17
Add Related projects section to README.md

balta2ar:
I think it's worth sharing knowledge about alternative implementations, similar solutions. This is basically a copy-paste from [percol's README](https://github.com/mooz/percol#related-projects).

### DIFF
--- a/README.md
+++ b/README.md
@@ -422,6 +422,18 @@ Use iconv to convert NFD paths to NFC:
 find . | iconv -f utf-8-mac -t utf8//ignore | fzf
 ```
 
+Related projects
+----------------
+
+- [canything by @keiji0](https://github.com/keiji0/canything)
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks but I think we should put this on a wiki page instead, so that anyone can edit. Besides these alternative projects that are similar to fzf, there are also projects that tackle the same set of problems with different approaches (e.g. pathpicker), and there are projects that depend on fzf. It would be nice to list them on a single wiki page.

balta2ar:
How about moving it to the wiki but adding a link to it in Related projects section of the README? Like you did in Examples section. After all, it's convenient to have links to other parts of the project in the README.

junegunn:
That's okay, but I don't want to talk about other projects in the README page, especially the ones that are not directly related to fzf, it feels too "meta".

PR #272, opened 2015-06-20
use vi-fetch-history on zsh to get history line

okapia:
In addition to being simpler and not relying on zsh options, it allows subsequent up/down history or accept-line-and-down-history widgets to work.

I also replaced your pipeline for extracting the number with what I think is simpler: assign the output to an array and then take the first element.

Furthermore, The cd widget didn't work in my configuration because of two aliases I have defined:
  alias find='noglob find'
  alias command='command '
The later enables alias expansion after command. It'd be better to use zsh's autoloadable functions: put the functions each in their own file, .fzf.zsh would then need to add that directory to fpath:
  fpath+=( $fzf_base/shell/fpath )
and autoload them with the -U option to disable alias expansion:
  autoload -Uz fzf-history-widget fzf-file-widget fzf-cd-widget __fzfcmd __fsel

### DIFF
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -26,7 +26,7 @@ bindkey '^T' fzf-file-widget
 
 # ALT-C - cd into the selected directory
 fzf-cd-widget() {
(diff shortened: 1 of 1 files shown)


junegunn:
Very nice, thanks!

PR #285, opened 2015-07-04
Fix g++: possible retry loop

evverx:
Hi!
There is a problem with a `g++` Bash autocompletion:
`g++`<kbd>space</kbd><kbd>tab</kbd> causes `g++ bash: warning: programmable_completion: g++: possible retry loop`

How to reproduce:
Ubuntu 14.04
Bash-4.3
bash-completion-2.1
`source ~/.fzf.bash` after `source /usr/share/bash-completion/bash_completion` in `~/.bashrc`

See http://unix.stackexchange.com/q/213432/120177

### DIFF
--- a/shell/completion.bash
+++ b/shell/completion.bash
@@ -74,9 +74,10 @@ _fzf_opts_completion() {
 }
 
 _fzf_handle_dynamic_completion() {
-  local cmd orig ret
+  local cmd orig ret orig_cmd
   cmd="$1"
   shift
+  orig_cmd="$1"
 
   orig=$(eval "echo \$_fzf_orig_completion_$cmd")
   if [ -n "$orig" ] && type "$orig" > /dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


junegunn:
Hi, thanks for looking into this. Can you squash the commits into one? Meanwhile, I'll set up ubuntu 14 environment and make sure that your fix solves the issue.

evverx:
Done.
@junegunn , thanks for the useful tool!

junegunn:
Downloading the vagrant box for trusty is taking forever, so I'm just going to merge as I don't see any issue with your fix. (So the problem was `sed` replacing `g++` to `g__`, right?) Thanks!

evverx:
Yes, you are right. `cmd` in `_fzf_handle_dynamic_completion` for `g++` was `g__`. `complete | grep '\-F.* g__` found nothing.

junegunn:
:+1:

PR #326, opened 2015-08-24
docs(): add note about using ag inside of vim

mhartington:
Added a small note about ignoring your `.gitignore` using fzf.
Not initially clear that you need to add that to your shells rc files

### DIFF
--- a/README.md
+++ b/README.md
@@ -346,6 +346,9 @@ fzf
 export FZF_CTRL_T_COMMAND="$FZF_DEFAULT_COMMAND"
 ```
 
+If you're using vim and wish to ignore directories listed in your `.gitignore`, you'll need to set fzf to use ag in you're shell's rc file.
+
+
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks, but I believe the current documentation already covers that. See: https://github.com/junegunn/fzf/blame/master/README.md#L284-L285. And the variables are mentioned in [the usage section](https://github.com/junegunn/fzf#usage) and in `--help` output as well so it's really hard to miss. FYI, you can define `FZF_DEFAULT_COMMAND` or `FZF_DEFAULT_OPT` in your .vimrc instead if you want to; e.g. `let $FZF_DEFAULT_COMMAND = 'find .'`.

mhartington:
Ahh, alright, my mistake. Sorry for the noise.

junegunn:
No problem. Thanks.

PR #336, opened 2015-08-29
fix fish streams

fazibear:
Fix fish commands to stream data to fzf, and don't wait.

### DIFF
--- a/shell/key-bindings.fish
+++ b/shell/key-bindings.fish
@@ -19,7 +19,7 @@ function fzf_key_bindings
       -o -type f -print \
       -o -type d -print \
       -o -type l -print 2> /dev/null | sed 1d | cut -b3-"
-    eval $FZF_CTRL_T_COMMAND | eval (__fzfcmd) -m > $TMPDIR/fzf.result
+    eval "$FZF_CTRL_T_COMMAND | "(__fzfcmd)" -m > $TMPDIR/fzf.result"
     and commandline -i (cat $TMPDIR/fzf.result | __fzf_escape)
     commandline -f repaint
     rm -f $TMPDIR/fzf.result


junegunn:
Thanks. It looks like I fixed the issue in https://github.com/junegunn/fzf/commit/305ec3b3ce16141ecac0b8657ad00b371fc7a133, but accidentally reintroduced the problem in 938151a834c1858ffdd6091cbcb42688a3784b6b. So we only need to fix CTRL-T, CTRL-R and ALT-C don't have the problem. Can you to update and force-push the commit?

shell/key-bindings.fish
```diff
@@ -19,14 +19,14 @@ function fzf_key_bindings
       -o -type f -print \
       -o -type d -print \
       -o -type l -print 2> /dev/null | sed 1d | cut -b3-"
-    eval $FZF_CTRL_T_COMMAND | eval (__fzfcmd) -m > $TMPDIR/fzf.result
+    eval "$FZF_CTRL_T_COMMAND |" (__fzfcmd)" -m > $TMPDIR/fzf.result"
```
junegunn: Just a minor nitpick, but I'd like the double-quotes around `(__fzfcmd)` to be symmetric.

fazibear:
Double quotes fixed. I changed all 3 function to make all eval consistance.

junegunn:
Thanks, but what I meant is that we should only fix CTRL-T, and not the other ones. Changing the code that doesn't have the problem makes it harder to understand what the problem was and what our workaround was in retrospect. And please squash the commits.

fazibear:
Done.

junegunn:
Thanks you!

Issue #359, opened 2015-09-27
Behavior of sink* when match not found


jebaum:
I'm attempting to create a fzf.vim function which will either load a session (managed by https://github.com/manuel-colmenero/vim-simple-session) or create a new one. vim-simple-session contains a function called `s:SwitchSession` which, given a string, will switch to a session if it exists or create a new one by that name if it doesn't. I would expect the below code to work, but from what I can tell due to the commented out `echom` lines, the sink isn't getting called in the case that the exit status of fzf is 1 (a query was entered but there was no match). So the function works for switching to an existing session, but not for creating a new one.

related: https://github.com/junegunn/fzf/issues/345

```
function! Fzfvimsessions()
  let sessions = session#list()
  " call add(sessions, 'idk')
  " call add(sessions, 'bro')
  " echom join(sessions, "::")

  call fzf#run({
  \ 'source':   sessions,
  \ 'sink*':    function('CallSimpleSessionSwitch'),
  \ 'window':  '10new',
  \ 'options': '--print-query -x --tiebreak=index --ansi --prompt="Sessions> "',
  \})
endfunction

function! CallSimpleSessionSwitch(name)
  " echom 'wrapper to SwitchSession called'
  " echom join(a:name, "--")
  " due to --print-query, first item in list will be the typed query
  let session_name = a:name[-1]
  call SCall('plugin/session.vim', 'SwitchSession', [session_name])
endfunction
```

jebaum:
problem is around line 238 of fzf.vim where it checks v:shell_error. not sure what the proper behavior should be. I don't see why an exit status of 1 shouldn't continue on to call the callback function, but the comment there makes it pretty explicit so I'm guessing you have a reason. perhaps a new option should be accepted in the dictionary you can pass to `fzf#run()` to determine how exit status 1 should be handled?

jebaum:
actually after a little more investigation this definitely seems like a bug. if calling the tmux or neovim execute funcions, if fzf exits with status 1 it still calls the sink function. so my code from my original comment works fine if you open fzf in a tmux pane or a neovim `:term`. it's just the case of vim with fzf not in tmux where the behavior is different.

junegunn:
Cool, thanks. Exit status of 1 was a recent change, and I think I forgot to update the vim plugin accordingly.

junegunn:
So it should be:
- 0 or 1 -> sink
- 130 -> cancel, ignore
- rest -> error message

junegunn:
[...]

PR #360, opened 2015-09-27
[vim] handle exit status 1 properly (#359)

jebaum:


jebaum:
I'm guessing using `index()` to check two values at once is a style preference, I tend to prefer things as straightforward as possible but I can switch it back to your way if you want

PR #361, opened 2015-09-30
[vim] handle SwapExists

justinmk:
The SwapExists dialog prevents multiple files from being opening if the
dialog occurs before all files are opened. Opening the files is more
important than showing the dialog, so choose "readonly" automatically
and continue opening files.

This problem exists in neovim and Vim 7.3. Not sure about others. To test, choose multiple files in the `:FZF` prompt; if a swap-exists dialog appears before all files are opened, those remaining files will not be opened.

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -405,6 +405,12 @@ function! s:cmd_callback(lines) abort
   endif
   let key = remove(a:lines, 0)
   let cmd = get(s:action, key, 'e')
+  if len(a:lines) > 1
+    augroup fzf_swap
+      autocmd SwapExists * let v:swapchoice='o'
+            \| call s:warn('fzf: E325: swap file exists: '.expand('<afile>'))
+    augroup END
+  endif
   try
     let autochdir = &autochdir
     set noautochdir
@@ -413,6 +419,7 @@ function! s:cmd_callback(lines) abort
     endfor
   finally
     let &autochdir = autochdir
+    silent! autocmd! fzf_swap
   endtry
 endfunction
 


plugin/fzf.vim
```diff
@@ -405,6 +405,10 @@ function! s:cmd_callback(lines) abort
   endif
   let key = remove(a:lines, 0)
   let cmd = get(s:action, key, 'e')
+  augroup fzf_swap
+    autocmd SwapExists * let v:swapchoice='o'
+          \| call s:warn('fzf: E325: swap file exists: '.expand('<afile>'))
```
justinmk: This does not appear in the statusline, but it will be helpful in case user wonders why a file was opened RO.

junegunn:
Thanks, makes good sense.
How about if we do this only when multiple files are selected?

(I need to apply this patch to fzf.vim project as well.)

justinmk:
>  only when multiple files are selected?

Is that a matter of checking the length of `a:lines`? 

I would lean towards it being consistent in both cases--otherwise users will be confused that the dialog sometimes appears and sometimes doesn't, when using FZF.

junegunn:
Yeah I see your point and I also don't like the inconsistency. But what I'm thinking is that considering that `:FZF` is *mostly* (no proof) used to open a single file (and I believe that many users aren't even aware that it can select multiple entries especially since it requires extra `-m` option), it makes sense to make it consistent with the default behavior of Vim just in that case even if it means introducing some inconsistency.

justinmk:
:+1:

junegunn:
Alright, could you update your commit? Simply surrounding autocmd with `if len(a:lines) > 1` statement will do.

junegunn:
Merged, thanks.

PR #362, opened 2015-09-30
Added password-store completion.

d4ndo:
fuzzy search for the the standard unix password manager (password entries). http://www.passwordstore.org/

```bash 
apt-get install pass 
```

Usage:

```bash 
pass **<TAB>
pass -c **<TAB>
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -234,6 +234,15 @@ ssh **<TAB>
 telnet **<TAB>
 ```
 
+#### Password-store
+
+fuzzy search for the standard unix password manager:
+
+```sh
+pass **<TAB>
+pass -c **<TAB>
+```
+
 #### Environment variables / Aliases
 
 ```sh
--- a/README.md
+++ b/README.md
@@ -236,7 +236,7 @@ telnet **<TAB>
 
 #### Password-store
 
-fuzzy search for the standard unix password manager:
+fuzzy search for the standard unix [```password manager```](http://www.passwordstore.org):
 
 ```sh
 pass **<TAB>
--- a/README.md
+++ b/README.md
@@ -236,7 +236,7 @@ telnet **<TAB>
 
 #### Password-store
 
-fuzzy search for the standard unix [```password manager```](http://www.passwordstore.org):
+fuzzy completion for [pass command](http://www.passwordstore.org):
 
 ```sh
 pass **<TAB>
--- a/shell/completion.bash
+++ b/shell/completion.bash
@@ -206,9 +206,11 @@ EOF
 }
 
 _fzf_pass_completion() {
-  pwdir=${PASSWORD_STORE_DIR-~/.password-store}
+  local pwdir=${PASSWORD_STORE_DIR-~/.password-store/}
+  local stringsize="${#pwdir}"
+  let "stringsize+=1"
   _fzf_list_completion '+m' "$@" << "EOF"
-  find "$pwdir" -name "*.gpg" -printf "%P\n" | sed -r 's,(.*)\.gpg,\1,'
+  find "$pwdir" -name "*.gpg" -print | cut -c "$stringsize"- | sed -e 's/\(.*\)\.gpg/\1/'
 EOF
 }
 
--- a/shell/completion.bash
+++ b/shell/completion.bash
@@ -205,6 +205,13 @@ _fzf_ssh_completion() {
 EOF
 }
 
+_fzf_pass_completion() {
+  pwdir=${PASSWORD_STORE_DIR-~/.password-store}
+  _fzf_list_completion '+m' "$@" << "EOF"
+  find "$pwdir" -name "*.gp"g -printf "%P\n" | sed -r 's,(.*)\.gpg,\1,'
+EOF
+}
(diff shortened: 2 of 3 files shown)


junegunn:
Hi, actually I'm not familiar with this pass. Since it's hard for me to maintain a feature that I do not know well, I'm going to need some time to learn the utility. Anyway, here are a few comments.

- Doesn't seem to work on my OS X, I'm getting
    - `sed: illegal option -- r`
    - `find: -printf: unknown primary or operator`
- `$pwdir` should be local
- It would be nice to make "the standard unix password manager" in the README file link to the project homepage since not everyone knows what it is
- Do you think it's possible to add a test case for the feature?
- Please squash the commits into one.

d4ndo:
I dont use OS X. This makes it hard for me to debug. I assume that OSX uses BSD unix commands.

The "printf" option of ```find``` is only supported by the GNU ```find``` command. see my first commit message:

https://github.com/d4ndo/fzf/commit/7e090295ae53515a572af6904d13e7b53b8fa56c

I guess i can fix that. Also the ```sed``` command. By using just (GNU and BSD) compatible options.

junegunn:
> I dont use OS X. 

I can help you with that, once I get to know better about pass. By the way, I'd like to know if this pass is a popular utility and it will be actively maintained in the future, what do you think? I'm concerned about feature bloat and don't want to expand fzf for tools that are not used by many.

netei:
I don't think that this should go inside the fzf repository, but rather in the wiki as a sample.

justinmk:
Though I applaud the general idea, calling pass "the standard unix password manager" is premature since it is not included by default in any distro I am aware of.

d4ndo:
It is the only password manager project that has understood the unix philosophy so far. So it is de facto the ```standard unix password manager```. It has potential and it will be the only password manager that will be compatible to fzf in the near future. To a lot of nerds, there is no alternative to this.

```pass ``` and ```fzf``` would just go well together.

The alternative would be some modification of pass, so it would output some raw text format. 
1 ) It could be called like this:

``` pass --raw | fzf ```

2 ) or some shell function:

```bash
passfzf() {
    local pwdir=${PASSWORD_STORE_DIR-~/.password-store}
    local path=$(find "$pwdir" -name "*.gpg" -printf "%P\n" | \
    sed -r 's,(.*)\.gpg,\1,' | \
    fzf +m);
    [[ ! -z "$path" ]] && pass -c "$path"
}
```

3) However, this is somewhat the best solution, in my opinion:

``` pass **<TAB> ```

The maintenance should be negligible. I hope. 

I will fix it anyway. Maybe you can use it as example @ wiki.

**Comment by d4ndo** [comment]
Tested it on FreeBSD. Does it work on OS X ?

**Comment by junegunn** [comment]
I upgraded to El Capitan today, and Homebrew now fails to compile pass :hankey: 
I'll test it on another macbook later in the day.

I also don't think it's appropriate to call it "the standard unix password manager". With the labelling I was tricked into believing that it was some kind of 30-year-old unix command that I shamefully did not know, but it was not. "`Fuzzy completion for [pass command](link)`" should be sufficient.

**Comment by junegunn** [comment]
Yes, it does work, thanks.

However, I'm not sure if I'm going to merge this. This is a small, simple addition and it's indeed a useful feature. But we can't endlessly add things just because they're useful and we have to draw the line somewhere to keep the repository maintainable in the long run. As you might have noticed, I haven't added any extra shell/vim features to this repository for quite a while. Actually I'm leaning towards the opposite, I want to make it smaller.

Take a look at [fzf.vim](https://github.com/junegunn/fzf.vim) project or the wiki pages. Instead of adding more stuff to this central repository, force-feeding them to every user, we're adding/posting features there only for the users who actually want them. Basically I think we should do the same for fuzzy completion. The problem is however, there is currently no easy, straightforward, and consistent way to add custom completion functions. Once we make it more extensible and come up with nicer API for creating custom fuzzy completion, we can share more useful stuff on a new wiki page.

**Comment by d4ndo** [comment]
No problem. 
It is no easy task to keep the source code lean and simple.

Thanks for testing.

**Comment by junegunn** [comment]
Thanks for understanding and I appreciate your effort. I'll try to come up with a better way to organize/extend completion functions and let you know of the progress.

**Comment by Amar1729** [comment]
(sorry to comment on such an old issue)  
Leaving this here for users on macOS: with the default `bsdfind`, there's a small inconsistency in how the output is displayed:  
```
.password-store//pw1.gpg
```

instead of:
```
.password-store/pw1.gpg
```

so i changed this function to use `sed` instead of `cut` to remove the prefix and slashes from the beginning of the completion candidate:
```
# pass completion suggested by @d4ndo (#362) (slightly modified)
_fzf_complete_pass() {
  _fzf_complete '+m' "$@" < <(
    local pwdir=${PASSWORD_STORE_DIR-~/.password-store/}
    find "$pwdir" -name "*.gpg" -print |
        sed -e "s#${pwdir}/\{0,1\}##" |
        sed -e 's/\(.*\)\.gpg/\1/'
  )
}
```

PR #364, opened 2015-10-01
Remove dependency on zsh/pcre module

### LINKED ISSUE #363: Mac OS X El Capitan, zsh, and fzf
fzf integration with zsh no longer works after having upgraded to Mac OS X 10.11.

I’m getting:

```
/Users/austin/.fzf.zsh:3: failed to load module `zsh/pcre': dlopen(/usr/lib/zsh/5.0.8/zsh/pcre.so, 9): image not found
/Users/austin/.fzf.zsh:3: -pcre-match not available for regex
```

The error is correct—/usr/lib/zsh/5.0.8/zsh/pcre.so does not exist. If I enable ~/.brew/bin/zsh as a valid shell, it all works because pcre.so is included.

halostatue:
Fixes #363.

Manual testing suggests that the changes work with bash as well as zsh.

### DIFF
--- a/install
+++ b/install
@@ -185,7 +185,7 @@ for shell in bash zsh; do
   echo -n "Generate ~/.fzf.$shell ... "
   src=~/.fzf.${shell}
 
-  fzf_completion="[[ \$- =~ i ]] && source \"$fzf_base/shell/completion.${shell}\" 2> /dev/null"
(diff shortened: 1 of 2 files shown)


junegunn:
For some reason, I can't reproduce the problem on my El Capitan (removed zsh from homebrew and rebooted but still can't, must be missing something) but your patch looks harmless anyway so I'll just merge it. Thanks!

PR #371, opened 2015-10-09
Trigger netrw autocommand when opening directory

wilywampa:
Attempting to edit a directory in a `try` block in Vim silently fails to trigger the autocommands used by file browsers (netrw, vimfiler, probably others). Adding a BufEnter autocommand after the edit command if the item was a directory allows netrw to work instead of opening a blank buffer.

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -416,6 +416,9 @@ function! s:cmd_callback(lines) abort
     set noautochdir
     for item in a:lines
       execute cmd s:escape(item)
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks, I didn't know that because the default command does not list directories.
Anyway, can you add `exists('#BufEnter')` to the condition? It can give error message: `No matching autocommands`. (e.g. `vim -Nu NONE`)

junegunn:
Thanks for the quick response, could you squash the commits?

junegunn:
Thanks :+1:

Issue #377, opened 2015-10-12
fzf on android?


acornejo:
I am happy fzf user in linux and os x. Today I tried to install it inside my android tablet (inside a terminal prompt provided by termux) and realized there is no linux armv71 version. Given that go supports cross compiling to other architectures, perhaps we could fzf here too?

junegunn:
Unfortunately, ncurses makes cross compilation impossible (or at least non-trival). I haven't been successful so instead I'm currently building [linux binaries using VM](https://github.com/junegunn/fzf/blob/0.10.8/src/Makefile#L80-L82) on OS X.

I have no experience or knowledge on the android platform you mentioned, but I suppose it's possible in a similar manner given that ncurses is available there.

acornejo:
Ahh, of course, for some reason I thought fzf was pure go, but it does make sense that it links to ncurses for term stuff.

Given that you are already using docker, I'll look into creating a Dockerfile that builds fzf from source. I already have a Dockerfile to build other ncurses based android programs, so I just need to figure out how to get ncurses to link statically in android. I'll try to work on this latter and put up a pull-request for adding a Dockerfile.android

junegunn:
Sound good!

acornejo:
ok, so i've made some progress on the dockerfile and building the cross compiler toolchain for android inside. However, I am getting the following error (looking for ideas):

curses/curses.go: In function 'swapOutput':
curses/curses.go:9:10: error: lvalue required as left operand of assignment
   stdout = stderr;
          ^
curses/curses.go:10:10: error: lvalue required as left operand of assignment
   stderr = temp;

Error seems pretty clear, the android C compiler doesn't like writing to stdout and stderr; presumably because they are defined as const. Can we do without this?

junegunn:
Yo, I think you're looking at an old revision.

acornejo:
Ok, just to give some updates. I got that older revision to compile by removing the swapOutput function (probably it broke functionality, but I just wanted it to see if that would work). After compiling I copied the binary over to my android device and tested it, and it worked!!

Next I saw your message and tried to compile the latest version. Here I am facing a slightly different problem, the error I get is:

/tmp/go-build054928491/github.com/junegunn/fzf/src/curses/_obj/_cgo_main.o:_cgo_main.c:_cgohack_stdin: error: undefined reference to 'stdin'
[...]

PR #380, opened 2015-10-13
add android build

acornejo:
closes issue #377

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -2,11 +2,13 @@ ifndef GOPATH
 $(error GOPATH is undefined)
 endif
 
+ifndef GOOS
 UNAME_S := $(shell uname -s)
 ifeq ($(UNAME_S),Darwin)
 	GOOS := darwin
 else ifeq ($(UNAME_S),Linux)
 	GOOS := linux
+endif
 endif
 
 ifneq ($(shell uname -m),x86_64)
@@ -16,11 +18,13 @@ endif
 SOURCES   := $(wildcard *.go */*.go)
 BINDIR    := ../bin
 
-BINARY32  := fzf-$(GOOS)_386
-BINARY64  := fzf-$(GOOS)_amd64
-VERSION    = $(shell fzf/$(BINARY64) --version)
-RELEASE32  = fzf-$(VERSION)-$(GOOS)_386
-RELEASE64  = fzf-$(VERSION)-$(GOOS)_amd64
+BINARY32    := fzf-$(GOOS)_386
+BINARY64    := fzf-$(GOOS)_amd64
+BINARYARM7  := fzf-$(GOOS)_arm7
+VERSION     := $(shell awk -F= '/version =/ {print $$2}' constants.go | tr -d "\" ")
+RELEASE32   = fzf-$(VERSION)-$(GOOS)_386
+RELEASE64   = fzf-$(VERSION)-$(GOOS)_amd64
+RELEASEARM7 = fzf-$(VERSION)-$(GOOS)_arm7
 
 all: release
 
@@ -30,6 +34,11 @@ release: build
 	rm -f $(RELEASE32) $(RELEASE64)
 
 build: test fzf/$(BINARY32) fzf/$(BINARY64)
+
+android-build:
+	cd fzf && GOARCH=arm GOARM=7 CGO_ENABLED=1 go build -a -ldflags="-extldflags=-pie" -o $(BINARYARM7)
+	cd fzf && cp $(BINARYARM7) $(RELEASEARM7) && tar -czf $(RELEASEARM7).tgz $(RELEASEARM7) && \
+	rm -f $(RELEASEARM7)
 
 test:
 	go get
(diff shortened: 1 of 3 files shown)


junegunn:
Great, I'll test it tomorrow. A few comments.
- Other dockerfiles use Go 1.4.2. Let's use the same version here.
- Is it possible to add makefile target so I can build android binary with `make android` a la `make linux`?
- If it's possible, I can upload the binaries to https://github.com/junegunn/fzf-bin/releases. Then we can update install script to download it on android platforms.

acornejo:
All your comments make sense, I may be using an older go version only because I started off an old branch where I copied one of the older dockerfiles as a starting point.

If we bump versions of go the sed patch I have there to remove pthread constants may no longer be necessary since that has been fixed upstream for some time now. It may also be that the stdin/stderr hacks in go have been cleaned up (not too hopeful on that).

We can certainly add make targets for android and upload binaries.

junegunn:
The build was successful with [your instruction](https://github.com/junegunn/fzf/issues/377#issuecomment-147773338) (although I had to remove `=` from https://github.com/junegunn/fzf/pull/380/files#diff-1011bc30beb7b7f9636f1e96746c44deR25 as I was using an old version of Docker), but how am I going to check if the binary actually works without an android tablet? Any idea? I need to be able to check if it works if I'm going to release official android binaries.

> If we bump versions of go, then the sed patch I have there to remove pthread constants may no longer be necessary since that has been fixed upstream for some time now. It may also be that the stdin/stderr hacks in go have been cleaned up (not too hopeful on that).

You can try Go 1.5 instead of 1.4.x for android build if it helps.

acornejo:
Cool. Thanks for testing this.

I tested the build with 1.4.2 (to be consistent with the other dockerfiles) and it worked; however all the patches are still necessary. If we go to 1.5 we can definitely remove at least one of the patches, maybe more. It is up to you to decide if we want to give it a shot, or stick with 1.4.2 for consistency with the other builds.

I just pushed an update to the branch with the changes to use 1.4.2, and to add the android targets to the makefile (they are like the centos target which drop you in a shell ready to type the build commands, if we want to automate the bit which actually calls the go compiler inside the container, then we will need to either a .sh script or make some rather intrusive changes to the Makefile).

I also changed the = line in the Dockerfile.

As for testing, there are several alternatives. The cleaner I think is to setup a testing environment inside the same container that we are building.

I have no experience with the "go test" framework though. What I can do is provide you a way for running arm binaries inside the docker container, so if that if that is sufficient to run the tests, then we can make it work, I think at the very least it would allow us to run test_go.rb, but not sure how go handles testing cross compiled things (if at all).

I have another idea which might make "go test" work. It involves emulating an arm architecture so that the go compiler is compiling from arm to arm (i.e. not cross compilation), I have something like that half working for another project, so if I make more progress on that front we can go in that direction. The downside is that compilation will be slower (obviously), and the Dockerfile will be a lot more complex.

junegunn:
`go test` runs simple unit tests that do not involve ncurses, so I wouldn't worry too much about it. The real "acceptance tests" are done by `test_go.rb` which simulates user interaction with tmux send-keys. But I think it's probably good enough that they are tested on OS X or Linux (travis-ci). I'll merge this after a few things ironed out.

> It involves emulating an arm architecture so that the go compiler is compiling from arm to arm

That's what I initially thought you were going to do, then I realized the new dockerfile is based on ubuntu image, but I'm totally okay with it. It works and you say it's simpler.

But we now seem to have a little inconsistency in the naming of makefile targets;

- `make centos`
    - starts centos shell
- `make arch`
    - starts arch linux shell
- `make ubuntu`
    - starts ubuntu shell
- `make linux`
    - **builds** linux binary
- `make android`
    - starts ubuntu shell for building android binaries

You see the point? Maybe we should have `make ubunut-android` for shell and `make android` for build. What do you think?

acornejo:
I went on a little tangent and tried to see if I could get fzf to build inside a virutalized rootfs hosted with qemu-arm. Turns out the qemu ARM emulation doesn't implement the VFPv3 floating point hardware, which then prevents the go compiler from being able to run inside qemu to provide an android arm binary.

On to the current pull-request, I have made the suggested changes, renaming android to ubuntu-android and adding a new android target that actually produces the binaries. I had to tweak the makefile a bit more than I would have wanted, but couldn't keep using the old targets in there since you can't build 32 and 64 bit binaries for arm7 (arm7 is 32 bits, no 64 bit support).

Updated branch, let me know if we want to make any further changes to get this upstream.

src/Makefile
```diff
@@ -81,5 +93,13 @@ linux: docker-centos
...
+ubuntu-android: docker-android
+	docker run -i -t -v $(GOPATH):/go junegunn/$@-sandbox \
+		sh -c 'cd /go/src/github.com/junegunn/fzf/src; /bin/bash'
+
+android: docker-android
+	docker run -i -t -v $(GOPATH):/go junegunn/android-sandbox \
+		/bin/bash -ci 'cd /go/src/github.com/junegunn/fzf/src; GOOS=android make android-build'
+
 .PHONY: all build release test install uninstall clean docker \
 	linux arch ubuntu centos docker-arch docker-ubuntu docker-centos
```
junegunn: LGTM. Can you add new targets to `.PHONY`?

acornejo: @junegunn done (sorry for being MIA, got distracted).

i've been using this for a while, seems to work well.

Unfortunately, the only way to test it is with either an android device, or using the android emulator (this is because go builds for the android target relying on the android specific runtime.. which is not available in, for example, vanilla linux on arm).

junegunn:
Merged, thanks!

PR #381, opened 2015-10-14
Add an improved file-widget for zsh

Vifon:
The new widget uses the already typed path as a root directory for the file search. It's not enabled by default.

### DIFF
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -25,6 +25,27 @@ fzf-file-widget() {
 zle     -N   fzf-file-widget
 bindkey '^T' fzf-file-widget
 
+fzf-file-widget-dwim() {
+  autoload -U modify-current-argument split-shell-arguments
+  local reply REPLY REPLY2
+
+  ((--CURSOR))
+  split-shell-arguments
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks, but this should instead be on the wiki page. I would like the behaviors of the three key bindings consistent across bash, zsh, and fish, and I decided not to add more shell stuff to this repo. Maintenance burden is just too much for me. By the way the widget doesn't seem to handle multiple selections.

Vifon:
I understand.

Yes, I've missed the multiple selection handling. I've managed to fix it but I needed to change some of your functions. Does it still make sense to add it to the wiki?

junegunn:
Thanks for understanding. I want to keep the core simple and small.

> but I needed to change some of your functions

I see, but I think you can still make it work without the code in this repo, make it self-contained, although it can cause some code duplication. It's up to you to put it on the wiki page or not.

Vifon:
Ok. I'll clean it up a bit and then add to the wiki.

junegunn:
Currently we don't have a dedicated page for custom key bindings. Consider creating `Examples (key bindings)` page.

PR #403, opened 2015-11-04
Not relying on exit status for CTRL-R

JackDanger:
In the case that fzf-tmux returns a user-selected result but with a
non-zero exit status (for reasons I haven't figured out quite yet) this
allows CTRL-R to continue working as expected.

Addresses #203 (Tranquility's comment)

### DIFF
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -37,7 +37,8 @@ bindkey '\ec' fzf-cd-widget
 # CTRL-R - Paste the selected command from history into the command line
 fzf-history-widget() {
   local selected num
-  if selected=( $(fc -l 1 | $(__fzfcmd) +s --tac +m -n2..,.. --tiebreak=index --toggle-sort=ctrl-r -q "$LBUFFER") ); then
+  selected=( $(fc -l 1 | $(__fzfcmd) +s --tac +m -n2..,.. --tiebreak=index --toggle-sort=ctrl-r -q "$LBUFFER") )
+  if [ -n "$selected" ]; then
     num=$selected[1]
     if [ -n "$num" ]; then
       zle vi-fetch-history -n $num


junegunn:
Thanks, but the suggested change is identical to #242. I'd like to know why you get non-zero exit status in the first place before merging it. I can't reproduce on my side. Does it always happen or just occasionally happen? Do you still have the problem if your .zshrc is minimal? Can you check which causes the problem? @lyeoh [mentioned that](https://github.com/junegunn/fzf/pull/242#issuecomment-129700522) git_prompt_short_sha was the problem.

/cc @robinro

robinro:
@junegunn @JackDanger 
I spend quite some time back in #242 to look for the reason but couldn't find it. It happened only occasionally and seemed to only happen after some usage of the shell (maybe history size is important? For me this is quite large often and I also increased the limits).

Nevertheless I would recommend to merge the pull request. For me it fixed the problem, apparently also for @JackDanger and there is __no__ downside. It only expliciity takes care of non-zero return code, which the original version ignores. Also I find the split-up `if` a bit more readable.

@junegunn knowing the reason for the non-zero return code would certainly be interesting for fzf. Maybe one can add a debug-section that checks for the return code and then saves out some more details about what happened in fzf before returning to zsh.

junegunn:
What irks me is that doing so simply hides the fundamental problem that can be or should be fixed in the first place and deprives us of a chance to get a better understanding of the situation. I'm not particularly against having various workarounds and hacks for issues that are out of my control, but not without proper understanding of the problem.

robinro:
@junegunn if the problem can not be found within a certain time/effort, still the workaround should be included. Fixing the original problem would be nicer, but not fixing a bug for a long time that keeps many people from using the tool is not reasonable.

junegunn:
@robinro I installed oh-my-zsh (since I couldn't reproduce with vanilla zsh) and noticed that zsh occasionally fails to update the prompt even when fzf-tmux properly completes (logged exit status and observed no anomaly there, always 0). I applied the patch, but the problem still persists. I don't think this is a proper fix to the problem.

@JackDanger Can you try logging the exit status? I don't understand that fzf-tmux exits with non-zero code (unless the selected item is empty).

robinro:
@junegunn Thanks for looking into this. For me the patch resolved the problem, but I'll also look into logging the return code and collecting more info what's going on. I will only be able to do this towards the end of next week.

junegunn:
One correction to my previous comment: I can't reproduce the problem with or without the patch. I was misguided because I was doing this:

- `CTRL-R`
- Select history entry X (OK)
- Delete the entry with `CTRL-U`
- `CTRL-R`
- Select Y (OK)
- `CTRL-R`
- Reselect X
- Since I previously edited (deleted) the entry X, my prompt is not updated.

Anyway, here's what I did:

```diff
diff --git a/bin/fzf-tmux b/bin/fzf-tmux
index 1085b96..47c38b1 100755
--- a/bin/fzf-tmux
+++ b/bin/fzf-tmux
@@ -134,5 +134,7 @@ else
   cat <&0 > $fifo1 &
 fi
 cat $fifo2
-exit "$(cat $fifo3)"
+x="$(cat $fifo3)"
+echo "fzf-tmux: $x" >> /tmp/x
+exit $x
 
diff --git a/shell/key-bindings.zsh b/shell/key-bindings.zsh
index c429d91..f192296 100644
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -40,8 +40,11 @@ fzf-history-widget() {
   if selected=( $(fc -l 1 | $(__fzfcmd) +s --tac +m -n2..,.. --tiebreak=index --toggle-sort=ctrl-r -q "$LBUFFER") ); then
     num=$selected[1]
     if [ -n "$num" ]; then
+      echo "$num" >> /tmp/x
       zle vi-fetch-history -n $num
     fi
+  else
+    echo "else" >> /tmp/x
   fi
   zle redisplay
 }
```

and I `tail -f /tmp/x` in another tmux pane to see what's going on. I never was able to see `else` in the file unless I interrupt fzf with CTRL-C.

JackDanger:
>  Can you try logging the exit status?

@junegunn It was `128`. I didn't get any more info about it than that but, like other folks here, I've noticed that the behavior is different depending on if my current working directory is a git repo. So there's something about ZSH + my `$PS1` and the git functions that are causing fzf to exit uncleanly. I'll be damned if I could figure out what, exactly was doing it though.

junegunn:
@JackDanger Can you share your `$PS1` and the functions used in it? I'll try to reproduce the problem from my machine with it.

fzf never exits with `128`, so if that was the case, it was probably caused by some code in `fzf-tmux` script. It would be really helpful if you could put `#!/bin/bash -x` on the top of the script and check what's going on when that happens.

JackDanger:
@junegunn Here's what I've got, I'll try to include as much as possible:

I've changed the key binding function to include some debug output:
```
fzf-history-widget() {
  local selected restore_no_bang_hist
  if selected=( $(fc -l 1 | $(__fzfcmd) +s --tac +m -n2..,.. --tiebreak=index --toggle-sort=ctrl-r -q "$LBUFFER") ); then
    num=$selected[1]
    if [ -n "$num" ]; then
      zle vi-fetch-history -n $num
    fi
  else
    echo last exit: $?
    echo __fzfcmd: $(__fzfcmd)
  fi
  zle redisplay
}
```

And, with `bash -x` set as the shebang on fzf-tmux, I open a new tmux pane:

```
jackdanger $ echo $PS1
%{$fg[red]%}$(pwd | xargs basename)%f%b %{$fg_bold[green]%}$(git_prompt_info)%f%b%{$fg[red]%}$%f%b
jackdanger $ which git_prompt_info
git_prompt_info () {
                ref=$(command git symbolic-ref HEAD 2> /dev/null)  || ref=$(command git rev-parse --short HEAD 2> /dev/null)  || return
                echo "$ZSH_THEME_GIT_PROMPT_PREFIX${ref#refs/heads/}$(parse_git_dirty)$ZSH_THEME_GIT_PROMPT_SUFFIX"
}
jackdanger $ which parse_git_dirty
parse_git_dirty () {
        local SUBMODULE_SYNTAX=''
        local GIT_STATUS=''
        local CLEAN_MESSAGE='nothing to commit (working directory clean)'
        if [[ "$(command git config --get oh-my-zsh.hide-status)" != "1" ]]
        then
                if [[ $POST_1_7_2_GIT -gt 0 ]]
                then
                        SUBMODULE_SYNTAX="--ignore-submodules=dirty"
                fi
                if [[ "$DISABLE_UNTRACKED_FILES_DIRTY" = "true" ]]
                then
                        GIT_STATUS=$(command git status -s ${SUBMODULE_SYNTAX} -uno 2> /dev/null | tail -n1)
                else
                        GIT_STATUS=$(command git status -s ${SUBMODULE_SYNTAX} 2> /dev/null | tail -n1)
                fi
                if [[ -n $GIT_STATUS ]]
                then
                        echo "$ZSH_THEME_GIT_PROMPT_DIRTY"
                else
                        echo "$ZSH_THEME_GIT_PROMPT_CLEAN"
                fi
        else
                echo "$ZSH_THEME_GIT_PROMPT_CLEAN"
        fi
}

```

The PS1 seems to work fine and the only real function call in the PS1 string has as it's last value an `echo` so nothing seems fishy to me.

Now I `CTRL-R` and I get a hell of a lot of output in the current tmux pane as a new fzf-tmux pane opens below:

```
jackdanger $ + args=()
+ opt=
+ skip=
+ swap=
+ close=
+ term=
+ '[' 9 -gt 0 ']'
+ arg=-d40%
+ case "$arg" in
+ '[' -n '' ']'
+ [[ -d40% =~ ^.[lrw] ]]
+ opt=
+ [[ -d40% =~ ^.u ]]
+ '[' 5 -gt 2 ']'
+ size=40%
+ [[ 40% =~ %$ ]]
+ size=40
+ '[' -n '' ']'
+ opt=' -p 40'
+ shift
+ '[' 8 -gt 0 ']'
+ arg=+s
+ case "$arg" in
+ args+=("$1")
+ shift
+ '[' 7 -gt 0 ']'
+ arg=--tac
+ case "$arg" in
+ args+=("$1")
+ shift
+ '[' 6 -gt 0 ']'
+ arg=+m
+ case "$arg" in
+ args+=("$1")
+ shift
+ '[' 5 -gt 0 ']'
+ arg=-n2..,..
+ case "$arg" in
+ args+=("$1")
+ shift
+ '[' 4 -gt 0 ']'
+ arg=--tiebreak=index
+ case "$arg" in
+ args+=("$1")
+ shift
+ '[' 3 -gt 0 ']'
+ arg=--toggle-sort=ctrl-r
+ case "$arg" in
+ args+=("$1")
+ shift
+ '[' 2 -gt 0 ']'
+ arg=-q
+ case "$arg" in
+ args+=("$1")
+ shift
+ '[' 1 -gt 0 ']'
+ arg=
+ case "$arg" in
+ args+=("$1")
+ shift
+ '[' 0 -gt 0 ']'
+ '[' -z %477 ']'
+ tmux list-panes -F '#F'
+ grep -q Z
+ set -e
+ id=31966
+ argsf=/var/folders/d2/7_pwqd3d7s9b20dh3pf14hr4000159/T//fzf-args-31966
+ fifo1=/var/folders/d2/7_pwqd3d7s9b20dh3pf14hr4000159/T//fzf-fifo1-31966
+ fifo2=/var/folders/d2/7_pwqd3d7s9b20dh3pf14hr4000159/T//fzf-fifo2-31966
+ fifo3=/var/folders/d2/7_pwqd3d7s9b20dh3pf14hr4000159/T//fzf-fifo3-31966
+ trap cleanup EXIT SIGINT SIGTERM
++ which fzf
+ fzf=/Users/jackdanger/bin/fzf
+ '[' -x /Users/jackdanger/bin/fzf ']'
+ envs='env TERM=screen-256color '
+ '[' -n '' ']'
+ '[' -n '' ']'
+ mkfifo /var/folders/d2/7_pwqd3d7s9b20dh3pf14hr4000159/T//fzf-fifo2-31966
+ mkfifo /var/folders/d2/7_pwqd3d7s9b20dh3pf14hr4000159/T//fzf-fifo3-31966
+ opts=
+ for arg in '"${args[@]}"'
+ arg=+s
+ arg=+s
+ opts=' "+s"'
+ for arg in '"${args[@]}"'
+ arg=--tac
+ arg=--tac
+ opts=' "+s" "--tac"'
+ for arg in '"${args[@]}"'
+ arg=+m
+ arg=+m
+ opts=' "+s" "--tac" "+m"'
+ for arg in '"${args[@]}"'
+ arg=-n2..,..
+ arg=-n2..,..
+ opts=' "+s" "--tac" "+m" "-n2..,.."'
+ for arg in '"${args[@]}"'
+ arg=--tiebreak=index
+ arg=--tiebreak=index
+ opts=' "+s" "--tac" "+m" "-n2..,.." "--tiebreak=index"'
+ for arg in '"${args[@]}"'
+ arg=--toggle-sort=ctrl-r
+ arg=--toggle-sort=ctrl-r
+ opts=' "+s" "--tac" "+m" "-n2..,.." "--tiebreak=index" "--toggle-sort=ctrl-r"'
+ for arg in '"${args[@]}"'
+ arg=-q
+ arg=-q
+ opts=' "+s" "--tac" "+m" "-n2..,.." "--tiebreak=index" "--toggle-sort=ctrl-r" "-q"'
+ for arg in '"${args[@]}"'
+ arg=
+ arg=
+ opts=' "+s" "--tac" "+m" "-n2..,.." "--tiebreak=index" "--toggle-sort=ctrl-r" "-q" ""'
+ '[' -n '' -o -t 0 ']'
+ mkfifo /var/folders/d2/7_pwqd3d7s9b20dh3pf14hr4000159/T//fzf-fifo1-31966
+ cat
+ tmux set-window-option -q synchronize-panes off ';' split-window -p 40 'env TERM=screen-256color  bash /var/folders/d2/7_pwqd3d7s9b20dh3pf14hr4000159/T//fzf-args-31966'
+ cat /var/folders/d2/7_pwqd3d7s9b20dh3pf14hr4000159/T//fzf-fifo2-31966
+ cat
```

I select one of the simple options in the fzf-tmux pane and hit enter and the fzf-tmux pane closes and the following is written to the main pane:

```
++ cat /var/folders/d2/7_pwqd3d7s9b20dh3pf14hr4000159/T//fzf-fifo3-31966
+ exit 0
+ cleanup
+ rm -f /var/folders/d2/7_pwqd3d7s9b20dh3pf14hr4000159/T//fzf-args-31966 /var/folders/d2/7_pwqd3d7s9b20dh3pf14hr4000159/T//fzf-fifo1-31966 /var/folders/d2/7_pwqd3d7s9b20dh3pf14hr4000159/T//fzf-fifo2-31966 /var/folders/d2/7_pwqd3d7s9b20dh3pf14hr4000159/T//fzf-fifo3-31966
jackdanger $ last exit: 128
__fzfcmd: fzf-tmux -d40%
```

Those last two lines (`last exit: 128` and `__fzfcmd: fzf-tmux -d40%`) are debug output from `fzf-history-widget`. If I remove the two echo commands from `fzf-history-widget` then I get nothing at all in my current shell line.

Interestingly, if I just run this in my terminal the exit code is fine:
```
x=( $(fc -l 1 | $(__fzfcmd) +s --tac +m -n2..,.. --tiebreak=index --toggle-sort=ctrl-r -q "$LBUFFER") )
echo $?
0
echo $x
10007 ./my-command
```

Any ideas?

JackDanger:
Oh, I really should mention that if I'm in a git repo that this behavior does _not_ occur. The exit status of  `( $(fc -l 1 | $(__fzfcmd) ... )` is 0 and my selection is written to the prompt. Weird, right?

junegunn:
@JackDanger Thanks for looking into it. It looks like that `rm -f` in `cleanup()` failed with exit status 128. Is your `rm` an alias of something? What does `type rm` say?

And can you check adding `set +e` to `cleanup()` as follows fixes the issue?

```sh
cleanup() {
  set +e
  rm -f $argsf $fifo1 $fifo2 $fifo3
}
```

JackDanger:
Sadly,

```
$ type rm
rm is /bin/rm
```

I added `-e` and got this:

```
++ cat /var/folders/d2/7_pwqd3d7s9b20dh3pf14hr4000159/T//fzf-fifo3-5677
+ exit 0
+ cleanup
+ set +e
+ rm -f /var/folders/d2/7_pwqd3d7s9b20dh3pf14hr4000159/T//fzf-args-5677
/var/folders/d2/7_pwqd3d7s9b20dh3pf14hr4000159/T//fzf-fifo1-5677
/var/folders/d2/7_pwqd3d7s9b20dh3pf14hr4000159/T//fzf-fifo2-5677
/var/folders/d2/7_pwqd3d7s9b20dh3pf14hr4000159/T//fzf-fifo3-5677
jackdanger $ last exit: 128
jackdanger $ __fzfcmd: fzf-tmux -d40%
```

No luck so far.

JackDanger:
Sadly,

```
$ type rm
rm is /bin/rm
```

I added `-e` and got this:

```
++ cat /var/folders/d2/7_pwqd3d7s9b20dh3pf14hr4000159/T//fzf-fifo3-5677
+ exit 0
+ cleanup
+ set +e
+ rm -f /var/folders/d2/7_pwqd3d7s9b20dh3pf14hr4000159/T//fzf-args-5677
/var/folders/d2/7_pwqd3d7s9b20dh3pf14hr4000159/T//fzf-fifo1-5677
/var/folders/d2/7_pwqd3d7s9b20dh3pf14hr4000159/T//fzf-fifo2-5677
/var/folders/d2/7_pwqd3d7s9b20dh3pf14hr4000159/T//fzf-fifo3-5677
jackdanger $ last exit: 128
jackdanger $ __fzfcmd: fzf-tmux -d40%
```

No luck so far.

JackDanger:
I added explicit checking of `rm`'s return value:
```
cleanup() {
  set +e
  rm -f $argsf $fifo1 $fifo2 $fifo3
  echo rm exit: $?
}
```

looked okay:

```
++ cat /var/folders/d2/7_pwqd3d7s9b20dh3pf14hr4000159/T//fzf-fifo3-27194
+ exit 0
+ cleanup
+ set +e
+ rm -f /var/folders/d2/7_pwqd3d7s9b20dh3pf14hr4000159/T//fzf-args-27194 /var/folders/d2/7_pwqd3d7s9b20dh3pf14hr4000159/T//fzf-fifo1-27194 /var/folders/d2/7_pwqd3d7s9b20dh3pf14hr4000159/T//fzf-fifo2-27194 /var/folders/d2/7_pwqd3d7s9b20dh3pf14hr4000159/T//fzf-fifo3-27194
+ echo rm exit: 0
jackdanger $ last exit: 128
jackdanger $ __fzfcmd: fzf-tmux -d40%

```

I'll keep investigating, would love some pointers for where else to look.

junegunn:
@JackDanger Thank you. So the whole log says that fzf-tmux correctly finished with `0` as expected. We'll have to find where that 128 comes from. Hmm. Can you check again with a very simple `$PS1`? Such as `PS1="> "`.

junegunn:
@JackDanger Found it. Changing `return` in `git_prompt_info` to `return 0` fixes the issue.

So for some reason, the exit status from `PS1` is affecting the return code of command substitution. Extremely strange. Not sure if it's a bug or an expected behavior, but finally I'm willing to apply the patch.

junegunn:
@JackDanger Can you rebase your commit?

JackDanger:
@junegunn Nice sleuthing! Sorry I couldn't find it myself, I got a little lost in all the moving pieces here. I've rebased and edited my comment to let the next person know why this might be necessary and what kinds of situations can cause this. Thank you!

junegunn:
Merged, thanks!

robinro:
thanks @junegunn @JackDanger for this fix! :+1:

PR #410, opened 2015-11-06
Add README instructions for the Cygwin installer

me-and:
fzf is now available via Cygwin's installer/package manager, so add
instructions for installing fzf using that route to README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,19 @@ brew install fzf
 /usr/local/opt/fzf/install
 ```
 
+#### Using Cygwin setup-\*.exe
+
+On Cygwin, you can use the [Cygwin installer](https://cygwin.com/install.html)
+to install fzf.  fzf is split into multiple packages:
+
+-   "fzf" under Utils: the core fzf executables and documentation.
+-   "fzf-bash" and "fzf-bash-completion" under Shells: the Bash key bindings
+    and completion script respectively.
+-   "fzf-zsh" and "fzf-zsh-completion" under Shells: the Z shell key bindings
+    and completion script respectively.
+-   "fzf-fish" under Shells: the fish key bindings.
+-   "fzf-vim" under Editors: the Vim plugin.
+
 #### Install as Vim plugin
 
 Once you have cloned the repository, add the following line to your .vimrc.


junegunn:
So you (or the maintainer of the package) succeeded to build Go binary for cygwin? I was trying to do it but was never been successful. Can you share how it is done?

me-and:
Me – I'm the maintainer of the package – although it's using the Ruby scripts rather than the Go binary, sadly. I'd been using fzf on Cygwin for a while, and since I'm already the Git maintainer for Cygwin, I figured I'd make it a little easier for others to use it too.

The packaging details, if you're interested, are all over at https://github.com/me-and/Cygwin-fzf.

junegunn:
Ah, that's unfortunate. To be honest, I wouldn't recommend using the Ruby version anymore since it's no longer updated and lacks too many new features that the Go version has. So I don't want to give the impression that I "officially" support Cygwin platform by mentioning it on the README page because it's not. Even if we mention Cygwin I'd like to make it clear that the new version is not available on it and instead of giving the full details, simply point to the wiki page for Cygwin.

me-and:
Okay, that makes sense. I've already mentioned the Cygwin installer on the wiki page.

As I understand it, Cygwin will still benefit from the changes made to the supporting scripts, such as the changes to the Bash completion script in 0.10.9, or are those changes wholly dependent on new features in the Go version?

me-and:
More directly related to the pull request: do you want a new PR that just says "there's a Cygwin package, but it's using the legacy Ruby code and not officially supported", or would you prefer to continue not mentioning it at all?

junegunn:
Some new features are already being used by those scripts, but I've been trying to keep the ruby version still usable by [doing this](https://github.com/junegunn/fzf/blob/master/fzf#L209-L214). I don't think it's sustainable.

I'd rather not put it on the README page at the moment. So I'll close the PR. But feel free to update the wiki page as you wish.

Note that I'm still very interested in providing Go binaries for windows and cygwin, but I don't own a Windows machine and I don't think I have enough experience or knowledge on that platform. I've been waiting for some help, but had no luck so far.

PR #423, opened 2015-11-18
zsh: fzf-completion: use noshwordsplit local option

blueyed:
This also fixes the completion causing a bell / flickering in case
"shwordsplit" was not set, because then the function would return false.


junegunn:
This is nice, thanks!

PR #425, opened 2015-11-18
usage: s/Default{s,} options/; fix/add punctuation

blueyed:
Not sure about the punctuation, but it should be consistent - so otherwise remove them from where they are already.


junegunn:
Thanks, I places periods only on options that has more than one sentences. Fixed the exceptions in ruby file and the go source.

PR #428, opened 2015-11-23
Add fzf.plugin.zsh for oh-my-zsh plugin style usage

Treri:
Maybe someone want to use fzf plugin in oh-my-zsh plugin style.
So I created a [fzf-zsh](https://github.com/Treri/fzf-zsh) plugin for my own use.

But I think if fzf itself support oh-my-zsh plugin, it would be better. So I opened the PR.

The user can clone fzf repo to oh-my-zsh/custom/plugins and execute
`oh-my-zsh/custom/plugins/fzf/install --bin`, then add fzf to zsh `plugins` list

In my zshrc, I found  fzf autocompletion is broken with zsh-autosuggestion plugin,
so I only enabled fzf keybindings.

If the user didn't use zsh-autosuggestion, he can enable fzf autocompletion manually.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,34 @@ Or you can have [vim-plug](https://github.com/junegunn/vim-plug) manage fzf
 Plug 'junegunn/fzf', { 'dir': '~/.fzf', 'do': './install --all' }
 ```
 
+#### Install as oh-my-zsh plugin
+
+Firstly clone the repository to `oh-my-zsh/custom/plugins`
+
+```sh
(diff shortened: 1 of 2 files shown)


junegunn:
Thanks for the suggestion. But I'd rather not add dependencies to the external projects. (And please note that I use neither zsh nor oh-my-zsh)

Treri:
Thanks.

Maybe I should put these configs in my fzf-zsh repo😊

junegunn:
Thanks for understanding :)

Issue #303, opened 2015-07-27
While tmux is zoomed, invoke fzf in vim will make tmux unzoomed 


trungphan:
Steps:
* Open tmux with multiple pane
* Open vim in one of the panes
* Zoom the pane (ctrl+b z)
* call fzf#run in vim => the tmux is unzoomed.

Maybe fzf should detect if tmux is in zoomed mode and use the current pane for fzf instead of creating a new one.

junegunn:
That is actually a very nice idea.

junegunn:
Patch pushed to master. I also applied the same trick to `fzf-tmux` script. Thanks.

junegunn:
Oops, I was in a haste and forgot that the result of `s:tmux_enabled()` is cached. Pushed another commit.

trungphan:
Perfect. Thanks

Issue #196, opened 2015-04-16
FZF breaking Tmux zoom


shkm:
Assuming I have a multi-pane window and vim is zoomed, executing FZF takes the window out of zoom.

Is this fixable? Perhaps via a tmux config change, if nothing else?

Thanks!

junegunn:
I don't think it's possible. You'll notice that a zoomed pane shrinks when you `tmux split`. You can try Neovim where fzf uses native vim splits instead of tmux panes.

shkm:
Yeah, thanks for responding.

I closed the issue because I realised that this just works in NeoVim :-). Unfortunately I'm having other issues with NeoVim, but I'll try to get around them because I really love FZF and the whole zoom-breaking thing is killing me right now.

DestyNova:
This seems to have been fixed recently...

junegunn:
Yeah, kind of. Now it does not split the window if it's zoomed, that's the best we can do.

DestyNova:
It's much better this way! :+1:

PR #433, opened 2015-12-06
Support fzf-tmux when zoomed

pokey:
Here's a slightly hacky approach to supporting zoomed windows:
1. Create temporary window and move the zoomed pane to the tmp window
2. Do the normal fzf-tmux thing
3. Move the zoomed pane back to the original window, kill the tmp window, and re-zoom

It's a bit fragile, for a couple reasons:
- If the user starts moving around windows while fzf-tmux is open they'll notice that something is odd
- If the user tries to select one of the hidden panes while fzf-tmux is open they'll notice that they are not there

I generally don't do much tmuxing around while fzf is open, though, so this seems good enough for my workflow.  Btw, the approach of creating a tmp window is based on this stack overflow answer:

http://superuser.com/a/357799

### DIFF
--- a/bin/fzf-tmux
+++ b/bin/fzf-tmux
@@ -82,9 +82,17 @@ while [ $# -gt 0 ]; do
   shift
 done
 
-if [ -z "$TMUX_PANE" ] || tmux list-panes -F '#F' | grep -q Z; then
+if [ -z "$TMUX_PANE" ]; then
   fzf "${args[@]}"
   exit $?
+fi
+
+# Handle zoomed tmux pane by moving it to a temp window
+if tmux list-panes -F '#F' | grep -q Z; then
+  zoomed=1
+  original_window=$(tmux display-message -p "#{window_id}")
+  tmp_window=$(tmux new-window -d -P -F "#{window_id}" "bash -c 'while :; do for c in \\| / - \\\\; do sleep 0.2; printf \"\\r\$c fzf-tmux is running\\r\"; done; done'")
+  tmux swap-pane -t $tmp_window \; select-window -t $tmp_window
 fi
 
 set -e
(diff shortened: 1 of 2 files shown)


pokey:
Also, for cross-reference this is an alternate fix to issues #196 and #303

junegunn:
Thanks, it's a clever hack and I like it.

Just a couple minor suggestions:
```diff
diff --git a/bin/fzf-tmux b/bin/fzf-tmux
index a5b51d0..db54ec7 100755
--- a/bin/fzf-tmux
+++ b/bin/fzf-tmux
@@ -87,11 +87,11 @@ if [ -z "$TMUX_PANE" ]; then
   exit $?
 fi
 
-# Handled zoomed tmux pane by moving it to a temp window
+# Handle zoomed tmux pane by moving it to a temp window
 if tmux list-panes -F '#F' | grep -q Z; then
   zoomed=1
   original_window=$(tmux display-message -p "#{window_id}")
-  tmp_window=$(tmux new-window -d -P -F "#{window_id}")
+  tmp_window=$(tmux new-window -d -P -F "#{window_id}" "bash -c 'while :; do for c in \\| / - \\\\; do sleep 0.2; printf \"\\r\$c fzf-tmux is running\\r\"; done; done'")
   tmux swap-pane -s $tmp_window \; select-window -t $tmp_window
 fi
 ```

So what do you think? If you think they're okay, please update your commit.

(I wonder why the test is failing though, I'll look into it)

pokey:
Nice I like the indicator :smile:.  Strange that it's now passing the tests...

junegunn:
Thanks, one last request: we no longer use `s:tmux_not_zoomed` function in the vim plugin. So can you remove the function and squash the commits into one?

pokey:
Yes, I had noticed that; wasn't sure if you wanted to keep it around for
potential future use.  Will remove and squash commits

junegunn:
Thanks :+1:

pokey:
Of course!  Amazing tool, so happy to be able to help in some way :smile:

PR #439, opened 2015-12-09
Correct fzf-tmux tmux checking bug

pokey:
I think this method of checking whether we're in a tmux session is more robust.  Using `$TMUX_PANE` doesn't work when I do something like `bind-key t run -b foo`, where `foo` is a script that uses `fzf-tmux`.


junegunn:
Merged, thanks!

pokey:
You're welcome!

Issue #436, opened 2015-12-08
fish: In ^r, multiline commands become multiple separate commands


tomfitzhenry:
## Steps to reproduce:

First execute a multiline command, e.g.

```
for i in *
    echo $i
end
```

Then try to retrieve that: ^r echo
## Expected:

fzf to return...

```
for i in *
    echo $i
end
```
## Actual:

fzf returns...
`echo $i`

tomfitzhenry:
Related: https://github.com/junegunn/fzf/issues/146

tomfitzhenry:
The problem is that fzf's fish plugin uses `history` (see https://github.com/junegunn/fzf/blob/a1db64e7b1588bc44dccd3b3895a075977f70ce4/shell/key-bindings.fish#L29 ) will print multiline commands as multiple separate commands, e.g.
```
$ for i in 1 2 3
    echo $i
end
1
2
3
$ history
for i in 1 2 3
echo $i
end
```

junegunn:
Thanks for the report, but unfortunately I don't know how to fix this, I don't use fish, so.. Got any idea?

tomfitzhenry:
`cat ~/.config/fish/fish_history | grep '^- cmd' | cut -d ' ' -f 3-` could be used as an alternative to `history`, but it is hacky.

This returns a string with escaped newline characters in, e.g
`for i in 1 2 3 \n echo $i \n end`
I need to find a way to un-escape those escaped newline charcaters, so it appears as:
```
for i in 1 2 3
echo $i
end
```

leiserfg:
In fish > fisn2.2 you can use `fish escape`.

leiserfg:
For un-escape just use printf.

leiserfg:
In fish the $history var holds all the the commands as an array. It could be used to solve it without hack the history file.

hauleth:
The best way to deal with that would be adding `-0` flag to Fish `history` that would split history entries using `\0` instead of newline.

amosbird:
hmm, does `fzf` support multiline items now? since fish `history -z` will produce `NUL` separated strings.

tomfitzhenry:
@amosbird sweet. https://github.com/fish-shell/fish-shell/blob/master/src/history.cpp#L1549

I'll prepare a PR.

tomfitzhenry:
Better link: https://github.com/fish-shell/fish-shell/commit/a26f68d63fb3780a1eb0c092b31b99bc67290933

[...]

PR #440, opened 2015-12-09
[WIP] Multiline Ctrl-r in fish

tomfitzhenry:
https://github.com/junegunn/fzf/issues/436

Todo:
- [ ] Don't hardcode `~/.config/fish/fish_history`

### DIFF
--- a/shell/key-bindings.fish
+++ b/shell/key-bindings.fish
@@ -26,7 +26,7 @@ function fzf_key_bindings
   end
 
   function __fzf_ctrl_r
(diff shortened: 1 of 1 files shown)


tomfitzhenry:
I'm beginning to think the poor multiline handling is a bug against fish's `history`, and that we shouldn't work around it in `fzf`. I'll raise this in `fish`.

junegunn:
Thanks for looking into this. Note that `tac` is not available on OS X and `tail -r` can be used instead.

leiserfg:
fish_history can   be on ~/.local/share/fish/fish_history   too.

tomfitzhenry:
Closing, in favour of https://github.com/junegunn/fzf/pull/954

PR #443, opened 2015-12-10
improving fish script 

leiserfg:


junegunn:
Thanks. We have to make the following prefixes work correctly with ctrl-r:
- `echo $PATH`
- `echo $(`
- `echo 'foo`
- `echo "foo`
- `echo -e "foo\nbar"`
- `echo -e "foo\rbar"`

leiserfg:
Last commit works with substitutions but not with quotes. I'm working on it.

leiserfg:
It's failing because  I remove the temporal dir on the script but not on the installer. Letter I'll write a way to use multilines completions with ctrl+r. Now it's time for me to lunch :D

junegunn:
No hurries :)

kopischke:
Heck of a lunch pause. Snark apart, how is the status on this?

PR #446, opened 2015-12-11
Fix CTRL-T in tmux.

chaoren:
- Don't assume ~/.fzf.bash exists.
- Source the current script for `__fzf_select__`.
- Forward `$PATH`.
- Dynamically select which `__fzf_select__` to use.

### DIFF
--- a/shell/key-bindings.bash
+++ b/shell/key-bindings.bash
@@ -25,7 +25,7 @@ __fzf_select_tmux__() {
   else
     height="-l $height"
   fi
-  tmux split-window $height "cd $(printf %q "$PWD"); FZF_CTRL_T_COMMAND=$(printf %q "$FZF_CTRL_T_COMMAND") bash -c 'source ~/.fzf.bash; tmux send-keys -t $TMUX_PANE \"\$(__fzf_select__)\"'"
+  tmux split-window $height "cd $(printf %q "$PWD"); PATH=$(printf %q "$PATH") FZF_CTRL_T_COMMAND=$(printf %q "$FZF_CTRL_T_COMMAND") bash -c 'source \"${BASH_SOURCE[0]}\"; tmux send-keys -t $TMUX_PANE \"\$(__fzf_select__)\"'"
 }
 
 __fzf_cd__() {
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks! I see no problem with the first commit. But the second one seems to change the behavior of the key bindings when used with tmux.

1. I see an extra space is appended before the cursor as soon as I press CTRL-T
2. The selected items are appended at the end of the command-line regardless of the cursor position.
    - Note that inserting at any position is currently [not implemented in vi-mode](https://github.com/junegunn/fzf/blob/0.11.1/shell/key-bindings.bash#L78)

chaoren:
Oh whoops. I thought those macros were the same. Is there a reason you're not using bind -x? Seems like it would make everything much easier.

chaoren:
I'll see if I can rewrite the other two with bind -x as well.

chaoren:
Ugh, sorry, that change breaks non-tmux CTRL-T.

junegunn:
No hurries :) Thanks for looking into this.

chaoren:
Okay, now the changes are strictly limited to tmux.

junegunn:
Hmm, for some reason, now bash redraws the same line when I press CTRL-T.

screen shot 2015-12-12 at 1 10 56 am

Actually this problem is not new to me. I don't remember exactly, but maybe it was the reason I decided not to use `bind -x`, but I'm not sure.

chaoren:
Weird. I'm not seeing that either with or without the tmux pane.
screen shot 2015-12-11 at 08 33 14
What version of bash are you using?

junegunn:
Oh right, it's only reproducible on bash 3.2, the system default bash on OS X (`/bin/bash`). I installed bash 4 using Homebrew and it doesn't have the problem.

I believe many OSX users are still using bash 3, and I do not want to force them to switch their default shells. How about if we keep the legacy implementation without `-x` as well and choose it depending on `$BASH_VERSION`? Maybe we can remove the part when OS X finally update its default bash version.

justinmk:
OSX may be stuck on bash 3 for a long time, because of GPL v3.

junegunn:
@justinmk Ha, didn't know that! :bulb:

junegunn:
@chaoren Works perfectly. Thanks!

PR #455, opened 2015-12-25
Pass FZF_DEFAULT_OPTS to non-interactive bash instance

frizinak:
FZF_DEFAULT_OPTS are lost when CTRL+T invokes `__fzf_select__` in a new tmux window.


junegunn:
Thanks!

Issue #456, opened 2015-12-25
Autocompletion for hidden files. [bash]


frizinak:
Assuming you have at least one dotfile in your homedir:

1: No hidden files
```
cd ~
ls **<TAB>
```

2: Leading `./` and hidden files
```
cd ~
ls .**<TAB>
```

There is a `./` safeguard ( `[ "$dir" = './' ] && dir=''` ) which obviously doesn't catch `.**<TAB>`.

To fix both issues I'd suggest adding three envs (FZF_COMP_PATH_COMMAND, FZF_COMP_DIR_COMMAND and FZF_COMP_FILE_COMMAND) to allow the same flexibility FZF_DEFAULT_COMMAND gives us.
(using ag or locate instead of find)

or

Adding these 3 options to fzf itself while also adding the above env vars but removing the need to pipe anything to it.

or

just fixing case 1 by not using `"$dir"*` and stripping the leading './'

junegunn:
Thanks for the suggestions.

I really am not considering adding more env vars. We already have too many of them, and I think we are at the point where adding more variables is not the answer to all kinds of customization needs. We should draw the line somewhere and tell the users to use their own forks of the script.

Anyway, the behavior you observed is clearly not consistent and needs to be fixed. Probably we should remove `*` suffix as you mentioned. That's what we did in 53d5d9d1624d23175eca4c1e1bd93bbbb7b41b0c. I'll look into it when I get some time.

frizinak:
Although I agree adding env vars isn't the most elegant solution, they do allow us to quickly customize fzf's input (e.g. excluding files/dirs, using locate or ag instead of find).  
Maintaining a fork and regularly merging with the upstream repo seems even less elegant.

Anyhoo created #457, which should resolve this issue.

junegunn:
It's been a while I haven't updated the scripts except for minor fixes for corner cases. If it works for you and you don't have an issue, you don't have to follow the upstream though I do understand your point of view. It's tempting to fix the current issues with seemingly innocuous small patches, however as the maintainer of the project, I should maintain a long term perspective and be cautious of any changes that might make the project bigger which make it less simpler and less flexible in the long term. Also note that the suggested options should handle a directory argument unlike the ones we already have and we'll have to explain the difference to the users. Anyway thanks for the PR!

frizinak:
works, tyvm!

PR #457, opened 2015-12-27
Make sure bash completion includes hidden files (see #456)

frizinak:
Not fixed for zsh, as it's such a trivial fix (and zsh scares me)

`ls .**` and `ls ..**` will respectively include `.` and `..`.  
but it seems pretty non-unixy to specifically exclude em.

### DIFF
--- a/shell/completion.bash
+++ b/shell/completion.bash
@@ -112,7 +112,7 @@ __fzf_generic_path_completion() {
         leftover=${leftover/#\/}
         [ "$dir" = './' ] && dir=''
         tput sc
-        matches=$(find -L "$dir"* $1 2> /dev/null | $fzf $FZF_COMPLETION_OPTS $2 -q "$leftover" | while read item; do
+        matches=$(find -L "${dir:-./}" $1 2> /dev/null | sed 's|^\./||' | $fzf $FZF_COMPLETION_OPTS $2 -q "$leftover" | while read item; do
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks, but `ls **<tab>` gives me something like this:



Also `../**<tab>` gives items with `..//` prefix which I don't think is pretty.
Let me take a stab at it later myself, I also want to add some accompanying test cases.

frizinak:
Hmm weird, I don't have the extra `/`.

```
find (GNU findutils) 4.4.2
GNU bash, version 4.3.42(1)-release (x86_64-unknown-linux-gnu)
```

`$dir` is
- empty when `ls **<T>`
- Downloads when `ls Downloads**<T>`
- Downloads/ when `ls Downloads/games**<T>`
- .. when `ls ..**`
- ../ when ls `../**`

When any of those are passed to `find -L` I get no leading or double slashes.

junegunn:
I think it's specific to OS X. I also don't see extra trailing slashes on Ubuntu. However your patch has a side effect of including the target directory itself, which leads to an empty entry when `$dir` is empty.
I'm working on a separate patch that is based on your idea and it will handle the original issues of missing hidden files without changing other behavior. The patch will also cover zsh.

junegunn:
Fixed in e1dd798. Thanks.

PR #458, opened 2015-12-28
Fix autocompletion for absolute paths

frizinak:
Should fix `ls /**` for both bash (tested) and zsh (not tested)

While testing I also noticed `ls dir**` is broken if both `dir` and `dirAbc` exist in your cwd (dirAbc[/*] not listed).

I'll have a look at it tomorrow, if not I'll create an issue.


junegunn:
Merged, thanks. `ls dir**` only showing the contents inside `dir` is the intended behavior (e.g. `ls ~USERNAME**`), but you raised an interesting point with it.

PR #486, opened 2016-02-10
Clean fish keybindings

hauleth:
It now uses `read` instead of temporary files.

### DIFF
--- a/shell/key-bindings.fish
+++ b/shell/key-bindings.fish
@@ -1,12 +1,6 @@
 # Key bindings
 # ------------
 function fzf_key_bindings
-  # Due to a bug of fish, we cannot use command substitution,
-  # so we use temporary file instead
-  if [ -z "$TMPDIR" ]
-    set -g TMPDIR /tmp
-  end
-
   function __fzf_escape
     while read item
       echo -n (echo -n "$item" | sed -E 's/([ "$~'\''([{<>})])/\\\\\\1/g')' '
@@ -19,18 +13,15 @@ function fzf_key_bindings
       -o -type f -print \
       -o -type d -print \
       -o -type l -print 2> /dev/null | sed 1d | cut -b3-"
-    eval "$FZF_CTRL_T_COMMAND | "(__fzfcmd)" -m > $TMPDIR/fzf.result"
-    and sleep 0
-    and commandline -i (cat $TMPDIR/fzf.result | __fzf_escape)
+    eval "$FZF_CTRL_T_COMMAND" | __fzfcmd -m | __fzf_escape | read -l select
+    and commandline -i "$select"
     commandline -f repaint
-    rm -f $TMPDIR/fzf.result
   end
 
   function __fzf_ctrl_r
-    history | eval (__fzfcmd) +s +m --tiebreak=index --toggle-sort=ctrl-r > $TMPDIR/fzf.result
-    and commandline (cat $TMPDIR/fzf.result)
+    history | __fzfcmd +s +m --tiebreak=index --toggle-sort=ctrl-r | read -l select
+    and commandline "$select"
     commandline -f repaint
-    rm -f $TMPDIR/fzf.result
   end
 
   function __fzf_alt_c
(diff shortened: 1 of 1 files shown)


junegunn:
Tests on Travis has been flaky lately, let's see if we can go green with a few retries.

junegunn:
There's a failing test case. Check if `ruby test/test_go.rb` on tmux passes with your patch.

```
TestFish#test_ctrl_t:
UncaughtThrowError: uncaught throw "timeout"
    test/test_go.rb:34:in `throw'
    test/test_go.rb:34:in `wait'
    test/test_go.rb:105:in `until'
    test/test_go.rb:1176:in `test_ctrl_t'
```

Questions:
- Fish version requirement? Does the new code run on slightly older versions of fish?
- Your patch removes [a recent fix](https://github.com/junegunn/fzf/commit/23244bb41087f4f833c33219bba7a7213710ab20) that is supposed to reduce the probability of intermittent errors on CTRL-T. It's a hack that I don't fully understand, but it did seem to help. Can you check if the new code doesn't suffer the problem?

bucaran:
> **@junegunn**: Does the new code run on slightly older versions of fish?

 No. The new code uses the `-a` option for read only available in fish 2.2.0. Still, you should merge as soon as tests pass since fish users are usually free to upgrade their version at will.

Source
* https://fishshell.com/release_notes.html
* https://github.com/kballard/fish-shell/commit/973dd6ffbdc189f22b634de0d684e92a9c160c9d

junegunn:
@bucaran 

> fish users are rarely locked to a particular version anyway

You're probably right. But does using temp files really affect the user experience? I know the new method is cleaner and theoretically more efficient, but if it doesn't make a noticeable difference from the user's point of view, I'm not sure if it's a good call to ignore some users who have to run older versions of fish.

bucaran:
Just my opinion as I don't have any data to backup my claim, but I am sure a solution without temp files will be faster, especially in machines that do not have flash storage.

What I meant by fish users not being locked to a particular version is that different from other shells, fish is rarely or _never_ installed by default anywhere, so users are never constrained to use an outdated version. In other words, it is very likely folks still using fish 2.1.2 (the one before 2.2.0) simply forgot or didn't realize there was a new version.

FWIW, if users with an older fish suddenly run into problems because their `fzf` breaks, assuming this PR gets merged, then we can recommend they upgrade fish, which helps the fish community move forward as `2.2.0` introduced several features which are difficult or impossible to replicate with `2.1.2`.

junegunn:
>  I am sure a solution without temp files will be faster, especially in machines that do not have flash storage.

Actually I'm not sure about that since Kernel does not flush to disk immediately.

> so users are never constrained to use an outdated version.

`apt-get` on ubuntu installs an old fish. But I know there's an easy workaround (PPA) so I guess the point is moot. How about if we print out a warning message when `key-bindings.fish` is loaded but the version requirement is not met? So that the users know what's going on.

bucaran:
> How about if we print out a warning message when key-bindings.fish is loaded but the version requirement is not met?

Can be done. Right in `key-bindings.fish` actually. 

```fish
if test "$version" != 2.2.0
   echo "fzf: Please update your fish version for better results."
end
```

junegunn:
Yeah, we should use a better operator than `!=` though. And the message should make it clear about the version requirement, e.g. `fzf: fish 2.2.0 or above is required`.

bucaran:
@junegunn We can write just a slightly more complex check, but anything more complex than this would involve forking at least `awk` once, which we want to avoid since `key_bindings` are loaded the start of the shell.

```fish
switch "$version"
    case 2.2.0
    case 2.1.\*
        echo "fzf: fish 2.2.0 or above is required"
end
```

jethrokuan:
I just found an edge case where files with `&` in their name don't get escaped properly. Perhaps the fish way is to use quotes to prevent these issues in the first place. Thoughts? @bucaran

bucaran:
@jethrokuan Can you provide a reproducible code snippet?

jethrokuan:
@bucaran : `touch "1 & 2.md"`

Ctrl-t -> select "1 & 2.bd" ->  `~/Source  1\ &\ 2.md`

The `&` is not escaped

shell/key-bindings.fish
```diff
@@ -1,12 +1,6 @@
...
 function fzf_key_bindings
-  # Due to a bug of fish, we cannot use command substitution,
-  # so we use temporary file instead
-  if [ -z "$TMPDIR" ]
-    set -g TMPDIR /tmp
-  end
-
   function __fzf_escape
     while read item
       echo -n (echo -n "$item" | sed -E 's/([ "$~'\''([{<>})])/\\\\\\1/g')' '
```
bucaran: What are you trying to do here?

jethrokuan: @bucaran this escapes the spaces in file names I think. It's not my PR :smile:

junegunn: @bucaran See #67 for more info.

bucaran: Thanks! I'll have a look.

jethrokuan: There's a single character fix:
https://github.com/jethrokuan/fish-fzf/commit/244e45b214b28a396ff5037009bf73912662de78

Ambrevar:
There is no need for using `read -a`: it is actually not what you want. A `while read` loop will do the job.
Is there any plan regarding the fish binding overhaul?
I've rewritten all of them to get rid of the TMP file and improve on the code a bit, let me know if I should push my changes.

junegunn:
Closed via #759

PR #488, opened 2016-02-11
Removed the useless `.R` macros

nhooyr:
If you do `man fzf > /dev/null`, you'll get the following output

`R' is a string (producing the registered sign), not a macro.`
`R' is a string (producing the registered sign), not a macro.`
`R' is a string (producing the registered sign), not a macro.`
`R' is a string (producing the registered sign), not a macro.`
`R' is a string (producing the registered sign), not a macro.`
`R' is a string (producing the registered sign), not a macro.`

Removing these `.R` macros has no effect on the rendering of the page but
gets rid of the error.

### DIFF
--- a/man/man1/fzf.1
+++ b/man/man1/fzf.1
@@ -71,7 +71,7 @@ e.g. \fBhistory | fzf --tac --no-sort\fR
 .BI "--tiebreak=" "CRI[,..]"
 Comma-separated list of sort criteria to apply when the scores are tied.
 .br
-.R ""
+
 .br
 .BR length "  Prefers item with shorter length"
 .br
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks, I didn't know that. However, AFAICT, the line does affect the rendering of the page. I used it to add an empty line. Maybe there's a proper way to do that?

nhooyr:
@junegunn yup, you're right. never mind. I just googled the error, found some RHEL bug report about how its useless. an empty line seems to work fine but I don't know if its the proper way.

junegunn:
Merged, thanks.

PR #489, opened 2016-02-12
Key bindings fixes

hughdavenport:
Hey,

This PR is a set of fixes to make this act more like bash does for CTRL-R mode. I start by adding variable names to all the readline commands, just to make it easier to read. Then I set some directional keybindings while in CTRL-R mode, and then also allow CTRL-g to abort unchanged. This changed the way the history replacement worked, and now uses $READLINE_LINE instead of shell expansion of $(__fzf_history__). Finally I added the CTRL-o mode, which lets you cycle through a bunch of commands. This also required overloading the up and down keys.

See commit messages for further details of each thing. Let me know if there is anything you would like me to change.

Cheers,

Hugh

### DIFF
--- a/shell/key-bindings.bash
+++ b/shell/key-bindings.bash
@@ -47,9 +47,15 @@ __fzf_cd__() {
 __fzf_history__() {
   local out key line num cmd keyseq
   shopt -u nocaseglob nocasematch
-  out=$(
-    HISTTIMEFORMAT= history |
-    $(__fzfcmd) +s --tac +m -n2..,.. --tiebreak=index --bind ctrl-r:up,ctrl-s:down --expect ctrl-g)
+  if [ -v _fzf_vi_mode ]; then
+    out=$(
+      HISTTIMEFORMAT= history |
+      $(__fzfcmd) +s --tac +m -n2..,.. --tiebreak=index --bind ctrl-r:up,ctrl-s:down --expect ctrl-g)
+  else
+    out=$(
+      HISTTIMEFORMAT= history |
+      $(__fzfcmd) +s --tac +m -n2..,.. --tiebreak=index --bind ctrl-r:up,ctrl-s:down --expect ctrl-g,ctrl-o)
+  fi
   key=$(head -1 <<< "$out")
   line=$(head -2 <<< "$out" | tail -1)
   num=$(sed 's/^ *\([0-9]*\)\** .*/!\1/' <<< "$line")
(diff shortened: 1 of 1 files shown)


hughdavenport:
I've noticed for the ctrl-o mode, ctrl-c doesn't stop it, and I can't seem to catch that binding...

hughdavenport:
Fixed the CTRL-c by adding a trap, see latest commit.

junegunn:
Thanks, I'll review the changes thoroughly when I get some time.

Some questions:
- Can we do the same for zsh and fish?
- Can we add test cases for the new behaviors?
- Looks like you're replacing `CTRL-R` binding which used to toggle sort. Even if we can't apply all the changes here to zsh and fish, this should at least be consistently applied, and It should be properly documented. 
- I do not wish to lose `toggle-sort` as I use it quite often. We should consider assigning another key for that, e.g. TAB.

hughdavenport:
Hey,

1. Yep, same for zsh and fish should be possible, though I am not as experienced in what features they have, so any patches I would do would be limited to just the variable names.

2. Sure thing, can you point me to some example test cases and I'll make some for the new stuff.

3. and 4. Yep, I would agree with consistency, which is why I replaced the CTRL-R to be consistent with what bash users would expect. I didn't realise at the time that this also broke consistency with the other shells. I would agree that another consistent keybinding should be used for that, and TAB could be the winner. It looks like it just aborts the search in normal bash, so shouldn't be getting in the way of other users habits (as it isn't documented to do so in bash/readline). Speaking of documentation, I agree, I can push another patch once we agree on what keybinding should be used for search.

Thanks for the quick response. Take your time reviewing, took me around a day to write and test so it is a bit complex!

Cheers,

Hugh

junegunn:
To run tests, you need Ruby and you have to be on tmux:

```sh
# Full test
ruby test/test_go.rb

# Run a single test
ruby test/test_go.rb --name test_ctrl_r
```

I just noticed that [`test_ctrl_r`](https://github.com/junegunn/fzf/blob/master/test/test_go.rb#L1222-L1238) currently does not cover `toggle-sort` (that's why this PR passed Travis CI). Test cases under `TestShell` is shared among bash, zsh, and fish. A bash-only test should be added under `TestBash`.

hughdavenport:
Cool, will do that this weekend hopefully :)

junegunn:
I applied your patch and noticed that it has some issues on bash 3. Unfortunately, the current OS X ships with bash 3 even to this day, and we're forced to stick with that version :(
If your patch requires bash 4, I can't merge it. Please check before putting more effort into this PR. Not that it's impossible to upgrade bash on OS X, but I really don't want to require that when installing fzf. Maybe we'll have to maintain an optional, separate repository that only provides shell extensions that are more advanced than the ones in the main repo for the time being.

hughdavenport:
Hmm ok, will see what happens on bash 3 and get that working as well.

antoniomo:
Well of those of us bash users, I think the sum of non-OS X + those on OS X who update their bash are more than those OS X bashers who stay on bash 3 (which avid terminal user does that?). Just food for though.

Alternatively this (and related "advanced bash bindings" could be a non-default `advanced-bindings.bash` that gets sourced if we have bash >= 4. By the way fzf bash keybindings have been broken on my linux box (with modern bash) since forever ;)

junegunn:
> By the way fzf bash keybindings have been broken on my linux box (with modern bash) since forever

Do you know why?

junegunn:
@antoniomo 

> By the way fzf bash keybindings have been broken on my linux box (with modern bash) since forever

Do you know why?

antoniomo:
Hi! Sorry my comment was so unhelpful. This fixed it: https://github.com/junegunn/fzf/issues/668
However I wasn't aware because I haven't upgraded in a while (precisely because I had some customization on that file), sorry. All good now!

PR #496, opened 2016-02-16
Go 1.3 compatibility

noscript:
Fixes the following error:
```
src/matcher.go:203: syntax error: unexpected range, expecting {
src/matcher.go:207: non-declaration statement outside function body
src/matcher.go:208: syntax error: unexpected }
```

While not all distributions ship Go 1.4 (Debian 8, Ubuntu 15.04) yet, this negligible fix makes it possible to build against 1.3.


junegunn:
I won't be testing the latest source against older versions of Go, so there's no guarantee of backward compatibility in the future. But since this looks harmless and simple, I'll merge it for now. Thanks.

PR #497, opened 2016-02-18
Build on OpenBSD (tested on amd64 and i386)

anaseto:
Added a build target for OpenBSD, and an additional test to avoid make error because `uname -m` returns `amd64` and not `x86_64` under OpenBSD (and build works on `i386` too).

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -12,7 +12,9 @@ endif
 endif
 
 ifneq ($(shell uname -m),x86_64)
-$(error "Build on $(UNAME_M) is not supported, yet.")
+ifneq ($(shell uname -s),OpenBSD)
+$(error "Build on $(shell uname -m) is not supported, yet.")
+endif
 endif
 
 SOURCES   := $(wildcard *.go */*.go)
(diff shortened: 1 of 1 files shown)


anaseto:
There is a failing test on OpenBSD, though. It seems to happen because of a difference of behavior in OpenBSD when accessing files with bad permissions, but I'm not sure about it.

     --- FAIL: TestHistory (0.04s)
         history_test.go:19: Error expected for: /etc
         history_test.go:19: Error expected for: /etc/sudoers

junegunn:
Thanks. Shouldn't we be changing `GOOS` on OpenBSD instead of adding a new target?

anaseto:
Cross-compilation didn't work for me with go on OpenBSD failing with a `go build runtime: openbsd/386 must be bootstrapped using make.bash`. I don't know much about `go`, but it seems that `OpenBSD` version doesn't come with cross-compilation capabilities out of the box (so I needed to compile directly for a i386 machine). And the `build` target depends on `test`, but this was blocking because of one failing test on OpenBSD.

junegunn:
I would like to properly fix the issues and release the binaries instead of just adding a target. I'll take a look when I get some time. For now, if you have Go installed on your system, you can simply do `go get github.com/junegunn/fzf/src/fzf`.

anaseto:
Fine, thanks. About cross-compiling issues, from what I have read, it
seems due to the fact that I have only go 1.4.2, and some work to get
cross-compilation working seems necesary. I will retry this when I
update to go 1.5 which improves on that.

-- 
Yon

yunake:
@anaseto do you plan on submitting this as a port?

anaseto:
Oh, I haven't really thought about that, I'm not very knowledgeable
about ports and just running -release, but if there's anything I can do
to help, I'll be glad to do it.

And you are certainly right that this would be a better fit for a port
patch. Though, it there is some special support for go code in the ports
makefiles, the patch would be perhaps unnecessary. And as junegunn
pointed out, it can be installed with "go get", and as it does not have
any dependencies (outside the standard library), a "go install" just
after download in the fzf/src/fzf directory would do the job too and
simplify the process.

yunake:
I don't think there's any special support for Go in ports — at least there
wasn't last time I checked. `go install` is a perfectly valid installation
method for me, I'm not a ports purist, thank you!

Issue #237, opened 2015-05-16
ctrl-r matches on history line number too


alexeymuranov:
Is there a reason someone would want to match the line number from history when using **fzf** with `ctrl-r`? Currently, if i use **fzf** with `ctrl-r` (in **zsh**) and type, for example, `123`, the returned list includes the item number 123 from the history, and may also include the item 12, or 192, containing `3`, etc. I wouldn't need them in the list, personally.

junegunn:
https://github.com/junegunn/fzf/blob/master/shell/key-bindings.zsh#L39

Changing `-n2..,..` to `-n2..` will do. There is no variable to customize the behavior so you'll have to edit the source directly or manage your own version. The reason it is so is that it's not practical to make every aspect of the code configurable so I want the users to see the code as a reference implementation that they can refer to when they want to write their own key bindings. Thanks.

alexeymuranov:
Thanks for the explanation.  However, IMO, a sensible default would be not to match the history line number.

junegunn:
Okay, but I actually type in the index number to pinpoint the item when there are many very similar entries in the list. I find it especially helpful when in `--extended` search mode, where I can first type in a part of the command, then append a number to greatly narrow down the list. e.g. `git 55`

PR #501, opened 2016-02-19
Suppress zsh history line numbers

noscript:
Before:
before

After:
after

### DIFF
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -37,13 +37,10 @@ bindkey '\ec' fzf-cd-widget
 
 # CTRL-R - Paste the selected command from history into the command line
 fzf-history-widget() {
-  local selected num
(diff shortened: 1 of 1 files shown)


noscript:
Apparently Travis is experiencing some issues with file system write permissions.
```
TestGoFZF#test_execute_shell:
Errno::ENOENT: No such file or directory @ rb_sysopen - /tmp/output-test_execute_shell-0.out
```

I tried to revert the commit and it fails as well. Would you please try to trigger Travis manually?

junegunn:
Thanks, but I actually like the numbers. See #237. It's a preference thing, so I can understand that you feel otherwise, but there's little reason to suddenly change it after two years. You'll have to override the function in your own configuration.

junegunn:
Also note that your version does not handle multi-line commands in history.

PR #506, opened 2016-02-23
s:callback: Always return list.

justinmk:
Fixes "E706: Variable type mismatch for: ret" when an exception is caught.

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -239,7 +239,7 @@ function! s:exit_handler(code, command, ...)
   return 1
 endfunction
 
-function! s:execute(dict, command, temps)
+function! s:execute(dict, command, temps) abort
   call s:pushd(a:dict)
   silent! !clear 2> /dev/null
   let escaped = escape(substitute(a:command, '\n', '\\n', 'g'), '%#')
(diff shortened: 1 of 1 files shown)


plugin/fzf.vim
```diff
@@ -239,7 +239,7 @@ function! s:exit_handler(code, command, ...)
   return 1
 endfunction
 
-function! s:execute(dict, command, temps)
+function! s:execute(dict, command, temps) abort
```
justinmk: I added `abort` to some functions. Let me know if you prefer not to.

plugin/fzf.vim
```diff
@@ -392,12 +391,12 @@ try
...
     silent! call delete(tf)
   endfor
-
-  return lines
 catch
   if stridx(v:exception, ':E325:') < 0
     echoerr v:exception
   endif
+finally
+  return lines
```
justinmk: When exception is caught, empty string is returned, which causes problem [here](https://github.com/junegunn/fzf/pull/506/files#diff-24d9d57e35df5afce7890f2400917919R349) because `ret` is initialized to a list.

justinmk:
CI error seems unrelated:

```
  1) Error:
TestGoFZF#test_execute_shell:
Errno::ENOENT: No such file or directory @ rb_sysopen - /tmp/output-test_execute_shell-0.out
    test/test_go.rb:907:in `readlines'
    test/test_go.rb:907:in `test_execute_shell'
```

junegunn:
Thanks. Some tests are flaky on Travis Ci.

PR #529, opened 2016-03-28
Fix typo in README.md

mjwestcott:
Based on previous pull requests, it seems that you appreciate these sort of corrections.

"The following" is treated as either singular or plural. See [here](http://www.oxforddictionaries.com/definition/english/following), section 'noun', number 2.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ Pros
 Installation
 ------------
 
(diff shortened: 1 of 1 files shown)


junegunn:
Merged, thank you!

PR #531, opened 2016-03-29
don't install zsh files if zsh doesn't exist

adam8157:
as $subject. keep bash files since it's almost preloaded everywhere.

### DIFF
--- a/install
+++ b/install
@@ -267,6 +267,10 @@ fi
 
 echo
 for shell in bash zsh; do
+  if [ $shell = "zsh" ] && [ ! -n "$(which zsh &>/dev/null)" ]; then
+    break;
+  fi
+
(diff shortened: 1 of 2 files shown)


netei:
it is recommended not to use which. You should rather use `command -v`

Like so : 

```
if [ $shell = "zsh" ] && ! command -v zsh >/dev/null; then
```

junegunn:
Thanks for the suggestion, I'll update the code myself to keep the style consistent across the file.

@netei Didn't know that. Thanks for the info.

junegunn:
Fixed it. Let me know if you see any issues, thanks.

adam8157:
Cool. Better to fix uninstall script too, thanks.

PR #540, opened 2016-04-12
Fix Bash+vimode pre-launch delay

WChargin:
Summary:
Fix from: https://adamheins.com/blog/fzf-vi-mode-and-fixing-delays
(Thanks, @adamheins!)
The basic problem is that
fzf presses <Esc> to enter vi-movement-mode
(as opposed to insert mode)
and then presses a bunch of keys to set up the buffer.
But the <Esc> keypress is also the prefix for a bunch of other commands,
so Bash will dutifully wait an excruciating half-second
before actually executing this command.
Instead, we bind <C-a>, which is unused by default,
to be another way to enter vi-movement-mode;
this binding is unambiguous, so fzf can use it without delay.

This change was made by just `:s/\\e/\\C-a/gc` in the relevant section,
after adding the actual binding and comment at the top.

### DIFF
--- a/shell/key-bindings.bash
+++ b/shell/key-bindings.bash
@@ -115,7 +115,7 @@ else
 
   # ALT-C - cd into the selected directory
   bind '"\ec": "\C-x\C-addi$(__fzf_cd__)\C-x\C-e\C-x\C-r\C-m"'
-  bind -m vi-command '"\C-ac": "i\C-ac"'
+  bind -m vi-command '"\ec": "ddi$(__fzf_cd__)\C-x\C-e\C-x\C-r\C-m"'
 fi
 
 unset -v __use_tmux __use_tmux_auto
--- a/shell/key-bindings.bash
+++ b/shell/key-bindings.bash
@@ -85,6 +85,15 @@ if [ -z "$(set -o | \grep '^vi.*on')" ]; then
   # ALT-C - cd into the selected directory
   bind '"\ec": " \C-e\C-u$(__fzf_cd__)\e\C-e\er\C-m"'
 else
+  # We'd usually use "\e" to enter vi-movement-mode so we can do our magic,
+  # but this incurs a very noticeable delay of a half second or so,
+  # because many other commands start with "\e".
+  # Instead, we bind an unused key, "\C-x\C-a",
+  # to also enter vi-movement-mode,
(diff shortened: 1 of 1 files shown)


junegunn:
Interesting, thank you. Looks like it only happens with bash 4+. How about if we choose a more obscure key that is less likely to be used by the users, such as `\C-x\C-a` or something? I'm saying this because there can be users who mapped `\C-a` to something else.

And `bind '"\C-ac":` seems like a mistake, right?

WChargin:
Yep, good call on both. (I didn't know `<Esc>c` would `cd`—nice!)

WChargin:
Actually, I'm not sure what the line `bind -m vi-command '"\ec": "i\ec"'` is supposed to do. Is it supposed to let you use `<Esc>c` when you're not in insert mode? This doesn't work for me even before this patch (on `origin/master` 935e986be5fe3d6c9ca6fa578d836500a3b31525).

junegunn:
> Is it supposed to let you use `<Esc>c` when you're not in insert mode? 

Yes.

> This doesn't work for me even before this patch

Interesting. CTRL-R and CTRL-T work as expected but not ALT-C. Do you have any clue?

WChargin:
Nope, I don't know; I've never touched `bind` before. I do note that inlining it, as follows, works:
```sh
  bind '"\ec": "\C-x\C-addi$(__fzf_cd__)\C-x\C-e\C-x\C-r\C-m"'
  bind -m vi-command '"\ec": "ddi$(__fzf_cd__)\C-x\C-e\C-x\C-r\C-m"'
```
(Note that I removed the extra `\C-x\C-a` at the beginning of the mapping.)

So my guess was that it was a non-recursive mapping, but that doesn't appear to be the case: I ran
```sh
bind '"\C-j": "abcde"'
bind -m vi-command '"\C-i": "i\C-j"'
```
and then pressing `<C-i>` in normal mode did cause the characters to be inserted.

If you're comfortable with the patch in the first snippet, I'll push it.

WChargin:
(Could you reproduce, by the way, on master?)

junegunn:
Yeah, that works :)

WChargin:
Okay, I went ahead and squashed down the commits to two (one for each fix).

junegunn:
Merged, thanks!

PR #549, opened 2016-04-23
Update completion.bash

### LINKED ISSUE #548: Loading completion.bash is so slow on Cygwin
- Category
    - [ ] fzf binary
    - [ ] fzf-tmux script
    - [ ] Key bindings
    - [X] Completion
    - [ ] Vim
    - [ ] Neovim
    - [ ] Etc.
- OS
    - [ ] Linux
    - [ ] Mac OS X
    - [X] Windows (Cygwin)
    - [ ] Etc.
- Shell
    - [X] bash
    - [ ] zsh
    - [ ] fish

On my Cygwin system, enabling the completion file (which adds `**` handling) really slows down bash startup. The completion script takes 0.5 s to execute (on Core i7!). I've done some digging and traced the problem down to one line - https://github.com/junegunn/fzf/blob/master/shell/completion.bash#L282
Specifically, in `eval "$(printf "$orig" "$func")"` the `$()` command substitution creates a subshell (I guess?), I've replaced this call with `time echo "$(printf "$orig" "$func")"` and each call to printf takes 0.012 ms, since there are many commands in a loop that call this function, total time is over 0.4 s on my laptop. I guess this is due to `fork` on Cygwin being so slow - on Linux `completion.bash` executes very quickly.

By the way, while I was inserting debugging `echo` into completion.bash, I've noticed the entire file is sourced every time a completion is invoked for a command it was not invoked before (again, producing a significant slowdown on Cygwin). E.g., I put `echo Adding FZF completions` just before the _fzf_defc invokations. When I run bash, I can see that message as bash_completion initializes. When I type `scp ` and press Tab twice, I can see that message appear again! How does it work? I couldn't find the code which sources the script itself again.

gene-pavlovsky:
Fixes #548. Avoid using a subshell in _fzf_defc().
Please also take a look and answer my comment in the last paragraph of #548

### DIFF
--- a/shell/completion.bash
+++ b/shell/completion.bash
@@ -272,14 +272,15 @@ if type _completion_loader > /dev/null 2>&1; then
 fi
 
 _fzf_defc() {
-  local cmd func opts orig_var orig
+  local cmd func opts orig_var orig def
   cmd="$1"
   func="$2"
   opts="$3"
   orig_var="_fzf_orig_completion_$cmd"
   orig="${!orig_var}"
   if [ -n "$orig" ]; then
-    eval "$(printf "$orig" "$func")"
+    printf -v def "$orig" "$func"
+    eval "$def"
   else
     complete -F "$func" $opts "$cmd"
   fi


junegunn:
Good one.

gene-pavlovsky:
Thanks

PR #550, opened 2016-04-23
Update key-bindings.bash

gene-pavlovsky:
Faster startup. Use internal bash globbing instead of external grep binary (adapted from Gentoo's `/etc/bash/bashrc` TERM checking). Insignificant on Linux, but on Cygwin this cuts startup time by 40 ms on my Core i7 laptop.


junegunn:
Thanks.

gene-pavlovsky:
Thank you! I can't believe even one fork takes so long on Cygwin...

PR #554, opened 2016-04-24
Fix missing reference to UNAME_M

gene-pavlovsky:
The `Build on $(UNAME_M) is not supported, yet` message was referencing an undefined UNAME_M. Fixed that.

Still I'm thinking is there sense to keep this warning? Apparently it builds just fine on my i686 system. So why should the Makefile be disabled forcefully?


junegunn:
Thanks, I answered to your questions in #555.

PR #559, opened 2016-04-26
[vim] use a window-local variable to find the previous window

blueyed:
When using a tab for `fzf_layout`, I've ended up in the wrong window,
which was caused by the `wincmd p` being executed since the previous and
final tab page numbers where the same.

This patch fixes it by using a window-local variable to mark the
originating window.

> let g:fzf_layout = { 'window': 'execute (tabpagenr()-1)."tabnew"' }


blueyed:
The build failure seems to be unrelated?!

junegunn:
Yeah, don't worry about it. Some test cases are flaky on travis CI.

Can we just do this? Maybe I wasn't thinking straight.

```patch
index 526f245..3c4525b 100644
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -343,9 +343,7 @@ function! s:execute_term(dict, command, temps) abort
         " there's no other listed buffer (nvim +'set nobuflisted')
         close
       endif
-      if pos.tab == s:ppos.tab
-        wincmd p
-      endif
+      execute s:ppos.win.'wincmd w'
     endif
 
     if !s:exit_handler(a:code, s:command, 1)
```

junegunn:
Actually, I noticed a couple more issues with `enew` as `fzf_layout` (`let g:fzf_layout = { 'window': 'enew' }`).

1. CTRL-C leaves `[Process exited 130]` message on screen`
2. Enter on empty result (some bogus query) gives `Can't re-enter normal mode from terminal mode`

I'm a bit surprised by the second one because I didn't experience it in the past. We have to use `:b #` instead of `execute "normal! \<c-^>"` though they are [not completely equivalent](https://github.com/junegunn/fzf/commit/355d004895eb8cb7e6ca7f8cf0b095f381d6e41b#commitcomment-11794616).

```diff
diff --git a/plugin/fzf.vim b/plugin/fzf.vim
index 526f245..8d366bf 100644
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -298,7 +298,7 @@ function! s:split(dict)
   \ 'down':  ['botright', 'resize', &lines],
   \ 'left':  ['vertical topleft', 'vertical resize', &columns],
   \ 'right': ['vertical botright', 'vertical resize', &columns] }
-  let s:ppos = s:getpos()
+  let ppos = s:getpos()
   try
     for [dir, triple] in items(directions)
       let val = get(a:dict, dir, '')
@@ -311,7 +311,7 @@ function! s:split(dict)
         endif
         execute cmd sz.'new'
         execute resz sz
-        return {}
+        return [ppos, {}]
       endif
     endfor
     if s:present(a:dict, 'window')
@@ -319,20 +319,30 @@ function! s:split(dict)
     else
       execute (tabpagenr()-1).'tabnew'
     endif
-    return { '&l:wfw': &l:wfw, '&l:wfh': &l:wfh }
+    return [ppos, { '&l:wfw': &l:wfw, '&l:wfh': &l:wfh }]
   finally
     setlocal winfixwidth winfixheight
   endtry
 endfunction
 
 function! s:execute_term(dict, command, temps) abort
-  let winopts = s:split(a:dict)
-
-  let fzf = { 'buf': bufnr('%'), 'dict': a:dict, 'temps': a:temps, 'name': 'FZF', 'winopts': winopts }
-  let s:command = a:command
+  let [ppos, winopts] = s:split(a:dict)
+  let fzf = { 'buf': bufnr('%'), 'ppos': ppos, 'dict': a:dict, 'temps': a:temps,
+            \ 'name': 'FZF', 'winopts': winopts, 'command': a:command }
+  function! fzf.switch_back(inplace)
+    if a:inplace && bufnr('') == self.buf
+      " FIXME: Can't re-enter normal mode from terminal mode
+      " execute "normal! \<c-^>"
+      b #
+      " No other listed buffer
+      if bufnr('') == self.buf
+        enew
+      endif
+    endif
+  endfunction
   function! fzf.on_exit(id, code)
     let pos = s:getpos()
-    let inplace = pos == s:ppos " {'window': 'enew'}
+    let inplace = pos == self.ppos " {'window': 'enew'}
     if inplace
       for [opt, val] in items(self.winopts)
         execute 'let' opt '=' val
@@ -343,12 +353,12 @@ function! s:execute_term(dict, command, temps) abort
         " there's no other listed buffer (nvim +'set nobuflisted')
         close
       endif
-      if pos.tab == s:ppos.tab
-        wincmd p
-      endif
+      execute 'tabnext' self.ppos.tab
+      execute self.ppos.win.'wincmd w'
     endif
 
-    if !s:exit_handler(a:code, s:command, 1)
+    if !s:exit_handler(a:code, self.command, 1)
+      call self.switch_back(inplace)
       return
     endif
 
@@ -356,14 +366,7 @@ function! s:execute_term(dict, command, temps) abort
     let ret = []
     try
       let ret = s:callback(self.dict, self.temps)
-
-      if inplace && bufnr('') == self.buf
-        execute "normal! \<c-^>"
-        " No other listed buffer
-        if bufnr('') == self.buf
-          bd!
-        endif
-      endif
+      call self.switch_back(inplace)
     finally
       call s:popd(self.dict, ret)
     endtry
```

junegunn:
Geez, I found another issue with `enew`, pressing CTRL-X, CTRL-V, CTRL-T does not clean up the terminal buffer. Sorry, but I think I'm going to apply my patch first to address these issues. Let me know if you still have the problem and think it needs more improvements.

justinmk:
> Enter on empty result (some bogus query) gives Can't re-enter normal mode from terminal mode

This was a cheesy fix to avoid worse problems (segfaults, loops). Hopefully we can find a more elegant approach in the future.

junegunn:
@justinmk Ah okay, noted. Thanks.

blueyed:
@junegunn 
The problem with (re)storing tab/window numbers is that they are not bound to the actual tab/window.

Feel free to apply your patch first though.

junegunn:
@blueyed Yes, the assumption here is that it's unlikely that a user opens more than one fzf buffer at a time, which I believe is a reasonable one and it simplifies the implementation.

justinmk:
Would https://github.com/vim/vim/commit/86edef664efccbfe685906c854b9cdd04e56f2d5 help solve this more elegantly/robustly? (That patch is not merged into neovim yet)

junegunn:
Yeah, probably, I don't want to tell the users to install the bleeding edge Vim.

kopischke:
> I don't want to tell the users to install the bleeding edge Vim.

No need: I wrote [a `win_id` function family shim](https://github.com/kopischke/vim-stay/blob/release/1.4.0/autoload/stay/win.vim) for the upcoming _vim-stay_ release that should work with all versions of Vim 7. Might be overkill for this usage case, but I felt I would be remiss in not mentioning it.

blueyed:
@kopischke 
Very interesting.  I think it would be useful for other plugins as well, and could therefore become a separate lib?!

kopischke:
@blueyed that is the idea once I get my plugin updates out of the door.

Issue #560, opened 2016-05-01
Env vars docs incomplete


pmazurek:
Documentation according environment variables is incomplete, both in readme and in --help. Took me a while to try and come up with my own way of cd'ing to a folder using fzf, and only when I got everything ready and wanted to customize the keybindings, I realized that you, in fact, can override the alt+c default command, because the variable is buried there, somewhere in the code.

Lets update it. Will save someone's time in future.

PR #561, opened 2016-05-01
#560 Updating environment vars documentation

pmazurek:
#560


junegunn:
Thanks. However, the key bindings and the fuzzy completion are optional and one can use fzf binary alone without installing them. The binary only knows FZF_DEFAULT_OPTS and FZF_DEFAULT_COMMAND so we should not add the other ones to `--help` output.

The README page is also organized as such; fzf binary, key bindings, and fuzzy completion. And each section describes a subset of the variables that are relevant to the topic.

I noticed that `FZF_ALT_C_COMMAND` is missing on the page. I'll add it to the document.

Issue #556, opened 2016-04-25
FZF in Neovim Laggy in Some Cases


sbdchd:
- Category
   - [x] Neovim
- OS
    - [x] Mac OS X
- Shell
    - [x] bash

```bash
export FZF_DEFAULT_COMMAND='ag --hidden -U --ignore .git -g ""'
```

In Neovim, when I run `:FZF` in a directory with a git repo, with the default command as shown above, the terminal buffer freezes until I switch to a different tmux window and switch back.

The issue also appears to be present without Neovim being in tmux.

After some testing, it seems that 095f31b is the cause of the issue.

junegunn:
I don't have the problem. Please check if the problem is reproducible with the following configuration

```vim
nvim -u <(cat << "EOF"
  call plug#begin('~/.vim/plugged')
  Plug 'junegunn/fzf', { 'dir': '~/.fzf', 'do': 'yes \| ./install' }
  call plug#end()
  let $FZF_DEFAULT_COMMAND = 'ag --hidden -U --ignore .git -g ""'
EOF)
```

sbdchd:
I am not sure why, but after removing vim-airline, the problem seems to be resolved.

Thanks for the help.

junegunn:
Related: https://github.com/junegunn/fzf.vim/issues/84

I believe it's a bug of neovim.

justinmk:
See https://github.com/junegunn/fzf.vim/issues/84#issuecomment-182404906. This bug probably is in Vim, but it doesn't happen there because tmux is used instead.

junegunn:
@justinmk Thanks for chiming in. But no, it's not reproducible on original Vim. The bang versions of the commands such as `:FZF!` do not use tmux and they are still fine.

You can use this config to reproduce the problem:

```vim
set nocompatible
set laststatus=2

if !filereadable('/tmp/plug.vim')
  silent !curl -fLo /tmp/plug.vim
    \ https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim
endif

autocmd VimEnter *
  \  if !empty(filter(copy(g:plugs), '!isdirectory(v:val.dir)'))
  \|   PlugInstall
  \| endif

source /tmp/plug.vim
call plug#begin('/tmp/plugged')
Plug 'junegunn/fzf', { 'do': './install --bin' }
Plug 'bling/vim-airline'
Plug 'tpope/vim-fugitive'
call plug#end()
```

justinmk:
[...]

Issue #520, opened 2016-03-03
Restore window layout when using Neovim's terminal mode


blueyed:
I've noticed that using FZF commands in NeoVim will mess with the window layout after being closed, especially if there's a location list / quickfix window involved (at the bottom).

Given a window layout of:

    +-------------------------------------------+
    |                       |                   |
    |                       |                   |
    |                       |                   |
    |                       |                   |
    +------------------------                   +
    |                       |                   |
    |                       |                   |
    |                       |                   |
    |                       |                   |
    |-----------------------|                   |
    |                       |                   |
    | location list         |                   |
    +-------------------------------------------+

After opening FZF:

    +-------------------------------------------+
    |                       |                   |
    |                       |                   |
    |                       |                   |
    +------------------------                   +
    |                       |                   |
    |                       |                   |
    |-----------------------|                   |
    | location list         |                   |
    +-------------------------------------------+
    |                                           |
    | FZF                                       |
    |                                           |
    +-------------------------------------------+

After closing FZF:

    +-------------------------------------------+
    |                       |                   |
    |                       |                   |
    |                       |                   |
    +------------------------                   +
    |                       |                   |
    |                       |                   |
    |-----------------------|                   |
    | location list         |                   |
    |                       |                   |
    |                       |                   |
    |                       |                   |
    |                       |                   |
    +-------------------------------------------+

I've tried this with just `nvim -u vimrc.minimal`.

It gets worse on a second invocation, probably because of the `winfixheight` setting of quickfix/location list windows.

It would be nice, if fzf would restore the window layout from before it was invoked.

netei:
Very similar to https://github.com/junegunn/fzf.vim/issues/42

justinmk:
There is an issue I'm Neovim repo for augmenting window API to support this. Though I guess mkview could be abused somehow...

junegunn:
Yeah, this is an issue with vim plugins that create temporary splits.

- `botright 10new`
- `botright new`
- `close`

Unfortunately I can't think of a simple solution to this problem. Suggestions are welcome.

justinmk:
Is there a reason `(tabpagenr()-1).'tabedit|term ...` is not the default? That would preserve layout. It also wasn't obvious how to switch to this mode using `g:fzf_layout`.

IMO https://github.com/neovim/neovim/issues/3933 is the general solution to this problem.

blueyed:
@justinmk 
[...]

PR #581, opened 2016-05-28
Prefer tmux for launching in Vim

jez:
There is currently a [bug in Neovim][1] where opening a terminal buffer
by exec'ing a command that has a pipe in it causes the buffer to be
blank until `<ESC>` is pressed. While not strictly an issue with fzf, it
affects users like me who use a custom `FZF_DEFAULT_COMMAND`.

This change makes fzf prefer tmux over Neovim for launching a fzf split.
It's reasonable to assume that if a Neovim user under tmux either
- Doesn't care about the type of split, or
- Prefers tmux panes to nvim embedded terminals, as panes the main
  purpose of tmux.

Thus, this change works around the Neovim issue in a non-temporary,
non-hacky way; looks like a win-win.

[1]: https://github.com/neovim/neovim/issues/4487

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -140,14 +140,17 @@ try
   else
     let prefix = ''
   endif
-  let tmux = !has('nvim') && s:tmux_enabled() && s:splittable(dict)
+  let tmux = s:tmux_enabled() && s:splittable(dict)
   let command = prefix.(tmux ? s:fzf_tmux(dict) : fzf_exec).' '.optstr.' > '.temps.result
 
-  if has('nvim')
+  if tmux
+    let ret = s:execute_tmux(dict, command, temps)
+  elseif has('nvim')
     return s:execute_term(dict, command, temps)
+  else
+    let ret = s:execute(dict, command, temps)
   endif
 
-  let ret = tmux ? s:execute_tmux(dict, command, temps) : s:execute(dict, command, temps)
   call s:popd(dict, ret)
   return ret
 finally


junegunn:
I also have `FZF_DEFAULT_COMMAND` but I don't have any issues on nvim 0.1.5. Can you check if you can still reproduce the problem with [this minimal Vim config](https://gist.githubusercontent.com/junegunn/6936bf79fedd3a079aeb1dd2f3c81ef5/raw/vimrc)?

And also, last time I checked, using tmux pane for fzf in neovim had issues with upper pane not displaying correctly.

justinmk:
I must have misunderstood https://github.com/neovim/neovim/issues/4487. If it is still an issue as described, would like to know.

jez:
Hm. Well, I didn't realize that you had to explicitly pass `--HEAD` to brew to get the head version of Neovim (it used to be implicit). I upgraded from 0.1.4 to 0.1.5-dev with my personal config and it behaves no differently.

I then tried it with the minimal vimrc (for posterity's sake, I had forgotten at first that for Neovim it's ~/.config/nvim/init.vim, not ~/.vimrc).

On neovim #4487, they linked to Shougo/unite.vim#1121, where one of the suggestions was to disable vim-airline. Doing so fixed the issue.

Any ideas why this would be the case? I'm going to poke around a bit, but I can't see how vim-airline comes into play here o.o

justinmk:
So this is the same case as https://github.com/junegunn/fzf.vim/issues/84 / https://github.com/junegunn/fzf/issues/556#issuecomment-214444885 .

Correct me if I'm wrong, but pipes don't seem to be a factor. @junegunn narrowed it down to this system() call in airline, where I don't see a pipe: https://github.com/vim-airline/vim-airline/blob/985f3d3ecada29a037cac87299c61a0a5f71fe31/autoload/airline/extensions/branch.vim#L76

jez:
Since it seems like this isn't the preferred solution, and because the conversation has been moved to the Neovim board, I'm going to close this.

For people finding this from web search results, see the Neovim issue linked above for workarounds.

junegunn:
There can be a few reasons one might want to use tmux instead of neovim terminal (not messing up vim splits (#520) is one) and we can consider providing an option to prefer tmux, but making it the default feels a bit radical. Neovim terminal allows us to use fzf in GUI nvim without an external terminal emulator window (once GUI nvim is stable, I haven't tried it) and one might prefer its "more integrated" look-and-feel.

jez:
Yep, I agree. Though, if they were using a GUI nvim, they probably wouldn't
be using tmux. One point you brought up earlier is that tmux scrolls the
page up, but a :term split doesn't.

edi9999:
> we can consider providing an option to prefer tmux,

I would use such an option :-)

jez:
> > we can consider providing an option to prefer tmux,
> I would use such an option :-)

I'd encourage you to implement it! You can look at this PR to basically see which code you'd need to touch :P Shouldn't be more than a handful of lines.

edi9999:
Done : https://github.com/junegunn/fzf/pull/591

PR #582, opened 2016-05-31
Change tmux pane title for fzf splits

akashin:
Hello!
I'm having the following problem while using fzf with custom tmux keybindings required to work with vim-tmux-navigator:
I use C-h/j/k/l shortcut to navigate between vim and tmux splits, but the same shortcuts are used to iterate over lists in fzf split. vim-tmux-navigator basically solves the problem of shortcuts clash between tmux and vim using *if-shell* statement inside tmux config that detects the current command of the pane and sends appropriate commands to current pane.
The same solution would apply to fzf if only there were a way to detect that current split is running fzf. Currently it's impossible because there is no identity of fzf in the tmux pane - the active command is bash.
This commit changes the tmux pane title to "fzf" which allows to differentiate it in tmux if-shell command.

### DIFF
--- a/bin/fzf-tmux
+++ b/bin/fzf-tmux
@@ -144,14 +144,14 @@ if [ -n "$term" -o -t 0 ]; then
   cat <<< "$fzf $opts > $fifo2; echo \$? > $fifo3 $close" > $argsf
   tmux set-window-option synchronize-panes off \;\
     set-window-option remain-on-exit off \;\
-    split-window $opt "cd $(printf %q "$PWD");$envs bash $argsf" $swap \
+    split-window $opt "printf '\033]2;fzf\033\\'; cd $(printf %q "$PWD");$envs bash $argsf" $swap \
     > /dev/null 2>&1
 else
   mkfifo $fifo1
   cat <<< "$fzf $opts < $fifo1 > $fifo2; echo \$? > $fifo3 $close" > $argsf
   tmux set-window-option synchronize-panes off \;\
     set-window-option remain-on-exit off \;\
-    split-window $opt "$envs bash $argsf" $swap \
+    split-window $opt "printf '\033]2;fzf\033\\'; $envs bash $argsf" $swap \
     > /dev/null 2>&1
   cat <&0 > $fifo1 &
 fi


junegunn:
Since you use C-h/j/k/l to navigate between splits, don't you want to use different set of keys for moving cursor up and down in fzf? <kbd>CTRL-N</kbd> and <kbd>CTRL-P</kbd> are idential to <kbd>CTRL-J</kbd> and <kbd>CTRL-K</kbd> by default, so you can use them instead, and you can further customize keys using `--bind` option.

Having said that, if you still think it makes sense to change the title of the pane, please update the commit to use `;` between the commands instead of `&&` just for consistency.

akashin:
In this particular case I would like to avoid learning new shortcuts just because
a) Jumping out of the fzf split is not the reasonable workflow for me
b) I don't use <kbd>CTRL-N</kbd> and <kbd>CTRL-P</kbd> anywhere
c) This is an easy fix that doesn't introduce any other inconsistencies for others

junegunn:
Okay, fair enough. Not using <kbd>CTRL-N/P</kbd> part was a bit surprising to me though as they're universal; shell, vim, emacs, any apps using readline.

junegunn:
Merged, thanks!

kieran-bamforth:
@junegunn Thanks for merging this. @akashin Now that FZF opens a new split with the "fzf" pane title, do I have to update my .tmux.conf to have a flag called $is_fzf (or similar) to conditionally send the change window command? Further, what would that look like? Sorry for my inexperience with Tmux (I am trying to learn!)

junegunn:
@kieran-bamforth Unfortunately this was reverted due to #586.

akashin:
@kieran-bamforth yes, you're right. You can find my tmux.conf here: https://github.com/akashin/dotfiles/blob/master/tmux/tmux.conf.symlink#L72

I'm still going to investigate the problem that appeared for some other shells, hopefully on this weekend.

adamyonk:
@akashin `#T` is showing just `localhost` for me if I open fzf from vim and run `:run-shell 'echo "#T"'`. Did you ever figure this out?

adamyonk:
This seems to be working for me:
```
not_tmux='`echo "#{pane_current_command}" | grep -iqE "(^|\/)g?(view|n?vim?x?)(diff)?$"` || `echo "#{pane_start_command}" | grep -iqE "fzf"` || `echo "#{pane_current_command}" | grep -iqE "fzf"`'
```

akashin:
I didn't know that there is "pane_start_command" available. Unfortunately for me without the PR it shows "env TERM=screen-256color" when I invoke fzf.
With the PR I get "printf '\033]2;fzf\033\';' env..." in pane_start_command.

With this PR #T shows fzf as expected when I invoke fzf-tmux.
I print #T in tmux status pane with 
```
set -g status-left '#T'
```

akashin:
Actually I was seeing only part of #{pane_start_command}.
After adding
```
set -g status-left-length 120
```
I indeed got
```
env TERM=screen-256color  bash /tmp/fzf-args-26570
```
which contains fzf as a substring. So this is a possible workaround without my PR.

mark-westerhof:
pane_start_command and pane_current_command no longer contain anything relating to fzf. Is there another workaround?

brockspratlen:
In case this helps for others looking for the same functionality, I noticed that [christoomey/vim-tmux-navigator](https://github.com/christoomey/vim-tmux-navigator) had been updated in June 2016 to suggest a new detection strategy using `ps` scoped to the current pane's TTY (https://github.com/christoomey/vim-tmux-navigator/pull/117). I thought that was a better solution (as it doesn't rely on pane titles), and I was able to adapt it to support `fzf` as well. I tested this using both `fzf-tmux` and regular `fzf`, and it seems to work perfectly.

```
# Smart pane switching with awareness of Vim splits and fzf.
# See: https://github.com/christoomey/vim-tmux-navigator
not_tmux="ps -o state= -o comm= -t '#{pane_tty}' | grep -iqE '^[^TXZ ]+ +(\\S+\\/)?(g?(view|n?vim?x?)(diff)?|fzf)$'"
bind-key -n C-h if-shell "$not_tmux" "send-keys C-h"  "select-pane -L"
bind-key -n C-j if-shell "$not_tmux" "send-keys C-j"  "select-pane -D"
bind-key -n C-k if-shell "$not_tmux" "send-keys C-k"  "select-pane -U"
bind-key -n C-l if-shell "$not_tmux" "send-keys C-l"  "select-pane -R"
bind-key -n C-\ if-shell "$not_tmux" "send-keys C-\\" "select-pane -l"
```

exosyphon:
I ran into this no longer working with C-h,j,k,l. C-n and C-p work. Did the configuration change recently on this?

kry00la:
I also ran with the same problem as @exosyphon

jackpipe:
me 3

uzxmx:
Try to add this in your shell:

```
export FZF_DEFAULT_OPTS="$FZF_DEFAULT_OPTS --bind 'ctrl-h:execute-silent(tmux select-pane -L),ctrl-j:execute-silent(tmux select-pane -D),ctrl-k:execute-silent(tmux select-pane -U),ctrl-l:execute-silent(tmux select-pane -R)'"
```

Or just run:

```
fzf --bind 'ctrl-h:execute-silent(tmux select-pane -L),ctrl-j:execute-silent(tmux select-pane -D),ctrl-k:execute-silent(tmux select-pane -U),ctrl-l:execute-silent(tmux select-pane -R)'
```

PR #584, opened 2016-06-01
Take SSH completion hints from known_hosts (bash)

jimbocoder:


junegunn:
Merged in b8a3ba16a256e6c83abc709828e65368163e3b9e. I also updated zsh completion. Thanks!

PR #591, opened 2016-06-09
Add possibility to force to use tmux with g:fzf_force_tmux

edi9999:


plugin/fzf.vim
```diff
@@ -140,7 +140,8 @@ try
   else
     let prefix = ''
   endif
-  let tmux = !has('nvim') && s:tmux_enabled() && s:splittable(dict)
+  let forcetmux = get(g:, 'fzf_force_tmux', 0)
+  let tmux = (!has('nvim') && s:tmux_enabled() && s:splittable(dict)) || forcetmux
```
justinmk: The option shouldn't "force". It should still check `s:tmux_enabled()` and fallback to `:term` if not running in tmux.

So a better name for the option might be `g:fzf_prefer_tmux`.

plugin/fzf.vim
```diff
@@ -140,7 +140,8 @@ try
...
   let command = prefix.(tmux ? s:fzf_tmux(dict) : fzf_exec).' '.optstr.' > '.temps.result
 
   if has('nvim')
```
junegunn: We're still using Neovim terminal even when it's not necessary.

junegunn:
Thanks. To address the issues in the comments, I suggest the following:

```diff
diff --git a/plugin/fzf.vim b/plugin/fzf.vim
index 9361181..ba5fb25 100644
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -140,10 +140,10 @@ try
   else
     let prefix = ''
   endif
-  let tmux = !has('nvim') && s:tmux_enabled() && s:splittable(dict)
+  let tmux = (!has('nvim') || get(g:, 'fzf_prefer_tmux', 0)) && s:tmux_enabled() && s:splittable(dict)
   let command = prefix.(tmux ? s:fzf_tmux(dict) : fzf_exec).' '.optstr.' > '.temps.result
 
-  if has('nvim')
+  if has('nvim') && !tmux
     return s:execute_term(dict, command, temps)
   endif
 ```

edi9999:
Done in https://github.com/junegunn/fzf/pull/593

PR #601, opened 2016-06-15
zsh: pass through exit code from fzf with fzf-file-widget

blueyed:
This allows to have a custom widget like the following, which would
additionally accept the line, but only in case of entries being
selected:

    fzf-file-widget-with-accept() {
      zle fzf-file-widget
      if [[ "$?" == 0 ]] && (( $#BUFFER )); then
        zle accept-line
      fi
    }
    zle     -N   fzf-file-widget-with-accept
    bindkey '\e^T' fzf-file-widget-with-accept

With this `<C-a>t` will launch fzf, and simulate the pressing of "Enter"
afterwards.

### DIFF
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -8,10 +8,13 @@ __fsel() {
     -o -type f -print \
     -o -type d -print \
     -o -type l -print 2> /dev/null | sed 1d | cut -b3-"}"
+  setopt localoptions pipefail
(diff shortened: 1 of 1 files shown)


junegunn:
Looks good. Should we do the same to history widget?

blueyed:
Added it for all widgets, but not tested.

junegunn:
Merged, thanks.

PR #608, opened 2016-07-04
Removing error checking to allow clean installation

hposca:
On the append_line method, when trying to discover if the pattern exists
or not the `grep -nF "$pat" "$file"` fails and the installation exits
due to the `set -e`.

### DIFF
--- a/install
+++ b/install
@@ -342,8 +342,6 @@ EOF
 fi
 
 append_line() {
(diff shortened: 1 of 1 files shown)


junegunn:
Ah, thanks. My bad. I fixed it by unsetting pipefail.

PR #610, opened 2016-07-04
[zsh] Re-initialize zle when widgets finish

### LINKED ISSUE #279: HOME and END stop working after using fzf
Like the title says, <kbd>HOME</kbd> and <kbd>END</kbd> keys stop working for me after I've started and exited `fzf`. 

This only happens for one command, though, so after I abort or perform the command, <kbd>HOME</kbd> and <kbd>END</kbd> start working again.

I'm on Ubuntu 14.10 with fzf 0.9.13. This happens for me in the terminals Sakura, and Gnome terminal, but *not* in Terminator.

eigengrau:
zle automatically calls zle-line-init when it starts to read a new line. Many
Zsh setups use this hook to set the terminal into application mode, since this
will then allow defining keybinds based on the $terminfo variable (the escape
codes in said variable are only valid in application mode).

However, fzf resets the terminal into raw mode, rendering $terminfo values
invalid once the widget has finished. Accordingly, keyboard bindings defined
via $terminfo won’t work anymore.

This fixes the issue by calling zle-line-init when widgets finish. Care is taken
to not call this widget when it is undefined.

Fixes #279

### DIFF
--- a/shell/completion.zsh
+++ b/shell/completion.zsh
@@ -58,6 +58,7 @@ __fzf_generic_path_completion() {
         LBUFFER="$lbuf$matches$tail"
       fi
       zle redisplay
+      typeset -f zle-line-init >/dev/null && zle zle-line-init
       break
     fi
     dir=$(dirname "$dir")
@@ -97,6 +98,7 @@ _fzf_complete() {
     LBUFFER="$lbuf$matches"
   fi
   zle redisplay
+  typeset -f zle-line-init >/dev/null && zle zle-line-init
   rm -f "$fifo"
 }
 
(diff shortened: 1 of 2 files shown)


junegunn:
Do we also need this in completion.zsh?

eigengrau:
Good catch. I wasn’t even aware of completion.zsh, since the Archlinux package currently installs it into the wrong location (I’ve filed a bug for this). Since the keybinding there also launches fzf, we’ll need to fix up the terminal mode there, too. Hold on, I’ll amend this commit.

junegunn:
I could confirm that this fixes the issue with the config you provided. Merged, thanks!

Issue #39, opened 2014-05-03
Doc that vi-mode requires "set -o vi"


stephenh:
In my .inputrc, I use "set editing-mode vi". Because this puts readline into vi-mode everywhere, I don't have a "set -o vi" in my .bashrc file, which AFAIK is for making "only bash" (and not other readline-using programs) use vi bindings.

Without "set -o vi", when .bashrc sourced .fzf.bash, the set -o | grep vi check returned empty, looking like vi mode was not setup, and so the non-vi key bindings where set.

However, after .bashrc was evaluated, my .inputrc settings took affect, and if I manually run "set -o | grep vi" in a terminal, then it does actually report as on.

Basically, AFAIK, .fzf.bash is running before .inputrc, so doesn't see the "set editing-mode vi" flag.

After struggling quite awhile with trying to figure out why Ctrl-T was doing bizarre things, I finally put "set -o vi" into my .bashrc, and now things work much better.

A hint in the docs might help future users avoid this confusion. I can add something to the readme if you like, unless it's terribly obvious that anyone using "set editing-mode vi" would of course also have "set -o vi" set. (Obviously it was not obvious to me.)

PR #613, opened 2016-07-06
Fix vi mode test (bash key bindings)

tdy:
Changed the test to `[[ -o vi ]]` since true=0. Now in vi mode, it returns 0 and jumps to the `else` block like we want.

### DIFF
--- a/shell/key-bindings.bash
+++ b/shell/key-bindings.bash
@@ -67,7 +67,7 @@ __fzf_use_tmux__() {
 [ $BASH_VERSINFO -gt 3 ] && __use_bind_x=1 || __use_bind_x=0
 __fzf_use_tmux__ && __use_tmux=1 || __use_tmux=0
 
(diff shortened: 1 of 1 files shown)


junegunn:
Exit status 0 evaluates to true in shell. Are you sure you're not confused?

tdy:
Ah sorry you're right. I confused myself and didn't test it thoroughly enough.

It seems my issue was actually due to enabling vi mode with `set editing-mode vi` in .inputrc, which isn't detected by a `-o vi` test. I guess it requires an explicit `set -o vi`, so I added it to .bashrc and now it's working on my end.

I'm not sure if there's a way to test for both settings, but in any case this patch is wrong. Sorry!

junegunn:
We haven't found the silver bullet. See #39. Please let me know if you discover a better way to handle this. Thanks.

PR #616, opened 2016-07-09
use tmux buffers for sending output to preserve character encoding

seanlaguna:
The command `tmux send-keys` is really not intended for text I don't believe. Instead, we should use a buffer. The buffer can be deleted right after it's used.

This solves the problem of if you select a filename with ctrl+t that has unicode characters in them. I noticed this when selecting song files that were in different languages. The output was corrupted, and could not be used in commands.

This should fix the problem, and I do not believe this problem is anywhere else in the code.

Actually I'm not sure why ctrl+t needs a special function like this in bash only, but I haven't looked at the other shells' implementations to see how it works for them.

(Also I am working on a keybindings port to mksh, if you are interested in merging that as well.)

### DIFF
--- a/shell/key-bindings.bash
+++ b/shell/key-bindings.bash
@@ -26,7 +26,7 @@ __fzf_select_tmux__() {
     height="-l $height"
   fi
 
-  tmux split-window $height "cd $(printf %q "$PWD"); FZF_DEFAULT_OPTS=$(printf %q "$FZF_DEFAULT_OPTS") PATH=$(printf %q "$PATH") FZF_CTRL_T_COMMAND=$(printf %q "$FZF_CTRL_T_COMMAND") FZF_CTRL_T_OPTS=$(printf %q "$FZF_CTRL_T_OPTS") bash -c 'source \"${BASH_SOURCE[0]}\"; tmux send-keys -t $TMUX_PANE \"\$(__fzf_select__)\"'"
+  tmux split-window $height "cd $(printf %q "$PWD"); FZF_DEFAULT_OPTS=$(printf %q "$FZF_DEFAULT_OPTS") PATH=$(printf %q "$PATH") FZF_CTRL_T_COMMAND=$(printf %q "$FZF_CTRL_T_COMMAND") FZF_CTRL_T_OPTS=$(printf %q "$FZF_CTRL_T_OPTS") bash -c 'source \"${BASH_SOURCE[0]}\"; tmux setb -b \"fzf\" \"\$(__fzf_select__)\"; tmux pasteb -b \"fzf\" -t $TMUX_PANE; tmux deleteb -b \"fzf\"'"
 }
 
 fzf-file-widget() {


justinmk:
Are you using some terminal encoding other than utf8?

seanlaguna:
Nope, using utf-8. Note that this only affects control+t in bash and tmux mode. Such text displays correctly in my terminal using any other textual program, and displays correctly in many graphical programs that are sensitive to locale settings.

seanlaguna:
Just double-checked, this issue persists for gnome-terminal, xfce4-terminal, xterm, and st, all of which are using utf-8.

junegunn:
Thanks, I've noticed this recently with the latest versions of tmux. I'll merge your commit with updated test code.

seanlaguna:
Awesome?

Any comments on why the script is needed for bash? Why can't we just use pipes to fzf or fzf-tmux when doing the ctrl+t command implementation?

Also, are you interested in mksh? Problem with any korn shell is there's no inline command expansion -- i.e., the return key _must be pressed_ and a command _must be entered into the prompt_ in order for anything to happen. We still use bind, but bind enters an enter/return in order to execute the necessary command. I have ctrl+r, ctrl+t, and alt+c working, and I'm working on porting over as much as possible from complete as well. (But, korn shell command line completion is very weak, so it will not be possible to support it all). The ways to get these commands working properly are extremely hackish, but they will work identically to the other shells' versions except for entering a single command into the terminal of the name

``` $(FZF_FUNC_XXX)```

or whatever. (Note the leading space -- mksh _unconditionally_ doesn't save to history if there is a leading space in a command).

There are also some extra environment variables that are set; this is unaviodable.

If this is the kind of thing that could feasibly be merged I would work to make it as clean and standardized as possible, so let me know when you have the chance.

junegunn:
Merged, thanks.

> Actually I'm not sure why ctrl+t needs a special function like this in bash only

zsh and fish use `fzf-tmux` script which greatly simplifies the implementation. However in bash, invoking `fzf-tmux` requires `shell-expand-line` which expands aliases on the command line, e.g. `ls` becomes `ls -G` immediately when you press CTRL-T.

seanlaguna:
Argh, alias expansion... that'll do it...

If korn shell had $() expansion it would be my perfect shell.

junegunn:
> are you interested in mksh?

Hmm, I'm a bit concerned that I'll have difficulty maintaining the code as I have no experience or knowledge on mksh. For now, let's keep it as a separate repository and maybe add a link to it on the README page?

seanlaguna:
That would be awesome! I'd love to see people stress-test it, but for now I've done a decent amount of comparing the output with bash version and it looks to behave the same way. It's hard, though, because of differences with the history file/recall in bash/mksh, they inherently do not behave the same way.

junegunn:
Nice, ping me when your repository is ready, I'll update the doc.

junegunn:
Hmm, I noticed an issue with the change. Older tmux releases do not support named paste buffers and `tmux setb -b fzf "something"` gives `buffer invalid` error. That's why the tests are failing on Travis CI environment where tmux 1.9 is installed. I do not wish to revert the change, but I do wish to support older tmux versions as well. I'll see what I can do.

PR #624, opened 2016-07-19
Mru File List Support

lvht:
Let **FZF** to remember our most recently used file list like **CtrlPMRU**.  So we can use the **FZF** to replace the **CtrlP** plugin.

All recently used file path will be save an global list variable called `g:fzf_mru_file_list`. And the member of `g:fzf_mru_file_list` will be concatenate by `;` to generate an string, which will be stored in to the `g:FZF_MRU_FILE_LIST` variable. 

Thanks to Vim's **viminfo** and NeoVim's **shada**, the `g:FZF_MRU_FILE_LIST` will be save to and restore from the disk **automatically**. So there is no need to make another file to store the list.

**CtrlP** is awesome, but **FZF** is more better.

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -486,7 +486,63 @@ function! s:cmd(bang, ...) abort
   call fzf#run(extend({'options': join(args), 'sink*': function('<sid>cmd_callback')}, opts))
 endfunction
 
+function! s:ListMruFile()
+	let files = map(copy(g:fzf_mru_file_list), 'fnamemodify(v:val, ":~:.")')
+	let file_len = len(files)
+	if file_len == 0
+		return
+	elseif file_len > 10
+		let file_len = 10
+	endif
+	let file_len = file_len + 2
+	call fzf#run({
+			\ 'source': files,
+			\ 'sink': 'edit',
+			\ 'options': '-m -x +s',
+			\ 'down': file_len})
+endfunction
+
+function! s:RecordMruFile()
+	let cpath = expand('%:p')
+	if !filereadable(cpath)
+		return
+	endif
+	if cpath =~ 'fugitive'
+		return
+	endif
+	let idx = index(g:fzf_mru_file_list, cpath)
+	if idx >= 0
+		call filter(g:fzf_mru_file_list, 'v:val !=# cpath')
+	endif
+	call insert(g:fzf_mru_file_list, cpath)
+endfunction
+
+function! s:ClearCurrentFile()
+	let cpath = expand('%:p')
+	let idx = index(g:fzf_mru_file_list, cpath)
+	if idx >= 0
+		call remove(g:fzf_mru_file_list, idx)
+	end
+endfunction
+
+if has('nvim')
+	rsh
+else
+	" todo
+endif
+
+if exists('g:FZF_MRU_FILE_LIST_JSON')
+	let g:fzf_mru_file_list = split(g:FZF_MRU_FILE_LIST_JSON, ";")
+else
+	let g:fzf_mru_file_list = []
+endif
+
+autocmd! BufEnter * call s:ClearCurrentFile()
+autocmd! BufWinLeave,BufWritePost * call s:RecordMruFile()
+autocmd! VimLeavePre * let g:FZF_MRU_FILE_LIST_JSON = join(g:fzf_mru_file_list, ";")
+
 command! -nargs=* -complete=dir -bang FZF call s:cmd(<bang>0, <f-args>)
+command! FZFMru call s:ListMruFile()
 
 let &cpo = s:cpo_save
 unlet s:cpo_save
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -528,18 +528,19 @@ endfunction
 if has('nvim')
 	rsh
 else
-	" todo
+	execute 'set viminfo=!,'.&viminfo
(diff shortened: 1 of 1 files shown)


junegunn:
Thank you, but we don't add features to the Vim plugin in this repository. It is supposed to provide only the basic `fzf#run()` function. Instead, we add stuff to [fzf.vim](https://github.com/junegunn/fzf) project, but I'm not sure if I want to put MRU feature in there either as the required bookeeping tends to be a bit too involved for my taste.

There is [fzf-filemru](https://github.com/tweekmonster/fzf-filemru) plugin you might want to check out. I haven't tried it myself, but I believe @tweekmonster can tell you more about it. Maybe he's also interested in taking a look at your implementation.

tweekmonster:
Thanks for the plug :)  While it would be nice to have the MRU built into fzf.vim, I've come to see @junegunn's perspective.  fzf.vim makes use of FZF's and Vim's builtin features, but doesn't add anything beyond that.  Building in MRU would require more than a few options and a separate script since opinions on what an MRU is varies.

fzf-filemru is easy enough to maintain separately as long as it's possible to grab a reference to fzf.vim's `s:common_sink`.

lvht:
Hello @tweekmonster . There is no need to make a `$XDG_CACHE_HOME/fzf_filemru` file to store the MRU file list.

lvht:
@junegunn What I want to do is just make the FZF to replace the original CtrlP out of box. As @tweekmonster say,
>  fzf.vim makes use of FZF's and Vim's builtin features, but doesn't add anything beyond that. 

What fzf.vim can do is just make Vim's builtin features more easy. But it is just for @junegunn personal use.

The FZF is an awesome cli tool which should deserve to be used in more and more other tool, eg Vim, Emacs, and so on. Maybe the FZF repository should be split into to parts like 

1. fzf-cli which only contains the binary's `go` code
1. fzf-vim which contains the VimL required for vim integration
1. fzf-emacs for emacs
1. more ...

And I use Homebrew to install the fzf binary, and use Vim-Plug to install [fzf](https://github.com/junegunn/fzf) again. For vim, what I only need is the https://github.com/junegunn/fzf/blob/master/plugin/fzf.vim. However, I have to download all other `*.go` file.

lvht:
Finally, I make an new repo [fzf-mru](https://github.com/lvht/fzf-mru). FYI @tweekmonster https://github.com/lvht/fzf-mru#story .

Good luck.

tweekmonster:
> There is no need to make a $XDG_CACHE_HOME/fzf_filemru file to store the MRU file list. The viminfo or shada is just enough.

I don't understand why you wouldn't want MRU data stored in a separate file.  IMO, using viminfo/shada is not safe for state information.  First, it could be considered unexpected behavior if you change the settings.  Second, if the user limits the size of entries, your data might not persist.

I consider the data in viminfo/shada critical to my work.  I personally wouldn't want a plugin to alter the settings.

junegunn:
@tweekmonster 
> fzf-filemru is easy enough to maintain separately as long as it's possible to grab a reference to fzf.vim's s:common_sink.

Do you want me to make it public? I can surely consider that.

@lvht 

> However, I have to download all other *.go file.

Homebrew package of fzf contains fzf.vim, so you can do this:

```vim
Plug '/usr/local/opt/fzf' | Plug 'junegunn/fzf.vim'
```

I recently [updated the installation instructions](https://github.com/junegunn/fzf.vim/commit/cd94fca6ce608c0f66086a393b53e56f33b27f28) to mention the above.

> fzf-emacs for emacs

https://github.com/bling/fzf.el

> What fzf.vim can do is just make Vim's builtin features more easy. But it is just for @junegunn personal use.

That was the original intention as I know it becomes more and more difficult to satisfy different needs as the feature set grows, and fzf.vim is a plugin that does maybe too many different things. But I'm aware that there are a lot of users actively using it, so I probably won't break things.

There are many [CtrlP extensions](https://github.com/search?utf8=%E2%9C%93&q=ctrlp) out there that provide extra features. I prefer the approach over having an inflexible, monolithic plugin and we can do the same for fzf, like you guys did.

Feel free to add your projects to [the wiki page](https://github.com/junegunn/fzf/wiki/Related-projects). I'm going to close this as wontfix. Thanks.

tweekmonster:
> Do you want me to make it public? I can surely consider that.

Only if you think it'll solve other issues.  I think it works fine as-is.  Maybe it'd be better if the caller's `*sink` was wrapped and if it returned `1`, run `s:common_sink`.

FWIW, I considered `fzf.vim` to be a very functional showcase.  I would expect public utility APIs to be in `fzf`'s script.

junegunn:
> I would expect public utility APIs to be in fzf's script.

Good point. You don't want your fzf-based vim plugin to depend on both fzf and fzf.vim.

FinallyFinancialFreedom:
i thought the :History command does the mru work, is it different?

tweekmonster:
@FinallyFinancialFreedom `:History` uses `v:oldfiles` and current buffers to give you a list.  The order isn't exactly based on how recently you used the file.  With FZF you can filter the list pretty easily to find the file you want.  With an MRU, you can do it in fewer keystrokes if you have a lot of similarly named files since the results are sorted based on how recent and frequently you used the file.

FinallyFinancialFreedom:
well, thank you for explaining, i'll look into it

PR #630, opened 2016-07-28
Remove `name` option from `termopen`.

kassio:
`termopen` no longer accepts a `name` option, instead we should suffix the
command with `;#NAME`.

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -204,7 +204,7 @@ function! fzf#run(...) abort
 try
   let oshell = &shell
   set shell=sh
-  if has('nvim') && bufexists('term://*:FZF')
+  if has('nvim') && len(filter(range(1, bufnr('$')), 'bufname(v:val) =~# ";#FZF"'))
     call s:warn('FZF is already running!')
     return []
   endif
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks, but does `bufexists('term://*FZF')` really work? I just realized that `:help bufexists` states that "If the {expr} argument is a string it must match a buffer name exactly." and there's no mention of wildcard expansion. Do we need something like `len(filter(range(1, bufnr('$')), 'bufname(v:val) =~# "#FZF"'))`?

kassio:
@junegunn You're right, I just replaced the `bufexists` for the `filter`.

junegunn:
Merged, thanks!

khalidchawtany:
I am following **HEAD** in both **neovim** and **FZF** and this breaks **FZF** in **neovim**. The first time I run `:FZF` inside **neovim**, everything is smooth. However, after that, I get `FZF is already running!` error message.

neovim version:  NVIM v0.1.5-659-gaf95037
FZF version:        0.13.3

If I revert to the commit before this everything is working again.

junegunn:
@khalidchawtany That's strange. I tested with the latest nvim (v0.1.5-826-g56f1780) before merging this. Can you check if the problem is still reproducible with this minimal vimrc? https://gist.github.com/junegunn/6936bf79fedd3a079aeb1dd2f3c81ef5

khalidchawtany:
You are right, something is wrong with my setup. Your minimal vimrc has no problems. Thanks.

junegunn:
@khalidchawtany Thanks for the investigation. I just pushed the fix to the problem.

PR #642, opened 2016-08-25
vim: reset window sizes on close

joshuarubin:
This is a small patch to restore window sizes in vim (and neovim) after fzf is closed. The code was largely taken from CtrlP: https://github.com/kien/ctrlp.vim/blob/master/autoload/ctrlp.vim#L305-L307

I believe that this resolves junegunn/fzf.vim#42 and #520


junegunn:
Nice patch, thanks! Merged in f025602841384fb10cc80d37492ec44c2b872860.

PR #645, opened 2016-08-28
Split default zsh binding at the correct place

ckafi:
The command substitution and following word splitting to determine the default
zle widget for ^I formerly only works if the IFS parameter contains a space. Now
it specifically splits at spaces, regardless of IFS.


junegunn:
Thanks!

PR #646, opened 2016-08-28
Fix grouped tmux session confusion

### LINKED ISSUE #643: FZF can open in wrong terminal/window with multiple group sessions
- Category
    - [ ] fzf binary
    - [x] fzf-tmux script
    - [ ] Key bindings
    - [ ] Completion
    - [ ] Vim
    - [ ] Neovim
    - [ ] Etc.
- OS
    - [ ] Linux
    - [x] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [ ] bash
    - [x] zsh
    - [ ] fish

To reproduce:

1. Open two terminals (I just use two iTerm windows).
2. Terminal 1> Run `tmux`. Note the session-id.
3. Terminal 1> Run tmux command `new-window` (or the corresponding hot-key).
4. Terminal 2> Run `tmux new-session -t <session-id>`. This makes the new session share windows with the session open in terminal 1.
5. Terminal 2> Run `fzf` (or anything that invokes `fzf`).

Result: The fzf split will open in terminal 1 (which is on window 2). It should open in terminal 2 on window 1.

image

I've reproduced this in both iTerm 2 and the built-in OS X Terminal. `$TERM` is set to `screen-256color`.

qiemem:
Fixes #643.

Uses the suggestion from https://github.com/tmux/tmux/issues/521#issuecomment-243000022

### DIFF
--- a/bin/fzf-tmux
+++ b/bin/fzf-tmux
@@ -161,14 +161,14 @@ done
 
 if [[ -n "$term" ]] || [[ -t 0 ]]; then
   cat <<< "\"$fzf\" $opts > $fifo2; echo \$? > $fifo3 $close" > $argsf
-  tmux set-window-option synchronize-panes off \;\
+  TMUX='' tmux set-window-option synchronize-panes off \;\
     set-window-option remain-on-exit off \;\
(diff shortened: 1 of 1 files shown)


junegunn:
Merged, thanks!

PR #652, opened 2016-09-05
fixed $MANPATH reference

zfogg:
:)

### DIFF
--- a/install
+++ b/install
@@ -303,8 +303,8 @@ fi
 
 # Man path
 # --------
-if [[ ! "\$MANPATH" == *$fzf_base/man* && -d "$fzf_base/man" ]]; then
-  export MANPATH="\$MANPATH:$fzf_base/man"
+if [[ ! "\$MANPATH" == *$fzf_base/share/man* && -d "$fzf_base/share/man" ]]; then
+  export MANPATH="\$MANPATH:$fzf_base/share/man"
 fi
 
 # Auto-completion


zfogg:
this fixes $MANPATH on my OS X (10.11.6) machine for homebrew's fzf

```zsh
$ brew update && brew install fzf
$ cd `brew --prefix` && pwd
/usr/local/opt/fzf
$ cd man
cd: no such file or directory: man
$ tree share
share
└── man
    └── man1
        ├── fzf-tmux.1
        └── fzf.1

2 directories, 2 files
```

potentially useful debug info:

```zsh
$ brew info fzf
fzf: stable 0.13.5 (bottled), HEAD
Command-line fuzzy finder written in Go
https://github.com/junegunn/fzf
/usr/local/Cellar/fzf/0.13.5 (16 files, 3.4M) *
  Poured from bottle on 2016-09-05 at 05:37:53
From: https://github.com/Homebrew/homebrew-core/blob/master/Formula/fzf.rb
```

while installing to `$fzf_base/`, do any systems install to `./man/` vs `./share/man/`?
i could add an `if 'OS_X'` check in there, if so

junegunn:
Thanks for looking into it. The path is invalid only when you installed fzf using Homebrew, otherwise (git clone) the path is correct. And in that case we don't have to add it to MANPATH as Homebrew creates symlinks to those files inside `/usr/local/share/man/man1` which is already in MANPATH. So I would say that the current code is correct in that it does not update MANPATH when the directory is not found.

PR #663, opened 2016-09-15
Global search functionality

m2march:
Added a binding for searching from your home independently of your current directory. Change was tested on bash and zsh. Was not made for the .fish file. 

Readme was not properly updated.


junegunn:
Thanks for your effort, but I decided long ago not to add more shell stuff to this repository (there has not been a new shell key binding since March 2014). I believe you can define it on your shell configuration files.

m2march:
Yes. I have, just thought I'd share in case it was of interest.

Thanks for reading.

Best regards,
March

PR #670, opened 2016-09-24
Fix #668

### LINKED ISSUE #668: -bash: _fzf_orig_completion_g++: bad substitution
- Category
    - [ ] fzf binary
    - [ ] fzf-tmux script
    - [ ] Key bindings
    - [x] Completion
    - [ ] Vim
    - [ ] Neovim
    - [ ] Etc.
- OS
    - [ ] Linux
    - [x] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [x] bash
    - [ ] zsh
    - [ ] fish

I am on the latest master 7fa5e6c. On a fresh terminal, I type `emacsclient ~/a` and press tab. Then `-bash: _fzf_orig_completion_g++: bad substitution` will be displayed, giving this terminal screen:
```
$ emacsclient ~/a-bash: _fzf_orig_completion_g++: bad substitution

[a list of filenames that start with an `a` inside my home dir; elided]
```

At this point if I press tab again, then the normal completion will continue and this error will not happen a second time in the same shell session.

I believe it has to do with lines 280 and 281 of `completion.bash`:
```bash
  orig_var="_fzf_orig_completion_$cmd"
  orig="${!orig_var}"
```
According to the bash 4 manual (I am on `4.4.0(1)-release`), a name is:

> A word consisting only of alphanumeric characters and underscores, and beginning with an alphabetic character or an underscore.

I guess that means `_fzf_orig_completion_g++` is indeed not a name?

maverickwoo:
Fix #668

Handle uppercase letters in program names. This also deals with `-` and
`.`, both of which are quite common in program names, e.g., `xdg-open`
and `foo.sh`.

### DIFF
--- a/shell/completion.bash
+++ b/shell/completion.bash
@@ -32,7 +32,7 @@ fi
 
 _fzf_orig_completion_filter() {
   sed 's/^\(.*-F\) *\([^ ]*\).* \([^ ]*\)$/export _fzf_orig_completion_\3="\1 %s \3 #\2";/' |
-  awk -F= '{gsub(/[^a-z0-9_= ;]/, "_", $1); print $1"="$2}'
+  awk -F= '{gsub(/[^A-Za-z0-9_= ;]/, "_", $1); print $1"="$2}'
 }
 
 _fzf_opts_completion() {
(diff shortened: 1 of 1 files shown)


junegunn:
Thank you!

maverickwoo:
No problem. I really like fzf. My thanks to you!

PR #673, opened 2016-09-27
use `fc -r` instead of `fzf --tac` in Zsh's CTRL-R

pjmtdw:
When we press CTRL-R to search from history using fsf, it scrolls the history from top to bottom.
When the history size is large,(e.g. > 100,000) it is annoying since the scroll occurs each time we hit CTRL-R.
To avoid it, we should append `-r` to `fc` command so that it shows the history in reverse order, and remove `--tac` from `fzf`.

### DIFF
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -49,7 +49,7 @@ bindkey '\ec' fzf-cd-widget
 fzf-history-widget() {
   local selected num
   setopt localoptions noglobsubst pipefail 2> /dev/null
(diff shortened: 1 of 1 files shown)


pjmtdw:
sorry it does not pass the test. I should close it until I find the reason.

junegunn:
No problem. I really like what you did here. Maybe the option requires a certain version of zsh that is not available on Travis CI? I'll also take a look later in the day.

junegunn:
The tests pass locally on my machine with zsh 5.2. Travis CI env has zsh 4.3.17 which is way too old. And in my opinion it makes not much sense to try to support it. I'll see if I can install a newer version of zsh there.

PR #695, opened 2016-10-06
Make curl and wget requests quiet

matthewfranglen:
Otherwise you end up with things like the following being issued to stderr:

```
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   603    0   603    0     0   1331      0 --:--:-- --:--:-- --:--:--  1334
100  913k  100  913k    0     0   571k      0  0:00:01  0:00:01 --:--:--  890k
```

### DIFF
--- a/install
+++ b/install
@@ -112,11 +112,11 @@ link_fzf_in_path() {
 }
 
 try_curl() {
-  command -v curl > /dev/null && curl -fL $1 | tar -xz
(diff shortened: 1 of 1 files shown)


junegunn:
Actually I like the progress report. It is especially nice for users with slow internet connection. If you don't want it, you can redirect stderr to /dev/null.

PR #704, opened 2016-10-15
add auto-margin option

acornejo:
This branch is not ready yet, but wanted to open the PR to get some comments.

There are two main problems.

1: The MaxY from curses doesn't work as I expected (i.e. it doesn't return the terminal height).
2. Even when using margin, it seems fzf clears the entire screen

I can think of a few solutions for (1), but I don't know what to do about (2).

### DIFF
--- a/src/options.go
+++ b/src/options.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/junegunn/fzf/src/algo"
 	"github.com/junegunn/fzf/src/curses"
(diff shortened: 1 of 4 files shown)


acornejo:
Read a bit about ncurses. There is no easy fix for (2) since there seems to be no portable way of querying the current state (i.e. characters displayed) on a terminal.

Without (2) I am not sure this PR makes sense (i.e. what use is auto-margin?). Closing.

PR #714, opened 2016-10-24
Termbox

kelleyma49:
Here are my initial changes.  

Note that preview windows works on MacOS termbox version, and seems to work under Windows 10.


junegunn:
It's not yet finished, right?

Here are a few issues I have noticed
- Colors look slightly different. Looks like ANSI color codes are shifted by 1.
- Preview window does not render text with ANSI colors
- Preview window does not wrap lines
- Process terminated on mouse scroll

screen shot 2016-10-30 at 2 16 18 pm
screen shot 2016-10-30 at 2 17 54 pm

kelleyma49:
@junegunn - regarding the issues that you mentioned:

1. I will look into this 
2. can you give me a test case for this?
3. should be fixed in my latest commit
4. there's a bug in mouse scrolling in termbox-go (see mentioned issue in my latest source code commit).  I disabled mouse input for now as this terminates fzf and I can't find an easy workaround.

Other than that it seems to be working fine for me, though I haven't stress tested all of fzf's functionality.

junegunn:
Correction: "Preview window does not render text with ANSI colors **correctly**"

Anyway, the content in the preview window looks still broken after the fix, in a different way.

I'm using `highlight` program to generated syntax-highlighted text. In case you don't have access the program, I uploaded the files so you can test with them. 

https://gist.githubusercontent.com/junegunn/fb0b084d0718a2f27793d7982f983e10/raw/131e64c7386f4d67285b6c884883fad263b6dbb3/algo.txt

https://gist.githubusercontent.com/junegunn/fb0b084d0718a2f27793d7982f983e10/raw/131e64c7386f4d67285b6c884883fad263b6dbb3/gistfile1.txt

screen shot 2016-10-31 at 3 07 27 pm

screen shot 2016-10-31 at 3 07 32 pm

Long lines are not wrapped. 

(`fzf --preview 'cat algo.txt' --preview-window right:10`)

screen shot 2016-10-31 at 3 09 51 pm

junegunn:
FYI, the difference between `Print` and `Fill` are as follows:

- `Print` is used to print a single line string in the candidate list. If strips non-printable characters (`r < 32`)
- `Fill` was introduced for preview window. It does not strip non-printable characters. Note that ncurses automatically handles line-wraps.
- `Fill` returns false if the cursor goes beyond the screen, so there's no point of calling it again. We use this information to stop rendering the preview window earlier.

kelleyma49:
Thanks for the explanation - ncurses documentation is as detailed as I'd like. :)

kelleyma49:
It took me longer than expected, but the main reason was my switch from termbox to tcell.  Termbox had the fast scrolling bug with the mouse, and I noticed a similar bug with arrow key scrolling.  Tcell seems to have none of these issues.

I've tested these commits under Mac and Windows.

kelleyma49:
Hm, I'm not sure why this is failing...

junegunn:
Great job, thanks! I'll review the code later in the day.

> I'm not sure why this is failing

There is a ZERO WIDTH NO-BREAK SPACE character at the start of tcell.go

```diff
diff --git a/src/tui/tcell.go b/src/tui/tcell.go
index eaf363c..5404726 100644
--- a/src/tui/tcell.go
+++ b/src/tui/tcell.go
@@ -1,4 +1,4 @@
-<U+FEFF>// +build tcell windows
+// +build tcell windows
```

I could compile the code after removing it.

kelleyma49:
Good find!  I've submitted a fix.  

One small bug under Windows: double click doesn't work (works on MacOS).

junegunn:
Thanks, is it something you're going to fix? Let me know when you're done. I'll finish up the work and merge.

Swoorup:
Does this mean that this would work in cygwin, msys2 and both neovim and vim plugins?

EdVanDance:
Proabably Not. See Windows at https://github.com/gdamore/tcell/blob/master/README.md

kelleyma49:
I've disabled single and double click for mouse as I believe this is not an issue with our code.  cmd.exe and conemu.exe consume these events for text selection, which means that single and double click don't really work under Windows.  Mouse scrolling seems to work without a problem, so I left that in.

I ran into another bug which you might be able to help me debug.  DOS doesn't like single quotes, so I changed the quoteEntry function to use double quotes instead:

```go
func quoteEntry(entry string) string {
	if runtime.GOOS == "windows" {
		return "\"" + strings.Replace(entry, "\"", "\\\"", -1) + "\""
	} else {
		return "'" + strings.Replace(entry, "'", "'\\''", -1) + "'"
	}
}
```

This almost works, except the double quotes at the head and tail of the string get escaped, which DOS doesn't like.

Here's my example command:

`.\fzf.exe --preview-window right:70 --preview "echo {} "`

And here's what gets printed:

` \"C:\go\src\github.com\kelleyma49\fzf\src\fzf\fzf.exe\"`

How do I stop fzf from escaping the double quotes?

kelleyma49:
I think you can merge this after the quoting issue is fixed (and after your final cleanup pass.)

kelleyma49:
Here's an appveyor.xml CI script to get the termbox branch of fzf built and tested:

```yaml
version: 1.0.{build}
branches:
  only:
  - termbox
image: WMF 5
clone_folder: c:\gopath\src\github.com\junegunn\fzf
environment:
  GOPATH: c:\gopath
install:
- cmd: >-
    go version

    go env

    go get -v -d -t github.com/junegunn/fzf/...
build_script:
- cmd: 
test_script:
- cmd: go test -v -tags tcell -cover -race ./...


```

junegunn:
Thanks for your efforts, it's really appreciated. I squashed your commits into one and fixed some minor issues I noticed including the quoting issue you mentioned (it seems to work now, but I'm not perfectly sure about it yet). The changes are pushed to [tcell branch](https://github.com/junegunn/fzf/tree/tcell). The acceptance tests are passing locally, but for some reason, I'm having a hard time making them pass on Travis CI environment. I'll merge the branch to master once it's fixed. Thanks again!

kelleyma49:
Thanks June.  I'm looking forward to seeing this merged.

One issue I noticed is that the outputted string in the placeholder string is missing all '\'.  I tried the change below and the '\'s will appear:

```go
func quoteEntry(entry string) string {
	if util.IsWindows() {
		//return `"` + strings.Replace(entry, "\"", "\\\"", -1) + `"`
		return strconv.Quote(strings.Replace(entry, "\"", "\\\"", -1))
	}
	return "'" + strings.Replace(entry, "'", "'\\''", -1) + "'"
}
```

junegunn:
Thank you, I applied your fix. Here is the updated list of issues I fixed in the tcell branch.

- Fix display of CJK wide characters
- Fix horizontal offset of header lines
- Add support for keys with ALT modifier, shift-tab, page-up and down
- Fix util.ExecCommand to properly parse command-line arguments
- Fix redraw on resize
- Implement Pause/Resume for execute action
- Remove runtime check of GOOS
- Change exit status to 2 when tcell failed to start
- TBD: Travis CI build for tcell renderer
    - Pending. tcell cannot reliably ingest keys from tmux send-keys

I was unaware of tcell before, but it looks very promising. It's actively developed and already supports 24-bit colors unlike ncurses. But unfortunately, I couldn't find a way to make the existing test cases based on `tmux send-keys` pass reliably with the tcell version. Another minor issue I noticed with tcell is that go-runewidth which tcell depends on does not cover every wide character. I should probably revisit tcell in the future, but for now we're going to stick with ncurses for non-windows binaries.

junegunn:
@kelleyma49 Released 0.15.6 and prebuilt binaries for Windows are up!
https://github.com/junegunn/fzf-bin/releases

Thanks again.

EdVanDance:
First, thanks for this awesome piece on windows :+1:

For me it works as expected within cmd.exe, but fails with "character set not supported" in ConEmu. Any idea what the problem might be here?

```
> chcp
Active code page: 850
```

kelleyma49:
@EdVanDance - it works for me under ConEmu.  What are your ConEmu settings?

EdVanDance:
@kelleyma49 Sorry for responding late, didn't have access to a windows box.

You're right, it actually works with vanilla ConEmu, I was using [cmder](http://cmder.net/) and there it failed.

EdVanDance:
@kelleyma49 I think I found the problem. Cmder is setting TERM=cygwin for git for windows, that's not supported by tcell.

Issue #484, opened 2016-02-07
Completion trigger for fish


nh2:
I noticed that `**` completion triggers don't work in `fish`. Would be awesome if they did!

junegunn:
Yes, fuzzy completion is only supported for bash and zsh.

https://github.com/junegunn/fzf#fuzzy-completion-for-bash-and-zsh

Unfortunately I'm not a fish expert and don't know to how to implement fuzzy completion on it. Let me know if you're interested in contributing.

hauleth:
AFAIK it is not possible now. Actually it would be better to open issue in fish-shell/fish-shell to allow external completion tool, so `cmd <TAB><TAB>` would open `fzf` instead of built in fuzzy-finder.

coderstephen:
This isn't available in the Fish core yet, but there is now a third-party script plugin that enables writing these types of completions called [`expand`](https://github.com/oh-my-fish/plugin-expand) (shameless plug, I wrote it). If you're looking for fancier fzf integration for Fish, likely a plugin could be written for it.

junegunn:
Closed via #759.

miyl:
This is 2016 - does `FZF_COMPLETION_TRIGGER` work in fish in 2021?
For me it seems it does not (e.g. `vim /etc/**<tab>` blinks and then nothing), but I don't see it mentioned anywhere that it shouldn't work?

bs10x:
Does this work now? Still the same for me..

> This is 2016 - does `FZF_COMPLETION_TRIGGER` work in fish in 2021? For me it seems it does not (e.g. `vim /etc/**<tab>` blinks and then nothing), but I don't see it mentioned anywhere that it shouldn't work?

janderssonse:
Same here, I guess it should not work in Fish as doc state "https://junegunn.github.io/fzf/shell-integration/#fuzzy-completion-for-bash-and-zsh"

bakai-taalaibek:
I made a function for myself where upon pressing TAB I run a function that checks previous character and runs one of two functions conditionally. If previous character is " ", then I run fzf, otherwise atuin for command completion. 

function act_on_prev_char
    # Get full current commandline before the cursor
    set -l line (commandline --current-buffer --cut-at-cursor)
    
    # Get the last character before the cursor
    set -l last_char (string sub -l 1 -s -1 -- $line)

    if test "$last_char" = " " -o "$last_char" = "/"
        fzc
    else
        _atuin_search
    end
end

maybe it can help somebody

Issue #546, opened 2016-04-21
Allow for alternate key bindings


kopischke:
- Category
    - [ ] fzf binary
    - [ ] fzf-tmux script
    - [x] Key bindings
    - [ ] Completion
    - [ ] Vim
    - [ ] Neovim
    - [ ] Etc.
- OS
    - [ ] Linux
    - [x] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [x] bash
    - [ ] zsh
    - [ ] fish

Two out of three of `fzf`’s key bindings conflict with OS X features:

- `Ctrl-T` clobbers the “transpose” function of the Cocoa text system ordinarily bound to that key; too useful a feature to simply lose for `fzf` usage. 
- `Alt-C` does not work at all, as Alt+letter combinations insert special characters (“ç” on a German keyboard); disabling that is not even an option for anybody writing in a language other than English on OS X.

Providing either a way to configure these bindings (not just opt out on install – the hotkeys **are** useful after all), or saner defaults on OS X would be much appreciated.

junegunn:
Technically CTRL-T is not an OS X feature, but a default key binding of bash that is mapped to "transpose-chars" (see `bind -p` to confirm). I was well aware of the feature but as I hardly used it – and I figure not many others use it judging from the fact you're the first one to mention it – so I decided to override the key. But I guess that it makes sense to provide an option or suggest a way to configure the key without too much hassle.

Excerpt from https://www.iterm2.com/faq.html:

> Q: How do I make the option/alt key act like Meta or send escape codes?
>
> A: Go to Preferences->Profiles tab. Select your profile on the left, and then open the Keyboard tab. At the bottom is a set of buttons that lets you select the behavior of the Option key. For most users, Esc+ will be the best choice.

I use `Esc+` setting as suggested above so it works for me, but since you said it's not an option, `ESC-C` should do the trick. If you weren't aware of this, I can say you are missing out. `ALT-B` (backward-word), `ALT-F` (forward-word), `ALT-Backspace` (backward-kill-word), `ALT-D` (kill-word), and `ALT-.` (insert-last-argument) are just indispensable. Likewise, they are available to you via `ESC-B`, `ESC-F`, etc. Note that iTerm allows you to choose the option for left option key and right option key separately, so it's something you can consider.

kopischke:
[...]

PR #747, opened 2016-11-29
Overhaul fish functions

Ambrevar:
The fish functions needed a good revamp. Here it is.

Use seperate files to auto-load functions (more idiomatic in fish).

Remove default bindings. Since they conflict with fish bindings it is
more work for the user to restore them.

Replace the "temp file" workaround with the "read" function: it's
simpler and faster.

Use proper escaping, remove the custom function.

The "file" widget uses last token as root for the "find" command.
This replaces the equivalent of '**' completion in bash/zsh.
The "$cwd" non-expanded variable can be used in FZF_CTRL_T_COMMAND to
set the root.

This closes #443, #484 and #486.


junegunn:
Thanks, but test cases are failing. Probably because you removed the default bindings? Improving the existing scripts is okay, but I don't think changing the code structure just to make it more idiomatic justifies breaking backward compatibility for the users.

Have you taken a look at projects like https://github.com/fisherman/fzf? The default key bindings provided here can be easily opted-out by saying no to the question `Do you want to enable key bindings? ([y]/n)`. And one can choose to use any third-party implementation instead.

Ambrevar:
Tests are failing indeed, I was aware of that, now let's discuss! :)

I do consider backward compatibility very seriously, but is backward compatibility _of the shell functions_ essential?

I've changed the following things that might break backward compatibility:

- Removing the key bindings: I believe it to be a sane behaviour, bindings should belong to the user dotfiles.

- Renaming the functions: not necessary _per se_, but since those functions cannot be invoked without a bindings, I thought that would make more sense, and infestation of "fzf-*" widgets in the global namespace.

- Splitting the file into one-function-per-file: actually, this should not break backward compatibility since the functions should be autoloaded by fish if installed properly (which is the whole point).

The `fish` functions have been in a poor shape for a while: as it is now, the way fzf is meant to be loaded by fish just feels broken. Breaking the broken is acceptable in my opinion.

If you are still relunctant to accept some of those 3 changes, I'll re-submit accordingly.

Ambrevar:
`fisherman` is not a solution to the binding problem, it is just merely a workaround.
I believe this should be fixed upstream, i.e. on fzf.

junegunn:
Thanks for sharing your thoughts. But there are points that I don't quite agree with.

> but is backward compatibility of the shell functions essential?

Yes, they are a part of the project so why should they be any different?

> bindings should belong to the user dotfiles.

I don't disagree. But by that logic, don't we have to consistently remove the default bindings from bash and zsh extensions as well? But it's a drastic change I can't make at this point. Like I said above, key bindings are opt-in features. If you don't like the settings, answer no to the installer question, and use your own bindings. And if the user explicitly said yes to the question, we should enable them, or what's the point of it?

> limiting the infestation of "fzf-*" widgets in the global namespace.

The names are exposed on purpose and have been a part of the public interface of the project. The names are carefully chosen to be consistent across all three shells as much as possible. One can use those "public" functions to set up alternative bindings, or to extend the functionality.

- https://github.com/junegunn/fzf/issues/546#issuecomment-213891483
- https://github.com/junegunn/fzf/wiki/Configuring-shell-key-bindings#directly-executing-the-command-ctrl-x-ctrl-r

> Splitting the file into one-function-per-file

I like that key binding implementation for each shell is contained in a single file: key-bindings.{bash,zsh,fish}. Easier to browse, easier to install/uninstall. Taking advantage of autoloading of fish is of course nice and probably more idiomatic, but the practical benefit of it is questionable (load time? the difference should be hardly noticeable), so I'm leaning towards the consistency/tidiness in the project layout.

> the way fzf is meant to be loaded by fish just feels broken

"Feels broken" does not mean it's broken. For many users who have no issue with the current implementation and have no interest in implementation details, your change simply fixes what's not broken, and breaks what has been working fine for the past couple of years with no good reason without a single word of caution. One day they'll suddenly realize that the feature they have been using for years no longer works after an update, and have to figure out what happened and how to make it work again by themselves with no friendly documentation. I honestly don't see the point of all the trouble. If the user does care about the idiomatic implementation and do not like the default version, one can always choose more sophisticated, better structured third-party implementation. I don't see a problem with that.

> fisherman is not a solution to the binding problem, it is just merely a workaround. I believe this should be fixed upstream

Well, I see it differently. I see the key bindings this repository provides as a reference implementation, a starter-kit for the users who don't have enough knowledge, time, or passion to set up their own bindings. That's what I've been telling people on the issue tracker. Those key bindings are after all glorified personal configurations. They are meant to be forked and adapted to different tastes and needs. I'm not going to try to make them one-size-fits-all solutions.

So my suggestion is that we simply improve the existing functions (e.g. `read` instead of using temp files) without changing file structure, function names, and default bindings. If you don't agree with the direction, your time is better spent working on third-party implementations.

Ambrevar:
> Yes, they are a part of the project so why should they be any different?

You see the API as the whole package, user configuration included. To me the API is just fzf, not the config, which is why I think it's no big deal to change them.

Other programs who make use of fzf (Vim, Emacs, etc.) do not rely on the shell functions. (Or do they? I might be wrong here.)

I look at the project as a several distrinct bricks (in the "Unix" sense), you look at it more as a bundle, this is where our views differ.

Beside `fzf` is in version 0.x, isn't the right moment to make API changes?

> don't we have to consistently remove the default bindings from bash and zsh extensions as well?

Absolutely :)

> I like that key binding implementation for each shell is contained in a single file: key-bindings.{bash,zsh,fish}. Easier to browse, easier to install/uninstall. Taking advantage of autoloading of fish is of course nice and probably more idiomatic, but the practical benefit of it is questionable (load time? the difference should be hardly noticeable), so I'm leaning towards the consistency/tidiness in the project layout.

You are talking installation details here. "Easy to browse / install" is the packager's concern only (or the users who install it manually).

The file split not impact the API: when the packager updates the package, or when the user installs fzf manually, the changes are made at install time: it does not impact the user configuration.

The practical benefit is more than that: from a fish user point of view, fzf should provide an extension that is configured / loaded in some expected way: it's not, and it is actually a bit tricky to get it right.

The current loading process:

- Source `key-bindings.fish`. At this point, no `fzf`-widget is loaded. Maybe confusing?
- Run `fzf_key_bindings` from `fish_user_key_bindings` or else it won't work.
This both loads the functions and sets the bindings.
- Optional: change the bindings.

The idiomatic process:

- Set the bindings from `fish_user_key_bindings`.

In the idiomatic process, `fzf` widgets are automatically made public.

> Well, I see it differently. I see the key bindings this repository provides as a reference implementation, a starter-kit for the users who don't have enough knowledge, time, or passion to set up their own bindings.

I believe that users using fzf are power-users and can update their configuration following upstream changes.

And of course we can document the changes!

Talking documentation, if you insist on keeping the fish functions as it is, then we should document the configuration process as it is not trivial (as I detailed it above).

> That's what I've been telling people on the issue tracker. Those key bindings are after all glorified personal configurations. They are meant to be forked and adapted to different tastes and needs. I'm not going to try to make them one-size-fits-all solutions.

Isn't it a bit drastic to fork fzf just to change the default bindings? This is my whole point: making fzf a bit more universal.

> So my suggestion is that we simply improve the existing functions (e.g. read instead of using temp files) without changing file structure, function names, and default bindings.

I'll re-submit the changes then.

Ambrevar:
See #759.

junegunn:
Sorry for not responding quickly to your comment. I've been pretty busy lately and it's hard for me to digest more information about the tools that I don't actively use. Anyway, you should probably know better about fish than me, so I'll revisit this when I get some time.

Issue #672, opened 2016-09-26
Customizable Ctrl+R keybinding


jimmytheneutrino:
- Category
    - [x] Key bindings
- OS
    - [x] Linux
- Shell
    - [x] bash

Hi! As much as I adore fzf's ctrl+R behaviour, sometimes it gives too many results and requires more effort to get to the needed command than the original ctrl+R of bash.

I ended up binding fzf's ctrl+R to ctrl+F instead (by modifying key-bindings.bash) and use original ctrl+R and fzf's search interchangeably now. 

What do you think, 
1) is it possible to make this keybinding easily configurable
or maybe
2) just setting it to ctrl+F by default can be discussed?

junegunn:
A few things you might want to consider first:
1. Sorting is by default disabled. You can enable it by pressing CTRL-R again.
2. If you want to enable sorting by default, set `export FZF_CTRL_R_OPTS='--sort'`
3. Prefix each search term with `'` (single-quote) to turn it into an exact (not fuzzy) term.
4. Or you can even add `--exact` to FZF_CTRL_R_OPTS as well. Then prefixing a term with `'` will make it fuzzy term.

Although fzf does not provide an option to customize the key, you can access the definition with `bind -s`, so you can use the information to customize the binding.

```sh
[ -f ~/.fzf.bash ] && source ~/.fzf.bash

bind "$(bind -s | grep fzf_history | sed 's/r/f/')"
bind '"\C-r": reverse-search-history'
```

jimmytheneutrino:
Thanks! 

It seems --exact does work pretty much like default ctrl+r behaviour.

1)
One more thing though.
Bash's reverse-search-history immediately invokes the command on enter. 
Fzf copies the command to the command line instead. 
Is it configurable too?

2) 
I think it is worth putting such a snippet (with binding customization) in the readme.

junegunn:
1) There's an open issue for that: #477. I have a solution for zsh (enter / ctrl-x), but not for bash. If you think you can hack it up let me know.

2) Agreed. I'll add a page to the wiki when I get some time.

PR #749, opened 2016-11-29
Implement FZF_CTRL_R_COMMAND for custom history

ahal:
I'd like the ability to specify a custom history command to be able to use something like:
`fc -l 1 | grep ${(q)LBUFFER}`

to limit the results to history that exactly matches the full text entered on the command line.

### DIFF
--- a/shell/key-bindings.bash
+++ b/shell/key-bindings.bash
@@ -53,9 +53,10 @@ __fzf_cd__() {
 
 __fzf_history__() (
   local line
+  local cmd="${FZF_CTRL_R_COMMAND:-"history"}"
   shopt -u nocaseglob nocasematch
   line=$(
-    HISTTIMEFORMAT= history |
+    HISTTIMEFORMAT= eval "$cmd" |
     FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} $FZF_DEFAULT_OPTS +s --tac -n2..,.. --tiebreak=index --toggle-sort=ctrl-r $FZF_CTRL_R_OPTS +m" $(__fzfcmd) |
     command grep '^ *[0-9]') &&
     if [[ $- =~ H ]]; then
--- a/shell/key-bindings.fish
+++ b/shell/key-bindings.fish
@@ -43,10 +43,11 @@ function fzf_key_bindings
   end
 
   function fzf-history-widget -d "Show command history"
+    set -q FZF_CTRL_R_COMMAND; or set -l FZF_CTRL_R_COMMAND "history"
     set -q FZF_TMUX_HEIGHT; or set FZF_TMUX_HEIGHT 40%
     begin
       set -lx FZF_DEFAULT_OPTS "--height $FZF_TMUX_HEIGHT $FZF_DEFAULT_OPTS +s --tiebreak=index $FZF_CTRL_R_OPTS +m"
-      history | eval (__fzfcmd) -q '(commandline)' | read -l result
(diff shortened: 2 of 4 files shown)


junegunn:
Can you check failing test cases?

ahal:
I rebased and updated the test, is there anything else I need to do before this can land?

junegunn:
Thanks, but before we decide to add a new option, have you checked the comments in https://github.com/junegunn/fzf/issues/672? I've noticed that many users don't know that sorting is disabled by default and can be turned on, and exact matching can be used instead. Please check if the suggestions given in the comments addresses your use case.

The main reason I'm reluctant to add an option for tweaking the command for CTRL-R, is that the script processing the output of the fzf command expects the output to conform to a certain format, 
- in bash, each line should start with an index number, then the full command without anything else
- in zsh, each line should start with an index number, the rest doesn't matter
- in fish, each line should be the full command

So if a user uses any command that doesn't produce lines in those exact formats, the binding will not work. And it's the confusion I'd certainly like to avoid as the maintainer of the project.

ahal:
No, `--exact` doesn't quite do what I'm looking for, though it comes close. I'd like to force the input text to be an exact match, while keeping fuzzy search enabled for everything typed after that point. For example, say my history is:

```
1    cd foo
2    cd bar
3    ls abc/def
```

Now if I type `cd <ctrl-r>`, I would expect only the first two entries to appear in the list, but the third does as well because there is a fuzzy match for `cd`. I want an exact match on `cd`, but then still have the ability to fuzzy match the rest of the command.

This contrived example doesn't do a good job showing why this behaviour is so useful, but when your history has *many* commands in it that are hundreds of characters long, the probability of there being a fuzzy match for a given command (especially short ones that can be spelt in hexadecimal like `cd`), is very high. I find when I type `cd <ctrl-r>` my search list is dominated by vcs commands with commit hashes in them.

I think your concerns about CTRL_R_COMMAND are reasonable, and I could just re-defined the ctrl-r keybinding to my own custom function, but I believe this behaviour is actually quite useful and worth trying to upstream somehow. Maybe there's another way to achieve it that doesn't use a custom CTRL_R_COMMAND.

junegunn:
Yeah, I can definitely see the point.

Have you tried enabling sort by default? i.e. `export FZF_CTRL_R_OPTS="--sort"`

ahal:
Yeah, `--sort` improves things but it's still not perfect. The real world example it falls apart on for me is:

```
1    mach try foobar
2    mach decision --project=try
```

I want to fuzzy search only through the `mach try` commands, and `--sort` won't give `1` precedence over `2`. (For context, mach is a command dispatch binary used at mozilla, and try is the name of a repository). I'm probably an edge case here, so if you think `--exact` and `--sort` are good enough for most people and want to close this, I can live with re-defining `<ctrl-r>`.

ahal:
For the record, as a user redefining `CTRL_R_COMMAND`, I would have no expectation of the main project maintaining backwards compatibility with weird and custom workflow.

junegunn:
> 1    mach try foobar
> 2    mach decision --project=try

In this particular case, `machtry` (without space) will give you better result, because of the smaller gap between the words. When you search with `mach try`, fzf will search for the words individually, so it does not consider the gap.

Anyway, I'm not against the idea of having `FZF_CTRL_R_COMMAND` per se, but I'd like to delay the addition until the inconsistency between the shells is ironed out.

ahal:
That's fair, I'll re-bind `ctrl-r` in the meantime.

junegunn:
Okay, if you don't mind, I'll leave this open.

PR #753, opened 2016-12-01
feat(key-bindings-zsh): ALT-SHIFT-C to cd into first child directory

yannvr:
Navigate straight into next directory with ALT-SHIFT-C. No need for ```cd``` anymore!
  
Note that it also contains ALT bindkey fixes (inclduing ALT-C)  to support OSX

### DIFF
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -44,6 +44,24 @@ fzf-cd-widget() {
 }
 zle     -N    fzf-cd-widget
 bindkey '\ec' fzf-cd-widget
+# OSX support
+bindkey 'ç' fzf-cd-widget
+
+# ALT-SHIFT-C - cd into the first child directory
+fzf-cd1-widget() {
+  local cmd="${FZF_ALT_C_COMMAND:-"command find -L . \\( -path '*/\\.*' -o -fstype 'dev' -o -fstype 'proc' \\) -prune \
+    -o -type d -maxdepth 1 -print 2> /dev/null | sed 1d | cut -b3-"}"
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks, but I decided not to add more key bindings to this repo due to maintenance burden. If I decided to add a key binding, I have to make sure that the same key works consistently across all three supported shells (bash, zsh, and fish) and then add more configuration variables for the new key binding, such as `$FZF_ALT_C1_OPTS`, `$FZF_ALT_C1_COMMAND` also with test cases for different set of options. That's already a lot of work and once added, there's no going back. I'll have to maintain those new features forever unless we decide to break backward compatibility at some point.

You know that you can simply set up `$FZF_ALT_C_COMMAND` to customize the default behavior of ALT-C, right? Well, the obvious downside of the approach is that you can't easily switch between the two behaviors.

So you can define a wrapper widget that reuses the code of fzf-cd-widget.

```sh
fzf-cd1-widget() {
  FZF_ALT_C_COMMAND="command find -L . \\( -path '*/\\.*' -o -fstype 'devfs' -o -fstype 'devtmpfs' -o -fstype 'proc' \\) -prune \
     -o -type d -maxdepth 1 -print 2> /dev/null | sed 1d | cut -b3-" fzf-cd-widget
}
zle     -N    fzf-cd1-widget
bindkey '\eC' fzf-cd1-widget
```

Regarding `bindkey 'ç' fzf-cd-widget` thing, many users including myself configure iterm2 to make option key send escape sequence. See https://www.iterm2.com/faq.html

Issue #432, opened 2015-12-03
Ctr-r shuld use the current command as query as default Ctr-r do in bash


leiserfg:
It's my implementation  on fish

``` fish
  function __fzf_ctrl_r
    set -l query (commandline|string trim)
    if test -z $query
      set query ''
    else
      set query "-q $query"
    end

    history | eval (__fzfcmd) +s +m --tiebreak=index --toggle-sort=ctrl-r  $query | read -l out
    and commandline $out
    commandline -f repaint
  end
```

junegunn:
So you're suggesting we make the behavior across all three shells, right? Unfortunately I haven't been able to find a way to achieve that in bash.

junegunn:
I applied your patch like follow

```diff
diff --git a/shell/key-bindings.fish b/shell/key-bindings.fish
index 3d3c2a7..36cbf17 100644
--- a/shell/key-bindings.fish
+++ b/shell/key-bindings.fish
@@ -26,7 +26,14 @@ function fzf_key_bindings
   end
 
   function __fzf_ctrl_r
-    history | eval (__fzfcmd) +s +m --tiebreak=index --toggle-sort=ctrl-r > $TMPDIR/fzf.result
+    set -l query (commandline|string trim)
+    if test -z $query
+      set query ''
+    else
+      set query "-q $query"
+    end
+
+    history | eval (__fzfcmd) +s +m --tiebreak=index --toggle-sort=ctrl-r $query > $TMPDIR/fzf.result
     and commandline (cat $TMPDIR/fzf.result)
     commandline -f repaint
     rm -f $TMPDIR/fzf.result
```

But I'm getting 

```
fish: Unknown command 'string'
~/.config/fish/functions/fish_user_key_bindings.fish (line 1): commandline|string trim
```

Maybe I did something wrong, I don't know fish too well.

leiserfg:
If you are using a fish <= fish2.2.0 (not an development version like me) you have not the string builtin yet (sorry) so you can replace it:
```fish
set -l query (commandline|string trim)
```
with:
```fish
set -l query (commandline|xargs) #xargs alone do the trimming work
```

junegunn:
I noticed the patch doesn't work with non-trivial string on the command prompt.

e.g. `seq 10 | wc -l<CTRL-R>` or `echo $PATH<CTRL-R>`

leiserfg:
[...]

Issue #286, opened 2015-07-07
[bash] <Ctrl-R> not working


jpcirrus:
Using bash with either `set +o histexpand` or `set +H` to turn off csh style history substitution, fzf no longer expands the returned history item since b4a05ff. This is the same issue as with zsh in #214. Possibly a similar fix could be made by testing for `H` in the bash variable `$-`, something like:

    [[ $- == *H* ]]; oldH=$?  # Save -H state
    set -H                    # Enable ! history substitution
    # history expansion code
    (( $oldH )) && set +H     # Reset -H if necessary
    unset -v oldH             # Clean up

assuming `shopt -u nocaseglob`

junegunn:
Thanks for the report. For zsh (#214), the code was later updated again to use `vi-fetch-history` instead of temporarily setting the option (see #272). I wonder if a similar trick is available for bash.

jpcirrus:
`vi-fetch-history` is available in bash's readline, but in quick testing I couldn't get it to accept the history number as an argument. Interactively, with `set -o vi`, entering command mode, the history number and then `G` (it's default key-binding) it works.

junegunn:
Hmm, that's unfortunate. Then I currently have no idea how we should restore the setting since in order to do that, we'll have to invoke another `shell-expand-line` which will cause unwanted expansion of the pasted command (e.g. `echo $PATH`)

We could fall back to the old method used before b4a05ff27ed8b1371dc3745ce02f48430345761b only when `histexpand` is not set, but then we'll not be able to handle multi-line commands.

jpcirrus:
I'll have another look at trying to get `vi-fetch-history`working over the weekend when I have some time. Will come back with the result.

jpcirrus:
I haven't been able to get `vi-fetch-history` to work in emacs mode, nor been able to set`histexpand`  without polluting the environment, so I think your suggestion when `histexpand` is not set to revert back to the behaviour prior to b4a05ff is the way to go, as at least recalling single-line history commands will work, which doesn't currently.

junegunn:
Okay, thanks for looking into it. So what do you think about the following? Seems to do the trick.

```sh
__fzf_history__() {
  local line
  line=$(
    HISTTIMEFORMAT= history |
    $(__fzfcmd) +s --tac +m -n2..,.. --tiebreak=index --toggle-sort=ctrl-r |
    \grep '^ *[0-9]') &&
    if [[ $- =~ H ]]; then
[...]

PR #757, opened 2016-12-03
bash keybinding: C-r remembers typed text.

### LINKED ISSUE #731: Start the history search with the word you typed at the prompt in bash
- Category
    - [ ] fzf binary
    - [ ] fzf-tmux script
    - [x] Key bindings
    - [ ] Completion
    - [ ] Vim
    - [ ] Neovim
    - [ ] Etc.
- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [x] bash
    - [ ] zsh
    - [ ] fish

I use hh and find it very useful to start the history search with the word that I typed before pressing `Ctrl+R`.

The hh command is defined as follows:
```
if [[$ - = ~. * I. *]]; Then bind ' "\ C-r": "\ C-a hh \ C-j"'; fi
```

I could do the same way with fzf overwriting the binding, but I would like to know if it is possible to get at this result using only the settings, without the need to override the binding.

acornejo:
This modifies the bash fzf bindings so that the C-r binding remebers the
text previously typed into the prompt. This should close #731 and #432.

### DIFF
--- a/shell/key-bindings.bash
+++ b/shell/key-bindings.bash
@@ -51,14 +51,15 @@ __fzf_history__() (
   shopt -u nocaseglob nocasematch
   line=$(
     HISTTIMEFORMAT= history |
-    eval "$(__fzfcmd) +s --tac +m -n2..,.. --tiebreak=index --toggle-sort=ctrl-r $FZF_CTRL_R_OPTS" |
-    command grep '^ *[0-9]') &&
-    if [[ $- =~ H ]]; then
-      sed 's/^ *\([0-9]*\)\** .*/!\1/' <<< "$line"
-    else
-      sed 's/^ *\([0-9]*\)\** *//' <<< "$line"
-    fi
+    eval "$(__fzfcmd) +s --tac +m -n2..,.. --tiebreak=index --toggle-sort=ctrl-r $FZF_CTRL_R_OPTS --query=$1" |
+    command grep '^ *[0-9]') && sed 's/^ *\([0-9]*\)\** *//' <<< "$line"
 )
+
+fzf-history-widget() {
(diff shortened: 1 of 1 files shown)


shell/key-bindings.bash
```diff
@@ -51,15 +51,16 @@ __fzf_history__() (
...
     HISTTIMEFORMAT= history |
-    eval "$(__fzfcmd) +s --tac +m -n2..,.. --tiebreak=index --toggle-sort=ctrl-r $FZF_CTRL_R_OPTS" |
-    command grep '^ *[0-9]') &&
-    if [[ $- =~ H ]]; then
-      sed 's/^ *\([0-9]*\)\** .*/!\1/' <<< "$line"
-    else
-      sed 's/^ *\([0-9]*\)\** *//' <<< "$line"
-    fi
+    eval "$(__fzfcmd) +s --tac +m -n2..,.. --tiebreak=index --toggle-sort=ctrl-r $FZF_CTRL_R_OPTS --query=$1" |
+    command grep '^ *[0-9]') && sed 's/^ *\([0-9]*\)\** *//' <<< "$line"
```
junegunn: Does this work with multi-line commands?

- https://github.com/junegunn/fzf/commit/b4a05ff27ed8b1371dc3745ce02f48430345761b
- https://github.com/junegunn/fzf/issues/286

acornejo: good catch, I don't think it does. Never noticed since I don't use those frequently. Closing this PR.

junegunn: I was wondering why you removed the conditional branch when it seems unrelated to the suggested change. Anyway, I like the idea. It would be nice if we could implement it with the other shells as well.

PR #759, opened 2016-12-05
Overhaul fish functions

Ambrevar:
Replace the "temp file" workaround with the "read" function: it's
simpler and faster.

Use proper escaping, remove the custom function.

The "file" widget uses last token as root for the "find" command.
This replaces the equivalent of '**' completion in bash/zsh.
The "$cwd" non-expanded variable can be used in FZF_CTRL_T_COMMAND to
set the root.

This closes #443, #484, #486 and #747.

### DIFF
--- a/README.md
+++ b/README.md
@@ -434,11 +434,36 @@ export FZF_DEFAULT_COMMAND='
 
 It's [a known bug of fish](https://github.com/fish-shell/fish-shell/issues/1362)
 that it doesn't allow reading from STDIN in command substitution, which means
-simple `vim (fzf)` won't work as expected. The workaround is to store the result
-of fzf to a temporary file.
+simple `vim (fzf)` won't work as expected. The workaround is to use the `read`
+fish command:
 
 ```sh
-fzf > $TMPDIR/fzf.result; and vim (cat $TMPDIR/fzf.result)
+fzf | read -l result; and vim $result
+```
+
+or, for multiple results:
+
+```sh
+fzf -m | while read -l r; set result $result $r; end; and vim $result
+```
+
+The globbing system is different in fish and thus `**` completion will not work.
+However, the `CTRL-T` command will use the last token on the commandline as the
+root folder for the recursive search. For instance, hitting `CTRL-T` at the end
+of the following commandline
+
+```sh
+ls /var/
+```
+
+will list all files and folders under `/var/`.
+
+When using a custom `FZF_CTRL_T_COMMAND`, use the unexpanded `$dir` variable to
+make use of this feature. `$dir` defaults to `.` when the last token is not a
+valid directory. Example:
+
+```sh
+set -l FZF_CTRL_T_COMMAND "command find -L \$dir -type f 2> /dev/null | sed '1d; s#^\./##'"
 ```
 
 License
--- a/shell/key-bindings.fish
+++ b/shell/key-bindings.fish
@@ -1,52 +1,60 @@
 # Key bindings
 # ------------
 function fzf_key_bindings
-  # Due to a bug of fish, we cannot use command substitution,
-  # so we use temporary file instead
-  if [ -z "$TMPDIR" ]
-    set -g TMPDIR /tmp
-  end
 
-  function __fzf_escape
-    while read item
-      echo -n (echo -n "$item" | sed -E 's/([ "$~'\''([{<>})])/\\\\\\1/g')' '
+  # Store last token in $dir as root for the 'find' command
+  function fzf-file-widget -d "List files and folders"
+    set -l dir (commandline -t)
+    # The commandline token might be escaped, we need to unescape it.
+    set dir (eval "printf '%s' $dir")
+    if [ ! -d "$dir" ]
+      set dir .
     end
-  end
+    # Some 'find' versions print undesired duplicated slashes if the path ends with slashes.
+    set dir (string replace --regex '(.)/+$' '$1' "$dir")
 
-  function fzf-file-widget
+    # "-path \$dir'*/\\.*'" matches hidden files/folders inside $dir but not
+    # $dir itself, even if hidden.
     set -q FZF_CTRL_T_COMMAND; or set -l FZF_CTRL_T_COMMAND "
-    command find -L . \\( -path '*/\\.*' -o -fstype 'devfs' -o -fstype 'devtmpfs' -o -fstype 'proc' \\) -prune \
-      -o -type f -print \
-      -o -type d -print \
-      -o -type l -print 2> /dev/null | sed 1d | cut -b3-"
-    eval "$FZF_CTRL_T_COMMAND | "(__fzfcmd)" -m $FZF_CTRL_T_OPTS > $TMPDIR/fzf.result"
-    and for i in (seq 20); commandline -i (cat $TMPDIR/fzf.result | __fzf_escape) 2> /dev/null; and break; sleep 0.1; end
+    command find -L \$dir \\( -path \$dir'*/\\.*' -o -fstype 'devfs' -o -fstype 'devtmpfs' \\) -prune \
+    -o -type f -print \
+    -o -type d -print \
+    -o -type l -print 2> /dev/null | sed '1d; s#^\./##'"
+
+    eval "$FZF_CTRL_T_COMMAND | "(__fzfcmd)" -m $FZF_CTRL_T_OPTS" | while read -l r; set result $result $r; end
+    if [ -z "$result" ]
+      commandline -f repaint
+      return
+    end
+
+    if [ "$dir" != . ]
+      # Remove last token from commandline.
+      commandline -t ""
+    end
+    for i in $result
+      commandline -it -- (string escape $i)
+      commandline -it -- ' '
+    end
     commandline -f repaint
-    rm -f $TMPDIR/fzf.result
   end
 
-  function fzf-history-widget
-    history | eval (__fzfcmd) +s +m --tiebreak=index --toggle-sort=ctrl-r $FZF_CTRL_R_OPTS -q '(commandline)' > $TMPDIR/fzf.result
-    and commandline -- (cat $TMPDIR/fzf.result)
+  function fzf-history-widget -d "Show command history"
+    history | eval (__fzfcmd) +s +m --tiebreak=index $FZF_CTRL_R_OPTS -q '(commandline)' | read -l result
+    and commandline -- $result
     commandline -f repaint
-    rm -f $TMPDIR/fzf.result
   end
 
-  function fzf-cd-widget
+  function fzf-cd-widget -d "Change directory"
     set -q FZF_ALT_C_COMMAND; or set -l FZF_ALT_C_COMMAND "
-    command find -L . \\( -path '*/\\.*' -o -fstype 'devfs' -o -fstype 'devtmpfs' -o -fstype 'proc' \\) -prune \
-      -o -type d -print 2> /dev/null | sed 1d | cut -b3-"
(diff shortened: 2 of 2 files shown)


Ambrevar:
CTRL_T test is failing. Could it come from that my patch uses the last token as root folder for find?
I've made this change to fill the lack of support for '**'.

junegunn:
Thanks for the follow-up. I'll review the code and get back to you.

junegunn:
The test passes locally, but I noticed that I had to upgrade my fish to make it work. Its probably related. Let me check.

junegunn:
The test passed on the second run. Sorry for the confusion. Some test cases are based on timeout so they are known to be flaky on Travis CI environment.

Can you respond to the comment I left above? I'll merge this after we address the issue. Thanks.

Ambrevar:
Agreed. I've rebased the change accordingly.

Ambrevar:
Talking tests, I'm having a hard time running time / understanding them. A few questions:

- Would the timeout thing be an issue on a very slow computer?
- Is there any documentation? Might be a stupid question, but I don't know Ruby at all.
- Any plan for a full switch to Go? That might simplify the whole suite.

junegunn:
> Would the timeout thing be an issue on a very slow computer?

The test script sends keystrokes to a tmux pane running fzf and waits for the expected output to appear on the pane. So yeah, it's possible. There can be other factors that might be causing flakiness of the tests. I've been trying to make the tests more robust (e.g. making sure that the driver sends keystrokes when the target pane is ready), and they almost always work fine on my machine, but I'm still seeing some issues on Travis CI from time to time. Interestingly, most failures come from the test cases for fish, so I vaguely suspected that it's partly related to the way fish behaves, like unreliable consumption of keystrokes. Maybe or maybe not.

> Is there any documentation? 

Briefly described in https://github.com/junegunn/fzf/tree/master/src#test, but `ruby test/test_go.rb`is all you need to know to run the suite (for a single test `ruby test/test_go.rb --name test_ctrl_t`). Make sure to run it while you're on tmux.

> Any plan for a full switch to Go? That might simplify the whole suite.

Go tends to be much more verbose than Ruby, so I don't want to bother rewriting the existing test cases in Go.


Anyway, thanks for the update. A minor issue I see is that if I press CTRL-T after a path ending with `/`, the entries in the list have double slashes.

```sh
vim /tmp/<CTRL-T>

  /tmp//foo
  /tmp//bar
```

Do you think it makes sense that we trim the extra slashes? (FYI https://github.com/junegunn/fzf/blob/16b5902aa278639fd99be89abe8ac50eb49b34aa/shell/completion.bash#L134)

Ambrevar:
Actually I've noticed that the `sed` suggestion is implemented in zsh/bash as a `cut -b3-` which spawns one more process compared to the `sed` solution. I'll fix that for all implementations at once.

Ambrevar:
Regarding the double-slash issue, I cannot reproduce... ? Let me investigate.

junegunn:
screen shot 2016-12-10 at 4 24 25 pm

This is what it looks like.

Ambrevar:
Got it: it's because of the cut! The `./` are not the leading characters anymore in that case...

I'm going to revert the `cut` change: the leading `./` should remain to let the user know which root was used in `find`. I'll add a comment to make this clear.

Ambrevar:
Still can't reproduce the double-slash issue after investigation. What's in your CTRL_T_COMMAND?

junegunn:
> What's in your CTRL_T_COMMAND?

It's not set.

> the leading ./ should remain to let the user know which root was used in find. If the root is not a 1 letter folder, its name gets trimmed. 

Hmm, I don't understand the problem you describe. The updated sed command only removes leading `./` not anything else.

Ambrevar:
Sorry, I mixed up my own words.

1. The `sed` command is correct. I was wondering if it would not be confusing, but after second thought, I don't think so.

2. `./` is cut with `cut -b3-` in other widgets and in bash/zsh. When piped over a `sed`, it spawns an additional process that could have been spared. I suggest we replace all the `sed 1d | cut -b3-` with `sed '1d; s#^\./##'`. Can do it later though.

shell/key-bindings.fish
```diff
@@ -1,52 +1,56 @@
...
-  function __fzf_escape
-    while read item
-      echo -n (echo -n "$item" | sed -E 's/([ "$~'\''([{<>})])/\\\\\\1/g')' '
+  # Use last token $cwd_esc as root for the 'find' command
+  function fzf-file-widget -d "List files and folders"
+    set -l cwd (commandline -t)
+    # The commandline token might be escaped, we need to unescape it.
+    set cwd (eval "printf '%s' $cwd")
+    if [ ! -d "$cwd" ]
+      set cwd .
```
Ambrevar: I still don't understand why you would need to do that: $cwd is used as a root for find, it is never passed to fzf, and thus the trailing slash should not matter.

Concretely, in the fzf project folder from the example on your screenshot, would you get different results from those two?

	command find -L shell/ -path '*/\\.*' -prune -o -type f -print -o -type d -print -o -type l -print
	command find -L shell -path '*/\\.*' -prune -o -type f -print -o -type d -print -o -type l -print

I don't.
(I've simplified the find command a bit, not that matters.)

junegunn: Yes, I get different results. Note that I use find command from macOS. I don't have the issue if I use gnu find from coreutils package.

Ambrevar: Oh! This is one portability issue I wasn't aware about!
Alright then, I'll use your suggested fix and add a comment.

Ambrevar:
Pushed the changes, updated the README and some comments.

Ambrevar:
FYI, I realized that GNU find automatically removes the trailing slash if there is only one, not if there is more than 1. So `find shell//` would result in duplicated slashes as well.

shell/key-bindings.fish
```diff
@@ -1,52 +1,61 @@
...
+    else if [ "$cwd" != '/' ]
+      # Some 'find' versions print undesired duplicated slashes if the path ends with slashes.
+      set cwd (string replace --regex '/+$' '' "$cwd")
```
junegunn: This unfortunately does not work with `ls //<CTRL-T>` though it's silly to type in `//` or `///`. So I think it's enough that we just remove one slash at the end (which is usually the result of tab completion).

README.md
```diff
@@ -434,11 +434,35 @@ export FZF_DEFAULT_COMMAND='
...
 
 ```sh
-fzf > $TMPDIR/fzf.result; and vim (cat $TMPDIR/fzf.result)
+fzf | read -l result; and vim $result
+```
+
+or, for multiple results:
+
+```sh
+fzf | while read -l r; set result $result $r; end; and vim $result
```
junegunn: I think it will be clearer if we add `-m` option to fzf here.

shell/key-bindings.fish
Ambrevar: We can also fix that with

	string replace --regex './+$' '$1' "$cwd"

griwes: Just FYI, `//` is *technically* different from `/` and `///` (and `////` and `/////` and so on): http://unix.stackexchange.com/a/12291

griwes: I'm not saying that it *needs* to be supported, and I do agree that it's ill-designed. Just dropping a piece of information that is relevant, so for example later on when someone reports a bug, you can just point to the above comment and say "it's ignored by design".

Do merge this soon! :D

Ambrevar: Of course, that's why I said "interesting". Did not mean to discard your comment at all. If that is what you understood, sorry for that.

junegunn: I'm okay with either way as long as the issue is properly addressed. By the way, is the command correct? I believe you meant `(.)/+$`

Ambrevar: Indeed!

README.md
junegunn: Please add `-m` to the fzf command so that it's clear that it can produce multiple lines.

README.md
```diff
@@ -434,11 +434,35 @@ export FZF_DEFAULT_COMMAND='
...
+ls /var/
+```
+
+will list all files and folders under `/var/`.
+
+When using a custom `FZF_CTRL_T_COMMAND`, use the unexpanded `$cwd` variable to
+make use of this feature. Example:
+
+```sh
+set -l FZF_CTRL_T_COMMAND "command find -L \$cwd -type f 2> /dev/null | sed '1d; s#^\./##'"
```
junegunn: I propose to use `$dir` instead of `$cwd`. "CWD" as far as I can tell is a short for "current working directory", so it can be confusing. We should also mention that `$dir` defaults to `.` when the last token is not a valid directory.

Ambrevar: Agreed.

junegunn:
Merged, thanks for your patience and persistence!

amosbird:
hmm, does this PR achieve what #484 needs? I'm trying to use `fish/z` plus `fzf` via fuzzy completion triggering. I still can't see a way to do it.

Ambrevar:
Not sure I quite understand your scenario.
Say you want to list files under foo/bar, just press the file-widget binding after you've written `foo/bar`. Note that the directory at point _must_ exist.

amosbird:
well, this isn't about file-widget IMO. Say I want to invoke `fzf` like this `kill <Tab>`

Ambrevar:
That's command completion, not file completion, or is it?
You'll find a completion widget here:
https://github.com/junegunn/fzf/wiki/Examples-(fish)

It works well except for variable expansion and a few other edge cases. That's a fish bug that will hopefully get fixed one day.
https://github.com/fish-shell/fish-shell/issues/3469

amosbird:
cool, thank you.

shell/key-bindings.fish
```diff
@@ -1,52 +1,60 @@
...
+    for i in $result
+      commandline -it -- (string escape $i)
+      commandline -it -- ' '
+    end
     commandline -f repaint
-    rm -f $TMPDIR/fzf.result
   end
 
-  function fzf-history-widget
-    history | eval (__fzfcmd) +s +m --tiebreak=index --toggle-sort=ctrl-r $FZF_CTRL_R_OPTS -q '(commandline)' > $TMPDIR/fzf.result
```
adriaanzon: Why was `--toggle-sort=ctrl-r` removed here?

junegunn: @adriaanzon Ugh, you're right. It was our oversight. Let me put it back.

hrouault:
Using the typed path as a root for a Ctrl-T search is really a great feature.
Would it be possible to have the very same feature for the cd-widget i.e. if I type `./foo/` and then alt-c, I would fzf folders `./foo/*`?

Ambrevar:
I like the idea. I'll submit a PR.

Ambrevar:
Thinking about it, what would be the use case?
Does it allow anything more than

    $ ./foo/<ENTER>
    $ <ALT-C>

?

hrouault:
It is often that I need to complete arguments that I know are folders. Then, I do not want to change to that folder.

Ambrevar:
Alt-C is bound to fzf-cd-widget: it is meant for changing directory, not completing them. I think your suggestion would deviate to much from the original intention.

You can easily implement that yourself in your configuration:
- Copy `fzf-file-widget` to a `file-folder-widget` in your fish configuration.
- Modify the new widget to keep folders only.
- Bind the widget to your favorite shortcut.

hrouault:
It makes sense. Thank you.

PR #771, opened 2016-12-14
Append '/' to folders in the file widget

Ambrevar:
This helps differentiating folders from files.

Besdies, writing a folder with an appended slash on the commandline is
usually what the user wants: the user can keep on completing/fzf-ing, it
saves one key stroke.

### DIFF
--- a/shell/key-bindings.bash
+++ b/shell/key-bindings.bash
@@ -1,10 +1,10 @@
 # Key bindings
 # ------------
 __fzf_select__() {
-  local cmd="${FZF_CTRL_T_COMMAND:-"command find -L . \\( -path '*/\\.*' -o -fstype 'devfs' -o -fstype 'devtmpfs' -o -fstype 'proc' \\) -prune \
+  local cmd="${FZF_CTRL_T_COMMAND:-"command find -L . -mindepth 1 \\( -path '*/\\.*' -o -fstype 'devfs' -o -fstype 'devtmpfs' -o -fstype 'proc' \\) -prune \
     -o -type f -print \
-    -o -type d -print \
-    -o -type l -print 2> /dev/null | sed 1d | cut -b3-"}"
+    -o -type d -exec printf '%s/\n' {} + \
+    -o -type l -print 2> /dev/null | cut -b3-"}"
   eval "$cmd | fzf -m $FZF_CTRL_T_OPTS" | while read -r item; do
     printf '%q ' "$item"
   done
--- a/shell/key-bindings.fish
+++ b/shell/key-bindings.fish
@@ -16,10 +16,10 @@ function fzf_key_bindings
     # "-path \$dir'*/\\.*'" matches hidden files/folders inside $dir but not
     # $dir itself, even if hidden.
     set -q FZF_CTRL_T_COMMAND; or set -l FZF_CTRL_T_COMMAND "
-    command find -L \$dir \\( -path \$dir'*/\\.*' -o -fstype 'devfs' -o -fstype 'devtmpfs' \\) -prune \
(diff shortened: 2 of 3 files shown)


junegunn:
Thanks, but there are actually failing test cases (`ruby test/test_go.rb --name test_ctrl_t_unicode`). Looks like the new command changes the order of the output which makes `sed 1d` delete the wrong line.

```
> command find -L . -type f -print -o -type d -print 2> /dev/null
.
./fzf-unicode 테스트1
./fzf-unicode 테스트2

> command find -L . -type f -print -o -type d -exec printf '%s/\n' {} + 2> /dev/null
./fzf-unicode 테스트1
./fzf-unicode 테스트2
./
```

Ambrevar:
Same behaviour with GNU find. Bummer!

From my understanding, `find` is supposed to perform a breadth-first-search (unless the `-depth` flag is used) following the filesystem structure. From run to run it should be the consistent, independently of the output formatting. So the root should always be the first result when we print folder. I am not sure what is happening here, I may have misunderstood something. Let's investigate.

Ambrevar:
OK, got it, that was quite obvious actually: I'm using the `+` argument, which means "all folders" here. So no folder (only files) gets printed till the search finishes, at which point the folder list gets passed to `printf`.

Easy fix: use `;` instead. This will spawn one printf per folder, maybe costly.
Other fix: filter out the root in the first place, so that the `sed` call isn't needed anymore.

Ambrevar:
I believe I've fixed the issue with `-mindepth 1`.

junegunn:
Thanks for looking into it.

> So no folder (only files) gets printed till the search finishes

Hmm, I'm not sure if I like that aspect. I sometimes use CTRL-T just to see the structure of a repository and directories not showing up at the right place but at the end of the list doesn't feel ideal.

screen shot 2016-12-17 at 10 08 44 am

In the above image, I prefer the output on the left. Alas, it would have been much simpler if macOS find supported `-printf`. If there's no simple way to have the right ordering, I would just suggest `-printf` option on the wiki page rather than changing the default.

Ambrevar:
Absolutely.

Last question, benchmarking one: would executing a `printf` per folder be too costly? (The system command) I tend to think so. If not, that would be a solution.

Also keep in mind that the output order of `find` is not supposed to be reliable. That you get what you expect in the above picture is mostly because this is the most practical way to "find" on that filesystem. Some filesystems might return very different results. But I agree, it is _generally_ more intuitive that way than leaving the folder to the end.

Let me know about the benchmark. Either way, I'll submit the `-mindepth 1` patch.

Ambrevar:
Further thoughts: this whole issue could be superseded by an optional coloring feature akin to `ls` or `ranger`: each line would be colored according to which regexp it matches or something like that.

Folders would need a trailing slash for that to work though.

junegunn:
Hmm, not looking good. It shouldn't be the default.

```sh
jg@jg-mbp:~/.vim> time find . -type d -print | wc -l
    4252

real    0m0.089s
user    0m0.013s
sys     0m0.079s
jg@jg-mbp:~/.vim> time find . -type d -exec printf '%s/\n' {} \; | wc -l
    4252

real    0m6.914s
user    0m2.576s
sys     0m4.036s
```

Ambrevar:
I've PRed the `-mindepth` fix: #779. Feel free to close this PR when you've added the `printf` trick to the wiki.

junegunn:
Thanks. Hmm, the default CTRL-T command is quite lengthy, it doesn't feel so elegant to have to write down such a long, multi-line command just for a little tweak. As you have mentioned, it would be nice if there's a simpler tool that can be used as an alternative, with ANSI colors based on file extension, `/` for directories, etc. I'll update the wiki page when I find one.

PR #776, opened 2016-12-18
s:execute_term: switch_back: check that self.pbuf exists

blueyed:
With a `bufhidden=wipe` buffer (e.g. vim-startify) the buffer would not
exist anymore, resulting in an error.


junegunn:
Makes sense. Thanks.

PR #779, opened 2016-12-23
shell: Use '-mindepth 1' to omit root folder in 'find' output

Ambrevar:
This removes the need for the 'sed' call. Faster, cleaner.


junegunn:
Merged, thanks!

PR #784, opened 2016-12-30
Support half page scrolling using Emacs key bindings.

amosbird:


junegunn:
Thanks for the suggestion. Are ctrl-v and alt-v standard bindings in Emacs? If it's not the case, I don't want to provide those bindings by default. Instead, we add those actions, and let the users set up their own bindings using `--bind`.

amosbird:
It's the default binding in emacs.

junegunn:
Half-page or whole-page? If those keys are for whole-page movement in Emacs, there's no reason we should bind them differently in fzf.

amosbird:
Oh, I remapped them a long time ago and it becomes my builtin default 😆 . My bad.

junegunn:
I see, then we can provide unbound `half-page-up` and `half-page-down` actions and let the users bind those actions to the keys of their choices via `$FZF_DEFAULT_OPTS`.

man/man1/fzf.1
```diff
@@ -444,6 +444,8 @@ e.g. \fBfzf --bind=ctrl-j:accept,ctrl-k:kill-line\fR
     \fBnext-history\fR          (\fIctrl-n\fR on \fB--history\fR)
     \fBpage-down\fR             \fIpgdn\fR
     \fBpage-up\fR               \fIpgup\fR
+    \fBhalfpage-down\fR
+    \fBhalfpage-up\fR
```
junegunn: Please use `half-page` instead of `halfpage`.

src/options.go
```diff
@@ -667,6 +667,10 @@ func parseKeymap(keymap map[int]actionType, execmap map[int]string, str string)
 			keymap[key] = actPageUp
 		case "page-down":
 			keymap[key] = actPageDown
+		case "halfpage-up":
```
junegunn: Same here.

src/terminal.go
```diff
@@ -1413,6 +1415,12 @@ func (t *Terminal) Loop() {
 			case actPageDown:
 				t.vmove(-(t.maxItems() - 1))
 				req(reqList)
+			case actHalfPageUp:
+				t.vmove(t.maxItems()/2 - 1)
+				req(reqList)
+			case actHalfPageDown:
+				t.vmove(-(t.maxItems()/2 - 1))
```
junegunn: `- 1` doesn't seem necessary here. The point of `t.maxItems() - 1` is to keep the scroll offset unchanged when you press pgup for the first time. It doesn't apply for half page movements.

amosbird:
updated.

junegunn:
Oops, I forgot to include this in the new release. I'll make sure it's included in the next version. Sorry for the delay.

junegunn:
Released 0.16.1 with your path. Thanks.

PR #785, opened 2016-12-31
[vim] Use cmd.exe in Windows

janlazo:
When running Vim on Windows,
- use cmd.exe as launcher and shell
- enable shellslash within s:cmd()
- don't escape ' ' and '\\' for paths

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -80,7 +80,13 @@ function! s:shellesc(arg)
 endfunction
 
 function! s:escape(path)
-  return escape(a:path, ' $%#''"\')
+  let l:escaped_chars = '$%#''"'
+
+  if has('unix')
+    let l:escaped_chars .= ' \'
+  endif
+
+  return escape(a:path, l:escaped_chars)
 endfunction
 
 " Upgrade legacy options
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks. Let me review the commit. By the way, we don't use `l:` prefix for local variables. Can you remove the prefixes for consistency?

junegunn:
I haven't tried running Vim plugin for fzf on Windows, and never expected it to work correctly. Can you share your overall experience?

janlazo:
With this PR, it works on file paths if without spaces or not network drives (it breaks on `C:\` ).
It works fine if running fzf, via the Vim plugin, within the home directory (because of `~` expansion).
`:FZF` redraws the entire screen for cmd.exe to run fzf and it behaves as expected when selecting files/directories (even with spaces) to open.

To run terminal Vim, I use ConEmu with Powershell (to minimize file path issues).
I used cmd.exe in the PR because Powershell has long startup time.

janlazo:
For now, in this PR, `FZF` and `FZF!` commands behave the same on terminal Vim in Windows because I did not set the geometry for cmd.exe similar to the xterm_launcher.

junegunn:
Thanks for the explanation. Merged, thanks!

PR #786, opened 2017-01-01
Vim plugin can use Windows filepaths with spaces

janlazo:
This is an update on my previous PR for Vim plugin on Windows #785.
This PR enables terminal Vim and Gvim to run fzf on filepaths with spaces.
It uses `noshellslash` on both `s:cmd()` and `fzf#run()`  for strict filename expansion on both.
It ensures that shellescape() works as expected for fzf.
For the launcher, the fzf command is double-quoted for cmd.exe. 

Tested on Windows 8.1, (g)vim 8, and Con Emu + Powershell (to run terminal vim).

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -80,7 +80,13 @@ function! s:shellesc(arg)
 endfunction
 
 function! s:escape(path)
-  return escape(a:path, ' $%#''"\')
+  let l:escaped_chars = '$%#''"'
+
+  if has('unix')
+    let l:escaped_chars .= ' \'
+  endif
+
+  return escape(a:path, l:escaped_chars)
 endfunction
 
 " Upgrade legacy options
@@ -234,7 +240,13 @@ endfunction
 function! fzf#run(...) abort
 try
   let oshell = &shell
-  set shell=sh
+
+  if has('win32') || has('win64')
+    set shell=cmd.exe
+  else
+    set shell=sh
+  endif
+
   if has('nvim') && len(filter(range(1, bufnr('$')), 'bufname(v:val) =~# ";#FZF"'))
     call s:warn('FZF is already running!')
     return []
@@ -353,7 +365,11 @@ function! s:xterm_launcher()
     \ &columns, &lines/2, getwinposx(), getwinposy())
 endfunction
 unlet! s:launcher
-let s:launcher = function('s:xterm_launcher')
+if has('win32') || has('win64')
+  let s:launcher = 'cmd.exe'
+else
+  let s:launcher = function('s:xterm_launcher')
+endif
 
 function! s:exit_handler(code, command, ...)
   if a:code == 130
@@ -370,7 +386,9 @@ endfunction
 
 function! s:execute(dict, command, temps) abort
   call s:pushd(a:dict)
-  silent! !clear 2> /dev/null
+  if has('unix')
+    silent! !clear 2> /dev/null
+  endif
   let escaped = escape(substitute(a:command, '\n', '\\n', 'g'), '%#')
   if has('gui_running')
     let Launcher = get(a:dict, 'launcher', get(g:, 'Fzf_launcher', get(g:, 'fzf_launcher', s:launcher)))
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks for the follow-up. Can you elaborate on the reasoning behind the changes?

A few questions:

- You added `set shellslash` in the previous PR, but changed it to `noshellslash` here. Can you tell me why?
- Does `let escaped = '"'.(escaped).'"'` work when `escaped` contains double-quote characters?
- Do we need temporary switching of `shellslash` in `s:cmd` when it's already handled in `fzf#run()`?

It would be much helpful if you can give some examples, such as steps to verify the correctness of the implementation. Thanks.

janlazo:
We need `noshellslash` in `s:cmd` and `fzf#run` for consistency on shell escaping.
I use `shellslash` in my vim config so `shellescape()` behaved the same in `fzf#run` and `s:cmd()` but 
`noshellslash` is default even on Windows Vim.
Having temporary switches of `shellslash`  in `s:cmd` and `fzf#run` guarantees the behaviour.

Examples:
- `C:/ ` (working directory, `set shellslash`)
   - [X] FZF
   - [X] FZF Program\ Files/
   - [ ] FZF 'Program\ Files/'
   - [ ] FZF "Program\ Files/"
- `C:/ Program\ Files/` (working directory, `set shellslash`)
   - [X] FZF
   - [X] FZF C:/Program\ Files/
   - [ ] FZF 'C:/Program\ Files/'
   - [ ] FZF "C:/Program\ Files/"

`noshellslash` forces double-quoting shell arguments such that cmd.exe can read it
**In this PR, If `escaped` contains double-quote characters, the command breaks.**
Double-quoted shell arguments have to escaped by wrapping them in **single-quotes**
but the command to pass to cmd.exe has to be double-quoted.
The Windows launcher for fzf is 'cmd.exe /C %s'
The string specifier does not have to wrapped in double-quotes because it can work without it.

Examples:
- `C:/ ` (working directory)
  - [X] `cmd /C fzf`
  - [X] `cmd /C cmd /C cmd /C fzf`
  - [ ] `cmd /C 'fzf'`
  - [X] `cmd /C "fzf"`
  - [ ] `cmd /C "'fzf'"`
  - [X] `cmd /C "cmd /C fzf"`
  - [X] `cmd /C "cmd /C fzf --prompt '"C:/Program\ Files/"'`
  - [X] `cmd /C "cmd /C fzf --prompt '"C:\Program Files\"'`

junegunn:
Thanks for the explanation. Hmm, so whenever we use `shellescape` function to build command-line options for fzf, we have to guarantee that `shellslash` is unset, so that single-quotes are not used as they are not compatible with `cmd.exe`, right?

`:FZF` is just one command built with `fzf#run`. It's a bummer that one should be aware of the above fact whenever writing a function or a command based on `fzf#run`. I think it makes sense that we provide a helper function `fzf#shellescape` that can be used instead. Let me look into it.

junegunn:
Squashed and merged your commits and slightly updated in 6c0fd7f. Let me know if you see any issues. Thanks.

janlazo:
cmd.exe internally uses single-quotes to escape commands so we have to resort to double-quotes
and escape the double-quotes if necessary.
```bat
>cmd.exe /C 'dir'
''dir'' is not recognized as an internal or external command, operable program, or batch file
>cmd.exe /C "'dir'"
''dir'' is not recognized as an internal or external command, operable program, or batch file
```
Commit https://github.com/junegunn/fzf/commit/6c0fd7f9ca663f2057a114f1aa86bfe5f6a78c68 works for me on terminal Vim and GVim.
The last issue remaining for Windows filepaths w/ brackets such as `C:\Program Files(x86)`.
Terminal Vim is fine but GVim is not because of the `escaped` string to pass to the launcher.

junegunn:
I see, would it help to create a temporary batch file that contains the command and use it to avoid escaping issue?

janlazo:
Executing cmd.exe outside of the terminal causes Windows to make a new terminal for cmd.exe
Since launcher and the shell are both cmd.exe, we can do one of the following:

1. add another if condition on s:execute() to check if running on Windows Vim.
    if running Windows Vim, then `let escaped = command`
2. if running Windows Vim, then `let s:launcher = "%s"` and remove the double-quoting for `escaped`

janlazo:
Code for Option 2 in Commit https://github.com/janlazo/fzf/commit/4572bdcf7d710d8dcc4460319c0ac0f0a1878216

junegunn:
@janlazo Thanks, the code looks good to me. Please send me a pull request.

Issue #775, opened 2016-12-17
FZF blocks Vim while opened in tmux pane


toiffel:
- Category
    - [ ] fzf binary
    - [ ] fzf-tmux script
    - [ ] Key bindings
    - [ ] Completion
    - [x] Vim
    - [ ] Neovim
    - [ ] Etc.
- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Windows Subsystem for Linux
    - [ ] Etc.
- Shell
    - [x] bash
    - [ ] zsh
    - [ ] fish

Steps to reproduce:

1. Run Vim inside tmux
2. Run FZF
3. Switch back to Vim pane
4. Type any characters

Expected: Vim consumes typed characters as usual
Observed: Vim is blocked and typed characters are echoed on the screen until FZF is closed

![fzf](https://cloud.githubusercontent.com/assets/1043953/21285819/7a6e8b76-c46d-11e6-9523-8d8b069f10d1.png)

This issue per se is not so important but it causes another bugs: https://github.com/junegunn/fzf/issues/434, https://github.com/tmux-plugins/vim-tmux-focus-events/issues/2. With `set -g focus-events on` tmux sends FocusIn/FocusOut escape sequences (`<Esc>[I` and `<Esc>[O`) to Vim instance. But since Vim is blocked by FZF, these escape sequences are echoed on the screen as `^[[I` and `^[[O`.

By default, Neovim doesn't have such issues because FZF is running asynchronously in separate terminal. But in Vim FZF is invoked with blocking system() call.

https://github.com/junegunn/fzf/blob/master/plugin/fzf.vim#L395

Would it be possible to have the same asynchronous FZF behavior in Vim as well, e.g. by using Vim 8 job API?

junegunn:
I'm not sure if I want to rewrite Vim plugin to work in asynchronous manner, as it complicates the implementation without obvious benefit (and also because I don't use the tmux plugin). Have you tried the workaround suggested in https://github.com/tmux-plugins/vim-tmux-focus-events/issues/2#issuecomment-87863831?

toiffel:
Yes, I've tried this workaround but unfortunately it doesn't work for me. Once I open FZF, leftover characters begin to appear.

Another affected plugin is https://github.com/christoomey/vim-tmux-navigator. Again, navigation works fine until I open FZF. While it's opened, I can't switch from Vim to other panes using vim-tmux-navigator because Vim is blocked.

[...]

Issue #782, opened 2016-12-29
Add `--height` option


junegunn:
- Category
    - [x] fzf binary

## Summary

Add `--height HEIGHT[%]` option to display fzf window below the cursor instead of opening it in fullscreen using ncurses or tcell.

## Why?

fzf-tmux provides flexible, non-fullscreen layouts, but it requires tmux and the experience can be suboptimal on large screens, where the tmux pane serving fzf can be far away from the cursor.

## A few things to note

- Supports preview window and mouse.
- 24-bit colors. We're not constrained by the limitation of ncurses.
- Requires `/dev/tty`, `stty`, and `tput`.
- Not sure about Windows.

The new option can be applied to the key bindings and fuzzy completion with the following environment variables.

```sh
FZF_TMUX=0
FZF_CTRL_T_OPTS='--height 40% --reverse'
FZF_CTRL_R_OPTS='--height 40%'
FZF_ALT_C_OPTS='--height 40% --reverse'
FZF_COMPLETION_OPTS='--height 40% --reverse'
```

Or it can be globally applied via `$FZF_DEFAULT_OPTS`:

```sh
export FZF_DEFAULT_OPTS='--height 40% --reverse'
```

## Sample images

<img width="1280" alt="fzf-height" src="https://cloud.githubusercontent.com/assets/700826/21546093/db47a20a-ce1f-11e6-92b1-bed3710dd616.png">

<img width="1280" alt="fzf-height-ctrl-r" src="https://cloud.githubusercontent.com/assets/700826/21546110/ec17df64-ce1f-11e6-813f-69edcfa5b202.png">

<img width="1280" alt="fzf-height-24bit" src="https://cloud.githubusercontent.com/assets/700826/21546114/f307352c-ce1f-11e6-95e7-c11cbe6dbeb7.png">

## Issues

- Windows support
- Manual handling of escape sequences might bring back #661

paulhybryant:
I am using the latest version of fzf installed with homebrew, and it complains about 

unknown option: --height

I guess the binary is not up to date but this option is already added to the zsh binding file.

I think we should check the version in the zsh to avoid this problem (e.g. only add --height when the fzf version is newer than x.y.z)?

junegunn:
@paulhybryant Rerun the install script to download the binary.

paulhybryant:
Thanks @junegunn for the quick reply. Not sure which install script you are referring to.

What I did is 
brew reinstall fzf

and the error is still there.
[...]

PR #792, opened 2017-01-06
[vim] Run fzf-tmux asynchronously using Vim8/Neovim jobs (#775)

toiffel:
Hi @junegunn,

This is a PR for accompanying issue #775. I've tried to make it the least invasive as possible. Basically it just adds a separate code path `s:execute_tmux_async` to run fzf-tmux asynchronously if Vim8 jobs or Neovim are available. `s:execute_tmux_async` is closely modeled after `s:execute_tmux` and `s:execute_term`.

Please check the commit message for more details.

Tested on both Linux and Mac using Vim 8.0.0086 and Neovim 0.1.7.

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -266,7 +266,8 @@ try
     set shell=sh
   endif
 
-  if has('nvim') && len(filter(range(1, bufnr('$')), 'bufname(v:val) =~# ";#FZF"')) || exists('s:curr_job')
+  if has('nvim') && len(filter(range(1, bufnr('$')), 'bufname(v:val) =~# ";#FZF"')) ||
+    \ exists('s:curr_job') && (has('nvim') || job_status(s:curr_job) == 'run')
     call s:warn('FZF is already running!')
     return []
   endif
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -75,6 +75,10 @@ function! s:tmux_enabled()
   return s:tmux
 endfunction
 
+function! s:has_job()
+  return has('nvim') || v:version >= 800 && has('job') && has('lambda')
+endfunction
+
 function! s:shellesc(arg)
   return '"'.substitute(a:arg, '"', '\\"', 'g').'"'
 endfunction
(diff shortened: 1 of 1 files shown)


plugin/fzf.vim
```diff
@@ -262,7 +266,7 @@ try
     set shell=sh
   endif
 
-  if has('nvim') && len(filter(range(1, bufnr('$')), 'bufname(v:val) =~# ";#FZF"'))
+  if has('nvim') && len(filter(range(1, bufnr('$')), 'bufname(v:val) =~# ";#FZF"')) || exists('s:curr_job')
```
junegunn: I think we should also check the return value of `job_status(s:curr_job)`. Or Vim will keep complaining once we kill the tmux pane running FZF.

junegunn:
By the way, I'm working on non-fullscreen renderer (#782). When this lands, we can use it on Vim to display fzf in non-fullscreen mode without using tmux as follows:

screen shot 2017-01-08 at 1 29 40 am

Then we probably don't need this extra code branch.

plugin/fzf.vim
toiffel: Thank you for the review! Pushed the fix using `job_status(s:curr_job)`.

Unfortunately FZF still complains about already running instance after killing tmux pane. But this happens due to another bug in fzf-tmux, not because of this PR. I opened a separate issue #796 for this.

Once that bug is fixed, this branch will work correctly after killing tmux pane as well.

toiffel:
This non-fullscreen renderer looks awesome! :+1: I'll definitely give it a try later.

Anyway, I believe my PR won't harm as long as you keep maintaining tmux support in Vim plugin.

junegunn:
> I'll definitely give it a try later.

Please do. It's a non-trivial change and feedbacks are welcome. You can check out devel branch and run the install script to download the alpha binary.

> I believe my PR won't harm as long as you keep maintaining tmux support in Vim plugin.

Yeah, `--height` can only cover `down` layout. fzf-tmux is still needed for `up`/`left`/`right` options, though they are rarely used in practice. I'm a little concerned that we may have a bit too many ways to start fzf within Vim.

toiffel:
Tried the latest 0.16.1 on Linux/Mac and it works smoothly w/ or w/out tmux. I don't need to run fzf in tmux pane anymore. Thank you for the great job!

Created a minor issue #814. It can be easily fixed by using non-silent mapping but it could be a potential bug.

I think this PR can be closed now.

junegunn:
Glad that you like the new option. Thanks for taking your time into this!

PR #793, opened 2017-01-06
Alt-o shortcut to open file

kmARC:
!!! DRAFT !!!

Similarly to Alt-c (*cd into directory*), Alt-o command opens up the FZF
file selector view and invokes the default associated application with
the selected file.

Current implementation is based on Alt-c functionality. It is tested on
freedesktop compliant linux system (Ubuntu 16.04) and opens the selected
entry with `xdg-open`. It implements configuration parameters of
FZF_ALT_O_OPTIONS and FZF_ALT_O_COMMAND.

TODOs:
* `/usr/bin/xdg-open` is hard-coded. Other platforms need different
  configuration
* automatic testing stub is empty.

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,6 +205,9 @@ fish.
 - `ALT-C` - cd into the selected directory
     - Set `FZF_ALT_C_COMMAND` to override the default command
     - Set `FZF_ALT_C_OPTS` to pass additional options
+- `ALT-O` - open selected file with default application (`xdg-open`)
+    - Set `FZF_ALT_O_COMMAND` to override the default command
+    - Set `FZF_ALT_O_OPTS` to pass additional options
 
 If you're on a tmux session, fzf will start in a split pane. You may disable
 this tmux integration by setting `FZF_TMUX` to 0, or change the height of the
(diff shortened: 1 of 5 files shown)


junegunn:
Thanks for the suggestion. But I'm not interested in adding more key bindings to this repository to avoid increased maintenance burden. It has been already painful for me to make sure that those 3 key bindings work without issues across different environments. i.e. {CTRL-T, CTRL-R, ALT-C} x {bash3,bash3/vi,bash4,bash4/vi,zsh,fish,fish/vi} x {macOS, Linux}. Adding another key binding also means that we have to add one or two more configuration variables that has to be tested, documented, and maintained. I'd like to only focus on improving the core fzf binary. Extra key bindings and applications can be published on the wiki page, or in a blog post, or even provided as third-party projects.

kmARC:
Ah okay. Then I'll prepare an entry for the [Examples wiki page](https://github.com/junegunn/fzf/wiki/Examples), how about that?

junegunn:
Sure, the write permission to the wiki page is open to anyone.

PR #812, opened 2017-01-19
Quote query value for zsh Ctrl-R

jamesoff:
This fixes pre-populating the query value from the typed command so far,
which broke in 208d4f2.

### DIFF
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -55,7 +55,7 @@ fzf-history-widget() {
   local selected num
   setopt localoptions noglobsubst pipefail 2> /dev/null
   selected=( $(fc -l 1 |
-    FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} $FZF_DEFAULT_OPTS +s --tac -n2..,.. --tiebreak=index --toggle-sort=ctrl-r $FZF_CTRL_R_OPTS +m --query=${(q)LBUFFER}" $(__fzfcmd)) )
+    FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} $FZF_DEFAULT_OPTS +s --tac -n2..,.. --tiebreak=index --toggle-sort=ctrl-r $FZF_CTRL_R_OPTS +m --query=\"${(q)LBUFFER}\"" $(__fzfcmd)) )
   local ret=$?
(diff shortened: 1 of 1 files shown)


junegunn:
Are you sure that it's broken? It works for me.

jamesoff:
Double-checked; without the fix, it breaks with a trailing space but is ok otherwise.

I tend to type "ssh " and then hit Ctrl-R :)

I see the test for test_ctrl_r is failing, but I don't immediately see in the output why :(

junegunn:
> it breaks with a trailing space

I see, that was the information I needed :)

Test cases sometimes fail of Travis CI due to timeout issues, I'll restart the build and see if it helps.

junegunn:
It is a genuine failure. When you press CTRL-R on empty command-line, you get `''` in fzf prompt.

The code not working when your command-line ends with a space turned out to be a bug of shellwords library fzf depends on. I'll fix it and include the patch in the next fzf release. But until then, a workaround would be to add `''` at the end.

```diff
diff --git a/shell/key-bindings.zsh b/shell/key-bindings.zsh
index c8705ad..4d34129 100644
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -55,7 +55,7 @@ fzf-history-widget() {
   local selected num
   setopt localoptions noglobsubst pipefail 2> /dev/null
   selected=( $(fc -l 1 |
-    FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} --reverse $FZF_DEFAULT_OPTS +s --tac --no-reverse -n2..,.. --tiebreak=index --toggle-sort=ctrl-r $FZF_CTRL_R_OPTS +m --query=${(q)LBUFFER}" $(__fzfcmd)) )
+    FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} --reverse $FZF_DEFAULT_OPTS +s --tac --no-reverse -n2..,.. --tiebreak=index --toggle-sort=ctrl-r $FZF_CTRL_R_OPTS +m --query=${(q)LBUFFER}''" $(__fzfcmd)) )
   local ret=$?
   if [ -n "$selected" ]; then
     num=$selected[1]
```

It does help, right?

junegunn:
Hi, I was able to workaround the issue by reordering the options. Thanks for the report and the fix.

PR #817, opened 2017-01-20
CTRL_T key-bindings respect ~/.ignore

DanielFGray:
it's a bit of a hack, but it should work. or at least, it works on my machine..

### DIFF
--- a/shell/key-bindings.bash
+++ b/shell/key-bindings.bash
@@ -2,7 +2,7 @@
 # ------------
 __fzf_select__() {
   local cmd d ignore_dirs=()
-  while IFS= read -r d; do
+  [[ -r "$HOME/.ignore" ]] && while IFS= read -r d; do
     ignore_dirs+=( -path "\*/$d" -o )
   done < "$HOME/.ignore"
   cmd="${FZF_CTRL_T_COMMAND:-"command find -L . -mindepth 1 \\( ${ignore_dirs[*]} -path '*/\\.*' -o -fstype 'devfs' -o -fstype 'devtmpfs' -o -fstype 'proc' \\) -prune \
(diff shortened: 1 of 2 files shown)


acornejo:
I am curious, isn't it cleaner and faster to set in your environment `FZF_CTRL_T_COMMAND='ag -l -g ""'`

There are plenty of other tools besides `ag` that fully support `~/.gitignore` and other exclude files, for instance look at `sk` or `ack`. Implementing this inside the fzf distribution seems unnecessary.

DanielFGray:
Well sure, you could rely on external tools for that, but this would work with just plain `find`..

But I understand if you'd prefer not to merge this. Actually I really should've `test`ed to see if the file exists before trying to parse it...

DanielFGray:
Oh I just realized this isn't in a feature patch.. going to re-submit.

acornejo:
I am not the maintainer of this repo, I am just one of many great full users ;), so please disregard my comments, I was just curious as to why one would want to parse ignore files with shellscripts when there are alternatives out there.

DanielFGray:
My reasoning was that while it may not be as fast as other alternatives, it's still pretty darn fast. And if you have another tool you'd prefer, you can use that, and if it's not available, with this, `find` will use your `~/.ignore`.

junegunn:
Hi, thanks for the suggestion, but I'm pretty much with @acornejo on this, for a number of reasons.

1. We'll have to implement the logic for 3 different shells (bash, zsh, and fish) and I'll have to maintain the code forever. Supporting 3 shells has been the number one pain point for me, because all I use is bash. Let's delegate the role to the tools that are obviously better at it (unix philosophy) and keep it simple.
2. I know `.gitignore`, but `.ignore` is new to me. But I'll just assume that it's similar to `.gitignore`. The complexity of `.gitignore` is monstrous (see `man gitignore`). It is obvious that the suggested solution won't be able to fully address the specification and it is not clear to the user what is supported and what is not. This can lead to an unnecessary confusion.

PR #828, opened 2017-01-31
Add current command name on a global variable

kassio:
This can be useful to add the current command name to the statusline,
like this:

```viml
au User FzfStatusLine call <SID>fzf_statusline()

function! s:fzf_statusline()
  let fzf_cmd_name = get(g:, 'fzf_cmd_name', 'FZF')
  let &l:statusline = '> '.fzf_cmd_name
endfunction
```

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -201,6 +201,10 @@ function! fzf#wrap(...)
   endfor
   let [name, opts, bang] = args
 
+  if len(name)
+    let opts.name = name
+  end
+
   " Layout: g:fzf_layout (and deprecated g:fzf_height)
   if bang
     for key in s:layout_keys
@@ -521,6 +525,7 @@ function! s:execute_term(dict, command, temps) abort
   let winrest = winrestcmd()
   let pbuf = bufnr('')
   let [ppos, winopts] = s:split(a:dict)
+  let b:fzf = a:dict
   let fzf = { 'buf': bufnr(''), 'pbuf': pbuf, 'ppos': ppos, 'dict': a:dict, 'temps': a:temps,
             \ 'winopts': winopts, 'winrest': winrest, 'lines': &lines,
             \ 'columns': &columns, 'command': a:command }


junegunn:
Thanks for the suggestion. I see your point, but calling `fzf#wrap` does not necessarily mean that the user will call `fzf#run` immediately with the result. It's designed to be a pure function that should not affect the global state. Also, the variable you suggested only makes sense in Neovim and we don't want to unnecessarily expose it in ordinary Vim.

So here's the idea. Let's make the dictionary returned from `fzf#wrap` contain the optional name argument, and expose the dictionary as a buffer-local variable called `b:fzf`.

```diff
diff --git a/plugin/fzf.vim b/plugin/fzf.vim
index 1cd6ff7..e1e509a 100644
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -201,6 +201,10 @@ function! fzf#wrap(...)
   endfor
   let [name, opts, bang] = args
 
+  if len(name)
+    let opts.name = name
+  endif
+
   " Layout: g:fzf_layout (and deprecated g:fzf_height)
   if bang
     for key in s:layout_keys
@@ -521,6 +525,7 @@ function! s:execute_term(dict, command, temps) abort
   let winrest = winrestcmd()
   let pbuf = bufnr('')
   let [ppos, winopts] = s:split(a:dict)
+  let b:fzf = a:dict
   let fzf = { 'buf': bufnr(''), 'pbuf': pbuf, 'ppos': ppos, 'dict': a:dict, 'temps': a:temps,
             \ 'winopts': winopts, 'winrest': winrest, 'lines': &lines,
             \ 'columns': &columns, 'command': a:command }
```

And you can use it like follows:

```vim
au User FzfStatusLine call <SID>fzf_statusline()

function! s:fzf_statusline()
  let &l:statusline = '> '.get(b:fzf, 'name', 'FZF')
endfunction
```

kassio:
Suggested changed done. 😄

junegunn:
Merged, thanks.

PR #838, opened 2017-02-08
fallback to fzf when tmux pane is zoomed.

amosbird:


junegunn:
Have you taken a look at #433? We already handle zoomed panes.

(By the way, I just noticed a bug with the loging handling `-l` `-u` on zoomed panes. Let me push the fix for it.)

amosbird:
well, it seems not working on my env. `fzf-tmux` just deletes all the panes when finished.

junegunn:
Can you check if 1ba7acf4bd4314f299b57cea73cbd376bde5aba3 helps?

amosbird:
no, it doesn't help.

amosbird:
I'm using latest tmux, configuration is in https://github.com/amosbird/tmuxconfig

junegunn:
Looks like you're a fish user, so I believe #891 fixes the issue you experience.

PR #848, opened 2017-02-19
Add support for FZF_CTRL_T_FILE_PREFIX

odedlaz:
I find it useful to inject a specific prefix when pressing "Ctrl-T" on files.
I made the default value `$EDITOR`, which means that by default "Ctrl-T" will return the selection, prefixed by the users editor.

### DIFF
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -27,7 +27,21 @@ __fzfcmd() {
 }
 
 fzf-file-widget() {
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks, but I have no plan to change the way the key bindings work. It's been really painful for me to make sure that they work consistently across three different shells (bash, zsh, and fish), and any new options will make the maintenance even more difficult.

I believe you'll be able to implement the suggested behavior by defining another widget that is built upon fzf-file-widget.

PR #858, opened 2017-02-26
Work with vanilla zsh settings

pjcj:
Zsh is very customisable and some of the customisations will break the
completion function.  Restore the state to its default for the duration
of the function.

### DIFF
--- a/shell/completion.zsh
+++ b/shell/completion.zsh
@@ -142,6 +142,7 @@ _fzf_complete_unalias() {
 }
 
 fzf-completion() {
+  emulate zsh
   local tokens cmd prefix trigger tail fzf matches lbuf d_cmds
   setopt localoptions noshwordsplit noksh_arrays
 


junegunn:
Thanks, looks good to me. Can you give me a set of options that breaks the function as an example? I would like to reproduce the problem locally, and confirm that the patch fixes the issue.

pjcj:
I wish I knew that off-hand.  I'll look round for some likely candidates, but my .zshrc is quite well customised by now...

pjcj:
I have found the culprit in my case: setopt posix_builtins

junegunn:
I tested with this extremely minimal zshrc

```sh
[ -f ~/.fzf.zsh ] && source ~/.fzf.zsh
setopt posix_builtins
```

but fuzzy-completion seems to work just fine. `vim ~/**<tab>`, `cd **<tab>`, etc. Am I missing something?

pjcj:
Ah, it needs a couple of options.  Try this:

```
setopt all_export posix_builtins
. ~/.fzf.zsh
```

junegunn:
Oh yeah, now I can reproduce. Looks like `CTRL-R` also seems to break with those options. Should we do the same for fzf-history-widget as well?

pjcj:
Ah, yes, fzf-history-widget needs it too.  And it probably makes sense to do the same for fzf-cd-widget and fzf-file-widget too, even though they work with this particular configuration, as they may well break with some other configuration, or if they get changed.

junegunn:
So, what do you think?

```diff
diff --git a/shell/completion.zsh b/shell/completion.zsh
index 9b9dba3..ccd18ce 100644
--- a/shell/completion.zsh
+++ b/shell/completion.zsh
@@ -46,7 +46,7 @@ __fzf_generic_path_completion() {
   tail=$6
   fzf="$(__fzfcmd_complete)"
 
-  setopt localoptions nonomatch
+  setopt nonomatch
   dir="$base"
   while [ 1 ]; do
     if [[ -z "$dir" || -d ${~dir} ]]; then
@@ -142,8 +142,8 @@ _fzf_complete_unalias() {
 }
 
 fzf-completion() {
+  emulate -L zsh
   local tokens cmd prefix trigger tail fzf matches lbuf d_cmds
-  setopt localoptions noshwordsplit noksh_arrays
 
   # http://zsh.sourceforge.net/FAQ/zshfaq03.html
   # http://zsh.sourceforge.net/Doc/Release/Expansion.html#Parameter-Expansion-Flags
diff --git a/shell/key-bindings.zsh b/shell/key-bindings.zsh
index 9b2353a..dbafafb 100644
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -27,6 +27,7 @@ __fzfcmd() {
 }
 
 fzf-file-widget() {
+  emulate -L zsh
   LBUFFER="${LBUFFER}$(__fsel)"
   local ret=$?
   zle redisplay
@@ -38,9 +39,10 @@ bindkey '^T' fzf-file-widget
 
 # ALT-C - cd into the selected directory
 fzf-cd-widget() {
+  emulate -L zsh
+  setopt pipefail 2> /dev/null
   local cmd="${FZF_ALT_C_COMMAND:-"command find -L . -mindepth 1 \\( -path '*/\\.*' -o -fstype 'sysfs' -o -fstype 'devfs' -o -fstype 'devtmpfs' -o -fstype 'proc' \\) -prune \
     -o -type d -print 2> /dev/null | cut -b3-"}"
-  setopt localoptions pipefail 2> /dev/null
   local dir="$(eval "$cmd" | FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} --reverse $FZF_DEFAULT_OPTS $FZF_ALT_C_OPTS" $(__fzfcmd) +m)"
   if [[ -z "$dir" ]]; then
     zle redisplay
@@ -57,8 +59,9 @@ bindkey '\ec' fzf-cd-widget
 
 # CTRL-R - Paste the selected command from history into the command line
 fzf-history-widget() {
+  emulate -L zsh
+  setopt pipefail 2> /dev/null
   local selected num
-  setopt localoptions noglobsubst pipefail 2> /dev/null
   selected=( $(fc -l 1 |
     FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} $FZF_DEFAULT_OPTS +s --tac -n2..,.. --tiebreak=index --bind=ctrl-r:toggle-sort $FZF_CTRL_R_OPTS --query=${(q)LBUFFER} +m" $(__fzfcmd)) )
   local ret=$?
```

(Reference: http://zsh.sourceforge.net/Guide/zshguide02.html)

pjcj:
That looks good to me (although I haven't run it). The -L option to emulate is certainly the right way to go.

junegunn:
Alright, let's push the fix and see if anyone complains.

PR #870, opened 2017-03-07
Add support for IPv6 addresses in ssh ** command

S4mw1s3:
Without this patch, IPv6 addresses are not parsed completely in the ssh known_hosts file.
Signed-off-by: Sam Van Den Berge <sam@drgt.net>

### DIFF
--- a/shell/completion.bash
+++ b/shell/completion.bash
@@ -234,7 +234,7 @@ _fzf_complete_telnet() {
 _fzf_complete_ssh() {
   _fzf_complete '+m' "$@" < <(
     cat <(cat ~/.ssh/config /etc/ssh/ssh_config 2> /dev/null | command grep -i '^host' | command grep -v '*') \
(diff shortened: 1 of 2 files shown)


junegunn:
Merged, thanks.

S4mw1s3:
@junegunn I'm affraid I made a mistake. With this commit, I broke support for hostnames containing a hyphen (e.g. my-host-name). This is because for the grep command, in bracket expressions, the hyphen acts as a range expression, unless it's located as the last character where it's interpreted as a literal.
So, for this patch to be correct, the colon should be placed in front of the hyphen. What do you suggest I should do?

junegunn:
> So, for this patch to be correct, the colon should be placed in front of the hyphen. 

That will do. Will you send me a new pull request for the master branch? I just realized that I made a mistake of merging this only to devel branch instead of master.

PR #871, opened 2017-03-10
make install script work with non GNU tar

reezer:
This change makes the install scripts work on non GNU tar systems as well (which at least includes FreeBSD).


junegunn:
Thanks!

reezer:
Hehe, no worries. Thanks for accepting so quickly. :)

PR #877, opened 2017-03-20
Add support for IPv6 addresses in ssh ** command

S4mw1s3:
In my previous pull request I made a mistake as discussed here: https://github.com/junegunn/fzf/pull/870
This is now the correct version where the colon is in front of the hyphen.


junegunn:
Thanks!

PR #882, opened 2017-03-24
[vim] Start new terminal for nvim in Windows

janlazo:
This PR starts allows the vim plugin to work for Neovim in Windows by starting a new and separate terminal via cmd.exe and the start command. This is a workaround for the missing `:terminal` support.

I used jobstart() and on_exit to mimic s:execute() for Vim in Windows.

Tested only in Windows 8.1 and a recent March build of Neovim for Windows.

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -479,11 +479,7 @@ function! s:execute(dict, command, use_height, temps) abort
   if has('unix') && !a:use_height
     silent! !clear 2> /dev/null
   endif
-  let escaped_chars = '%#!'
-  if s:is_win
-      let escaped_chars .= '\'
-  endif
-  let escaped = escape(substitute(a:command, '\n', '\\n', 'g'), escaped_chars)
+  let escaped = escape(substitute(a:command, '\n', '\\n', 'g'), '%#!')
   if has('gui_running')
     let Launcher = get(a:dict, 'launcher', get(g:, 'Fzf_launcher', get(g:, 'fzf_launcher', s:launcher)))
     let fmt = type(Launcher) == 2 ? call(Launcher, []) : Launcher
@@ -718,7 +714,8 @@ let s:default_action = {
 function! s:shortpath()
   let short = pathshorten(fzf#fnamemodify(fzf#getcwd(), ':~:.'))
   let slash = s:is_win ? '\' : '/'
-  return empty(short) ? '~'.slash : short . (short =~ '/$' ? '' : slash)
+  let path = empty(short) ? '~'.slash : short . (short =~ '/$' ? '' : slash)
+  return s:is_win ? escape(path, ' \') : path
 endfunction
 
 function! s:cmd(bang, ...) abort
@@ -729,7 +726,8 @@ function! s:cmd(bang, ...) abort
     if s:is_win
         let opts.dir = substitute(opts.dir, '/', '\\', 'g')
     endif
-    let opts.options .= ' --prompt '.fzf#shellescape(opts.dir)
+    let path = s:is_win ? escape(opts.dir, ' \') : opts.dir
+    let opts.options .= ' --prompt '.fzf#shellescape(path)
   else
     let opts.options .= ' --prompt '.fzf#shellescape(s:shortpath())
   endif
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -83,7 +83,11 @@ function! s:tmux_enabled()
 endfunction
 
 function! s:shellesc(arg)
-  return '"'.substitute(a:arg, '"', '\\"', 'g').'"'
+  let escaped = substitute(a:arg, '"', '\\"', 'g')
+  if s:is_win
+      return escaped
+  endif
+  return '"'.escaped.'"'
 endfunction
 
 function! s:escape(path)
@@ -344,6 +348,9 @@ try
   let command = prefix.(use_tmux ? s:fzf_tmux(dict) : fzf_exec).' '.optstr.' > '.temps.result
 
   if use_term
+    if s:is_win
+        return s:execute_win_term(dict, command, temps)
+    endif
     return s:execute_term(dict, command, temps)
   endif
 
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks. As far as I know, Windows support of Neovim is experimental and is not officially announced yet. I'm not quite sure if I want to add code for it at this point.

> I used jobstart() and on_exit to mimic s:execute() for Vim in Windows.

So Neovim for Windows will not have `:terminal`? Then can't we just unset `use_term` and use the same code for ordinary Vim on Windows? Any reason `s:execute` can't run on Neovim?

justinmk:
> So Neovim for Windows will not have :terminal ?

It's planned, and was working at one point, but no timeframe on it.

janlazo:
`:!some_shell_command`  is non-interactive and is shellescaped by Neovim. `:!fzf` works but the UI is frozen and the user cannot see fzf running. Shellquoting in Windows is not resolved yet and `system()` is non-interactive. `jobstart(['executable', 'args', ...], callback_dict)` is the only choice left to run a separate process and terminal for fzf and get around the shellquoting issue. Doing this does not freeze the Neovim UI so putting the jobstart in `s:execute()` would return an empty array. If we move the code for the fzf in `s:execute()`, would this break any assumptions in the code meant for ordinary Vim?

Nightly builds of Neovim for Windows include nvim-qt for a GUI because there's no TUI officially available yet so we can limit the scope of the new code for Neovim GUIs in Windows only.

justinmk:
TUI is "sort of working" in neovim, and the related PR will be merged soon.

janlazo:
`:FZF C:/Program\ Files\ (x86)\` fails in Neovim for Windows. Paths with no spaces work. Toggling `shellslash` changes some slashes but the result is the same. Vim does not have this issue.

Full command:
`fzf  --multi  --prompt 'C:/Program Files (x86)/'  --expect=ctrl-v,ctrl-x,ctrl-t --no-height > C:/Users/JANEDM~1/AppData/Local/Temp/nvimzOaZXl/0`

Does https://github.com/neovim/neovim/pull/6359 fix this issue?

janlazo:
@junegunn How far can I go for handling the filepaths on Windows?
I need to add `fzf#fnamemodify()` and `fzf#getcwd()` for `s:shortpath()` because of `shellslash`

justinmk:
@janlazo try https://github.com/neovim/neovim/pull/6359 or just wait for it to be merged.

janlazo:
`set noshellslash` should cause shellescape() to double-quote but it always returns a single-quoted string.
`fzf --prompt 'C:\Program Files (x86)\'` fails on cmd.exe but works on powershell. Tradeof is the startup of the new powershell process.
Using the batchfile didn't matter because the prompt issue is on cmd.exe

janlazo:
I used powershell in this commit, https://github.com/janlazo/fzf/commit/7a752fbb0371cdcd13ae2c952c910611c4c7cba6, to get around the single-quote escaping. I passed `-NoProfile` to speed up startup time similar to `vim -u NONE`.
`:FZF path_with_spaces` work on that commit for Neovim on Windows.

janlazo:
@junegunn This PR has a mixture of commits meant for Vim but is compatible with Neovim via powershell. I can make a separate PR for using backslashes in Windows filepaths such that, with `set noshellslash`, the finalized command passed to `:!` should  use backslash for all filepaths.

junegunn:
Thanks for your work @janlazo. To be honest, it's a little hard for me to follow this thread; what are the exact problems we are trying to solve? What are the steps to reproduce each problem and how do we validate the correctness of the fixes? Are we talking about Vim, Neovim, or both? Is this the right time for us to make changes for unannounced Neovim? Is it okay for us to proceed with the Neovim issues mentioned above unresolved? Do we have to fix the code for `:FZF` command? Because then we also have to fix some commands in fzf.vim repository.

So basically what I'm saying is that I'd like to see this PR broken down into smaller ones each tackling a single problem.

janlazo:
There are at least 2 separate problems we're trying to solve:
- (Vim/Neovim) ambiguity in shellescaping when running shell commands via cmd.exe in Windows
   -  `shellslash` affects any VIm behaviour dealing with filepaths in Windows.
   In this PR, wrapper functions are used to handle `getcwd()`, `fnamemodify()` in `s:shortpath()`.
   We have `fzf#shellescape()` for `shellescape()` to escape the prompt but it only works with `:!` because the prompt filepath uses forward slash. Switching to backslashes on the prompt alongside `fzf#shellescape()` breaks the prompt, output redirection, and the rest of the command.
We can use a batch file but `:FZF` is already working without it.
   - When using backslashes for Windows filepaths that will be used in `system()` or `:!`, `fzf#shellescape()` may not be necessary because cmd.exe allows `C:\Program Files (x86)` unquoted. Using `fzf#shellescape()` risks breaking the prompt because the added quotes would have to be escaped as well.
- (Neovim) no stable workaround for missing `:terminal` support in Windows
  - Using your master branch, running `:FZF` on Neovim in Windows doesn't work because `shellescape()` doesn't work as expected with `set noshellslash` and the Vim plugins assumes that, if Neovim is used, then `:terminal` must be available.
   To reproduce the error, init.vim only contains `set rtp+=~/.fzf` for the Vim plugin and run `:FZF`. It should always fail no matter what filepath is used, including `s:shortpath()`.
  - What we can do is mimic the behaviour of `:FZF` for Vim GUI and leave it at that. I can only provide support for the Neovim GUIs (nvim-qt) as an end-user.
      As you suggested, we'll unset `use_term` and run `s:execute()`. I'll limit the differences to swapping `:!` for `jobstart()` for the callback when fzf exits and trivial shellescaping. If some of the common filepaths in Windows don't work, then there's no point expending effort for Neovim support in Windows until the official release.

junegunn:
> There are at least 3 separate problems we're trying to solve:

Can we have separate PR for each subproblem? I'd like to delay merging Neovim-related changes and wait for the related issues are cleared.

janlazo:
https://github.com/neovim/neovim/pull/6497 is merged to master so I'll open another PR just for jobstart in Neovim. Anything else comes after #896

PR #885, opened 2017-03-26
fzf#run: improve "is already running" message

blueyed:
This displays the buffer(s) in this case, which is useful when FZF got
stuck, and you have to manually remove the buffer.


junegunn:
Thanks!

PR #896, opened 2017-04-02
[vim] Use backslash for Windows filepaths

janlazo:
- Category
    - [ ] fzf binary
    - [ ] fzf-tmux script
    - [ ] Key bindings
    - [ ] Completion
    - [X] Vim
    - [X] Neovim (starting https://github.com/junegunn/fzf/pull/896/commits/3adea892331b158b2e0290cea519670d8670d739)
    - [ ] Etc.
- OS
    - [ ] Linux
    - [ ] Mac OS X
    - [X] Windows
    - [ ] Etc.
- Shell
    - [ ] bash
    - [ ] zsh
    - [ ] fish

PR for the shellescaping issue I described in #882.

I added wrapper functions to handle filepath functions in vim.
- `set noshellslash` for backslash in filepaths
- `set shellslash` for `shellescape()` to single quote first, and then double quote its output   (ex. path --> "'path'")
     - Vim runs cmd.exe twice, one is a wrapper proxy, and the other is the actual command
     - cmd.exe strips the double quotes of each argument
     - `C:\Program Files (x86)\` (no quotes) breaks on `s:execute()` so it must be escaped with single/double quotes in cmd.exe and powershell

I wasn't sure where to put the new functions so I placed them just after if guard at the top of the file because I used `fzf#expand` for `s:base_dir`.

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -51,6 +51,10 @@ function! s:fzf_expand(fmt)
   return s:fzf_call('expand', a:fmt)
 endfunction
 
+function! s:fzf_tempname()
+  return s:fzf_call('tempname')
+endfunction
+
 let s:default_layout = { 'down': '~40%' }
 let s:layout_keys = ['window', 'up', 'down', 'left', 'right']
 let s:is_win = has('win32') || has('win64')
@@ -311,7 +315,7 @@ try
     endif
   endif
   let dict   = exists('a:1') ? s:upgrade(a:1) : {}
-  let temps  = { 'result': tempname() }
+  let temps  = { 'result': s:fzf_tempname() }
   let optstr = get(dict, 'options', '')
   try
     let fzf_exec = s:fzf_exec()
@@ -324,7 +328,7 @@ try
   endif
 
   if !has_key(dict, 'source') && !empty($FZF_DEFAULT_COMMAND)
-    let temps.source = tempname().(s:is_win ? '.bat' : '')
+    let temps.source = s:fzf_tempname().(s:is_win ? '.bat' : '')
     call writefile((s:is_win ? ['@echo off'] : []) + split($FZF_DEFAULT_COMMAND, "\n"), temps.source)
     let dict.source = (empty($SHELL) ? &shell : $SHELL) . (s:is_win ? ' /c ' : ' ') . s:shellesc(temps.source)
   endif
@@ -335,7 +339,7 @@ try
     if type == 1
       let prefix = source.'|'
     elseif type == 3
-      let temps.input = tempname()
+      let temps.input = s:fzf_tempname()
       call writefile(source, temps.input)
       let prefix = (s:is_win ? 'type ' : 'cat ').s:shellesc(temps.input).'|'
     else
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -26,7 +26,7 @@ if exists('g:loaded_fzf')
 endif
 let g:loaded_fzf = 1
 
-function! s:without_shellescape(fn, ...)
+function! s:fzf_call(fn, ...)
   if s:is_win
     let shellslash = &shellslash
     try
@@ -40,15 +40,15 @@ function! s:without_shellescape(fn, ...)
 endfunction
 
 function! s:fzf_getcwd()
-  return s:without_shellescape('getcwd')
+  return s:fzf_call('getcwd')
 endfunction
 
 function! s:fzf_fnamemodify(fname, mods)
-  return s:without_shellescape('fnamemodify', a:fname, a:mods)
+  return s:fzf_call('fnamemodify', a:fname, a:mods)
 endfunction
 
 function! s:fzf_expand(fmt)
-  return s:without_shellescape('expand', a:fmt)
+  return s:fzf_call('expand', a:fmt)
 endfunction
 
 let s:default_layout = { 'down': '~40%' }
@@ -288,16 +288,7 @@ function! fzf#wrap(...)
 endfunction
 
 function! fzf#shellescape(path)
-  if s:is_win
-    let shellslash = &shellslash
-    try
-      set shellslash
-      return '"'.shellescape(a:path).'"'
-    finally
-      let &shellslash = shellslash
-    endtry
-  endif
-  return shellescape(a:path)
+  return s:fzf_call('shellescape', a:path)
 endfunction
 
 function! fzf#run(...) abort
@@ -716,7 +707,7 @@ function! s:cmd(bang, ...) abort
   else
     let prompt = s:shortpath()
   endif
-  let opts.options .= ' --prompt '.fzf#shellescape(prompt)
+  let opts.options .= ' --prompt '.fzf#shellescape(s:is_win ? escape(prompt, '\') : prompt)
   let opts.options .= ' '.join(args)
   call fzf#run(fzf#wrap('FZF', opts, a:bang))
 endfunction
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -26,39 +26,71 @@ if exists('g:loaded_fzf')
 endif
 let g:loaded_fzf = 1
 
-function! s:fzf_call(fn, ...)
-  if s:is_win
-    let shellslash = &shellslash
-    try
-      set noshellslash
-      return call(a:fn, a:000)
-    finally
-      let &shellslash = shellslash
-    endtry
-  endif
-  return call(a:fn, a:000)
-endfunction
+let s:is_win = has('win32') || has('win64')
+if s:is_win && &shellslash
+  set noshellslash
+  let s:base_dir = expand('<sfile>:h:h')
+  set shellslash
+else
+  let s:base_dir = expand('<sfile>:h:h')
+endif
+if s:is_win
+  function! s:fzf_call(fn, ...)
+    if s:is_win
+      let shellslash = &shellslash
+      try
+        set noshellslash
+        return call(a:fn, a:000)
+      finally
+        let &shellslash = shellslash
+      endtry
+    endif
+    return call(a:fn, a:000)
+  endfunction
 
-function! s:fzf_getcwd()
-  return s:fzf_call('getcwd')
-endfunction
+  function! s:fzf_getcwd()
+    return s:fzf_call('getcwd')
+  endfunction
 
-function! s:fzf_fnamemodify(fname, mods)
-  return s:fzf_call('fnamemodify', a:fname, a:mods)
-endfunction
+  function! s:fzf_fnamemodify(fname, mods)
+    return s:fzf_call('fnamemodify', a:fname, a:mods)
+  endfunction
 
-function! s:fzf_expand(fmt)
-  return s:fzf_call('expand', a:fmt)
-endfunction
+  function! s:fzf_expand(fmt)
+    return s:fzf_call('expand', a:fmt)
+  endfunction
 
-function! s:fzf_tempname()
-  return s:fzf_call('tempname')
-endfunction
+  function! s:fzf_tempname()
+    return s:fzf_call('tempname')
+  endfunction
+
+  function! s:fzf_shellescape(path)
+    return s:fzf_call('shellescape', a:path)
+  endfunction
+else
+  function! s:fzf_getcwd()
+    return getcwd()
+  endfunction
+
+  function! s:fzf_fnamemodify(fname, mods)
+    return fnamemodify(a:fname, a:mods)
+  endfunction
+
+  function! s:fzf_expand(fmt)
+    return expand(a:fmt)
+  endfunction
+
+  function! s:fzf_tempname()
+    return tempname()
+  endfunction
+
+  function! s:fzf_shellescape(path)
+    return shellescape(a:path)
+  endfunction
+endif
 
 let s:default_layout = { 'down': '~40%' }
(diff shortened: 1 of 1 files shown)


plugin/fzf.vim
```diff
@@ -26,10 +26,62 @@ if exists('g:loaded_fzf')
...
+    try
+      set noshellslash
+      return expand(a:fmt)
+    finally
+      let &shellslash = shellslash
+    endtry
+  endif
+  return expand(a:fmt)
+endfunction
+
```
janlazo: I used `fzf#` prefix because I thought that [fzf.vim] can use it to handle `shellslash`. However, these new functions are  not suppose to change and are integral to [fzf]'s Vim/Neovim plugin. Outside of [fzf] and [fzf.vim], users shouldn't rely on these new functions. Do you prefer to minimize Vimscript dependencies between [fzf] and [fzf.vim]?

[fzf]: https://github.com/junegunn/fzf
[fzf.vim]: https://github.com/junegunn/fzf.vim

junegunn: Yes. Ideally, users of the core fzf plugin, such as fzf.vim, should only rely on `fzf#run` and `fzf#wrap` but nothing else, and `fzf#run` should handle all the escaping issues automatically. So I'm not happy that we have to change the implementation of `:FZF` command.

Can we do better? If it's not feasible to properly escape a raw options string for different platforms, we can consider extending `fzf#run` so that it can process array arguments like so:

```vim
call fzf#run({'options': [['--prompt', getcwd()], '--reverse', ['--expect', 'ctrl-a']]})
```

janlazo: What about `fzf#shellescape()`? Make it a script-local function as well?
For `:FZF`, what should I do to not change the implementation? Move it to `fzf#wrap` or `fzf#run`?

junegunn: I'll think about it and let you know. I need to find time to test the plugin on Windows before making the decision. Thanks.

janlazo:
We can move the shellescaping to `s:execute` by putting the command in a temporary shell script (or batch file) to get around cmd.exe. We can refactor `fzf#shellescape` to use `s:without_shellescape` and continue using strings for the fzf options because the prompt won't lose its quoting. Side benefit is that we bypass shellescaping issues for Neovim in Windows and have more flexibility with the launcher format (e.g. Vim allows `:!start cmd /c` but not `:!start /WAIT cmd /c`).

It's the simplest workaround with minimal changes to `:FZF`

junegunn:
If it means that it fixes the issue on Windows without requiring user scripts to change to deal with escaping issues, then I'm all for it.

janlazo:
If the prompt is escaped with single quotes, then the space character has to be escaped via `escape()`.
If the prompt is escaped with double quotes, then the double quote character has to be escaped via `escape()`.
`escape(path, ' \')` as suggested in `:help escape()` catches both cases but we don't have to do that since we're defaulting to backslash.

We can still use the batch file to avoid anymore complications. I can't use `:!start /WAIT` directly. It has to be `!cmd /c start /WAIT `. Seems to be an edge case in Vim especially on gVim.

Do you want a separate PR just for the launcher?

janlazo:
@junegunn Can you test if the latest commit requires no changes with FZF.vim? I only tested `:FZF`, `:Files`, and `:Colors` on both Vim and Neovim (I mentioned that it only supports builds starting neovim/neovim#6497 and I added the bare minimum code to launch it).

junegunn:
@janlazo Thanks for the update and sorry for the delay. I'll test it this weekend.

junegunn:
@janlazo Hi, I finally got a chance to test it on Windows box and I made a few changes to your version.

1. `fzf_expand` doesn't work with `<sfile>` (see `:help <sfile>` for details), so I had to call `expand` at the top-level
1. Remove runtime check of `s:is_win` in `s:fzf_*` functions by defining different versions of functions depending on `s:is_win` up-front.
1. `:FZF` works regardless of `shellslash`, but `:Files` breaks with `shellslash` since it does not use `fzf#shellescape`.
1. But instead of exposing/using `fzf#shellescape`, I decided that it's better to allow fzf functions to take `options` in list type. `:Files` will still have to be updated, but it doesn't have to know about `fzf#shellescape`.

The code is here:
https://gist.github.com/junegunn/6929954a64bead0e0a2aa1955cf0afd7

Can you review the changes and update your pull request if it makes sense?

janlazo:
@junegunn Thanks for the feedback. I'll review your gist and update the PR during the week. I'll implement the list type for options if you don't mind. I didn't see your implementation in the gist.

junegunn:
> I didn't see your implementation in the gist.

It's there. See s:evaluate_opts.

- junegunn

janlazo:
How can I know which options to escape if we use a list instead of a dict?

junegunn:
I'm not sure if I follow you. We allow `options` entry of the dictionary to be a list instead of a string. And the user code simply does not escape the arguments in the list, fzf#run or fzf#wrap does.
So one can write, `call fzf#run({'options': ['--prompt', 'Prompt "string">']})` without worry about with escaping issues. i.e. `call fzf#run({'options': '--prompt "Prompt \"string\">"'})`.

I suggest that you apply the entire code in the gist to your branch and see what I have changed.

justinmk:
FWIW since https://github.com/junegunn/fzf/pull/882 was created, Windows support for `:terminal`  materialized in https://github.com/neovim/neovim/pull/6383. I haven't tried it with fzf yet.

janlazo:
@junegunn Did `fzf_expand` break because of `fzf_call`? `fzf_expand` can work if it uses expand directly (I assume this is why your gist had it as well).

```viml
let s:is_win = has('win32') || has('win64')
if s:is_win
  function! s:fzf_expand(fmt)
    let shellslash = &shellslash
    try
      set noshellslash
      return expand(a:fmt)
    finally
      let &shellslash = shellslash
    endtry
  endfunction
else
  function! s:fzf_expand(fmt)
    return expand(a:fmt)
  endfunction
endif
let s:base_dir = s:fzf_expand('<sfile>:h:h')     " it worked with :p as well
```

@justinmk As long as it works in cmd.exe (colorless or not) along with with the terminal UI, I can test it. Do I have to compile it manually or I can get the appveyor build from https://github.com/neovim/neovim/pull/6383?

justinmk:
AppVeyor build should work.

junegunn:
@janlazo No, like I mentioned above, you can't properly expand `<sfile>` in a function call, see `:help <sfile>`. If you remove fzf executable from your %PATH% and put it under `bin` directory of fzf repo, you'll notice that the code will fail. And your branch still misses a few important points of my change (expanding `s:fzf_go`, not using `s:fzf_expand` in `s:cmd`). Why don't you just overwrite the code entirely with my gist and start from there? Or do you want me to commit my change at this point? Agree with [your comment](https://gist.github.com/junegunn/6929954a64bead0e0a2aa1955cf0afd7) here though.

junegunn:
@janlazo To elaborate, I'd like to prove that we can implement `:FZF` without using any of `s:fzf_*` functions as it's a reference implementation of FZF-based commands. `s:shortpath` still contains calls to those functions, but I believe we can remove them as well, as it only affects the prompt string, and I think it's okay to have `/`s in it if the user set `shellslash`.

janlazo:
@junegunn I overwrote the vim plugin with your gist as you requested. I added a few changes for `s:shortpath` and added a fallback for `s:fzf_call`. Is this enough or should we go further?

junegunn:
@janlazo 

> added a fallback for s:fzf_call 

Hmm, do we need `s:fzf_call` when `s:is_win` is false? Am I missing something?

I like what you suggested in here: https://gist.github.com/junegunn/6929954a64bead0e0a2aa1955cf0afd7

Having two versions of `s:fzf_call` only.

janlazo:
@junegunn I thought you wanted to skip a function call since we're defining the functions upfront. I prefer we stick with `s:fzf_` functions where we can for consistency so I added `fzf_call`. If something similar to `expand('<sfile>')` happens, we'll handle it upfront similar to `s:base_dir`.

janlazo:
Testing it in https://github.com/janlazo/fzf/commit/9d8ef7d42a3740711e7cfc7fea34a746db1ff7ae

plugin/fzf.vim
```diff
@@ -675,19 +727,24 @@ let s:default_action = {
...
+    if s:is_win && !&shellslash
+      let opts.dir = substitute(opts.dir, '/', '\\', 'g')
+    endif
+    let prompt = opts.dir
   else
-    let opts.options .= ' --prompt '.fzf#shellescape(s:shortpath())
+    let prompt = s:shortpath()
   endif
-  let opts.options .= ' '.join(args)
+  call extend(opts.options, ['--prompt', (s:is_win && !&shellslash) ? escape(prompt, '\') : prompt])
```
junegunn: One question: do we need escape call here? I thought `s:evaluate_opts` will take care of it using `shellescape`.

janlazo: It's a hack.
`'\\'` is a literal backslash and `\"` is a literal double quote in cmd.
We need an additional backslash to escape the last backslash of the prompt such that the double quotes can be removed by cmd to process the prompt.
Instead of just appending a backslash because it's a directory filepath, I escaped all the backslash of the filepath. You can test by replacing the ternary expression with
```viml
prompt.((s:is_win && !&shellslash) ? '\' : '')
```

janlazo:
@justinmk I want to have a TUI to compare with a GUI when running `:terminal`. TUI is not required but it gives me a visual assurance of the `:terminal` support in Windows even though Neovim abstracts the terminal issues for any UI anyway.

justinmk:
I don't understand. `:terminal` (programs running inside nvim) is orthogonal to TUI (nvim running in console.exe).

junegunn:
@janlazo So shellescape fails to handle trailing `\`, hmm. How about if we directly tackle the problem in `fzf_shellescape`?

```diff
diff --git a/plugin/fzf.vim b/plugin/fzf.vim
index 72be9f2..e4b388d 100644
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -44,10 +44,18 @@ if s:is_win
       let &shellslash = shellslash
     endtry
   endfunction
+
+  function! s:fzf_shellescape(path)
+    return substitute(s:fzf_call('shellescape', a:path), '[^\\]\zs\\"$', '\\\\"', '')
+  endfunction
 else
   function! s:fzf_call(fn, ...)
     return call(a:fn, a:000)
   endfunction
+
+  function! s:fzf_shellescape(path)
+    return shellescape(a:path)
+  endfunction
 endif
 
 function! s:fzf_getcwd()
@@ -66,10 +74,6 @@ function! s:fzf_tempname()
   return s:fzf_call('tempname')
 endfunction
 
-function! s:fzf_shellescape(path)
-  return s:fzf_call('shellescape', a:path)
-endfunction
-
 let s:default_layout = { 'down': '~40%' }
 let s:layout_keys = ['window', 'up', 'down', 'left', 'right']
 let s:fzf_go = s:base_dir.'/bin/fzf'
@@ -743,7 +747,7 @@ function! s:cmd(bang, ...) abort
   else
     let prompt = s:shortpath()
   endif
-  call extend(opts.options, ['--prompt', (s:is_win && !&shellslash) ? escape(prompt, '\') : prompt])
+  call extend(opts.options, ['--prompt', prompt])
   call extend(opts.options, args)
   call fzf#run(fzf#wrap('FZF', opts, a:bang))
 endfunction
```

Seems to work in all cases, while keeping `:FZF` code simple.

- `noshellslash`
    - [x] `:FZF`
    - [x] `:FZF C:\Program\ Files\`
    - [x] `:FZF C:/Program\ Files/`
- `shellslash`
    - [x] `:FZF`
    - [x] `:FZF C:/Program\ Files/`
    - [x] `:FZF C:\Program\ Files\`

junegunn:
@janlazo Merged. Thanks for your persistence on this 👍

janlazo:
@junegunn Thank you for your assistance throughout this PR. We can move forward to the docs (fzf.txt) and fzf.vim with minimal changes now because of `s:evaluate_opts` and list type for options.

junegunn:
@janlazo No problem. Let me update vim doc and README-VIM.md. I think we should update fzf.vim a little later, maybe after a new release of fzf, for the users who don't use the master branch of fzf.

janlazo:
@junegunn I forgot to handle `sxe` for `s:fzf_shellescape`. It includes `"` so `C:\Program "Files\` has to be escaped with `^`, which has to escaped with `\` as well. `s:fzf_shellescape` should output `^"C:\Program \^"Files\^\^"`. `shellescape` by itself doesn't handle this.

Reference:
(Check the "A better method of quoting" section for escaping the double quotes with `^`)
https://blogs.msdn.microsoft.com/twistylittlepassagesallalike/2011/04/23/everyone-quotes-command-line-arguments-the-wrong-way/

junegunn:
@janlazo Thanks for the heads-up. Hmm, is it trivial to fix?

janlazo:
Non-trivial if using `^` to escape `"` because `shellescape` and `escape` are out.
Trivial if literal quotes are allowed via `\"` for the prompt. We can use `substitute` or `escape` to replace `shellescape` in `s:fzf_shellescape`.

janlazo:
@justinmk Windows TUI for Neovim for 0.2.x works with the vim plugin for fzf (via jobstart) but the scroll up bug is still unresolved. Can neovim/neovim#6383 for `:terminal` be merged for 0.2.1? If not, can the Appveyor build be restarted?

Issue #908, opened 2017-04-29
Use mattn's go-isatty, go-runewidth and go-shellwords


tomfitzhenry:
- Category
    - [x] fzf binary
- OS
    - [x] Linux

I'm a big fan of fzf, and so am currently packaging fzf for Debian: https://github.com/tomfitzhenry/pkg-fzf

It's working, but there's a slight issue.

fzf depends on:
* https://github.com/junegunn/go-isatty
* https://github.com/junegunn/go-runewidth
* https://github.com/junegunn/go-shellwords

This isn't necessarily a problem, but I noticed that:
* each of those libraries is a fork: [go-isatty](https://github.com/mattn/go-isatty), [go-runewidth](https://github.com/junegunn/go-runewidth), [go-shellwords](https://github.com/junegunn/go-shellwords)
* your forks seem to have no commits that aren't in upstream
* upstream has commits that aren't in your forks
* Debian has already packaged the three upstreams: [golang-github-mattn-go-isatty-dev](https://packages.debian.org/sid/golang-github-mattn-go-isatty-dev), [https://packages.debian.org/sid/golang-github-mattn-go-runewidth-dev](https://packages.debian.org/sid/golang-github-mattn-go-runewidth-dev), [https://packages.debian.org/sid/golang-github-mattn-go-shellwords-dev](https://packages.debian.org/sid/golang-github-mattn-go-shellwords-dev)

Consequently does it make more sense to have fzf depend on mattn's libraries?

junegunn:
I just wanted to secure the correctness of the build by using my own forks. When I started rewriting fzf in Go, there was no good way to manage the dependencies and that was my choice. Theoretically, upstream repositories can make breaking changes at any time that might affect the behaviour of fzf and they can even be deleted from GitHub by their authors. By using my own forks, I can be always be sure that fzf is unaffected by such events.

I know that those events are probably not going to happen, and now I use a script (https://github.com/junegunn/fzf/blob/master/src/deps) for checking out the exact revisions, but with forks, simple `go get` build will always be reproducible.

tomfitzhenry:
Thanks for the prompt response.

I can see the benefits to depending only on your code, especially if  upstream starts making breaking changes.

[...]

PR #909, opened 2017-04-30
Manage dependencies with godep

tomfitzhenry:
As described in issue #908, here's a PR to manage dependencies via godep.

See [syncthing](https://github.com/syncthing/syncthing) and [pgweb](https://github.com/sosedoff/pgweb) for two other projects that use godep.


tomfitzhenry:
Personally I'm not keen on the idea of including the source of dependencies in VCS, as is what happens with godep. Glide appears to be an alternative, that doesn't rely on this.

junegunn:
Thanks, but I really don't like having to check in all the dependencies. Is it a hard requirement of Godeps? That's really strange to me since most other dependency tools I have used only tell you to maintain a single file specifying the dependencies. (e.g. pom.xml, Gemfile, project.clj, Cargo.toml, etc.)

Another thing I'd like to mention is that I would like to keep backward compatibility of build scripts, i.e. I don't want to care about `GOPATH`, I just want to `git clone` the repo and `cd src && make install` to get the binary.

I'll see if Glide fits the use case.

junegunn:
Closed by #934. Thanks.

PR #916, opened 2017-05-07
[vim] Replace s:fzf_shellescape and s:shellesc with fzf#shellescape

janlazo:
Continuation of #896.

Previous PR cannot escape double quotes properly in `s:evaluate_opts` because of escaping assumptions made by cmd.exe. Instead on relying `shellescape` which relies on `shellslash`, use `s:shellesc` to escape with `\` in Unix and `\^` in Windows. For now, only `"` is escaped this way. The last backslash is escaped the same way as in `s:fzf_shellescape`.

The outer double quotes are escaped via `^` so cmd.exe knows where the escaping starts and ends. Without it, shell redirection with `>` fails even if the double quote character is escaped.

```dosbatch
> fzf --prompt "C:/Program \" Files (x86)/" "--multi" > tmp.txt
unknown option: >
```

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -46,7 +46,8 @@ if s:is_win
   endfunction
 
   function! s:shellesc(arg)
-    let escaped = substitute(a:arg, '[">]', '^&', 'g')
+    let escaped = substitute(a:arg, '>', '^&', 'g')
+    let escaped = substitute(escaped, '"', '\\^&', 'g')
     let escaped = substitute(escaped, '\\\\^', '\\^&', 'g')
     return '^"'.substitute(escaped, '[^\\]\zs\\$', '\\\\', '').'^"'
   endfunction
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -50,23 +50,20 @@ else
   endfunction
 endif
 
-function! s:shellesc_sh(arg)
-  return '"'.substitute(a:arg, '"', '\\"', 'g').'"'
-endfunction
-
 function! s:shellesc_cmd(arg)
   let escaped = substitute(a:arg, '[&|<>()@^]', '^&', 'g')
+  let escaped = substitute(escaped, '%', '%%', 'g')
   let escaped = substitute(escaped, '"', '\\^&', 'g')
   let escaped = substitute(escaped, '\\\+\(\\^\)', '\\\\\1', 'g')
   return '^"'.substitute(escaped, '[^\\]\zs\\$', '\\\\', '').'^"'
 endfunction
 
-function! fzf#shellesc(arg, ...)
+function! fzf#shellescape(arg, ...)
   let shell = get(a:000, 0, &shell)
   if shell =~# 'cmd.exe$'
     return s:shellesc_cmd(a:arg)
   endif
-  return s:shellesc_sh(a:arg)
+  return shellescape(a:arg)
 endfunction
 
 function! s:fzf_getcwd()
@@ -119,7 +116,7 @@ function! s:fzf_exec()
       throw 'fzf executable not found'
     endif
   endif
-  return s:is_win ? s:exec : fzf#shellesc(s:exec)
+  return s:is_win ? s:exec : fzf#shellescape(s:exec)
 endfunction
 
 function! s:tmux_enabled()
@@ -257,7 +254,7 @@ endfunction
 
 function! s:evaluate_opts(options)
   return type(a:options) == type([]) ?
-        \ join(map(copy(a:options), 'fzf#shellesc(v:val)')) : a:options
+        \ join(map(copy(a:options), 'fzf#shellescape(v:val)')) : a:options
 endfunction
 
 " [name string,] [opts dict,] [fullscreen boolean]
@@ -304,7 +301,7 @@ function! fzf#wrap(...)
     if !isdirectory(dir)
       call mkdir(dir, 'p')
     endif
-    let history = s:is_win ? fzf#shellesc(dir.'\'.name) : s:escape(dir.'/'.name)
+    let history = s:is_win ? fzf#shellescape(dir.'\'.name) : s:escape(dir.'/'.name)
     let opts.options = join(['--history', history, opts.options])
   endif
 
@@ -356,7 +353,7 @@ try
   if !has_key(dict, 'source') && !empty($FZF_DEFAULT_COMMAND)
     let temps.source = s:fzf_tempname().(s:is_win ? '.bat' : '')
     call writefile((s:is_win ? ['@echo off'] : []) + split($FZF_DEFAULT_COMMAND, "\n"), temps.source)
-    let dict.source = (empty($SHELL) ? &shell : $SHELL) . (s:is_win ? ' /c ' : ' ') . fzf#shellesc(temps.source)
+    let dict.source = (empty($SHELL) ? &shell : $SHELL) . (s:is_win ? ' /c ' : ' ') . fzf#shellescape(temps.source)
   endif
 
   if has_key(dict, 'source')
@@ -367,7 +364,7 @@ try
     elseif type == 3
       let temps.input = s:fzf_tempname()
       call writefile(source, temps.input)
-      let prefix = (s:is_win ? 'type ' : 'cat ').fzf#shellesc(temps.input).'|'
+      let prefix = (s:is_win ? 'type ' : 'cat ').fzf#shellescape(temps.input).'|'
     else
       throw 'Invalid source type'
     endif
@@ -431,7 +428,7 @@ function! s:fzf_tmux(dict)
     endif
   endfor
   return printf('LINES=%d COLUMNS=%d %s %s %s --',
-    \ &lines, &columns, fzf#shellesc(s:fzf_tmux), size, (has_key(a:dict, 'source') ? '' : '-'))
+    \ &lines, &columns, fzf#shellescape(s:fzf_tmux), size, (has_key(a:dict, 'source') ? '' : '-'))
 endfunction
 
 function! s:splittable(dict)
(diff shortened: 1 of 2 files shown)


janlazo:
I can't think of better way to escape the backslash for `\\"` so the double backslashes are back.

janlazo:
@junegunn Any issues with using `s:shellesc` instead of `shellescape`? I can move on to the non-Windows version of `s:shellesc` if there are no issues.

junegunn:
Yeah, looks like we don't need both `s:fzf_shellescape` and `s:shellesc`. But I'm not familiar with Windows and I can't verify the new escaping code for windows, can you show me some examples?

janlazo:
The following examples mainly focus on the escape characters: `\, ^, "`
- `^` is to tell cmd.exe to interpret the character literally and not make any assumption yet.
- `\` is for escaping escape characters, primarily `\` and `^`
- `"` is for wrapping the argument as a whole string

Metacharacters (`:h sxe`) run as expected in cmd.exe inside `^"^"` so they have to be escaped with `^`.
Since `^` is already used for the outer double quotes, inner double quotes need their `^` to be escaped and `^^"` doesn't work which leaves `\^`.
Same as `s:fzf_shellescape`, the final backslash has to be escaped because `\^"` is valid and we need the closing `^"` to not be escaped with `\`.

```dosbatch
:: Following comments before each command are the resulting prompt
:: If fzf doesn't run, output the error below the command

:: hello>
> fzf "--prompt" "hello>" "--multi" "--query" "world" > foobar.txt

:: he\^l^^l\\o>
> fzf "--prompt" "he\^l^^l\\o>" "--multi" "--query" "world" > foobar.txt

> fzf "--prompt" "he"llo>" "--multi" "--query" "world" > foobar.txt
The filename, directory name, or volume label syntax is incorrect

:: he"llo> --multi --query world
> fzf "--prompt" "he""llo>" "--multi" "--query" "world" > foobar.txt

> fzf "--prompt" "he\"llo>" "--multi" "--query" "world" > foobar.txt
The filename, directory name, or volume label syntax is incorrect

> fzf "--prompt" "he^"llo>" "--multi" "--query" "world" > foobar.txt
The filename, directory name, or volume label syntax is incorrect

> fzf "--prompt" "he\^"llo>" "--multi" "--query" "world" > foobar.txt
The filename, directory name, or volume label syntax is incorrect

> fzf "--prompt" ^"hello>" "--multi" "--query" "world" > foobar.txt
The filename, directory name, or volume label syntax is incorrect

:: hello>^
> fzf "--prompt" "hello>^" "--multi" "--query" "world" > foobar.txt

:: hello
> fzf "--prompt" ^"hello>^" "--multi" "--query" "world" > foobar.txt

:: hello>
> fzf "--prompt" ^"hello^>^" "--multi" "--query" "world" > foobar.txt

:: hello^>^ --multi --query world
> fzf "--prompt" ^"hel"lo^>^" "--multi" "--query" "world" > foobar.txt

:: hello> --multi --query world
> fzf "--prompt" ^"hel^"lo^>^" "--multi" "--query" "world" > foobar.txt

:: hel"lo>
> fzf "--prompt" ^"hel\^"lo^>^" "--multi" "--query" "world" > foobar.txt

:: hel"lo>" --multi --query world
> fzf "--prompt" ^"hel\^"lo^>\^" "--multi" "--query" "world" > foobar.txt

:: hel"lo>\
> fzf "--prompt" ^"hel\^"lo^>\\^" "--multi" "--query" "world" > foobar.txt

:: dir
> fzf "--prompt" ^"dir && ag fzf %USERPROFILE%\vimfiles ^" "--multi" "--query" "world" > foobar.txt
ERR: Error opening directory --multi : No such file or directory
```

junegunn:
Thank you very much for the detailed explanation, do you think this is ready for merge?

janlazo:
I'm testing this PR with https://github.com/junegunn/fzf.vim/pull/372 so I want to make sure that both versions of `s:shellesc` are resolved and any changes to fzf.vim are minimized if using list type for options.
For the Windows version of `s:shellesc`, it cannot handle something like `\\\\\^"' which should be simplified to `\^\\^"` and finalized in cmd.exe as `\"`. Because anything can happen with `^` and `\`, and each can break each other or affect the outer `^"`, it needs more testing.
For the default version, it has to escape the same set of characters to not lose features. The user shouldn't have to use `tempname` unless necessary (`Helptags` in Windows) to run the command.

I recommend that we keep both PRs open until you want to make a new release to minimize upgrade issues for the users.

janlazo:
Because of `s:evaluate_opts`, we can test `s:shellesc` or `s:fzf_shellescape` for the prompt with vader but I don't see any test cases for the prompt in https://github.com/junegunn/fzf/blob/master/test/fzf.vader. We don't have appveyor setup so we can't automate testing for the Windows version of `s:shellesc` for `FZF` and fzf.vim.

Do you have time to setup appveyor so we can convert the test cases in https://github.com/junegunn/fzf/pull/916#issuecomment-302292067 to vader tests?

junegunn:
Well I have no prior experience with appveyor and honestly I'm not sure if I can find time (and motivation) to look into that any time soon.

junegunn:
Anyway, I like the idea of setting up Vader test cases for the function. And I can live with having the function visible for testing, i.e. `fzf#_shellesc`, or maybe both `fzf#_shellesc_unix` and `fzf#_shellesc_win`)

janlazo:
How about using `shell` instead?

```viml
function! s:shellesc_sh(arg)
  return '"'.substitute(a:arg, '"', '\\"', 'g').'"'
endfunction

function! s:shellesc_cmd(arg)
  let escaped = substitute(a:arg, '[&|<>()@^]', '^&', 'g')
  let escaped = substitute(escaped, '"', '\\^&', 'g')
  let escaped = substitute(escaped, '\\\+\(\\^\)', '\\\\\1', 'g')
  return '^"'.substitute(escaped, '[^\\]\zs\\$', '\\\\', '').'^"'
endfunction

function! fzf#shellesc(arg, ...)
  let shell = exists('a:1') ? a:1 : &shell
  if shell =~# 'cmd.exe$'
    return s:shellesc_cmd(a:arg)
  endif
  return s:shellesc_sh(a:arg)
endfunction
```

junegunn:
I like that.

junegunn:
LGTM, ready for merge?

test/fzf.vader
```diff
@@ -147,6 +147,23 @@ Execute (fzf#wrap):
   let opts = fzf#wrap({})
   Assert opts.options =~ '^--color=fg:'
 
+Execute (fzf#shellesc with sh):
+  AssertEqual '""', fzf#shellesc('', 'sh')
+  AssertEqual '"\"\""', fzf#shellesc('""', 'sh')
+  AssertEqual '"foobar>"', fzf#shellesc('foobar>', 'sh')
+  AssertEqual '"\\""', fzf#shellesc('\"', 'sh')
```
janlazo: The inner double quote is unescaped but this case is handled in `s:shellesc_cmd`.
Need to handle backslashes preceding the escaped double quote.

junegunn: I don't remember why I added `s:shellesc` when there's builtin `shellescape`. What do you think about removing `s:shellesc_sh` and just using `shellescape` instead?

plugin/fzf.vim
```diff
@@ -44,20 +44,31 @@ if s:is_win
...
     return call(a:fn, a:000)
   endfunction
-
-  function! s:fzf_shellescape(path)
-    return shellescape(a:path)
-  endfunction
 endif
 
+function! s:shellesc_sh(arg)
+  return '"'.substitute(a:arg, '"', '\\"', 'g').'"'
```
junegunn:
```vim
return '"'.escape(a:arg, '\"').'"'
```

What do you think?

janlazo: I can't test in a linux box atm but I had no issues with this change in cygwin (using ConEmu, not mintty). It didn't collide with `s:escape`. Tested with this branch, https://github.com/janlazo/fzf/tree/vim/shellesc_cygwin_sh

If there's no issues on your end (including fzf.vim), I'll commit your suggestion (not the cygwin changes) and update the test cases.

test/fzf.vader
janlazo: Do you remember the issue with `shellescape` for `$ENV` vars and tmux integration?

plugin/fzf.vim
junegunn: It works well. Please update as above. Thanks.

test/fzf.vader
junegunn: No I was not referring to tmux integration as it's no longer the default (thanks to `--height`).

Maybe I was confused, I thought something like `:call fzf#run({'source': 'echo $PATH'})` wouldn't work if we use builtin `shellescape`, but it is unaffected.

plugin/fzf.vim
junegunn: No, wait, I guess we should probably just use builtin `shellescape` if it's okay.

test/fzf.vader
janlazo: `shellescape` in cygwin vim uses single quotes so it doesn't expand env vars.

janlazo: Say that the user wants to use `$VIMRUNTIME` as their prompt. What should appear, `$VIM` or the resolved path after variable expansion?

junegunn: Yeah, actually this was not properly discussed when we added support for list typed options. Currently `call fzf#run({'options': ['--prompt', '$USER']})` gives literal `$USER` due to the use of builtin `shellescape`, and I think it makes sense as one can manually expand the var when needed: `call fzf#run({'options': ['--prompt', expand('$USER')]})`.

I briefly looked at the other places where `s:shellesc` is used and it appeared that not expanding env vars anymore will not break anything.

janlazo: This PR becomes **Use fzf#shellescape for escaping shell arguments** to deprecate `s:shellesc` and the Window's version of `s:fzf_shellescape`. To be consistent with the built-in `shellescape`, I assume Windows-style env vars should  be escaped.

janlazo: The first `escape` in `s:execute` to handle `%#!` breaks `FZF` because fzf is run in the batchfile.
`%` are escaped as `%%` in batchfiles and `^%` in the interactive shell.

janlazo:
The first escape in s:execute to handle %#! breaks FZF because fzf is run in the batchfile.
% are escaped as %% in batchfiles and ^% in the interactive shell.

junegunn:
It's a different issue, right? Escaping of `%#!` is only needed when we execute the command using `:!`, so I think the following fix should do.

```vim
let escaped = (a:use_height || s:is_win) ? a:command : escape(substitute(a:command, '\n', '\\n', 'g'), '%#!')
```

janlazo:
@junegunn It's done with the test cases updated.

junegunn:
Thank you 👍

PR #921, opened 2017-05-12
Move vim function definitions to autoload dir

lucc:
All function definitions are moved into the autoload namespace fzf#core.
The previously defined and publicly visible functions fzf#run and
fzf#wrap still exist for compatibility and just wrap the corresponding
function in fzf#core.

The main purpose of this is to reduce startup time.  Most of the code of
this plugin will now only be loaded when it is called.

The functions where not simply moved to the fzf autoload namespace
because in this case the loading of the plugin/fzf.vim file from the
junegunn/fzf.vim repository would have triggered the autoloading
mechanism for them.  This would have defeated the purpose mentioned
above.

For me this did reduce the startup time for the files from this repo from 3.7 to 0.14+0.16=0.3 ms.


davidosomething:
this is nice, i would use some of the autoload fns myself, like the shortpath one since i have duplicated it in my own vimrc

lucc:
rebased

z0rc:
@junegunn any update? This PR is 6 months old. Do you have any concerns or issues with it?

junegunn:
Thanks, but as the maintainer of the project, I don't want to change the project structure at this moment unless it's necessary. There are several fzf packages out there (Homebrew/Linuxbrew, Arch linux packages, etc) that I'm not in charge of (probably there are some more that I'm not even aware of), and I have no way to ensure that the maintainers of those packages to update their package specs accordingly whenever I change the structure of the project. And I have noticed that some users [manually source `fzf.vim`](https://github.com/junegunn/fzf.vim/issues/456#issuecomment-340408714) instead of changing `&runtimepath`. If anything breaks, people will come to ask me why it stopped working, and I have to do my own research to answer them, and honestly, it's a tiring process.

On my laptop, loading fzf.vim takes less than a millisecond which is less 1% of the total startup time of Vim, so it's really not an issue to me. And I don't want to spend my time dealing with any unnecessary confusion or chores that this change might cause just to reduce a small startup time that I can't even notice.

So I'm going to close this. If you really need this, I suggest that you use your fork.

shaunbrady:
@junegunn First, thank you for the project.  Your justification for not fixing this seems sound at a glance.  What I think would be a good compromise is to indicate that fzf.vim (the repo) relies on the base plugin of the .fzf/ install (fzf repo).  I built the binary from scratch, got it working on the command line, so I figured that was all the reason you were having the fzf repo as part of the vim plugin install.  It seemed redundant, so I omitted it.  I think a docs update (of fzf.vim/README.md) would be low hanging fruit, and make sure that people get their setup fast so they can fall in love with the project.  I'll do a little research if something doesn't work out of the box when it's something I want to try out; I won't do a lot.

PR #929, opened 2017-05-21
[vim] use utf-8 for cmd.exe

janlazo:
Some commands in fzf.vim use utf-8 characters but cmd.exe doesn't use utf-8 by default.
Permanently changing the encoding in cmd.exe or powershell affects other terminal programs so a temporary variable is required to store the current encoding and reverting to it after fzf runs successfully.
Issue is if the command fails, then terminal users are left in a bad state with a permanently changed encoding.

This change is limited to Windows and GUIs that use external terminals for fzf should be unaffected if the fzf fails.

This PR should resolve fzf.vim commands that use utf-8 characters such as `Commands`

Reference:
https://ss64.com/nt/chcp.html

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -46,7 +46,7 @@ if s:is_win
   endfunction
 
   " Use utf-8 for fzf.vim commands
-: " Return array of shell commands for cmd.exe
+  " Return array of shell commands for cmd.exe
   function! s:wrap_cmds(cmds)
     return ['@echo off', 'for /f "tokens=4" %%a in (''chcp'') do set origchcp=%%a', 'chcp 65001 > nul'] +
           \ (type(a:cmds) == type([]) ? a:cmds : [a:cmds]) +
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -48,9 +48,9 @@ if s:is_win
   " Use utf-8 for fzf.vim commands
 : " Return array of shell commands for cmd.exe
   function! s:wrap_cmds(cmds)
-    return ['@echo off', 'for /f "tokens=4" %%a in (''chcp'') do set origchcp=%%a', 'chcp 65001'] +
+    return ['@echo off', 'for /f "tokens=4" %%a in (''chcp'') do set origchcp=%%a', 'chcp 65001 > nul'] +
           \ (type(a:cmds) == type([]) ? a:cmds : [a:cmds]) +
-          \ ['chcp %origchcp%']
+          \ ['chcp %origchcp% > nul']
   endfunction
 else
   function! s:fzf_call(fn, ...)
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -44,9 +44,21 @@ if s:is_win
       let &shellslash = shellslash
     endtry
   endfunction
+
+  " Use utf-8 for fzf.vim commands
+: " Return array of shell commands for cmd.exe
+  function! s:wrap_cmds(cmds)
+    return ['@echo off', 'for /f "tokens=4" %%a in (''chcp'') do set origchcp=%%a', 'chcp 65001'] +
+          \ (type(a:cmds) == type([]) ? a:cmds : [a:cmds]) +
+          \ ['chcp %origchcp%']
+  endfunction
 else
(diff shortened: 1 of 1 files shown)


plugin/fzf.vim
```diff
@@ -506,7 +518,7 @@ function! s:execute(dict, command, use_height, temps) abort
   endif
   if s:is_win
     let batchfile = s:fzf_tempname().'.bat'
```
junegunn: Unrelated to the issue, shouldn't we have to add this to `temps` so that it's deleted at the end? (`let temps.batchfile = ...`)

janlazo: Yes, if the cleanup happens after fzf runs.

junegunn:
I'm seeing "Active code page: 65001" message flashing in the cmd window before fzf starts. Is it possible to suppress the message?

janlazo:
Yes, by redirecting output to NUL for both `chcp` commands. Do you want to suppress stderr as well?

```dosbatch
:: Redirect stdout to Windows's /dev/null
chcp 65001 > nul
:: Redirect stdout and stderr to nul
chcp 65001 > nul 2>&1
```

junegunn:
> Do you want to suppress stderr as well?

I don't know, what do you think? (I'm not a Windows user)

janlazo:
In `:h 'shellredir'` for Vim and Neovim, `>%s 2>&1` is used for cmd.exe

I assume that Windows users prefer the GUI over the command line unless they use ConEmu/Cmder or a similarly capable terminal. Vim and Neovim ship a GUI for Windows builds and from where I work, Windows users use GVim. ConEmu users are likely to be familiar with cmd/powershell/cygwin quirks and can use `chcp 65001` on startup because it's recommended (look for `chcp utf-8` in Settings -> Startup -> Environment). However, setting up truecolor support for terminal vim in ConEmu is a PITA (requires some research to setup correctly) and is non-standard (to this day, I can't get emacs to use 256 colors without outputing garbled characters). The path of least resistance for users that use 256 or truecolor colorschemes and want to use utf-8 without breaking other programs is the GUI. Thus, the external terminal (cmd.exe) is likely where fzf will run in Windows.

I think most Windows users prefer an all-or-nothing approach; users see fzf only. If there is an error, then the terminal shouldn't instantly disappear so they can copy/paste the debug info. I suppose we can bypass this terminal problem by redirecting to Vim/Neovim via a tempfile so they can leave it in a buffer.

janlazo:
I rebased on master to resolve conflicts for `fzf#shellescape`.

Is it safe to assume that, in Linux/Mac, most terminals support/use UTF-8?
In my bashrc, I set `LANG`, `LANGUAGE`, `LC_ALL` to `en_US.UTF-8` so I don't know what to do for the default version of `s:wrap_cmds`.

plugin/fzf.vim
junegunn: Yes it does.

junegunn:
> Is it safe to assume that, in Linux/Mac, most terminals support/use UTF-8?

I think so. But it doesn't matter anyway since Linux and Mac are unaffected by this change, am I right?

plugin/fzf.vim
janlazo: Let's add it to `temps` then.
The code change should be a minimal (one-liner?) so can you commit the fix?
I'd rather not handle it in this pull request.

janlazo:
Yes. Default version of `s:wrap_cmds` returns the passed argument.

plugin/fzf.vim
junegunn: Sure, will do it myself.

junegunn:
I'll merge this tonight after testing it on Windows, thanks.

plugin/fzf.vim
```diff
@@ -44,10 +44,22 @@ if s:is_win
       let &shellslash = shellslash
     endtry
   endfunction
+
+  " Use utf-8 for fzf.vim commands
+: " Return array of shell commands for cmd.exe
```
junegunn: A minor nit. `:` at the beginning of this line.

janlazo: Done.

junegunn:
Merged, thanks!

Issue #910, opened 2017-05-01
fzf doesn't accept input


tw4452852:
- Category
    - [x] fzf binary
    - [ ] fzf-tmux script
    - [ ] Key bindings
    - [ ] Completion
    - [ ] Vim
    - [ ] Neovim
    - [ ] Etc.
- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Windows Subsystem for Linux
    - [ ] Etc.
- Shell
    - [ ] bash
    - [x] zsh
    - [ ] fish

Today I just upgrade the fzf to 2d61691 and run fzf on current directory, but it doesn't accept any input..

I just bisect it, found that this commit is the culprit: 6ccc12c332f97f6be9487d8289c46964e2d41bd3,
Any help? Thanks.

junegunn:
Please provide more details about your environment: OS distro, terminal emulator, `fzf --version`, etc. Do you have the same issue when you do `fzf --height 30%`?

6ccc12c332f97f6be9487d8289c46964e2d41bd3 was made in January, and no one has reported this issue so far. Hmm.

tw4452852:
@junegunn 

```
> fzf --version
0.16.7
> uname -a
Linux totorow 4.4.6-gentoo #1 SMP Wed Apr 26 15:08:41 CST 2017 x86_64 Intel(R) Core(TM) i5-4590 CPU @ 3.30GHz GenuineIntel GNU/Linux
```

And I use `urxvt`, its version is: `rxvt-unicode (urxvt) v9.21 - released: 2014-12-31`

> Do you have the same issue when you do fzf --height 30%?

Yes, it doesn't work as expected, it just hang:

```
> fzf --height 30%

```

And `ctrl-c` doesn't work, I have to kill it in another terminal.

Do you need any other information?

junegunn:
> it just hang

1. So you see nothing? Same issue with `echo foo | fzf --height 30%`?

2. Does the following command report the position of your cursor?

```sh
echo -e "\x1b[6n"
```

3. And can you test it with a different terminal emulator? (I'm not saying that `urxvt` is not supported. I can't reproduce the problem with urxvt installed on my mac.)

tw4452852:
> So you see nothing? Same issue with echo foo | fzf --height 30%?
Yes.

> Does the following command report the position of your cursor?

```
/home/tw/golib/src/github.com/junegunn/fzf [tw@totorow] 
[...]

PR #931, opened 2017-05-23
use read syscall directly to get character

tw4452852:
Fix issue #910 .

### DIFF
--- a/src/tui/light.go
+++ b/src/tui/light.go
@@ -251,8 +251,9 @@ func (r *LightRenderer) updateTerminalSize() {
 
 func (r *LightRenderer) getch(nonblock bool) (int, bool) {
 	b := make([]byte, 1)
+	fd := r.fd()
 	util.SetNonblock(r.ttyin, nonblock)
-	_, err := r.ttyin.Read(b)
+	_, err := util.Read(fd, b)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


junegunn:
Thanks, I'd prefer to have the comment as the commit message, can you move it?

tw4452852:
Done.

junegunn:
Oh well, it doesn't compile for Windows.

```
> GOOS=windows make release
cd /Users/jg/.vim/plugged/fzf/gopath/src/github.com/junegunn/fzf/src && go get -tags ""
# github.com/junegunn/fzf/src/tui
tui/light.go:256: cannot use fd (type int) as type syscall.Handle in argument to syscall.Read
make: *** [deps] Error 2
```

We might need to move getch code to util_unix and util_windows.

tw4452852:
Hmm, I just work out `syscall.Read` for different platforms instead.

BTW, `light.go` isn't for windows apparently, because it gets input from `/dev/tty` which doesn't exist on windows. For now, we disable `--height` option on windows, so `LightRenderer` won't be used on windows. I think we should add `+build !windows` to `light.go`.

junegunn:
Thanks, you're right, I think it's good enough for now. I'll merge your commit to devel branch first.

PR #940, opened 2017-06-07
Updates CHANGELOG with unix-line-discard+top

uri:
Also changed the example binding for `unix-word-rubout` to *ctrl-w* instead of *ctrl-u*


junegunn:
Yeah thanks, how about if we combine the two examples?

uri:
Cool, made the change.

junegunn:
Thanks!

PR #941, opened 2017-06-08
[vader] uncomment test case to escape % in cmd.exe

janlazo:
I commented out a `shellescape` test case for cmd.exe in PR #916.
This PR uncomments it and fixes the runaway single quote.


junegunn:
Thanks!

PR #944, opened 2017-06-12
[fish] Accept starting dir for <M-c> key binding

ipwnponies:
This is an enhancement for fish shell key bindings. Currently the <M-c> keybinding only searches subdirectories of cwd. This allows for specifying the root directory to start search from.

```fish
ipwnponies$ pwd                                                                                                              
/Users/ipwnponies/repos/fzf/src/algo

ipwnponies$ ~/repos/fzf<M-C>
 >
   11/11
   /Users/ipwnponies/repos/fzf/bin
   /Users/ipwnponies/repos/fzf/doc
   /Users/ipwnponies/repos/fzf/man
   /Users/ipwnponies/repos/fzf/man/man1
   /Users/ipwnponies/repos/fzf/plugin
   /Users/ipwnponies/repos/fzf/shell
   /Users/ipwnponies/repos/fzf/src
   /Users/ipwnponies/repos/fzf/src/algo
   /Users/ipwnponies/repos/fzf/src/tui
   /Users/ipwnponies/repos/fzf/src/util
 > /Users/ipwnponies/repos/fzf/test

ipwnponies$ ..<M-C>
>
  3/3
> ../algo
  ../tui
  ../util
```

### DIFF
--- a/shell/key-bindings.fish
+++ b/shell/key-bindings.fish
@@ -4,14 +4,9 @@ function fzf_key_bindings
 
   # Store current token in $dir as root for the 'find' command
   function fzf-file-widget -d "List files and folders"
-    set -l dir (commandline -t)
-    # The commandline token might be escaped, we need to unescape it.
-    set dir (eval "printf '%s' $dir")
-    if [ ! -d "$dir" ]
-      set dir .
-    end
-    # Some 'find' versions print undesired duplicated slashes if the path ends with slashes.
-    set dir (string replace --regex '(.)/+$' '$1' "$dir")
+    set -l commandline (__fzf_parse_commandline)
+    set -l dir $commandline[1]
+    set -l fzf_query $commandline[2]
 
     # "-path \$dir'*/\\.*'" matches hidden files/folders inside $dir but not
     # $dir itself, even if hidden.
@@ -24,7 +19,7 @@ function fzf_key_bindings
     set -q FZF_TMUX_HEIGHT; or set FZF_TMUX_HEIGHT 40%
     begin
       set -lx FZF_DEFAULT_OPTS "--height $FZF_TMUX_HEIGHT --reverse $FZF_DEFAULT_OPTS $FZF_CTRL_T_OPTS"
-      eval "$FZF_CTRL_T_COMMAND | "(__fzfcmd)" -m" | while read -l r; set result $result $r; end
+      eval "$FZF_CTRL_T_COMMAND | "(__fzfcmd)' -m --query "'$fzf_query'"' | while read -l r; set result $result $r; end
     end
     if [ -z "$result" ]
       commandline -f repaint
@@ -51,15 +46,26 @@ function fzf_key_bindings
   end
 
   function fzf-cd-widget -d "Change directory"
+    set -l commandline (__fzf_parse_commandline)
+    set -l dir $commandline[1]
+    set -l fzf_query $commandline[2]
+
     set -q FZF_ALT_C_COMMAND; or set -l FZF_ALT_C_COMMAND "
-    command find -L . -mindepth 1 \\( -path '*/\\.*' -o -fstype 'sysfs' -o -fstype 'devfs' -o -fstype 'devtmpfs' \\) -prune \
-    -o -type d -print 2> /dev/null | cut -b3-"
+    command find -L \$dir -mindepth 1 \\( -path \$dir'*/\\.*' -o -fstype 'sysfs' -o -fstype 'devfs' -o -fstype 'devtmpfs' \\) -prune \
+    -o -type d -print 2> /dev/null | sed 's@^\./@@'"
     set -q FZF_TMUX_HEIGHT; or set FZF_TMUX_HEIGHT 40%
     begin
       set -lx FZF_DEFAULT_OPTS "--height $FZF_TMUX_HEIGHT --reverse $FZF_DEFAULT_OPTS $FZF_ALT_C_OPTS"
-      eval "$FZF_ALT_C_COMMAND | "(__fzfcmd)" +m" | read -l result
-      [ "$result" ]; and cd $result
+      eval "$FZF_ALT_C_COMMAND | "(__fzfcmd)' +m --query "'$fzf_query'"' | read -l result
+
+      if [ -n "$result" ]
+        cd $result
+
+        # Remove last token from commandline.
+        commandline -t ""
+      end
     end
+
     commandline -f repaint
   end
 
@@ -82,4 +88,47 @@ function fzf_key_bindings
     bind -M insert \cr fzf-history-widget
     bind -M insert \ec fzf-cd-widget
   end
+
+  function __fzf_parse_commandline -d 'Parse the current command line token and return split of existing filepath and rest of token'
+    # eval is used to do shell expansion on paths
+    set -l commandline (eval "printf '%s' "(commandline -t))
+
+    if [ -z $commandline ]
+      # Default to current directory with no --query
+      set dir '.'
+      set fzf_query ''
+    else
+      set dir (__fzf_get_dir $commandline)
+
+      if [ "$dir" = "." -a (string sub -l 1 $commandline) != '.' ]
+        # if $dir is "." but commandline is not a relative path, this means no file path found
+        set fzf_query $commandline
+      else
+        # Also remove trailing slash after dir, to "split" input properly
+        set fzf_query (string replace -r "^$dir/?" '' "$commandline")
+      end
+    end
+
+    echo $dir
+    echo $fzf_query
+  end
+
+  function __fzf_get_dir -d 'Find the longest existing filepath from input string'
+    set dir $argv
+
+    # Strip all trailing slashes. Ignore if $dir is root dir (/)
+    if [ (string length $dir) -gt 1 ]
+      set dir (string replace -r '/*$' '' $dir)
+    end
+
+    # Iteratively check if dir exists and strip tail end of path
+    while [ ! -d "$dir" ]
+      # If path is absolute, this can keep going until ends up at /
+      # If path is relative, this can keep going until entire input is consumed, dirname returns "."
+      set dir (dirname "$dir")
+    end
+
+    echo $dir
+  end
+
 end


shell/key-bindings.fish
```diff
@@ -84,4 +86,26 @@ function fzf_key_bindings
     bind -M insert \cr fzf-history-widget
     bind -M insert \ec fzf-cd-widget
   end
+
+  function __fzf_get_starting_dir
+    set -l dir (commandline -t)
+
+    # Default to current dir if no argument input
+    if [ -z "$dir" ]
+        set dir '.'
```
junegunn: A minor nit: indent with 2 spaces.

shell/key-bindings.fish
```diff
@@ -84,4 +86,26 @@ function fzf_key_bindings
...
+    end
+
+    # The commandline token might be escaped, we need to unescape it.
+    set dir (eval "printf '%s' $dir")
+
+    if [ ! -d "$dir" ]
+      return 1
```
junegunn: I think we would want to `set dir '.'` like above in this case. It makes sense as the current version simply ignores the token.

shell/key-bindings.fish
```diff
@@ -84,4 +86,26 @@ function fzf_key_bindings
...
+  function __fzf_get_starting_dir
+    set -l dir (commandline -t)
+
+    # Default to current dir if no argument input
+    if [ -z "$dir" ]
+        set dir '.'
+    end
+
+    # The commandline token might be escaped, we need to unescape it.
+    set dir (eval "printf '%s' $dir")
```
junegunn: Not required if the above condition is met, and we can put it in the else clause.

shell/key-bindings.fish
```diff
@@ -84,4 +86,26 @@ function fzf_key_bindings
...
+
+    if [ ! -d "$dir" ]
+      return 1
+    end
+
+    # Some 'find' versions print undesired duplicated slashes if the path ends with slashes.
+    set dir (string replace --regex '(.)/+$' '$1' "$dir")
```
junegunn: Can go into the else clause of the above if statement. Correct me if I'm wrong :)

shell/key-bindings.fish
ipwnponies: This will shell expand to absolute path, ie. $HOME<C-t>. If this command works without expansion, do you think there is value or good UX to leave unexpanded completion? I'm indifferent between finding `$HOME/.vimrc` or `/Users/ipwnponies/.vimrc`.

shell/key-bindings.fish
ipwnponies: Yes, you're correct. If `$dir` is '.', then no need to do any more magic on it.

shell/key-bindings.fish
junegunn: It would be nice if we could preserve the original form, but since the command for `alt-c` is configurable in any way, it will be too complicated not to expand the token and properly process the output.

shell/key-bindings.fish
junegunn: > This is a helper function and, imo, it should defer responsibility to the file/cd widget to respond to invalid dir.

But the error is not properly handled outside the function and that's why I pointed it out.

> which results in a user-intuitive behaviour of no fzf completions

Actually I'm a bit torn on this. Because currently `alt-c` ignores whatever is on the command line, so changing it to produce an empty list somehow breaks backward compatibility. (e.g. a user types in `git add`, then realizes that she wants to change the directory so immediately presses `alt-c`, without having to delete the command line.) `alt-c` of bash and zsh also have the same behavior. 

> I think this might result in invalid filepaths for file widget.

Similarly, that's how file widget currently behaves.

junegunn: Your point is valid and I can surely understand it. I'm just concerned that there might be users who will not [appreciate the change](https://xkcd.com/1172/).

Anyway, not showing any result is surely not the best feedback we can give to the users in that scenario. One thing we can consider is to use the last token, that is not directory, as the default `--query` of the fzf command. That's how fuzzy completion of bash and zsh currently works, for example:

```sh
# Scanning ~/github directory with the default query set to "non-existent"
vim ~/github/non-existent**<tab>

# Scanning current directory with the default query "non-existent"
vim non-existent**<tab>
```

If this apply this scheme to `alt-c`

```sh
git add ../test<alt-c>
  # This will open fzf finder scanning the parent directory with "test" as the default query
  # Then we select a directory (like tests) and the current directory is changed
  # and we're left with "git add " in the command line

git add<alt-c>
  # This will start fzf with "add" as the default query, which is probably
  # not the intention of the user, but at least the user knows what's going on.
```

I'm actually not familiar with fish scripting and not sure if it's easy to implement this in fish. What do you think?

shell/key-bindings.fish
junegunn: I'm not sure about the intention of the comment (/ping @Ambrevar), but as you have noticed, without the line , `vim $HOME<CTRL-T>` or `vim ~jg<CTRL-T>` won't work.

Ambrevar: If I recall correctly, the comment was a copy-paste of a similar expansion in a different context. I could be wrong but I don't think there is anything to unescape here.

Feel free to correct the comment, something like:

```
    # The commandline token might need expansion to be tested for existence.
```

ipwnponies: To clarify, the current behaviour is that `vim $HOME<ctrl-t>` will resolve to `vim /Users/ipwnponies/foo.txt`. This is because the `$dir` input is passed to `find $dir`: if `$dir` is a relative path, the results are relative; otherwise the results are absolute paths.

Another clarification as to why I said this `eval "printf '%s' $dir"` is not required, `find $HOME` (no `eval`) and `find /Users/ipwnponies/` (with `eval`) output the same values and are functionally equivalent.

shell/key-bindings.fish
ipwnponies: I like the idea. So we would be looking for the longest substring that is an existing path, the rest of the token is passed as query to `fzf`. These are some test cases to document my understanding:

Command line token | Test case | `find` + `fzf` behaviour
-- | -- | --
../exist/exist | whole path exists | `find ../exist/exist \| fzf`
../exist/not_exist/not_exist | first part of path exists  | `find ../exist \| fzf --query 'not_exist/not_exist'`
not_exist/non_exist | not a path, first part does not exist | `find . \| fzf --query not_exist/non_exist`
/not_exist/non_exist | not a path, starts with / | `find / \| fzf --query not_exist/non_exist`

The code logic would be:
1. Get current command line token
2. Check if it is an existing directory. If not, run `dirname` and check again. Repeat until you find a directory that exists. This will be `$dir`.
3. The remainder of token will be used as `--query` input.

Fish scripting has [commandline](https://fishshell.com/docs/current/commands.html#commandline) builtin which allows for reading and editing command line. From there, you have the usual tools to do string manipulation and file path checks. This is all doable, seems to be reasonably straightforward.

P.S. I appreciate your feedback and attention to detail regarding this. I like fish and fzf and hoped to contribute back with this PR. Due to the

ipwnponies:
I've addressed the comments, the primary change was to modify the behaviour when given invalid input. Now the key bindings will split the command line input into two items: root directory for `find`; remainder is passed to `fzf --query`.

I've tested this manually with relative paths, ~, absolute paths, etc. I know of only one edge case ('//') but this is very minor and edge-y.

junegunn:
Thanks, looks good to me. By the way, can you also check if `FZF_CTRL_T_COMMAND` works as expected? I tried something like `set -l FZF_CTRL_T_COMMAND "echo \$dir"` but the code doesn't seem to pick it up. I noticed that it doesn't work even without your patch, so it's obviously not your fault, but I wonder if you could help debugging the issue.

ipwnponies:
This is a fish shell feature and departure from bash, You are setting the variable with a local scope, using `set -l`. When you call a function, variables with local scope disappear: https://fishshell.com/docs/current/index.html#variables-functions

You should use `set -g`, which will make it behave equivalent to bash variables:

Shell | Variable Assignment Syntax
-- | -
fish | set -g FZF_CTRL_T_COMMAND 'echo $dir'
bash | FZF_CTRL_T_COMMAND='echo $dir'

junegunn:
Thanks for the explanation. So that means the current documentation is incorrect.

https://github.com/junegunn/fzf#fish-shell

I believe I tested the example when I merged #759, hmm, maybe fish has changed since then?
 Anyway, thanks for this, I'll merge it after testing it some more.

shell/key-bindings.fish
```diff
@@ -4,27 +4,22 @@ function fzf_key_bindings
...
-    set dir (string replace --regex '(.)/+$' '$1' "$dir")
+    set -l commandline (__fzf_parse_commandline)
+    set -l dir $commandline[1]
+    set -l fzf_query $commandline[2]
 
     # "-path \$dir'*/\\.*'" matches hidden files/folders inside $dir but not
     # $dir itself, even if hidden.
     set -q FZF_CTRL_T_COMMAND; or set -l FZF_CTRL_T_COMMAND "
-    command find -L \$dir -mindepth 1 \\( -path \$dir'*/\\.*' -o -fstype 'sysfs' -o -fstype 'devfs' -o -fstype 'devtmpfs' \\) -prune \
+    command find -L $dir -mindepth 1 \\( -path \$dir'*/\\.*' -o -fstype 'sysfs' -o -fstype 'devfs' -o -fstype 'devtmpfs' \\) -prune \
```
junegunn: You might want to use `\$dir` instead of `$dir` to avoid issues with paths containing spaces.

```sh
mkdir -p "/tmp/a b c"
touch "/tmp/a b c"/d
vim /tmp/a\ b\ c<CTRL-T>
```

shell/key-bindings.fish
```diff
@@ -51,15 +46,26 @@ function fzf_key_bindings
...
 
   function fzf-cd-widget -d "Change directory"
+    set -l commandline (__fzf_parse_commandline)
+    set -l dir $commandline[1]
+    set -l fzf_query $commandline[2]
+
     set -q FZF_ALT_C_COMMAND; or set -l FZF_ALT_C_COMMAND "
-    command find -L . -mindepth 1 \\( -path '*/\\.*' -o -fstype 'sysfs' -o -fstype 'devfs' -o -fstype 'devtmpfs' \\) -prune \
-    -o -type d -print 2> /dev/null | cut -b3-"
+    command find -L $dir -mindepth 1 \\( -path \$dir'*/\\.*' -o -fstype 'sysfs' -o -fstype 'devfs' -o -fstype 'devtmpfs' \\) -prune \
```
junegunn: Here as well.

shell/key-bindings.fish
ipwnponies: Great catch, I have fixed this.

shell/key-bindings.fish
ipwnponies: Also fixed with `\`.

junegunn:
Merged, thanks!

There is a minor issue though, the script complains when the last token in the command line has unbalanced quotes. But the issue was there before this change, so we will address that in another issue/PR.

```sh
vim foo'<alt-c>
vim foo'<ctrl-t>
vim 'foo<alt-c>
vim 'foo<ctrl-t>
```

ipwnponies:
The issue with quotes is caused by the use of `eval` to do shell expansion, i.e. convert `~`, env vars. This is equivalent to an unmatched quote on the command line, they must be escaped first:
```shell
vim foo\'<alt-c>
vim \'foo<alt-c>
```

If you want don't want this behaviour, then we would need input string escaping. Probably simplest is to only escape `'` and `"` but continue to allow `~` and `$` to pass through. IMO, I don't think it's worth it to do any additional magic and manipulation, it's better to keep things consistent and predictable with fish's behaviour of shell expansion. Document that user input is simply evaled, so the all the same rules for shell expansion apply.

If you do open an issue for that, I can chime in with my opinion above.

junegunn:
Yeah, I don't think it's really worth it. Thanks for the comment.

Issue #953, opened 2017-06-22
Multiline item's separator doesn't use '\n'


amosbird:
- Category
    - [ ] fzf binary
    - [ ] fzf-tmux script
    - [ ] Key bindings
    - [x] Completion
    - [ ] Vim
    - [ ] Neovim
    - [ ] Etc.
- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Windows Subsystem for Linux
    - [ ] Etc.
- Shell
    - [ ] bash
    - [ ] zsh
    - [x] fish

I'm trying to get multiline history completion work by `history -z | fzf --read0`. However it generates items that are not separated by `\n` but `\?` which breaks command line substitution.

![](https://la.wentropy.com/p7hy.png)

junegunn:
fzf does not officially support multi-line entries, and with the premise, a new line character is not a valid character fzf can properly display on screen, and fzf prints it as `?` character instead. I know it is not ideal, but it's how it works right now.

fzf displays those characters as `?`, but that does not affect the final output of fzf, so you can still depend on it.

```sh
> echo -en "foo\nbar" | fzf --read0 --select-1 | wc -l
       2
```

Notice that we have no issues with multi-line commands in bash and zsh, as `history` and `fc -l 1` commands of them are smart enough to transform a multi-line command into a single-line entry.

```
$ a() {
> 1
> 2
> }
$ HISTTIMEFORMAT= history | tail -2
94918  a() { 1; 2; }
94919  HISTTIMEFORMAT= history | tail -2
$
```

Maybe fish can do something similar?

amosbird:
ah, fish's read command also needs a `-z` flag.  This works `history -z | eval (__fzfcmd) --read0 -q '$str' | read -lz result`

PR #954, opened 2017-06-22
[fish] Support multiline commands

### LINKED ISSUE #436: fish: In ^r, multiline commands become multiple separate commands

tomfitzhenry:
Tested to work against fish 2.4 on Debian Stretch. I will be able to test against OS X tomorrow.

Fix found by @amosbird at https://github.com/junegunn/fzf/issues/953#issuecomment-310309055

closes #436

### DIFF
--- a/shell/key-bindings.fish
+++ b/shell/key-bindings.fish
@@ -44,7 +44,7 @@ function fzf_key_bindings
     set -q FZF_TMUX_HEIGHT; or set FZF_TMUX_HEIGHT 40%
     begin
       set -lx FZF_DEFAULT_OPTS "--height $FZF_TMUX_HEIGHT $FZF_DEFAULT_OPTS --tiebreak=index --bind=ctrl-r:toggle-sort $FZF_CTRL_R_OPTS +m"
-      history | eval (__fzfcmd) -q '(commandline)' | read -l result
+      history -z | eval (__fzfcmd) --read0 -q '(commandline)' | perl -pe 'chomp if eof' | read -lz result
       and commandline -- $result
     end
     commandline -f repaint


shell/key-bindings.fish
```diff
@@ -44,7 +44,7 @@ function fzf_key_bindings
     set -q FZF_TMUX_HEIGHT; or set FZF_TMUX_HEIGHT 40%
     begin
       set -lx FZF_DEFAULT_OPTS "--height $FZF_TMUX_HEIGHT $FZF_DEFAULT_OPTS --tiebreak=index --bind=ctrl-r:toggle-sort $FZF_CTRL_R_OPTS +m"
-      history | eval (__fzfcmd) -q '(commandline)' | read -l result
+      history -z | eval (__fzfcmd) --read0 -q '(commandline)' | perl -pe 'chomp if eof' | read -lz result
```
tomfitzhenry: `chomp if eof` will remove the last character, which is always a newline character, when `history -z` is used.

Ideally we'd be able to do this with just coreutils or fish, rather than perl. Anyone know how?

junegunn: Thanks, the patch looks good to me. Just note that fzf currently prints a new line character as `?`, this can be improved in the future.

Yeah, it would be nice if we don't have to use perl here, but if there's no other solution, I'm okay with it. (FWIW, `head -c -1` doesn't work on macOS)

junegunn:
Merged, thanks. I'll consider changing how fzf displays new line character in the new version.

PR #962, opened 2017-07-02
Allow forcing use of Neovim terminal emulation via g:fzf_use_term

### LINKED ISSUE #958: [neovim] File completion on Windows seems to fail if FZF is launched in the same console Window.
- Category
    - [x] fzf binary
    - [ ] fzf-tmux script
    - [ ] Key bindings
    - [x] Completion
    - [ ] Vim
    - [x] Neovim
    - [ ] Etc.
- OS
    - [ ] Linux
    - [ ] Mac OS X
    - [x] Windows
    - [ ] Etc.
- Shell
    - [ ] bash
    - [ ] zsh
    - [ ] fish

Hi, I'm having a bit of a weird problem. I made a very quick and dirty modification (mikesmiffy128@e5f2d2cb83d4fee3e27adf6a9b661d4414d4e86c) so that FZF will fill the Windows console when using Vim, rather than popping up another window. This seems to work, except that `:FZF` and `:Files` generate a blank list. This only happens if I launch it directly, and fixes itself when using `start`. Even weirder, it works fine with fzf.vim functionality like `:GFiles`, `:Commands` etc, just not files.

Sorry if I'm overlooking something obvious, I'm just kind of hoping there'll be some quick hackish way to fix this. Although if it does end up being unfixable I can of couse deal with the separate pop up.

Thanks for making this tool, by the way, it works great with the exception of this small issue.

mikesmiffy128:
This allows manually enabling the use of Neovim's terminal emulator on Windows. This isn't supported in Neovim's master branch yet, however after rebasing neovim/neovim#6383 onto master I was able to launch FZF via `:term:` without any major issues. The only annoyance is that gdamore/tcell#158 still needs fixed, but that just means you have to press enter twice for now.

Also, the use of `;#FZF` to identify the terminal buffer doesn't work with `cmd` since its comment syntax is different, so I put in a check which will use `&::FZF` instead on Windows.

This can also close #958, as it makes FZF usable on Windows without a popup.

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -35,6 +35,8 @@ else
   let s:base_dir = expand('<sfile>:h:h')
 endif
 if s:is_win
+  let s:term_marker = '&::FZF'
+
   function! s:fzf_call(fn, ...)
     let shellslash = &shellslash
     try
@@ -53,6 +55,8 @@ if s:is_win
           \ ['chcp %origchcp% > nul']
   endfunction
 else
+  let s:term_marker = ";#FZF"
+
   function! s:fzf_call(fn, ...)
     return call(a:fn, a:000)
   endfunction
@@ -343,7 +347,7 @@ try
   endif
 
   if has('nvim')
-    let running = filter(range(1, bufnr('$')), "bufname(v:val) =~# ';#FZF'")
+    let running = filter(range(1, bufnr('$')), "bufname(v:val) =~# '".s:term_marker."'")
     if len(running)
       call s:warn('FZF is already running (in buffer '.join(running, ', ').')!')
       return []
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks, it looks promising, but I'd prefer to wait until the related issues are resolved.

mikesmiffy128:
Fair enough, I'm happy using my own fork in the meantime. :) I'm also currently taking a look at tcell to see if I can figure out how to fix it, but I've never used Go before nor looked through any of the code so I might have to mess around a little.

mikesmiffy128:
Just filed gdamore/tcell#159 which fixes the exiting issue. 😃 Hopefully it will be merged soon, but if not you can also feel free to use my fork. 👍

janlazo:
Vim has `:term` as of patch [8.0.693](https://github.com/vim/vim/commit/e4f25e4a8db2c8a8a71a4ba2a68540b3ab341e42) but does not use winpty, used by neovim.

Can this manual enabling be a no-op for versions/builds of Vim/Neovim that don't support `:term`?

mikesmiffy128:
The plugin already detects terminal support using `has("neovim") && !s:is_win`. The main point of the ability to force-enable it is to be able to use it even it isn't otherwise detected as supported - in my case, I manually built Neovim on Windows after rebasing neovim/neovim#6383 onto it. Therefore, it makes more sense in my opinion to detect terminal support in Vim too, and leave the variable for forcing it manually. I'm not sure, but after some Googling I think it's possible to use something along the lines of `has("patch-8.0.0693")`. I haven't checked if that works though.

janlazo:
`:terminal` PR is merged in Neovim so there's no need for a config variable anymore. Enable it by default by checking `has('nvim-0.2.1')`.
`+terminal` is still not enabled by default in Windows releases so there's no point in supporting the Vim 8 terminal in Windows.

mikesmiffy128:
Oh, good news! In that case, you're right, this is somewhat redundant. It might still be an idea to pick out a9bf29b though, since it fixes the comment syntax for cmd.

PR #967, opened 2017-07-07
Allow user to specify path to find binary

jeffwilliams:
This change allows the user to define the FZF_FIND_COMMAND environment variable
to specify which find binary to use. If the variable is unset, `find` is used
(the old default behaviour).

Basically on my system I need to source an environment for building older software, and as part of that the PATH is modified to point to an older find that doesn't work properly with the flags fzf uses. This change will let me use the working `/usr/bin/find` for fzf while letting the other build scripts use the older find.


jeffwilliams:
I realized I didn't correct the go code as part of this change. Will make a new pull request.

Issue #552, opened 2016-04-23
Override $SHELL when running fzf


dalizard:
- Category
    - [x] fzf binary
    - [ ] fzf-tmux script
    - [ ] Key bindings
    - [ ] Completion
    - [x] Vim
    - [x] Neovim
- OS
    - [ ] Linux
    - [x] Mac OS X
    - [ ] Windows
- Shell
    - [ ] bash
    - [ ] zsh
    - [x] fish

I am using neovim (using vim however would not make a difference) and fish shell. I have the following configuration:
```
~ $ echo $SHELL
/usr/local/bin/fish
~ $ cat ~/.config/nvim/init.vim | grep shell
set shell=/usr/local/bin/fish
~ $ fzf --version
0.12.0
```

The problem is when running fzf within neovim/vim (or the shell) there is a startup penalty because of the shell:
```
~ $ time /bin/bash -c "uname"
Darwin
        0.00 real         0.00 user         0.00 sys
~ $ time /usr/local/bin/fish -c "uname"
Darwin
        0.06 real         0.03 user         0.02 sys
~ $ 
```

This is not a huge difference, but I would to be able to set a different shell for fzf for performance reasons. When using fzf.vim this small startup penalty does make a bit of a difference. I found [this commit](https://github.com/junegunn/fzf/commit/e95d82748fc8fe5a05d93b30388ba37adb7dbac8), which make fzf use the default shell. Would it be possible to change the used shell just for fzf?

junegunn:
`$FZF_DEFAULT_COMMAND` is only used in `:Files` command where we don't explicitly specify its source. Please let me know if the following patch helps in your case.

```diff
diff --git a/autoload/fzf/vim.vim b/autoload/fzf/vim.vim
index 99c78ec..3f8b730 100644
--- a/autoload/fzf/vim.vim
+++ b/autoload/fzf/vim.vim
@@ -188,6 +188,9 @@ function! fzf#vim#files(dir, ...)
   else
     let args.options .= ' --prompt '.shellescape(pathshorten(getcwd())).'/'
   endif
+  if exists('$FZF_DEFAULT_COMMAND')
+    let args.source = $FZF_DEFAULT_COMMAND
+  endif
 
   return s:fzf(fzf#vim#wrap(args), a:000)
 endfunction
```

junegunn:
Tested with a bogus $SHELL (`sleep 1; sh "$@"`) and it helps.

PR #969, opened 2017-07-08
[vim] Don't pipe FZF_DEFAULT_COMMAND in Windows

janlazo:
cmd.exe leaves the pipe hanging for the source command after fzf finishes so powershell and ag keep running and cmd.exe does not close. Putting `FZF_DEFAULT_COMMAND` in a batchfile hides the issue from cmd.exe so Vim is stuck waiting and Neovim does not run the exit callback. Letting fzf run `FZF_DEFAULT_COMMAND` passes the responsibility of stopping the source program to fzf executable so Vim/Neovim don't have to wait for the source command to finish executing.

This issue does not happen when running the fzf executable directly in the shell if `FZF_DEFAULT_COMMAND` is using powershell. ag still gets stuck.

I did not consider ack because it requires perl and is too slow as a default command.

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -365,10 +365,10 @@ try
     let dict.dir = fnamemodify(dict.dir, ':p')
   endif
 
-  if !has_key(dict, 'source') && !empty($FZF_DEFAULT_COMMAND)
-    let temps.source = s:fzf_tempname().(s:is_win ? '.bat' : '')
+  if !has_key(dict, 'source') && !empty($FZF_DEFAULT_COMMAND) && !s:is_win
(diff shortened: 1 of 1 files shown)


janlazo:
The external cmd.exe window does not close for gvim and neovim-qt (or any Neovim GUI) because the source command is still running. Running powershell or ag consecutively on large directories slows down the system if they're not killed so they're not viable yet for default commands.

junegunn:
I see, it was introduced to avoid possible slowness of $SHELL (#552), which should be a non-issue on Windows platform.

junegunn:
Merged, thanks!

PR #970, opened 2017-07-09
[neovim] Fix lcd when fzf job exits on Windows

janlazo:
#896 did not consider relative filepaths in the exit handler because `dir /s/b` outputs absolute paths. I changed dict and temp to be part of the fzf job config to minimize issues with the exit handler when running `:FZF` consecutively in Neovim.

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -529,12 +529,15 @@ function! s:execute(dict, command, use_height, temps) abort
     let command = batchfile
     let a:temps.batchfile = batchfile
     if has('nvim')
-      let s:dict = a:dict
-      let s:temps = a:temps
       let fzf = {}
+      let fzf.dict = a:dict
+      let fzf.temps = a:temps
       function! fzf.on_exit(job_id, exit_status, event) dict
-        let lines = s:collect(s:temps)
-        call s:callback(s:dict, lines)
+        if s:present(self.dict, 'dir')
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks, can you post a scenario that reproduces the problem for future reference?

plugin/fzf.vim
```diff
@@ -529,12 +529,15 @@ function! s:execute(dict, command, use_height, temps) abort
...
-      let s:temps = a:temps
       let fzf = {}
+      let fzf.dict = a:dict
+      let fzf.temps = a:temps
       function! fzf.on_exit(job_id, exit_status, event) dict
-        let lines = s:collect(s:temps)
-        call s:callback(s:dict, lines)
+        if s:present(self.dict, 'dir')
+          execute 'lcd' s:escape(self.dict.dir)
+        endif
```
junegunn: Can you check if we can use `call s:pushd(self.dict)` instead as in https://github.com/junegunn/fzf/blob/68bd410159500f67e069a5d909f2d3062115cffc/plugin/fzf.vim#L677-L679?

janlazo: Didn't work. `self.dict.prev_dir` is set by the earlier `s:pushd(a:dict)`

junegunn: I see, thanks for the confirmation.

junegunn:
Thanks!

Issue #960, opened 2017-07-01
Use powershell for default command on Windows


janlazo:
- Category
    - [X] fzf binary
    - [ ] fzf-tmux script
    - [ ] Key bindings
    - [ ] Completion
    - [ ] Vim
    - [ ] Neovim
    - [ ] Etc.
- OS
    - [ ] Linux
    - [ ] Mac OS X
    - [X] Windows
    - [ ] Windows Subsystem for Linux
    - [ ] Etc.
- Shell
    - [ ] bash
    - [ ] zsh
    - [ ] fish

`dir /s/b` does not filter hidden directories/files and its results cannot be piped to another `dir` command.
I can't get `dir /b/a:-d-h | dir /s/b` to work in cmd.exe to filter hidden directories before recursing on subfolders.

Powershell does not have these issues because of object piping and output customization.
The following command works in cmd.exe and runs powershell on default settings to recursively get all non-hidden files. This takes up more time but removes files in hidden directories such as `.git`. Additional parameters are added to the powershell.exe to reduce the startup time.

```dosbatch
powershell.exe -NoLogo -NoProfile -Noninteractive -Command "Get-ChildItem -File -Recurse | ForEach-Object {echo $_.FullName}"
```

**Edit:**
`Get-ChildItem` supports `-Name` for relative path so object piping is not required.

```dosbatch
powershell.exe -NoLogo -NoProfile -Noninteractive -Command "Get-ChildItem -File -Recurse -Name"
```

kelleyma49:
Unfortunately PowerShell is really slow.  I love PowerShell, but dir is way faster, and that's what I use for my PowerShell wrapper module: https://github.com/kelleyma49/PSFzf/blob/master/PSFzf.psm1#L346.

Another option is to create and compile a small app written in C# or C++ that can do the directory listings.

janlazo:
It's speed vs accuracy.
a) Get 10000+ results but require non-trivial search syntax to reduce 10%. Hope that the path is short enough to see the file in the rightmost side of the cmd.exe terminal.
b) Wait 3-5 seconds but the results are down to 100 and you can see what you need. Trival search syntax can reduce number of entries to single digits.

> Without STDIN pipe, fzf will use find command to fetch the list of files excluding hidden ones.
The updated powershell command achieves this with similar output to the `find` command.
The slow startup issue can be minimized by using `cmd` as the main shell and not piping the command.

[...]

PR #971, opened 2017-07-09
Default command outputs relative filepath for preview to work in Windows

janlazo:
Modified the default command in Windows to match the description in the README

> Without STDIN pipe, fzf will use find command to fetch the list of files excluding hidden ones. (You can override the default command with FZF_DEFAULT_COMMAND)

### DIFF
--- a/src/constants.go
+++ b/src/constants.go
@@ -59,7 +59,7 @@ func init() {
 	} else if os.Getenv("TERM") == "cygwin" {
 		defaultCommand = `sh -c "command find -L . -mindepth 1 -path '*/\.*' -prune -o -type f -print -o -type l -print 2> /dev/null | cut -b3-"`
 	} else {
-		defaultCommand = `dir /s/b/a:-d-h`
+		defaultCommand = `dir /s/b/a:-d-h 2> nul`
 	}
 }
 
--- a/src/constants.go
+++ b/src/constants.go
@@ -59,7 +59,7 @@ func init() {
 	} else if os.Getenv("TERM") == "cygwin" {
 		defaultCommand = `sh -c "command find -L . -mindepth 1 -path '*/\.*' -prune -o -type f -print -o -type l -print 2> /dev/null | cut -b3-"`
 	} else {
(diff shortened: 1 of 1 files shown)


janlazo:
The following command can remove the current directory of the path to display relative paths

```dosbatch
cmd /q/V:ON/c "for /f %A in ('dir /s/b/a:-d-h') DO set RELPATH=%A & echo !RELPATH:%cd%\=!"
```

janlazo:
https://github.com/junegunn/fzf/pull/971#issuecomment-313971408 is necessary for preview to work to get around the network drive. `/a:-d-h` is optional.

janlazo:
@junegunn Can you test if it works for you? I compiled using mingw32-make.exe from http://tdm-gcc.tdragon.net/download and preview works for me because the network drive is removed.

You can test it out in fzf.vim after forcing forward slashes (or escaping the backslash) on the filepath of the preview script
Reference: https://github.com/janlazo/fzf.vim/commit/3657fc6f72661c54ab84cc8ffb342f16abce06f3#diff-7274b7c6c87526542ef9087f2e997dfaR38

janlazo:
I give on this.
For loop begins after `dir` is done so it is slow in large projects.
There's no native windows solution I can think of that has comparable speed to `dir`, can be configured and piped, and won't slow down the system.

junegunn:
I see, thanks for looking into it.

> There's no native windows solution

Many users of fzf on *nix use `ag` or `rg` instead of `find` as the default command, if there are equivalent solutions on Windows, we can suggest those to the users.

janlazo:
From my previous [note](https://github.com/junegunn/fzf/issues/960#issuecomment-313952566) about powershell for relative paths, I recommend rg because of [UTF-16 support](https://github.com/BurntSushi/ripgrep/blob/master/CHANGELOG.md#050-2017-03-12) and sift because only the prebuild binary is required.

junegunn:
It would be great if you could mention your findings here: https://github.com/junegunn/fzf/wiki/Windows

Issue #966, opened 2017-07-07
^R broken with fish: "history: invalid option -- -z"


ludovicaelbrecht:
- Category
    - [x] fzf binary
    - [ ] fzf-tmux script
    - [x] Key bindings
    - [x] Completion
    - [ ] Vim
    - [ ] Neovim
    - [ ] Etc.
- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Windows Subsystem for Linux
    - [ ] Etc.
- Shell
    - [ ] bash
    - [ ] zsh
    - [x] fish

I've been using fzf with fish (version 2.3.1, latest on Fedora 25) for a couple of months now. Was mostly using the completion it offers when I hit ^R. 

Decided to update fzf (git pull in ~/.fzf, then ~/.fzf/install), and now it's broken: when I hit ^R in fish, I get this: "history: invalid option -- -z". Still works fine in bash.

Here's what it looks like:

```
~/.fzf (master) $ git pull
Already up-to-date.
~/.fzf (master) $ ~/.fzf/install
Downloading bin/fzf ...
  - Already exists
  - Checking fzf executable ... 0.16.8
Do you want to enable fuzzy auto-completion? ([y]/n) y
Do you want to enable key bindings? ([y]/n) y

Generate ~/.fzf.bash ... OK
Update fish_user_paths ... OK
Symlink /home/ludovic/.config/fish/functions/fzf_key_bindings.fish ... OK

Do you want to update your shell configuration files? ([y]/n) y

Update /home/ludovic/.bashrc:
  - [ -f ~/.fzf.bash ] && source ~/.fzf.bash
    - Already exists: line #55 

Update /home/ludovic/.config/fish/functions/fish_user_key_bindings.fish:
  - fzf_key_bindings
    - Already exists: line #3 

Finished. Restart your shell or reload config file.
   source ~/.bashrc  # bash
   fzf_key_bindings  # fish

Use uninstall script to remove fzf.

For more information, see: https://github.com/junegunn/fzf
~/.fzf (master) $ history: invalid option -- -z
```
This last line, with the error message, shows up each time I hit ctrl-R.

Am I doing something wrong? Maybe the -z version is only supported in more recent versions of fish?

Thanks!
Ludovic

junegunn:
> Maybe the -z version is only supported in more recent versions of fish?

I just googled it and it turns out fish 2.4 is required.

[...]

PR #972, opened 2017-07-10
[fish] Fix ctrl-r regression in versions <2.4

tomfitzhenry:
... but multiline won't be supported in versions <2.4

Issue #966

I have yet to test this on version 2.3, but will.

### DIFF
--- a/shell/key-bindings.fish
+++ b/shell/key-bindings.fish
@@ -47,11 +47,11 @@ function fzf_key_bindings
       # history's -z flag was added in fish 2.4.0, so don't use it for versions
       # before 2.4.0.
       if [ "$FISH_MAJOR" -gt 2 -o \( "$FISH_MAJOR" -eq 2 -a "$FISH_MINOR" -ge 4 \) ];
-          history -z | eval (__fzfcmd) --read0 -q '(commandline)' | perl -pe 'chomp if eof' | read -lz result
-          and commandline -- $result
+        history -z | eval (__fzfcmd) --read0 -q '(commandline)' | perl -pe 'chomp if eof' | read -lz result
+        and commandline -- $result
       else
-          history | eval (__fzfcmd) -q '(commandline)' | read -l result
-          and commandline -- $result
+        history | eval (__fzfcmd) -q '(commandline)' | read -l result
+        and commandline -- $result
       end
     end
     commandline -f repaint
--- a/shell/key-bindings.fish
+++ b/shell/key-bindings.fish
@@ -1,7 +1,7 @@
 # Key bindings
 # ------------
 function fzf_key_bindings
-  
+
   # Store current token in $dir as root for the 'find' command
   function fzf-file-widget -d "List files and folders"
     set -l commandline (__fzf_parse_commandline)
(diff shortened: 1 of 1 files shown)


shell/key-bindings.fish
```diff
@@ -1,7 +1,9 @@
 # Key bindings
 # ------------
 function fzf_key_bindings
-
+  set -g __FZF_FISH_MAJOR (echo $FISH_VERSION | cut -f1 -d.)
```
tomfitzhenry: This creates a global variable, which isn't ideal. Ideally we'd be able to set a variable that fzf-history-widget has access to, but so far I've been unable to find a way.

tomfitzhenry:
I reproduced the bug on Debian Jessie with fish 2.2.0 and fzf 0.16.8, and can confirm that this PR fixes the issue.

shell/key-bindings.fish
junegunn: How about just doing this inside `fzf-history-widget`? We can create local variables instead only when they're needed, which makes us repeat it whenever the widget's called, but it won't cause any noticeable delay.

tomfitzhenry: Ok. I had expected people would prefer a global variable to unnecessary computation, but I'm happen with either.

junegunn: I'm fine with either actually. I'm not even a fish user 😆 

Maybe fish has some string manipulation functions we can use in this case, so we don't have to start two subprocesses?

tomfitzhenry: According to Wikipedia, fish is an "exotic" shell!

Fish does have string manipulation functions, but only as of 2.3 (see https://fishshell.com/docs/current/commands.html#string ). I figured supporting fish <2.3 was more important saving subprocess invocations.

shell/key-bindings.fish
```diff
@@ -1,7 +1,7 @@
 # Key bindings
 # ------------
 function fzf_key_bindings
-
+
```
junegunn: Everything looks good except this line :)

tomfitzhenry: Fixed!

shell/key-bindings.fish
```diff
@@ -39,8 +39,20 @@ function fzf_key_bindings
...
+      # history's -z flag is needed for multi-line support.
+      # history's -z flag was added in fish 2.4.0, so don't use it for versions
+      # before 2.4.0.
+      if [ "$FISH_MAJOR" -gt 2 -o \( "$FISH_MAJOR" -eq 2 -a "$FISH_MINOR" -ge 4 \) ];
+          history -z | eval (__fzfcmd) --read0 -q '(commandline)' | perl -pe 'chomp if eof' | read -lz result
+          and commandline -- $result
+      else
+          history | eval (__fzfcmd) -q '(commandline)' | read -l result
+          and commandline -- $result
```
junegunn: I'm sorry, but I just realized that you used 4-space indentation instead of 2. Can you fix it as well?

junegunn:
Thanks!

PR #984, opened 2017-07-22
Allow user to specify path to find binary

jeffwilliams:
This change allows the user to define the FZF_FIND_COMMAND environment variable
to specify which find binary to use. If the variable is unset, `find` is used
(the old default behaviour).

Basically on my system I need to source an environment for building older software, and as part of that the PATH is modified to point to an older find that doesn't work properly with the flags fzf uses. This change will let me use the working /usr/bin/find for fzf while letting the other build scripts use the older find.

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,6 +224,10 @@ or `py`.
 - `FZF_DEFAULT_OPTS`
     - Default options
     - e.g. `export FZF_DEFAULT_OPTS="--reverse --inline-info"`
+- `FZF_FIND_COMMAND:`
+    - The `find` command to use when autocompleting files and directories.
+    - When unset, defaults to `find`.
+    - e.g. `export FZF_FIND_COMMAND="/usr/local/bin/find"`
 
 #### Options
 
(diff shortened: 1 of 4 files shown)


junegunn:
Thanks, but I don't think this is a necessary addition, since we already have `$FZF_DEFAULT_COMMAND` that gives you more control. I can understand you don't want to replicate the whole command with only a small adjustment, but the default command is not likely to change in the future, so don't worry about keeping it in sync with the updated source.

Also many users including myself use ag or rg instead of find to [filter out git-ignored files](https://github.com/junegunn/fzf#respecting-gitignore-hgignore-and-svnignore), and I don't think one should try to stick to the default set of options for find, they don't matter that much really.

I can think of other workarounds too (like having an alias or wrapper script that temporarily prepends a path to $PATH), but I think `$FZF_DEFAULT_COMMAND` is the best way to do it. Having both `$FZF_DEFAULT_COMMAND` and `$FZF_FIND_COMMAND` can also be confusing.

jeffwilliams:
I see what you mean. If the default command is unlikely to change then this is probably unnecessary.

PR #998, opened 2017-07-28
Add MinGW on Windows to install script

ahal:
Add install support for http://www.mingw.org/

Running uname -sm yields:
MINGW32_NT-6.2 i686

The prebuilt windows 386 binary works well in this environment.


junegunn:
Thanks!

PR #1000, opened 2017-07-30
[vim] escape slash in s:shortpath

janlazo:
Fix for the regexp comparison for checking if directory filepath ends in a slash.
I need this issue resolved before porting it to `Files` in fzf.vim

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -756,7 +756,7 @@ let s:default_action = {
 function! s:shortpath()
   let short = pathshorten(fnamemodify(getcwd(), ':~:.'))
   let slash = (s:is_win && !&shellslash) ? '\' : '/'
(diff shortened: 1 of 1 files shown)


plugin/fzf.vim
```diff
@@ -756,7 +756,7 @@ let s:default_action = {
 function! s:shortpath()
   let short = pathshorten(fnamemodify(getcwd(), ':~:.'))
   let slash = (s:is_win && !&shellslash) ? '\' : '/'
-  return empty(short) ? '~'.slash : short . (short =~ slash.'$' ? '' : slash)
+  return empty(short) ? '~'.slash : short . (short =~ '\'.slash.'$' ? '' : slash)
```
junegunn: We don't need to escape forward slash. Wouldn't it be more clearer if we do `escape(slash, '\')`?

junegunn:
I just updated the commit and pushed to master. Thanks!

PR #1004, opened 2017-08-01
ignore versioned binaries

aureooms:


junegunn:
Thank you, but I don't think it's needed as now we simply store "fzf" inside the release tarballs.

See https://github.com/junegunn/fzf/commit/ee5aeb80a49fbf7bb958e09aa641ef17fb885884

aureooms:
Nice! Sorry for the trouble.

junegunn:
No problem, thanks.

PR #1015, opened 2017-08-13
Add MinGW 64 to install fzf in Windows 64-bit

janlazo:
Need it for installing fzf in Windows 64-bit if running sh.exe (msysgit) from cmd.exe or powershell.exe

```dosbatch
:: within cmd.exe terminal
sh .\install --bin
```


junegunn:
Thanks!

PR #1021, opened 2017-08-19
[vim] escape backslashes in fzf#shellescape

janlazo:
Forgot to handle the single backslash case which breaks the surrounding `^"` to wrap the entire string.
A use case is searching for backslash in `Lines`, `BLines`.
Requires `s:q` using `fzf#shellescape` in fzf.vim (commit: https://github.com/janlazo/fzf.vim/commit/32756d271a580c9d189dcc8dd053f8dd0731d9c8)


janlazo:
With the new `fzf#shellescape`, `Lines` supports the multiple backslash query (ie. `\\\`) in fzf.vader.

junegunn:
Thanks!

PR #1029, opened 2017-08-27
add parens around piped source commands

oconnor663:
Previously a command like `echo a && echo b` would get transformed into
`echo a && echo b | fzf`, which only pipes the output of the second
command. Adding parentheses around the source command avoids this issue,
and works on both Unix and Windows.

Caveat: I haven't tested this change on Windows, just played with some example
commands in the cmd.exe terminal.

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -366,7 +366,7 @@ try
     let source = dict.source
     let type = type(source)
     if type == 1
-      let prefix = source.'|'
+      let prefix = '( '.source.' )|'
     elseif type == 3
       let temps.input = s:fzf_tempname()
       call writefile(source, temps.input)


junegunn:
Thanks, this is sensible and is consistent with how `FZF_DEFAULT_COMMAND` works.

There are some corner cases we have to consider though, for example, if the command is already wrapped with parentheses it will not work as expected.

```sh
((echo 1 && echo 2)) | cat
# -bash: ((: echo 1: syntax error in expression (error token is "1")
```

One option would be to wrap the command with `(` and `<SPACE>)`

```sh
# Both work
(echo 1 && echo 2 ) | fzf
((echo 1 && echo 2) ) | fzf
```

But I'm not sure if it's a valid solution on Windows.

@janlazo, you're my go-to guy when it comes to Windows :) Can you comment on this?

janlazo:
> One option would be to wrap the command with `(` and `<SPACE>)`

That's fine.
Regular Vim adds parens in Windows (see `:h shellxquote`) and parens are escaped in [fzf#shellescape](https://github.com/junegunn/fzf/commit/8aab0fc1894facf9ac51132ff8a832b18bc115ac#diff-24d9d57e35df5afce7890f2400917919R54)

oconnor663:
Added the extra spaces. Though now I'm wondering, would The Right Thing be to run with `"/bin/sh -c " . shellescape(source)` on Unix and `"cmd.exe /C " . shellescape(source)` (or `%COMSPEC%`) on Windows? That way no parsing issues should be possible?

junegunn:
Yeah, that was the first thing that came to my mind. But we have learned that escaping the command properly on Windows is very hard (simply using `shellescape` breaks in every imaginable way), and I'd like to avoid creating another temporary batch file / shell script just for that purpose.

I'll merge this PR as it is now, thanks.

oconnor663:
Thanks. This is where I wish that vim exposed a simple "exec a command given a list of arguments" instead of forcing us to escape everything through the shell. We could do that if the plugin was written in Python though :)

PR #1031, opened 2017-08-29
quick and dirty hack to make fzf work better with fish

raichoo:
Hi,

I'm having some issues using `fzf` with `neovim` and `fish`. Turns out `fish` doesn't use parenthesis as delimiters but `begin` and `end`. Here is a quick and dirty workaround. There has got to be a better way but I didn't want to report this without proving at least a workaround.

Kind regards,
raichoo

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -366,7 +366,11 @@ try
     let source = dict.source
     let type = type(source)
     if type == 1
-      let prefix = '( '.source.' )|'
+      if (empty($SHELL) ? &shell : $SHELL) =~ "fish$"
+        let prefix = 'begin; '.source.'; end|'
+      else
+        let prefix = '( '.source.' )|'
+      end
     elseif type == 3
       let temps.input = s:fzf_tempname()
       call writefile(source, temps.input)


junegunn:
Thanks, I'm having a hard time understanding why fish users have this issue when we temporarily change `&shell` to `sh` before starting the terminal (https://github.com/junegunn/fzf/pull/1031/files#diff-24d9d57e35df5afce7890f2400917919L340). Any idea why?

Related: https://github.com/junegunn/fzf.vim/issues/302#issuecomment-325630306

/cc @cvlmtg

cvlmtg:
@junegunn there's something strange. I added `echom &shell` right before `call termopen(a:command . ';#FZF', fzf)` and when I remove `'down': '~40%'` from the dictionary, &shell is 'sh', but if I don't remove that key, &shell is still '/usr/local/bin/fish'

cvlmtg:
just another little test in nvim:

:echo &shell -> /usr/local/bin/fish
:set shell=sh
:echo &shell -> sh
(wait 1 sec or 2)
:echo &shell -> /usr/local/bin/fish

this doesn't happen if I unset $SHELL before running nvim. no idea if this helps :)

cvlmtg:
more strange things... this 'reset' happens if I `set shell=sh`, but doesn't happen if I `set shell=bash`

cvlmtg:
I made this little change and it seems to have fixed the problem for me at the moment. I guess not every system might have bash installed (e.g. *BSD i think) so it isn't really the correct solution. probably this is something that needs to be investigated on the nvim side

```
index 0f4fc77..2310c6d 100644
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -337,7 +337,7 @@ try
     set shell=cmd.exe
     set noshellslash
   else
-    set shell=sh
+    set shell=bash
   endif

   let dict   = exists('a:1') ? s:upgrade(a:1) : {}
```

junegunn:
That's strange. Can you check if the value of `&shell` changes around `s:split(a:dict)` like so? This should print two lines of `sh`.

```diff
diff --git a/plugin/fzf.vim b/plugin/fzf.vim
index 0f4fc77..2e5ef29 100644
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -624,7 +624,9 @@ endfunction
 function! s:execute_term(dict, command, temps) abort
   let winrest = winrestcmd()
   let pbuf = bufnr('')
+  echom &shell
   let [ppos, winopts] = s:split(a:dict)
+  echom &shell
   let b:fzf = a:dict
   let fzf = { 'buf': bufnr(''), 'pbuf': pbuf, 'ppos': ppos, 'dict': a:dict, 'temps': a:temps,
             \ 'winopts': winopts, 'winrest': winrest, 'lines': &lines,
```

cvlmtg:
```
sh
/usr/local/bin/fish
Error running ( git ls-files  )|'/Users/matteo/.fzf/bin/fzf'  -m --prompt "GitFiles> " --expect=ctrl
-v,ctrl-x,ctrl-t --no-height > /var/folders/zp/1cddchs97zxgtshn1gmcq0_c0000gn/T/nvim3AkxJy/2
```

cvlmtg:
as I wrote `&shell` change value automatically when set to 'sh', but I've made another test with `nvim -u NONE --noplugin` and it doesn't happen, so there's some plugin / config that does something strange, but it seems it doesn't happen only to me, so it shouldn't be something weird in my config. I'll try to find out

cvlmtg:
@junegunn looks like it's vim-gitgutter that every N seconds resets &shell when it's equal to "sh". I'll try to do some more in depth debug, but probably won't be able to do it soon

junegunn:
Hmm, yeah, that's what I suspected, you probably have an `autocmd CursorHold` that changes shell value, and another autocmd that changes the value when a new window is open.

What I think we can do is to `set shell=sh` again right before `termopen` (or at least after `s:split`).

```diff
diff --git a/plugin/fzf.vim b/plugin/fzf.vim
index 0f4fc77..d3673b3 100644
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -328,17 +328,20 @@ function! fzf#wrap(...)
   return opts
 endfunction
 
-function! fzf#run(...) abort
-try
-  let oshell = &shell
-  let useshellslash = &shellslash
-
+function! s:use_sh()
+  let [shell, shellslash] = [&shell, &shellslash]
   if s:is_win
     set shell=cmd.exe
     set noshellslash
   else
     set shell=sh
   endif
+  return [shell, shellslash]
+endfunction
+
+function! fzf#run(...) abort
+try
+  let [shell, shellslash] = s:use_sh()
 
   let dict   = exists('a:1') ? s:upgrade(a:1) : {}
   let temps  = { 'result': s:fzf_tempname() }
@@ -405,8 +408,7 @@ try
   call s:callback(dict, lines)
   return lines
 finally
-  let &shell = oshell
-  let &shellslash = useshellslash
+  let [&shell, &shellslash] = [shell, shellslash]
 endtry
 endfunction
 
@@ -625,6 +627,7 @@ function! s:execute_term(dict, command, temps) abort
   let winrest = winrestcmd()
   let pbuf = bufnr('')
   let [ppos, winopts] = s:split(a:dict)
+  call s:use_sh()
   let b:fzf = a:dict
   let fzf = { 'buf': bufnr(''), 'pbuf': pbuf, 'ppos': ppos, 'dict': a:dict, 'temps': a:temps,
             \ 'winopts': winopts, 'winrest': winrest, 'lines': &lines,
```

Does this fix the issue?

cvlmtg:
yes it does

junegunn:
@cvlmtg Thanks for the confirmation.
@raichoo Let me know if you still have the issue. Thanks!

Issue #1032, opened 2017-08-30
fzf-tmux fails with tmux's `run-shell` command


mike-hearn:
- Category
    - [ ] fzf binary
    - [x] fzf-tmux script
    - [ ] Key bindings
    - [ ] Completion
    - [ ] Vim
    - [ ] Neovim
    - [ ] Etc.
- OS
    - [x] Linux
    - [x] Docker!
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Windows Subsystem for Linux
    - [ ] Etc.
- Shell
    - [x] bash
    - [ ] zsh
    - [ ] fish

-----

I'm running into an issue using `fzf-tmux` as part of tmux's `run-shell` command.

If I run `fzf-tmux` from either tmux's built-in command prompt, or inside a tmux session using `tmux run-shell fzf-tmux`, it fails with the following error on 0.17.0 (e89eebb):

```
root@baf17140c5fe:/# tmux run-shell fzf-tmux
'fzf-tmux' returned 2
```

I downgraded to 0.15.9 to try it, and it gives a different (but presumably related) error:

```
root@5f43672d3b38:/# tmux run-shell fzf-tmux
Failed to open /dev/tty
'fzf-tmux' returned 2
```

It's definitely possible I'm missing something super obvious, but I spent some time today and couldn't figure it out.

I made a [Dockerfile](https://github.com/mike-hearn/fzf-tmux-issue/blob/master/Dockerfile) with fzf 0.17.0 and tmux 2.5 to help reproduce the issue. The repo is here with instructions:

https://github.com/mike-hearn/fzf-tmux-issue

Quick screencast of the issue (expected result first, then demo of the error using the Dockerfile):
![Error demo](https://i.imgur.com/ybGWJHp.gif)
[Higher quality MP4 version](https://zippy.gfycat.com/SarcasticFatalBlacknorwegianelkhound.mp4)

mike-hearn:
Update: I just went through my own reproduction steps, using the same Dockerfile, on a remote Linode Debian 8 box, and... it worked as expected.

So there must be some environmental difference between my local Macbook pro, where fzf-tmux fails inside tmux inside Docker (heh), and Debian 8, where fzf-tmux works inside tmux inside Docker. Will keep investigating.

mike-hearn:
[...]

PR #1037, opened 2017-09-01
Add fzf-tmux pane_height/pane_width fallback

mike-hearn:
Based on [this issue](https://github.com/junegunn/fzf/issues/1032).

Adds a fix to the fzf-tmux script for newer distros that use a particular version of `tput`, specifically Fedora 26 and Debian 10, which causes fzf-tmux to fail when run inside tmux's `run-shell` command.

The fix falls back to using tmux's `pane_width` and `pane_height` command to set the `$lines` and `$columns` variable, but only if `tput lines` or `tput cols` fails.

Dockerfile for verifying that this works:

```Dockerfile
FROM docker.io/fedora:26

RUN dnf install -y tmux git curl

RUN git clone https://github.com/mike-hearn/fzf ~/.fzf && \
        cd ~/.fzf && \
        git checkout fedora26-tput-fix && \
        ~/.fzf/install --all
```

### DIFF
--- a/bin/fzf-tmux
+++ b/bin/fzf-tmux
@@ -16,8 +16,8 @@ skip=""
 swap=""
 close=""
 term=""
-[[ -n "$LINES" ]] && lines=$LINES || lines=$(tput lines)
-[[ -n "$COLUMNS" ]] && columns=$COLUMNS || columns=$(tput cols)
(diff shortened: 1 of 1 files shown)


junegunn:
Looks good to me, thanks!

PR #1040, opened 2017-09-02
exclude bash when zsh_only=1

kaldown:
`:PlugInstall!`
won't affect `.bashrc && .fzf.bash` when `--zsh-only` applied


junegunn:
Thanks, but I think this is incomplete for the following reasons:

- No `--bash-only` or `--fish-only`
- Still updates fish stuff when `--zsh-only` is set
- What happens when both `--bash-only` and `--zsh-only` are specified. The behavior is undefined.

So instead of this, let me add `--no-{bash,zsh,fish} flags.

PR #1043, opened 2017-09-06
[vim] do not pathshorten prompt in cygwin

janlazo:
Cygwin converts unix-style paths to windows-style paths so `s:shortpath` shouldn't use `pathshorten` if `has('win32unix')`. `Files` has the same issue.

To reproduce
```sh
cd /usr/bin
fzf --prompt '/u/bin'
```

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -772,7 +772,10 @@ let s:default_action = {
   \ 'ctrl-v': 'vsplit' }
 
 function! s:shortpath()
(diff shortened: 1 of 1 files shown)


junegunn:
I see, thanks. We should port https://github.com/junegunn/fzf.vim/commit/e7928d154aad71e530daf5c17733275da5732680 for narrow-width terminals. Let me do that first.

junegunn:
5edc3f755ca5f1e5512679cc8e7f995de7e12c3b

junegunn:
Merged, thanks!

PR #1047, opened 2017-09-08
Exclude fzf binary from repo

llwu:
Not necessary but a bit convenient for me because `bin/fzf-0.13.5-darwin_amd64` is dirtying my index.


junegunn:
Thanks, but this is no longer relevant. See #1004.

PR #1048, opened 2017-09-09
Don't touch dotfiles if not requested

llwu:
If `.bashrc`/`.zshrc` doesn't exist, the install script will create an empty `.bashrc`/`.zshrc` even if `n` is answered to `Do you want to update your shell configuration files?`. This is a bit of unexpected behavior, e.g. a script which installs `fzf` and then symlinks `.zshrc` from somewhere else will fail.

Ex:
```
[ -f ~/.zshrc] && mv ~/.zshrc ~/.zshrc.bak
printf 'y\ny\nn\n' | ~/fzf/install
ln -s ~/dotfiles/.zshrc ~/.zshrc
```

### DIFF
--- a/install
+++ b/install
@@ -305,20 +305,22 @@ append_line() {
   line="$2"
   file="$3"
   pat="${4:-}"
+  lno=""
 
   echo "Update $file:"
   echo "  - $line"
-  [ -f "$file" ] || touch "$file"
-  if [ $# -lt 4 ]; then
-    lno=$(\grep -nF "$line" "$file" | sed 's/:.*//' | tr '\n' ' ')
-  else
-    lno=$(\grep -nF "$pat" "$file" | sed 's/:.*//' | tr '\n' ' ')
+  if [ -f "$file" ]; then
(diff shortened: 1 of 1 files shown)


install
```diff
@@ -305,19 +305,22 @@ append_line() {
...
+    else
+      lno=$(\grep -nF "$pat" "$file" | sed 's/:.*//' | tr '\n' ' ')
+    fi
   fi
   if [ -n "$lno" ]; then
     echo "    - Already exists: line #$lno"
   else
     if [ $update -eq 1 ]; then
+      touch "$file"
       echo >> "$file"
```
junegunn: Thanks. I think we should rewrite these two lines to

```sh
[ -f "$file" ] && echo >> "$file"
```

- Because we don't need to add a blank line above the line we're about to add if the file doesn't exist at the moment
- and `touch` is not necessary as `>>` will create the file anyway

llwu: Thanks for the suggestion, I changed the code to that.

junegunn:
Merged, thanks!

PR #1050, opened 2017-09-12
[vim] Use string type so term_start does not escape command

janlazo:
Workaround for Vim 8 terminal. Neovim does not have this issue.

Don't merge this yet.
I'm checking if this is actually required since the Vim plugin uses a temporary batchfile in Windows.
It's definitely required if there are no batchfiles.

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -699,7 +699,8 @@ function! s:execute_term(dict, command, temps) abort
     if has('nvim')
       call termopen(command, fzf)
     else
-      call term_start([&shell, &shellcmdflag, command], {'curwin': fzf.buf, 'exit_cb': function(fzf.on_exit)})
+      let cmds = [&shell, &shellcmdflag, command]
+      call term_start(s:is_win ? join(cmds) : cmds, {'curwin': fzf.buf, 'exit_cb': function(fzf.on_exit)})
     endif
   finally
     if s:present(a:dict, 'dir')


janlazo:
Actually, is there any issue with merging it as is? vim-plug use string type for the command in Vim 8 jobs.

junegunn:
Thanks, but I'm a bit lost on this. What problem does this solve? Or is it a step to simplify the code?

> vim-plug use string type for the command in Vim 8 jobs in Windows.

I believe you're referring to this line: https://github.com/junegunn/vim-plug/blob/cd39fa61332c4324d733a3bf0847cbb4dbba2a17/plug.vim#L1204

I believe that it is okay because we already "shell-escaped" parts of the command before calling `s:spawn`.

janlazo:
It avoids escaping double quotes in cmd.exe. I said not to merge it yet because `tempname` outputs a shortened path via `fnamemodify(path, ':8')` in regular Vim.

in Vim 8
```viml
let cmd = ['cmd', '/c', 'echo "Hey"]
call term_start(cmd)
" > \"Hey\"
call term_start(join(cmd))
" > "Hey"
````

in cmd.exe,
```dosbatch
> echo "Hey"
"Hey"
```

janlazo:
I'll revisit this after resolving https://github.com/junegunn/vim-plug/pull/676 because it uses a temporary batchfile for `s:spawn` in Windows.

PR #1053, opened 2017-09-14
Make fzf embedable

untoldwind:
Not so much a Pull-Request but rather a request for comment.

I wanted to embed/include the fzf-"ui" in one of my projects and encountered several quirks.

What I wanted:
* Invoke fzf.Run with Options created/set programmatically. (Unluckily plenty of things are not accessible from other packages)
* Feed the searcher programmatically
* Have a preview without spawning a process
* React to some specially keybindings (e.g. Ctrl-Space should invoke a go function rather than execute something)

Changes I made to achieve this:
* Reader is now an interface so that fzf can be feeded by some other means than stdin or the output of an executable. There is now an additional implementation using a go channel (chan []byte)
* There is now a Command interface so that fzf can trigger some go func rather than "just" running an executable. This applies to actions as well as the preview.
* All settings and inner types of Option are now public. Actually that is the bulk of the the changes

I admit that the result is not a very clean API, but nevertheless it is usable.

### DIFF
--- a/src/chunklist_test.go
+++ b/src/chunklist_test.go
@@ -9,7 +9,7 @@ import (
 
 func TestChunkList(t *testing.T) {
 	// FIXME global
-	sortCriteria = []criterion{byScore, byLength}
+	sortCriteria = []Criterion{CriterionByScore, CriterionByLength}
 
 	cl := NewChunkList(func(item *Item, s []byte) bool {
 		item.text = util.ToChars(s)
--- a/src/command.go
+++ b/src/command.go
@@ -1,0 +1,125 @@
+package fzf
+
+import (
+	"os"
+	"strings"
+
+	"github.com/junegunn/fzf/src/util"
+)
+
+type Command interface {
+	GetPreview(stripAnsi bool, delimiter Delimiter, query string, allItems []*Item) string
+	HasPlusFlag() bool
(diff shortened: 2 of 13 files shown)


junegunn:
Interesting, thanks for the suggestion. Unfortunately, I currently don't have time and mental bandwidth to think carefully about the direction; e.g. if I really want to support this use case when it'll increase maintenance burden for me, how we should design the API, how to revise the code that was written with the assumption that fzf is a short-lived, one-off process, etc. I'll revisit this when I find myself actually wanting it. fzf is quite stable now and I'm not planning to introduce major changes anytime soon, so I suggest that you use your fork for the time being if you really need this feature.

For what it's worth, starting fzf as a child process should be enough for most use cases. You can still programmatically feed the input, language selection is not limited to Golang, and there should be no noticeable performance difference. Granted, you can't statically link fzf to your binary, but on the flip side, this allows users to upgrade fzf separately (it's more like dynamic linking).

https://junegunn.kr/2016/02/using-fzf-in-your-program

ebrannin-bw:
For people who come here later and find the broken link to @junegunn's blog, here it is in the Wayback machine:

https://web.archive.org/web/20161221125259/http://junegunn.kr/2016/02/using-fzf-in-your-program/

PR #1062, opened 2017-09-26
_fzf_complete_ssh: include host aliases

joshpencheon:
With an entry in `~/.ssh/config` like:

```
Host my-server example-server my-example-server
  User josh
  Hostname example.com
```

...using FZF's completion on `ssh **`, I only get the primary name listed. With this patch, the other aliases should also be selectable:

```
$ ssh **
>
  3/3
> my-server
  example-server
  my-example-server
```

Thanks,
Josh


junegunn:
Merged, thanks!

PR #1068, opened 2017-09-30
[vim](fix): revert changes that break  fzf#call

### LINKED ISSUE #1066: [neovim]  fzf#call not working anymore
- Category
    - [ ] fzf binary
    - [ ] fzf-tmux script
    - [ ] Key bindings
    - [ ] Completion
    - [ ] Vim
    - [X] Neovim
    - [ ] Etc.
- OS
    - [X] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [ ] bash
    - [X] zsh
    - [ ] fish

Hi,
Not sure this is the right place since it's related to vim/neovim but I think `fzf#call` is defined in this place.

Anyway I had a command define with `fzf#call` but since last update however I call that function I get
FZF#call bug
I really don't know what's going on.

Since last update I had a new slightly annoying bug on `:File` command: whenever the `find` command throw an error  (like it is unable to enter a folder due to permissions) it prints the warning on screen.
Not sure if this is related to the previous...

marcmo:
fzf#call stopped working with the latest commit. this fixes it again.

fixes #1066

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -389,8 +389,7 @@ try
   let use_height = has_key(dict, 'down') && !has('gui_running') &&
         \ !(has('nvim') || s:is_win || has('win32unix') || s:present(dict, 'up', 'left', 'right', 'window')) &&
         \ executable('tput') && filereadable('/dev/tty')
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks, I properly updated the condition.

PR #1072, opened 2017-10-02
Fix ExecCommandWith for cmd.exe in Windows

### LINKED ISSUE #1018: Shell parsing for preview command in Windows is incompatible with fzf#shellescape and cmd.exe
- Category
    - [X] fzf binary
    - [ ] fzf-tmux script
    - [ ] Key bindings
    - [ ] Completion
    - [ ] Vim
    - [ ] Neovim
    - [ ] Etc.
- OS
    - [ ] Linux
    - [ ] Mac OS X
    - [X] Windows
    - [ ] Etc.
- Shell
    - [ ] bash
    - [ ] zsh
    - [ ] fish

Backslashes in the filepath of the preview command are escaped so they don't run in Windows.
The preview script itself (ruby or bash) has no issues with filepaths with backslashes.

https://github.com/junegunn/fzf.vim/pull/418#issuecomment-322641946 is caused by the network drive conflicting with the colon delimiter. I'll open a separate issue/PR for that.

janlazo:
Close #1018 

Run the command as is in cmd.exe with no parsing and escaping.
Explicity set cmd.SysProcAttr so execCommand does not escape the command.
Technically, the command should be escaped with ^ for special characters, including ".
This allows cmd.exe commands to be chained together.

See https://github.com/neovim/neovim/pull/7343#issuecomment-333350201

However, this requires a new shellescape function that is specific to one of the following:
- interactive prompt
- batchfile
- command name

fzf#shellescape in the Vim plugin can handle only the batchfile.

### DIFF
--- a/src/terminal.go
+++ b/src/terminal.go
@@ -1105,7 +1105,10 @@ func keyMatch(key int, event tui.Event) bool {
 
 func quoteEntry(entry string) string {
 	if util.IsWindows() {
-		return strconv.Quote(strings.Replace(entry, "\"", "\\\"", -1))
+		r, _ := regexp.Compile("[&|<>()@^%\"]")
+		return r.ReplaceAllStringFunc(strconv.Quote(entry), func (match string) string {
+		    return "^" + match
+		})
 	}
 	return "'" + strings.Replace(entry, "'", "'\\''", -1) + "'"
 }
--- a/src/terminal.go
+++ b/src/terminal.go
@@ -1105,8 +1105,10 @@ func keyMatch(key int, event tui.Event) bool {
 
 func quoteEntry(entry string) string {
 	if util.IsWindows() {
+		escaped := strings.Replace(entry, "\\", "\\\\", -1)
+		escaped = "\"" + strings.Replace(escaped, "\"", "\\\"", -1) + "\""
 		r, _ := regexp.Compile("[&|<>()@^%\"]")
-		return r.ReplaceAllStringFunc(strconv.Quote(entry), func (match string) string {
+		return r.ReplaceAllStringFunc(escaped, func (match string) string {
 		    return "^" + match
 		})
 	}
--- a/src/terminal.go
+++ b/src/terminal.go
@@ -1105,9 +1105,9 @@ func keyMatch(key int, event tui.Event) bool {
 
 func quoteEntry(entry string) string {
 	if util.IsWindows() {
-		escaped := strings.Replace(entry, "\\", "\\\\", -1)
-		escaped = "\"" + strings.Replace(escaped, "\"", "\\\"", -1) + "\""
-		r, _ := regexp.Compile("[&|<>()@^%\"]")
+		escaped := strings.Replace(entry, `\`, `\\`, -1)
+		escaped = `"` + strings.Replace(escaped, `"`, `\"`, -1) + `"`
+		r, _ := regexp.Compile(`[&|<>()@^%"]`)
 		return r.ReplaceAllStringFunc(escaped, func (match string) string {
 		    return "^" + match
 		})
--- a/src/terminal.go
+++ b/src/terminal.go
@@ -1107,7 +1107,7 @@ func quoteEntry(entry string) string {
 	if util.IsWindows() {
 		escaped := strings.Replace(entry, `\`, `\\`, -1)
 		escaped = `"` + strings.Replace(escaped, `"`, `\"`, -1) + `"`
-		r, _ := regexp.Compile(`[&|<>()@^%"]`)
+		r, _ := regexp.Compile(`[&|<>()@^%!"]`)
 		return r.ReplaceAllStringFunc(escaped, func (match string) string {
 		    return "^" + match
 		})
--- a/src/util/util_windows.go
+++ b/src/util/util_windows.go
@@ -6,8 +6,6 @@ import (
 	"os"
 	"os/exec"
 	"syscall"
-
-	"github.com/mattn/go-shellwords"
 )
 
 // ExecCommand executes the given command with cmd
@@ -18,11 +16,13 @@ func ExecCommand(command string) *exec.Cmd {
 // ExecCommandWith executes the given command with cmd. _shell parameter is
 // ignored on Windows.
 func ExecCommandWith(_shell string, command string) *exec.Cmd {
-	args, _ := shellwords.Parse(command)
-	allArgs := make([]string, len(args)+1)
-	allArgs[0] = "/c"
-	copy(allArgs[1:], args)
-	return exec.Command("cmd", allArgs...)
+	cmd := exec.Command("cmd")
+	cmd.SysProcAttr = &syscall.SysProcAttr{
+	    HideWindow: false,
+	    CmdLine: fmt.Sprintf(` /s /c "%s"`, command),
+	    CreationFlags: 0,
+	}
+	return cmd
 }
 
 // IsWindows returns true on Windows


junegunn:
Thanks. First off, let's make it clear that fzf is a command-line program that can be used with or without Vim (many users of fzf [do not even use Vim](https://github.com/bling/fzf.el)) and there is no way we can enforce anything on bind/preview expression users specify. The specifics of `fzf#shellecape` should be out of the discussion here.

Having said that, I have a few comments.

1. I see that you use `"%s"`, so what happens if the command contains `"`s? Is it really okay? I'm specifically asking about this because placeholder expression `{}`  is currently replaced to double-quote string of the current line on Windows; see https://github.com/junegunn/fzf/blob/0.17.0/src/terminal.go#L1111-L1116
1. Should we create a temporary batch file inside Go code instead if we eventually can't avoid non-trivial escaping of the command?

janlazo:
I mentioned `fzf#shellescape` because it is tested already in Windows and I'm interested in the internal `s:shellesc_cmd`. I like to rewrite `s:shellesc_cmd` in Go such that it works for the interactive prompt but I'm haven't gotten to Go regex yet. What users specify is mostly what they get because I don't know how to escape all the backslashes.

As for your questions,
1. It's okay to have multiple, uneven `"` as long as it works in the interactive prompt. Problem is how cmd.exe parses a command in the prompt. From previous experience, double-quoting works only if there are no nested double quotes and the last character is not a backslash. What's actually required is a double escape, one for each token and another for the entire command when passing to cmd.exe.
Idea is to make it similar to `sh -c 'user_command'`.
Instead, it will be `cmd.exe /s /c ^"escaped_user_command^"`
2. Yes, but I don't think we'll get there yet. Unlike Vim, this is your project so we can adjust the escape code as we go.

junegunn:
What is "interactive prompt"?

janlazo:
I see what you mean now for the double-quoted string for the placeholder.
The entire command can break if there's another double quote after the placeholder.
This is a problem for piping and redirection if the placeholder is not the last token in the preview command.

janlazo:
Is there anything else that the placeholder string needs to be escaped correctly?

janlazo:
Is `strings.replace` in `quoteEntry` for `shellwords.Parse`? I thought `strconv.Quote` handled this already for `"` and `\`.

janlazo:
I kept `strconv.Quote` for convenience in `quoteEntry` but it should explicitly escape the placeholder now in a cmd.exe session.

junegunn:
> I thought strconv.Quote handled this already for " and \.

I think you're right. Also, it seems that it was a mistake to use `strconv.Quote` as it does more than just quoting the string.

According to https://golang.org/pkg/strconv/#Quote

> The returned string uses Go escape sequences (\t, \n, \xFF, \u0100) for control characters and non-printable characters

I don't think we want this behavior in this context.

src/terminal.go
```diff
@@ -1105,7 +1105,12 @@ func keyMatch(key int, event tui.Event) bool {
 
 func quoteEntry(entry string) string {
 	if util.IsWindows() {
-		return strconv.Quote(strings.Replace(entry, "\"", "\\\"", -1))
+		escaped := strings.Replace(entry, "\\", "\\\\", -1)
+		escaped = "\"" + strings.Replace(escaped, "\"", "\\\"", -1) + "\""
+		r, _ := regexp.Compile("[&|<>()@^%\"]")
```
junegunn: Let's use raw string literals (using backticks) in these lines to make them more readable.

https://golang.org/ref/spec#String_literals

janlazo:
What about parsing `FZF_DEFAULT_OPTS`? https://github.com/junegunn/fzf/blob/f3177305d5572b26f135fc045481358b4eb1bf69/src/options.go#L270

junegunn:
What about it? I don't see how it's related to the current discussion.

janlazo:
The parser is incompatible with `cmd.exe` or `powershell.exe` for non-trivial parsing so the user must set it to work on `sh`.

janlazo:
Anyway, I can escape now in `FZF_DEFAULT_COMMAND` for cmd.exe. I couldn't customize it well for rg, sift, or powershell because of `shellwords.Parse`.

janlazo:
If there's no issues with the changes, then I will update the docs for `--preview` and `FZF_DEFAULT_COMMAND`.

junegunn:
> The parser is incompatible with cmd.exe or powershell.exe for non-trivial parsing so the user must set it to work on sh.

Yes, I'd like to leave this as an open issue, since 1. I believe it's rare to have non-trivial options in $FZF_DEFAULT_OPTS, 2. and it's easy to come up with a sh-compatible version of it.

> then I will update the docs for --preview and FZF_DEFAULT_COMMAND

What needs to be mentioned?

janlazo:
> What needs to be mentioned?

The user has to do the escaping themselves because fzf will run their command as is on cmd.exe, not `SHELL`. Updating the README is enough but `fzf --help` could mention it in Windows.

junegunn:
> The user has to do the escaping themselves

Can you give an example where escaping is needed?

> their command as is on cmd.exe, not SHELL

Umm, is `SHELL` actually used on Windows? I didn't expect Windows users would think that fzf would ever use `$SHELL`.

janlazo:
> Can you give an example where escaping is needed?

- prompt (https://github.com/junegunn/fzf/pull/916#issuecomment-302292067)
- `FZF_DEFAULT_COMMAND` for relative paths https://github.com/junegunn/fzf/wiki/Windows#relative-filepaths
  - if the command contains cmd.exe's `FOR` command, the user must escape it, especially for delayed variable expansion with `!ENV_VAR!` syntax (https://ss64.com/nt/delayedexpansion.html)
- nested `sh` or `powershell` because each shell have their own escaping rules
  - scripting in powershell is disabled by default so users will likely call powershell from command line
  - sh has escape issues with nested double quotes
- Format specifiers using `%` (https://github.com/junegunn/vim-plug/pull/681, `GFiles?`) because of `%ENV_VAR%`.
- any commands that has 2+ doubles quotes because of cmd.exe dequoting

> Umm, is SHELL actually used on Windows?

fzf unsets `TERM` to run in `sh` (aliased to `bash`). `SHELL` is set to `/usr/bin/bash`.

junegunn:
Thanks for the comment. But I was also curious how a command would look like after manual escaping. So the commands shown in https://github.com/junegunn/fzf/wiki/Windows#relative-filepaths will have to change?

> fzf unsets TERM to run in sh (aliased to bash). SHELL is set to /usr/bin/bash.

This is the point I don't understand. `/usr/bin/bash` is not a valid path on Windows, is it? Is SHELL a thing on Windows?

junegunn:
Are we discussing Cygwin environment?

janlazo:
> So the commands shown in https://github.com/junegunn/fzf/wiki/Windows#relative-filepaths will have to change?

For cmd.exe only for consistency with this PR (ie. add `/s` flag and wrap the command with double quotes)
The rest have little/no escaping so they're unaffected.

> Is SHELL a thing on Windows?

No, `COMSPEC` (or `ComSpec`) is but its value is almost always the absolute path of `cmd.exe`.

> Are we discussing Cygwin environment?

For `SHELL` or running fzf on sh or bash.

janlazo:
> /usr/bin/bash is not a valid path on Windows, is it?

By default, no.
Forward slash is supported and the network drive is optional for paths within the current network drive.
If the user made a directory `C:\usr\bin\`, then `cd /usr/bin` is allowed.

junegunn:
Thanks for the clarification and for being patient with me. So for the users running fzf on cmd.exe, there's no confusion, but Cygwin or bash users should be aware that SHELL is not used by fzf and FZF_DEFAULT_COMMAND should be written for cmd.exe.

You can briefly mention this on the README page or we can just add a link to the Windows wiki page.

janlazo:
We can update the Windows wiki page and then add a link in the README and Cygwin wiki page to consolidate the explanation for why `SHELL` is ignored and how `FZF_DEFAULT_COMMAND` should be set.

I don't mind doing more work to clarify the changes introduced.

janlazo:
On second thought, `!abc!` must be escaped as `^!abc^!`. The Vim plugin needs the update as well.
Test case is a prompt asking a user to define their prompt as follows:

```dosbatch
cmd /v:ON /s /c "set /p my_prompt= & fzf --prompt !my_prompt!"
```

junegunn:
We should probably add terminal_test.go and implement test cases for the Windows version of quoteEntry.

janlazo:
Will it run on Travis? I thought `util.IsWindows` is defined on compile time in src/util/util_windows.go

junegunn:
`util.IsWindows` returns a static value, but we can extract the block as `quoteEntryCmd` and test it separately.

```go
if util.IsWindows() {
  return quoteEntryCmd(entry)
} 
return ...
```

junegunn:
Can you add a few test cases?

janlazo:
I can add unescaped commands from https://github.com/junegunn/fzf/wiki/Windows#relative-filepaths,  some shell commands I used in https://github.com/neovim/neovim/pull/7343, and basic statements in batchfiles and powershell scripts.

It helps if I can isolate these test cases into file so I can test with `go run quote_test.go`.
This is what I have on one file to test `quoteEntry`

```golang
package main

import (
	"fmt"
	"os/exec"
	"syscall"
	"strings"
	// "strconv"
	"regexp"
)

func main() {
	tests := [...]string{
		"\"",
		"\\",
		"\\\"",
		"\"\\\\\\\"",
		"&|<>()@^%!",
		"%USERPROFILE%",
		"C:\\Program Files (x86)\\",
	}

	for _, v := range tests {
		escaped := Escape(v)
		out, _ := ExecCommandWith("echo " + escaped).Output()
		out2, _ := ExecCommandWith("\"echo\" " + escaped).Output()
		fmt.Printf("%s\n%s\n%s%s\n", v, escaped, out, out2)
	}
}

func Escape(entry string) string {
	escaped := strings.Replace(entry, `\`, `\\`, -1)
	escaped = `"` + strings.Replace(escaped, `"`, `\"`, -1) + `"`
	r, _ := regexp.Compile(`[&|<>()@^%!"]`)
	return r.ReplaceAllStringFunc(escaped, func (match string) string {
		return "^" + match
	})
}

func ExecCommandWith(command string) *exec.Cmd {
	cmd := exec.Command("cmd")
	cmd.SysProcAttr = &syscall.SysProcAttr{
	    HideWindow: false,
	    CmdLine: fmt.Sprintf(` /s /c "%s"`, command),
	    CreationFlags: 0,
	}
	return cmd
}
```

junegunn:
Functional test cases for `quoteEntry` should suffice for now.

```diff
diff --git a/src/terminal.go b/src/terminal.go
index 1890a2a..98d837c 100644
--- a/src/terminal.go
+++ b/src/terminal.go
@@ -1103,14 +1103,18 @@ func keyMatch(key int, event tui.Event) bool {
 		event.Type == tui.Mouse && key == tui.DoubleClick && event.MouseEvent.Double
 }
 
+func quoteEntryCmd(entry string) string {
+	escaped := strings.Replace(entry, `\`, `\\`, -1)
+	escaped = `"` + strings.Replace(escaped, `"`, `\"`, -1) + `"`
+	r, _ := regexp.Compile(`[&|<>()@^%!"]`)
+	return r.ReplaceAllStringFunc(escaped, func(match string) string {
+		return "^" + match
+	})
+}
+
 func quoteEntry(entry string) string {
 	if util.IsWindows() {
-		escaped := strings.Replace(entry, `\`, `\\`, -1)
-		escaped = `"` + strings.Replace(escaped, `"`, `\"`, -1) + `"`
-		r, _ := regexp.Compile(`[&|<>()@^%!"]`)
-		return r.ReplaceAllStringFunc(escaped, func (match string) string {
-		    return "^" + match
-		})
+		return quoteEntryCmd(entry)
 	}
 	return "'" + strings.Replace(entry, "'", "'\\''", -1) + "'"
 }
diff --git a/src/terminal_test.go b/src/terminal_test.go
index d42d2b8..60f2b1a 100644
--- a/src/terminal_test.go
+++ b/src/terminal_test.go
@@ -91,3 +91,22 @@ func TestReplacePlaceholder(t *testing.T) {
 	result = replacePlaceholder("echo {}/{1}/{3}/{2..3}", true, Delimiter{regex: regex}, false, "query", items1)
 	check("echo '  foo'\\''bar baz'/'f'/'r b'/''\\''bar b'")
 }
+
+func TestQuoteEntryCmd(t *testing.T) {
+	tests := map[string]string{
+		`"`:                       `^"\^"^"`,
+		`\`:                       `^"\\^"`,
+		`\"`:                      `^"\\\^"^"`,
+		`"\\\"`:                   `^"\^"\\\\\\\^"^"`,
+		`&|<>()@^%!`:              `^"^&^|^<^>^(^)^@^^^%^!^"`,
+		`%USERPROFILE%`:           `^"^%USERPROFILE^%^"`,
+		`C:\Program Files (x86)\`: `^"C:\\Program Files ^(x86^)\\^"`,
+	}
+
+	for input, expected := range tests {
+		escaped := quoteEntryCmd(input)
+		if escaped != expected {
+			t.Errorf("Input: %s, expected: %s, actual %s", input, expected, escaped)
+		}
+	}
+}
```

junegunn:
Closed by c4185e8. I added the test cases to your commits and pushed to master. Let me know if there's anything else you want to address. Thanks!

janlazo:
I will send another PR to add more test cases from the wiki to the map. I was busy this week so I didn't have time to include your changes myself so thanks for closing this with your changes.

@junegunn can you update the README or wiki for me to mention `SHELL` issue for Bash/Cygwin users so they can set `FZF_DEFAULT_COMMAND` accordingly?

junegunn:
No hurries. I already did update the wiki page, so please take a look.

https://github.com/junegunn/fzf/wiki/Windows

I'll also add a link to the page on the README page.

janlazo:
That's fine. Thanks you.

PR #1073, opened 2017-10-02
Allow custom sink*

lfmet:
As a wrapper, I think it should support `g:fzf_action` when `--expect` isn't given, avoiding [this](https://github.com/junegunn/fzf.vim/blob/master/autoload/fzf/vim.vim#L122) when using `fzf#wrap`.

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -320,11 +320,12 @@ function! fzf#wrap(...)
   endif
 
   " Action: g:fzf_action
-  if !s:has_any(opts, ['sink', 'sink*'])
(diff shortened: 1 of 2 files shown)


junegunn:
> I think it should support g:fzf_action when --expect isn't given

The commands in `g:fzf_action` are only for opening files, and they just don't make sense for other types of lists, e.g. names of color schemes. And I don't think if I like how the expectation of the function signature for `sink*` changes after "wrapping" in this proposal.

What is currently being done in fzf.vim definitely needs some improvement, but I need to think more on how.

PR #1074, opened 2017-10-02
Allow custom sink*

lfmet:
As a wrapper, I think it should support `g:fzf_action` when `--expect` is not given, avoiding [this](https://github.com/junegunn/fzf.vim/blob/master/autoload/fzf/vim.vim#L122) when using fzf#wrap.

Notes:
- `sink*` could receive `(actions, action, lines)` instead of `(actions, [action, lines])`.
- Perhaps there is no need for `actions` if [this](https://github.com/lfmet/fzf/blob/55cea7d2c02bafbce7aa6e129f4580d568936c6e/plugin/fzf.vim#L324) is defined in the script source scope.

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -320,11 +320,12 @@ function! fzf#wrap(...)
   endif
 
   " Action: g:fzf_action
-  if !s:has_any(opts, ['sink', 'sink*'])
-    let opts._action = get(g:, 'fzf_action', s:default_action)
-    let opts.options .= ' --expect='.join(keys(opts._action), ',')
+  if !has_key(opts, 'sink') && opts.options !~# '--expect'
+    let opts._action   = get(g:, 'fzf_action', s:default_action)
+    let opts.options  .= ' --expect='.join(keys(opts._action), ',')
+    let opts._sink_ref = has_key(opts, 'sink*') ? opts['sink*'] : function('s:common_sink')
     function! opts.sink(lines) abort
-      return s:common_sink(self._action, a:lines)
+      return self._sink_ref(self._action, a:lines)
     endfunction
     let opts['sink*'] = remove(opts, 'sink')
   endif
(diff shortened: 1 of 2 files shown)


junegunn:
See my comment in #1073. Also, it is documented that `sink*` only takes one argument that is the list of lines produced by fzf.

lfmet:
> it is documented that sink* only takes one argument that is the list of lines produced by fzf.

I mimic the `s:common_sink()`, but the pure `sink*` is better without sugar, more explicit.

> The commands in g:fzf_action are only for opening files, and they just don't make sense for other types of lists, e.g. names of color schemes.

It felt definitely weird trying to arrange this, but `fzf.vim` supports it most as a common case (Buffers, Lines, GFiles, Files, Marks, ...). As I write functions for `fzf#wrap`, I see that `fzf.vim` has a lot of "development tools" that could be in `fzf plugin`, but yeah, it needs more thought before improving it.

lfmet:
Also, what is the reason for `g:fzf_action` to be not defined in the source scope?
A custom `sink*` that wants to support `g:fzf_action` would need to copy the `s:default_action`.

junegunn:
I know some things are not clear, that's partly because the code in fzf.vim and `fzf#wrap` were not well thought through from the beginning. There was no `fzf#wrap` when I wrote most commands in fzf.vim project. The function was started as a script-local function in the project but later migrated back to the main repo and became a public API. `fzf_action` was also not well thought through. It was designed only for commands that produce simple lists of file paths. I ended up supporting the option in many commands in fzf.vim, but each command implements its own sink function with no code reuse, making `g:fzf_action` effectively a convention rather than a standard.

I'd like to see if we can do better than just exposing `s:default_action` and let the users implement their own sink functions using it. Maybe we could come up with a higher-order function that takes a custom sink function as the argument and returns an augmented version that can handle `g:fzf_action`.

PR #1079, opened 2017-10-08
Don't do shell quoting for weird chars

ionelmc:
This would prevent tabs from being escaped as `$'\t'` (definitely not what I would want to see as initial value in the search).

### DIFF
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -60,7 +60,7 @@ fzf-history-widget() {
   local selected num
   setopt localoptions noglobsubst noposixbuiltins pipefail 2> /dev/null
   selected=( $(fc -l 1 |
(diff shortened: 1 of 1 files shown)


shell/key-bindings.zsh
```diff
@@ -60,7 +60,7 @@ fzf-history-widget() {
   local selected num
   setopt localoptions noglobsubst noposixbuiltins pipefail 2> /dev/null
   selected=( $(fc -l 1 |
-    FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} $FZF_DEFAULT_OPTS --tac -n2..,.. --tiebreak=index --bind=ctrl-r:toggle-sort $FZF_CTRL_R_OPTS --query=${(q)LBUFFER} +m" $(__fzfcmd)) )
+    FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} $FZF_DEFAULT_OPTS --tac -n2..,.. --tiebreak=index --bind=ctrl-r:toggle-sort $FZF_CTRL_R_OPTS --query=${(q-)LBUFFER} +m" $(__fzfcmd)) )
```
junegunn: Can you change it to `(qqq)` instead? So that single quotes are preserved. e.g. `echo "I'd like that"`

ionelmc: Changed.

junegunn:
Merged, thanks!

Issue #1067, opened 2017-09-29
Suggestion for documentation


chrisjohnson:
- Category
    - [x] Completion
- OS
    - [x] Mac OS X
- Shell
    - [x] zsh

This isn't a bug, just a suggestion for the documentation. Right now, in the [settings section](https://github.com/junegunn/fzf#settings), you mention a solution to use ag for `_fzf_compgen_path` and call out the fact that ag only works on files not directories. This is great because by default path and dir globbing in homedir take a long time on my machine with tons of files, and being able to use a `.ignore` would greatly improve that. However, it doesn't solve the issue for dir globbing.

I would propose adding the following function definition (or a comparable ag definition) to also work for directory completion:

```
_fzf_compgen_dir() {
	rg --hidden --files --null "$1" 2>/dev/null | xargs -0 dirname | awk '!h[$0]++'
}
```
I found the suggestion in ripgrep's issues and they proposed using `| uniq` instead of this awk command, but uniq requires duplicates to be adjacent to one another, which results in duplicate results. I tried prepending `| sort` which removes duplicates but makes the command take about 3x longer. Finally I stumbled on this awk snippet which doesn't require the duplicates to be adjacent, and has the happy benefit of actually being a decent bit faster than the `| uniq` approach.

----

**Note**: For anybody reading this, you need to use `gdirname` from the coreutils homebrew package on OS X instead of dirname. I used this block to determine which dirname to use:

```
if [[ $(uname) == "Darwin" ]]; then
	# dirname on OS X behaves funky, get gdirname via
	# brew install coreutils
	export dirname_command="gdirname"
else
	export dirname_command="dirname"
fi
```

I understand this dirname business is probably too much for the fzf documentation though

junegunn:
Thanks for the suggestion. You're right that `_fzf_compgen_dir` is not mentioned on the page and I should add it.

I like how you implemented `_fzf_compgen_dir` with standard utilities, it's simple and concise. But it's sad to see that simplicity is undermined by the inconsistency of dirname on macOS. I'll see if there are simpler options.

Note that ripgrep may not be a good choice until a new version is released. See https://github.com/BurntSushi/ripgrep/issues/200

chrisjohnson:
I posted the same snippet into a ripgrep issue and somebody responded with a zsh approach which used the `h` modifier but unfortunately zsh would wait for the entire list to be generated before extracting the dirname which made it infeasible. Also it consumed a huge amount of memory.

[...]

PR #1083, opened 2017-10-12
Update README to add _fzf_compgen_dir in addition to _fzf_compgen_path

chrisjohnson:
As discussed here: https://github.com/junegunn/fzf/issues/1067

### DIFF
--- a/README.md
+++ b/README.md
@@ -369,10 +369,13 @@ export FZF_COMPLETION_OPTS='+c -x'
 
 # Use ag instead of the default find command for listing candidates.
 # - The first argument to the function is the base path to start traversal
-# - Note that ag only lists files not directories
 # - See the source code (completion.{bash,zsh}) for the details.
 _fzf_compgen_path() {
   ag -g "" "$1"
+}
+# Ag only lists files not directories, so we can use awk to get dirnames
+_fzf_compgen_dir() {
+  ag -g "" "$1" | awk 'function dirname(fn) { if (fn == "") return ".";  if (fn !~ "[^/]") return "/"; sub("/*$", "", fn); if (fn !~ "/") return "."; sub("/[^/]*$", "", fn); if (fn == "") fn = "/"; return fn } {$0 = dirname($0)} !a[$0]++'
 }
 ```
 


junegunn:
Thanks for the pull request. I think we still have some room for improvement here.

IMO, the code is not very readable. I'd prefer to present only short, easy-to-read code snippets that the users can easily understand without putting too much effort, as I don't want to encourage them to copy and paste some code without understanding it first.

Also, it does not print directories which don't have any files on them. For example, in the following case,

```sh
mkdir -p a/b/c/d
touch a/b/c/d/e
```

it only prints `a/b/c/d` skipping the intermediate directories. For that reason, I wrote a couple of filter scripts and have used them with other commands.

- https://github.com/junegunn/dotfiles/blob/master/bin/with-dir
- https://github.com/junegunn/dotfiles/blob/master/bin/only-dir

```sh
rg --files | only-dir
rg --files | with-dir
```

The actual scripts are not shorter than your awk script and they requires Ruby, but if the purpose of the README page is to show the users how they can customize the default behavior and give some examples, rather than to provide the best practices they can blindly follow, it seems to me that the shorter example using a separate filter script conveys the concept in more straightforward way although users can't simply copy the code and be done with it.

chrisjohnson:
Ha, that ruby script is even quicker than the awk one, that surprises me. It also adds many dirs that the awk approach does not because it traverse back up the path chain to add intermediate dirs without files. It's still subject to missing dirs that have no files in them or in any subdir but seems better than what I had.

I like this approach better. What would you say to copying the `only-dir` script out of your dotfiles into a gist (for a permanent un-changing url) and linking to it and using your proposed `rg --files | only-dir` in the README?

junegunn:
Sounds like a plan. Let me put up a gist.

> only-dir doesn't uniq

I believe it does by maintaining a unique set of visited directories. Let me know if it's not the case.

<pre>
diff --git a/README.md b/README.md
index f7ebb73..dc26307 100644
--- a/README.md
+++ b/README.md
@@ -367,15 +367,26 @@ export FZF_COMPLETION_TRIGGER='~~'
 # Options to fzf command
 export FZF_COMPLETION_OPTS='+c -x'
 
-# Use ag instead of the default find command for listing candidates.
+# Use rg instead of the default find command for listing path candidates.
 # - The first argument to the function is the base path to start traversal
-# - Note that ag only lists files not directories
 # - See the source code (completion.{bash,zsh}) for the details.
+# - rg only lists files, so we use with-dir script to augment the output
 _fzf_compgen_path() {
-  ag -g "" "$1"
+  rg --files "$1" | with-dir "$1"
+}
+
+# Use rg to generate the list for directory completion
+_fzf_compgen_dir() {
+  rg --files "$1" | only-dir "$1"
 }
 ```
 
+`only-dir` and `with-dir` scripts can be found [here][dir-scripts]. They are
+written in Ruby, but you should be able to rewrite them in any language you
+prefer.
+
+[dir-scripts]: https://gist.github.com/junegunn/8c3796a965f22e6a803fe53096ad7a75
+
 #### Supported commands
 
 On bash, fuzzy completion is enabled only for a predefined set of commands
</pre>

So, what do you think?

(As I write this, I noticed an issue with only-dir and with-dir. They traverse directories upwards until they find `.`, but the scheme doesn't work with absolute paths and they end up printing `/`. So I update both scripts to take an optional argument which denotes the base directory where they should stop.)

junegunn:
Maybe we should still suggest ag since the patch for https://github.com/BurntSushi/ripgrep/issues/200 is not yet included in an official release of ripgrep.

chrisjohnson:
Aha it's a Set, that makes sense then. And I agree with favoring `ag` over `rg` until that bug is fixed. Looks like a solid approach, I agree with it entirely

junegunn:
Updated the page in cb8e972. Thanks for the suggestions and comments!

chrisjohnson:
Cheers!

partounian:
What is the difference with setting compgen and FZF_DEFAULT_COMMAND? Skimming through code seems completely different but description wise they seem to be described the same, so I'm confused.

chrisjohnson:
I am guessing based on my experience, so somebody else might correct me, FZF_DEFAULT_COMMAND is used when you invoke fzf the command, and the compgen functions are invoked when you're trying to use fzf to complete command line arguments with **

junegunn:
So I updated the examples again, after trying out [fd](https://github.com/sharkdp/fd), which is a tool specifically designed for this use case. No need for extra scripts.

See: https://github.com/junegunn/fzf/commit/a6d2ab33609ad20ce107e66d9648d908800fa3ad

Issue #1085, opened 2017-10-14
vim: sets local dirs (via :lcd)


blueyed:
- Category
    - [ ] fzf binary
    - [ ] fzf-tmux script
    - [ ] Key bindings
    - [ ] Completion
    - [x] Vim
    - [x] Neovim
    - [ ] Etc.
- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Windows Subsystem for Linux
    - [ ] Etc.
- Shell
    - [ ] bash
    - [ ] zsh
    - [ ] fish

The Vim plugin uses `:lcd` to push/pod dirs, but should rather use the most appropriate `cd` command (`let cd = haslocaldir() ? 'lcd' : (exists(':tcd') == 2 && haslocaldir(-1, 0)) ? 'tcd' : 'cd'`).
Therefore the window where fzf has been used with gets a local dir (`haslocaldir()`).

Any thoughts on this, otherwise I might provide a PR which remembers the prev dir together with the command being used (and use that also for popping).

junegunn:
> fzf has been used with gets a local dir (haslocaldir())

I see, but I'm not sure what the issue is. What problem do you have when it's set? Also, `haslocaldir`  in Vim 7.4 does not take arguments, so I don't want to use it that way.

blueyed:
I am using only `haslocaldir()` with Vim in the PR.

The problem is that `:lcd` is meant to be special, and that I normally have a global or tab-local working dir.  If a local window dir is set, this will not adjust.
When saving views (e.g. via vim-stay) you have to put additional workarounds in place to not store this (https://github.com/kopischke/vim-stay/issues/10).

PR #1086, opened 2017-10-14
vim: do not use lcd always when temporarily changing dir

blueyed:
Fixes https://github.com/junegunn/fzf/issues/1085.

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -457,9 +457,9 @@ function! s:pushd(dict)
     if get(a:dict, 'prev_dir', '') ==# cwd
       return 1
     endif
-    let a:dict.prev_dir = cwd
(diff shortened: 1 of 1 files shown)


blueyed:
@junegunn 
Ping.
I've noticed this issue again after switching aways from this branch/patch, mostly since I have a `haslocaldir()` indicator in my statusline.  It is not nice to set a local dir, which is sticky (and inherited) to/from a window.

blueyed:
It might get improved to skip the "cd" altogether if it is not necessary, but that should be done separately (i.e. I have not looked into it regarding this now).

junegunn:
Is this a work in progress? I can see a few obvious problems in the patch.

blueyed:
@junegunn 
Please comment then - the patch is rather old and I've been using it since.

PR #1092, opened 2017-10-18
Fix known_hosts completion for custom port number

andreychernih:
My known_hosts file has a lot of records like this:

[20.20.7.168]:9722 ssh-rsa ...

This is a standard format for servers running on custom port according to http://man.openbsd.org/sshd.8#SSH_KNOWN_HOSTS_FILE_FORMAT

    A hostname or address may optionally be enclosed within ‘[’ and ‘]’ brackets then followed by ‘:’ and a non-standard port number.


junegunn:
Looks good to me, thanks!

PR #1097, opened 2017-10-19
Use fc -r instead of fzf --tac to speed up loadtime

dsh2:
Using fc -r instead of fzf --tac in the zsh history widget allows for faster load times and causes less screen flicker.

### DIFF
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -59,8 +59,8 @@ bindkey '\ec' fzf-cd-widget
 fzf-history-widget() {
   local selected num
   setopt localoptions noglobsubst noposixbuiltins pipefail 2> /dev/null
-  selected=( $(fc -l 1 |
-    FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} $FZF_DEFAULT_OPTS --tac -n2..,.. --tiebreak=index --bind=ctrl-r:toggle-sort $FZF_CTRL_R_OPTS --query=${(qqq)LBUFFER} +m" $(__fzfcmd)) )
+  selected=( $(fc -rl 1 |
+    FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} $FZF_DEFAULT_OPTS -n2..,.. --tiebreak=index --bind=ctrl-r:toggle-sort $FZF_CTRL_R_OPTS --query=${(qqq)LBUFFER} +m" $(__fzfcmd)) )
   local ret=$?
   if [ -n "$selected" ]; then
     num=$selected[1]


junegunn:
Merged, thanks!

tokikanno:
@dsh2 : Hi I got problem from the change. When I pressed ctrl + r, fzf shows

```
fc: history events are in wrong order, aborted
```

Then I found it's caused by

```
$ fc -rl
fc: history events are in wrong order, aborted
```

After reverting to last version (the fc -l version), the ctrl + r function works fine.

tokikanno:
More info: 

I asked my friend for doing the same `fc -rl` test. They got the right output

```
~ ❯❯❯❯ fc -rl
16439  fc -rl
16438  fg ; /usr/local/bin/emacs --resume-layouts -nw
16437  tmux at
16436  emacs -nw
```

But we got the different version of zsh

zsh 5.2 (x86_64-ubuntu-linux-gnu) vs zsh 4.3.17 (x86_64-unknown-linux-gnu)

Will try to find out if related to the zsh versions 🤔

junegunn:
Hi, zsh 4.3.17 is very old (http://zsh.sourceforge.net/News/), you might want to consider upgrading it. Although not clearly stated, I have never tested the scripts with zsh 4, and have no plan of making them compatible with it.

tokikanno:
After upgrade to zsh 5.4.2, still get the fc -rl problem

```
$ zsh --version
zsh 5.4.2 (x86_64-unknown-linux-gnu)

$ fc -rl
fc: history events are in wrong order, aborted
```

Maybe this is nothing about the zsh version.

junegunn:
You might want to check if your history file has some inconsistent, out-of-order entries as the error message suggests.

tokikanno:
I've tried delete my ~/.zsh_history file before but still failed.

Finally I found I need to run `fc -p` to pop all of history data out as a force refresh.

Think the problem should be solved 👍

tokikanno:
Update:

after `fc -p`, it still get broke after the next several command input .... 
I think l'll need to use the fc -l version manually till I found the root cause :-/

dsh2:
@tokikanno I am sorry you are having trouble with the change. To me it seems the update of your outdated zsh did not succeed. The cited message is not contained in zsh for a few years, cf. https://github.com/zsh-users/zsh/commit/d60de3001#diff-4168e84085fbdc944a0837ed78337962L1656. Maybe you have parallel installations of zsh in `/usr/bin` and `/usr/local/bin`?

tokikanno:
@dsh2 : You're damn right!!! 👍  
I found I didn't update the shell configuration in tmux config. So I'm still using the old zsh under tmux.

Seems the `fc -r` problem did related to the zsh version, and upgrade to a newer version will fix it.

Thx for pointing out the parallel installations problem.

PR #1098, opened 2017-10-20
Ensure ps called as command

z0rc:
When `ps` is aliased for something uncommon, like `alias ps=grc ps` which colorizes ps output, the output of `ps` can be unexpected and/or undesired.

This change makes ps to be always executed as command, even if it's aliased.


junegunn:
Can you update bash file as well?

z0rc:
Here you go. Though I'm not using bash, this was done without testing.

junegunn:
Merged, thanks!

Issue #1099, opened 2017-10-20
Upcoming fish change (removal of $FISH_VERSION) creates errors in fzf-history-widget


oschrenk:
- Category
    - [ ] fzf binary
    - [ ] fzf-tmux script
    - [x] Key bindings
    - [ ] Completion
    - [ ] Vim
    - [ ] Neovim
    - [ ] Etc.
- OS
    - [ ] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Windows Subsystem for Linux
    - [ ] Etc.
- Shell
    - [ ] bash
    - [ ] zsh
    - [x] fish

Just a heads up that https://github.com/fish-shell/fish-shell/commit/fb8ae04f80c3a129f789e7b718464d014508315f creates error output with  `fzf-history-widget` as the current implementation relies on that env variable.

You will see something like

```
test returned eval errors:
        invalid integer ''
        invalid integer ''
```

before your results.

I don't think it's urgent and I can work on a fix if you agree.

netmute:
Not strictly related to the issue, but for everyone on fish master who came here by searching for that error:

`set --export FISH_VERSION $version` is a temporary workaround you can do in your `config.fish` until this is resolved in upstream.

PR #1100, opened 2017-10-20
fish: Use $version instead of $FISH_VERSION

oschrenk:
This is the most simple fix I could think of for #1099 

If this is not robust enough for you or pre-mature (we can wait for fish 2.7 to be out), let me know.

Starting with fish 2.7 uppercase var are lowercases, see

  https://github.com/fish-shell/fish-shell/issues/4414

or in case of $FISH_VERSION renamed, see also

  https://github.com/fish-shell/fish-shell/commit/fb8ae04f80c3a129f789e7b718464d014508315f

### DIFF
--- a/shell/key-bindings.fish
+++ b/shell/key-bindings.fish
@@ -40,8 +40,8 @@ function fzf_key_bindings
     begin
       set -lx FZF_DEFAULT_OPTS "--height $FZF_TMUX_HEIGHT $FZF_DEFAULT_OPTS --tiebreak=index --bind=ctrl-r:toggle-sort $FZF_CTRL_R_OPTS +m"
 
-      set -l FISH_MAJOR (echo $FISH_VERSION | cut -f1 -d.)
-      set -l FISH_MINOR (echo $FISH_VERSION | cut -f2 -d.)
+      set -l FISH_MAJOR (echo $version | cut -f1 -d.)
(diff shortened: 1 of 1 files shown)


oschrenk:
Re-based with an even simpler version that should be robust.

`$version` is supported in every version of fish, while `$FISH_VERSION` is going away.

junegunn:
Thanks, I'll read the links and merge this before too late. It seems a little strange to me that they decided to use `$version` which doesn't sound like a special, reserved variable automatically set by fish. I can easily imagine users overriding the variable causing unexpected breakage while it's extremely unlikely that anyone would try to redefine `$FISH_VERSION`.

junegunn:
I just installed fish 2.7.0 using Homebrew, and it looks it still has `$FISH_VERSION`.

```
jg@jg-mbp ~/.v/p/fzf> echo $version
2.7.0
jg@jg-mbp ~/.v/p/fzf> echo $FISH_VERSION
2.7.0
```

oschrenk:
Commit https://github.com/fish-shell/fish-shell/commit/fb8ae04f80c3a129f789e7b718464d014508315f is not part of 2.7.0. See https://github.com/fish-shell/fish-shell/blob/2.7.0/src/env.cpp

I'm don't know why since https://github.com/fish-shell/fish-shell/issues/4414 marks it as part of it.

PR #1106, opened 2017-10-24
Allow differentiation between ESC and CTRL-C

mmorearty:
Added an action `interrupt` which exits with exit code 131 instead of 130. By default, this action is not bound to any key; but if it is bound to a key (e.g. `fzf --bind=ctrl-c:interrupt`), then it can be used to differentiate between the user pressing ESC and the user typing CTRL-C.

The action name `interrupt` seemed like a reasonable name given its similarity to `SIGINT`.

Note: I did not add a test for this, because it wasn't clear to me how to test the exit code. Also I did not update `CHANGELOG.md`. In the man page, all I did was add a line for the `interrupt` action, with no explanation of what it does.

Fixes junegunn/fzf/issues/1105

### DIFF
--- a/man/man1/fzf.1
+++ b/man/man1/fzf.1
@@ -505,6 +505,7 @@ triggered whenever the query string is changed.
     \fBforward-char\fR          \fIctrl-f  right\fR
     \fBforward-word\fR          \fIalt-f   shift-right\fR
     \fBignore\fR
+    \fBinterrupt\fR
     \fBjump\fR                  (EasyMotion-like 2-keystroke movement)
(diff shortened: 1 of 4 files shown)


SamHasler:
Has this been removed? If I set `FZF_DEFAULT_OPTS="--bind ctrl-c:interrupt"` on git bash on windows, and then at the bash prompt press CTRL+R to invoke FZF history search it just outputs:
```unknown action: interrupt```
and drops me back at the command prompt?

Should I raise this as a bug or is perhaps interrupt not available on windows?

PR #1111, opened 2017-10-26
Add bin/fzf.exe to .gitignore

me-and:
On Cygwin and MinGW, the fzf binary will have a .exe extension, so
ignore that binary if it exists as well as the bare binary.


junegunn:
Thanks!

PR #1113, opened 2017-10-28
Clear grep options

faf0:
Clear the environment variable `GREP_OPTIONS` to avoid incorrect results with grep and to potentially increase performance.

### DIFF
--- a/shell/completion.bash
+++ b/shell/completion.bash
@@ -124,7 +124,7 @@ _fzf_handle_dynamic_completion() {
   elif [ -n "$_fzf_completion_loader" ]; then
     _completion_loader "$@"
     ret=$?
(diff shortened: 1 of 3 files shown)


junegunn:
Thank you, I didn't know about GREP_OPTIONS. But a quick googling showed me that the option was [deprecated two years ago](http://savannah.gnu.org/forum/forum.php?forum_id=8152) and the use of it is now heavily discouraged, so I'll just assume that it is no longer used and ignore that it can interfere with our script, because I don't want to sprinkle `GREP_OPTIONS=''` all over the place. I've already gotten tired of repeating `command grep` or `\grep` and `GREP_OPTIONS='' command grep` just feels too much :)

PR #1115, opened 2017-10-29
_fzf_complete_ssh: support Include in ssh configs

ericmdw:
Adds support for the "Include" keyword when building the host list for `ssh **` completion.

See https://man.openbsd.org/ssh_config#Include

I have several external config files that are auto-generated with Host entries (one each for dev, staging, and production hosts). These are referenced from my `~/.ssh/config` file using `Include`, and having these hosts show up in `fzf` completion is a huge time saver.

Happy to address any feedback to make this acceptable for merge. Thanks

### DIFF
--- a/shell/completion.bash
+++ b/shell/completion.bash
@@ -233,7 +233,22 @@ _fzf_complete_telnet() {
 
 _fzf_complete_ssh() {
   _fzf_complete '+m' "$@" < <(
-    cat <(cat ~/.ssh/config /etc/ssh/ssh_config 2> /dev/null | command grep -i '^host' | command grep -v '*' | awk '{for (i = 2; i <= NF; i++) print $1 " " $i}') \
+    scan_config() {
+      shopt -s nocasematch; file="$1"; base="${2:-$(dirname "$1")}"; IFS=$'\n'
+      for line in $([[ -e "$file" ]] && cat "$file"); do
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks. Unfortunately, the code is non-trivial enough that I'm reluctant to add without test cases.

There is an ongoing discussion on leveraging external completion functions so we don't have to do the heavy lifting. See #1128.

barlik:
@ericmdw Please have a look at https://github.com/junegunn/fzf/pull/1149. It's still WIP, but once done you will be able to utilize external completion functions. Existing bash completions from https://github.com/scop/bash-completion,  that are already available in most distributions, already support `Include` statements for the ssh host completions.

ericmdw:
Hey @junegunn, thanks for taking a look. I like the idea of using external completion functions; that seems like a better solution instead of re-writing the same thing into fzf everywhere.

In the meantime, should we move forward with this PR or consider it declined? I agree that this is complex enough to need test cases, but I'm not sure how you would like test cases specified... Informally written up here, or in code somewhere? Any relevant examples I can follow?

@barlik I'll take a look at #1149, thanks! I actually didn't realize bash-completion supports `Include` since I've been using the bash3 compatible version which is pretty old, apparently. Would be nice to just delegate to bash-completion instead of putting this in fzf directly in my opinion.

PR #1122, opened 2017-11-04
[vim] Require tmap for :terminal on Windows

janlazo:
Vim's backspace doesn't work in ConEmu by default so `tnoremap` is required to fix it for winpty.
Alternative is to not use `:terminal` on ConEmu because https://conemu.github.io/en/VimXterm.html doesn't mention `tnoremap`.

It's a big jump from `8.0995` to `8.0.1108` but I can't use fzf without backspace.

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -389,7 +389,7 @@ try
   let use_height = has_key(dict, 'down') && !has('gui_running') &&
         \ !(has('nvim') || s:is_win || has('win32unix') || s:present(dict, 'up', 'left', 'right', 'window')) &&
         \ executable('tput') && filereadable('/dev/tty')
-  let has_vim8_term = has('terminal') && has('patch-8.0.995')
+  let has_vim8_term = has('terminal') && has('patch-8.0.995') && (!s:is_win || has('patch-8.0.1108'))
   let has_nvim_term = has('nvim-0.2.1') || has('nvim') && !s:is_win
   let use_term = has_nvim_term ||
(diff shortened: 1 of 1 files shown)


junegunn:
Is it okay that the users that do not use ConEmu also affected by this change?

janlazo:
It's not okay for those that run terminal vim on default Windows terminal for cmd.exe and powershell but I cannot confirm this case is limited to ConEmu only. `tmap` gives them the option to workaround Vim `:terminal` issues for terminals that use ANSI https://github.com/vim/vim/issues/2052#issuecomment-326903997

janlazo:
As long as gvim users do not run gvim from ConEmu, then detecting `$ConEmuANSI` is enough to know that Vim is running inside ConEmu. I can restrict the patch requirement to ConEmu this way. Gvim users have to unset `$ConEmuANSI` themselves.

junegunn:
I still don't see why this is necessary. If users complain about backspace key, we can just tell them it's a limitation of the GVim version they're using and they have to upgrade.

janlazo:
You mean this is something the user has to resolve themselves because the Vim plugin assumes only that the user is running terminal Vim in Window's default terminal?

GVim users don't face this backspace issue and I fixed this problem in my own config to fix keymap issues.

junegunn:
> because the Vim plugin assumes only that the user is running terminal Vim in Window's default terminal

No, my point is that, if I understood correctly, the issue, which makes it harder to use fzf, but does not completely break it, can be resolved by upgrading Vim, and that's what we should tell the users when they run into the problem.

PR #1128, opened 2017-11-06
Remove fallback to bash default completion

barlik:
This doesn't work correctly when there's 'globstar' option set
in bash. Running the completion with '**' and then cancelling it stills
triggers bash globstar search, which might be very expensive.

### DIFF
--- a/shell/completion.bash
+++ b/shell/completion.bash
@@ -310,15 +310,15 @@ done
 unset _fzf_defc
 
 # Kill completion
-complete -F _fzf_complete_kill -o nospace -o default kill
+complete -F _fzf_complete_kill -o nospace kill
 
 # Host completion
-complete -F _fzf_complete_ssh -o default ssh
-complete -F _fzf_complete_telnet -o default telnet
+complete -F _fzf_complete_ssh ssh
+complete -F _fzf_complete_telnet telnet
 
 # Environment variables / Aliases
-complete -F _fzf_complete_unset -o default unset
-complete -F _fzf_complete_export -o default export
-complete -F _fzf_complete_unalias -o default unalias
+complete -F _fzf_complete_unset unset
+complete -F _fzf_complete_export export
+complete -F _fzf_complete_unalias unalias
 
 unset cmd d_cmds a_cmds x_cmds
--- a/shell/completion.bash
+++ b/shell/completion.bash
@@ -310,15 +310,15 @@ done
 unset _fzf_defc
 
 # Kill completion
-complete -F _fzf_complete_kill -o nospace -o default -o bashdefault kill
+complete -F _fzf_complete_kill -o nospace -o default kill
 
 # Host completion
-complete -F _fzf_complete_ssh -o default -o bashdefault ssh
-complete -F _fzf_complete_telnet -o default -o bashdefault telnet
+complete -F _fzf_complete_ssh -o default ssh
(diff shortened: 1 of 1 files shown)


junegunn:
I don't think this can be a solution. We still want to use bash default completion with those commands. e.g. `ssh -i ~/.ssh/<tab>`. fzf completion should be an exception, that only happens when completion trigger is given, not the default.

barlik:
This will work, but I would still like to keep the default `**` though.

I've noticed that the problem is caused by adding the `-o bashdefault` to the completion. On my system, the default completion for ssh is: `complete -o default ssh` (Note: there's no `-o bashdefault`).
So, this still works with globstart because bash doesn't try to evaluate `**` as a globstar. FZF is adding `-o bashdefault` to the picture which causes the different (unwanted) behavior.

I've just tried it with: `complete -F _fzf_complete_ssh -o default` and `ssh -i ~/.ssh/<tab>` still works and `ssh **` works as expected for me as well.

junegunn:
>  On my system, the default completion for ssh is: complete -o default ssh (Note: there's no -o bashdefault).

I see, I'd like to know if it's universally true. Do you think it's safe to make this change?

barlik:
> I see, I'd like to know if it's universally true. Do you think it's safe to make this change?

I must admit that I don't really understand the different between 'default' and 'bashdefault'. The bash man page is not very clear about that. I'll try to investigate more the difference between them and also try it with different distributions. So far, I've tried Archlinux and Fedora and both work better without the 'bashdefault'.

barlik:
I've looked deeper into the fzf and bash completions to have a better understanding how it works. I think that my previous solution of removing `-o bashdefault` is not enough.  

I believe that both `-o default` and `-o bashdefault` should be removed because they are changing the behaviour of installed bash completions.

Here's a demonstration:

```
# Create a file called '-zabc'
cd $(mktemp)
touch -- -zabc
```
1. First test scenario: No bash completion installed, no fzf completions:
```
ssh -z<tab> -> -zabc
# There's no "smart" completion so bash defaults to filename completion
# This is fine, at least some completion is better than nothing
```
2. Second scenario: Installed bash completions without fzf completions:
```
ssh -z<tab> ->   # no completion
# This is actually correct as there's no -z flag for ssh!

ssh -i -z<tab> -> -zabc
# This is correct as -i parameters expects a file
```

3. Third scenario: Installed bash completions + fzf completion:
```
ssh -z<tab> -> -zabc
# fzf firstly tries original bash completion. That correctly returns an empty
# response. However, fzf then incorrectly tries filename completion as well
```

### My proposed solution:
* Remove `-o default -o bashdefault` from the completions. That will fix the case when bash completions are installed. I've looked at Archlinux, Fedora and Ubuntu and the default completion for ssh (and other commands) is: `complete -F _ssh ssh`. There's no `-o default` nor `-o bashdefault`
  * I assume that bash completions once installed are the same on every distribution.
* Modify `_fzf_handle_dynamic_completion` and add
```
else
  COMPREPLY=( $(compgen -o default -o bashdefault -- "$cur") )
  return 0
fi
```
at the end of that function. This will handle the cases when the bash completions are not installed. In that case we want to fall-back to the default filename completion.

barlik:
Proof of concept work at: https://github.com/junegunn/fzf/pull/1149

Hopefully, I'll find some time during the weekend to work on it.

PR #1136, opened 2017-11-11
Fire zsh precmd functions after cd

### LINKED ISSUE #915: Alt-c binding does not fire precmd_functions in zsh, so things like direnv do not work
- Category
    - [ ] fzf binary
    - [ ] fzf-tmux script
    - [x] Key bindings
    - [ ] Completion
    - [ ] Vim
    - [ ] Neovim
    - [ ] Etc.
- OS
    - [ ] Linux
    - [x] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [ ] bash
    - [x] zsh
    - [ ] fish

If I use Alt-c to change directories, the directory change happens, but precmd_functions (like the one installed by direnv) do not fire. This means I need to hit return one more time to get a new prompt and trigger direnv.

I originally reported this here: https://github.com/urbainvaes/fzf-marks/issues/6 but it is an issue here as well. I don't know enough about zsh/zle/etc to fix it myself, so I'm hoping someone here has an idea. Thanks!

aaronjensen:
Fixes #915

### DIFF
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -36,6 +36,16 @@ fzf-file-widget() {
 zle     -N   fzf-file-widget
 bindkey '^T' fzf-file-widget
 
+# Ensure precmds are run after cd
+fzf-redraw-prompt() {
+  local precmd
+  for precmd in $precmd_functions; do
+    $precmd
+  done
+  zle reset-prompt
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks!

PR #1141, opened 2017-11-14
Add Fedora installation information.

QuLogic:
Hi! fzf is now available in Fedora 26+; here are some instructions for installing that way.


junegunn:
Cool, thanks!

PR #1147, opened 2017-11-15
Add the ctrl-o for search in current git branches

javidh:
This enables the users to search in the current directory git branches. 
Handy moving between branches.

### DIFF
--- a/shell/key-bindings.bash
+++ b/shell/key-bindings.bash
@@ -11,6 +11,14 @@ __fzf_select__() {
   echo
 }
 
+__fzf_git_branches__(){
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks, but I decided not to add more key bindings to this repository due to excessive maintenance burden. It's really painful to make sure that they work consistently on all three different shells (bash 3, bash 4, zsh, and fish) on different platforms.

FYI, here are git key bindings I use: https://junegunn.kr/2016/07/fzf-git

javidh:
alright. Thanks for the link.

PR #1151, opened 2017-11-21
update plug option in readme

pruett:
i received an error when using the settings in the readme. changing the `dir` option to `rtp` fixed installation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,7 +112,7 @@ But it's recommended that you use a plugin manager like
 [vim-plug](https://github.com/junegunn/vim-plug).
 
 ```vim
-Plug 'junegunn/fzf', { 'dir': '~/.fzf', 'do': './install --all' }
+Plug 'junegunn/fzf', { 'rtp': '~/.fzf', 'do': './install --all' }
 ```
 
 ### Fedora


junegunn:
No, `dir` is right in this case.

- `dir` designates the directory where the plugin will be installed. If not given, it's `g:plug_home` (which is usually `~/.vim/plugged`) + the name of the plugin.
- `rtp` is only used when the plugin does not follow the standard directory structure.

pruett:
thanks for the response,

i'm still confused...
`brew install fzf` installs fzf, so wouldn't I want to point to where homebrew installed it, or is it different when installing a vim plugin?

pruett:
The following seems to work, although it installs the first plugin to `~/.fzf` even though `g:plug_home` is `~/.vim/plugged`

```vim
call plug#begin()
Plug 'junegunn/fzf'
Plug 'junegunn/fzf.vim'
call plug#end()
```

junegunn:
I guess the description can be confusing to the users without enough knowledge of Vim.

I just updated the README page, hope it helps.

https://github.com/junegunn/fzf#as-vim-plugin

junegunn:
> although it installs the first plugin to ~/.fzf

I think you're somehow confused, it doesn't.

pruett:
thanks @junegunn ! That clears it up a bit.

PR #1156, opened 2017-11-22
Functionalize zsh completion script.

QuLogic:
Let me first preface this by saying I don't use zsh so please let me know if this is a bad idea. On Fedora with bash, I install the `completion.bash` to the global directory and it works automatically once installed. With zsh, I try to install the `completion.zsh` to `site-functions`, but this doesn't work.

What seems to be missing is the `compdef` at the beginning (the shebang seems unnecessary since it's expected to be `source`d) and the top-level code needs to be in a function (and install as `_fzf`.) This allows it to be loaded from the zsh `site-functions` directory automatically. The `fzf` call at the end allows it to be `source`d just like before.

Anyway, like I said, I don't really use zsh, so I've only done minimal testing but this seems to work in both scenarios.

### DIFF
--- a/shell/completion.zsh
+++ b/shell/completion.zsh
@@ -1,4 +1,4 @@
-#!/bin/zsh
+#compdef fzf
 #     ____      ____
 #    / __/___  / __/
 #   / /_/_  / / /_
@@ -9,25 +9,7 @@
 # - $FZF_TMUX_HEIGHT        (default: '40%')
 # - $FZF_COMPLETION_TRIGGER (default: '**')
 # - $FZF_COMPLETION_OPTS    (default: empty)
-
 # To use custom commands instead of find, override _fzf_compgen_{path,dir}
-if ! declare -f _fzf_compgen_path > /dev/null; then
-  _fzf_compgen_path() {
-    echo "$1"
-    command find -L "$1" \
-      -name .git -prune -o -name .svn -prune -o \( -type d -o -type f -o -type l \) \
-      -a -not -path "$1" -print 2> /dev/null | sed 's@^\./@@'
-  }
-fi
-
-if ! declare -f _fzf_compgen_dir > /dev/null; then
-  _fzf_compgen_dir() {
-    command find -L "$1" \
-      -name .git -prune -o -name .svn -prune -o -type d \
-      -a -not -path "$1" -print 2> /dev/null | sed 's@^\./@@'
-  }
-fi
-
 ###########################################################
 
 __fzfcmd_complete() {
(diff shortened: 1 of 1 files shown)


junegunn:
I'm a bash user and I don't really know about `site-functions`, I'll have to do some research before merging this. Any comments from zsh users are welcome.

Anyway, should we use a more descriptive name than `_fzf`? Such as `_fzf_init_completion`.

QuLogic:
I was basing the name on [this article](https://askql.wordpress.com/2011/01/11/zsh-writing-own-completion/) using, e.g., `_hello` to complete `hello`. Additionally, most of the completions in `site-functions` followed thi pattern (except `_mercurial` for `hg`, for some reason). Of course, fzf completion works for more commands than just `fzf`, so I'm not sure how much the same name is needed.

midchildan:
> With zsh, I try to install the completion.zsh to site-functions, but this doesn't work.

Files in `site-functions` aren't meant to be used as initialization scripts. Instead, each file in `site-functions` represent a single function that can be automatically loaded the first time you use it. It is intended to be used to speed up shell initialization. (See the [Functions Section](http://zsh.sourceforge.net/Doc/Release/Functions.html) in the zsh docs)

> What seems to be missing is the compdef at the beginning (the shebang seems unnecessary since it's expected to be sourced) and the top-level code needs to be in a function (and install as _fzf.) This allows it to be loaded from the zsh site-functions directory automatically. The fzf call at the end allows it to be sourced just like before.

By doing this, you're defining a completion function for the `fzf` command. By taking a quick look at the changes made in this PR, it looks like the initialization commands for fzf completion won't run until you hit `fzf<tab>`.

QuLogic:
I think you're right; I must have been testing by running `fzf <tab>` first every time. This version definitely continues to work with the old method of sourcing the file manually. Do you have any suggestions on how this could be made to work with all commands supported by fzf?

midchildan:
I'm not sure that creating a specialized completion function for a set of commands is a good idea, because what we're trying to do isn't exactly an extension of the zsh completion system. Instead, we want to create a whole new completion system that is independent from the stock completion system.  `completion.zsh` already does this just fine by defining a [ZLE widget](http://zsh.sourceforge.net/Doc/Release/Zsh-Line-Editor.html), which is a function that can be bound to a key sequence.

There are good reasons for placing these widgets in `site-functions` , though. By placing widgets in a well-defined location, users will be able to assign them to any key sequence they like without having to `source` the scripts. It might even result in a slightly faster shell initialization time, since functions in `site-functions` won't be loaded until they are called. You might want to see my [.zshrc](https://github.com/midchildan/dotfiles/blob/8df547a6204d04b84e04ae08636e3fd5310d87a4/home/.zshrc#L96-L99) and [site-functions](https://github.com/midchildan/dotfiles/tree/8df547a6204d04b84e04ae08636e3fd5310d87a4/home/.local/share/zsh/site-functions) from my dotfiles if you want to go down this route. It contains modified widgets from `completion.zsh`.

Issue #1145, opened 2017-11-15
How to append history command instead of replace?


eyalev:
- Category
    - [ ] fzf binary
    - [x] fzf-tmux script
    - [x] Key bindings
    - [ ] Completion
    - [ ] Vim
    - [ ] Neovim
    - [ ] Etc.
- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Windows Subsystem for Linux
    - [ ] Etc.
- Shell
    - [ ] bash
    - [x] zsh
    - [ ] fish

I would like to append a command from history instead of replacing the whole line.

For example:

```
$ first-command && [launch fzf, choose second command to be appended] second-command && [launch fzf, same..] third-command
```

How can I do this?

Maybe with `--expect`? or `FZF_CTRL_T_COMMAND`?

junegunn:
It's not supported by the key bindings this repository provides and I have no idea how to implement it. Feel free to share your solution when you find one, thanks.

eyalev:
This seems to work:

```
# In .zshrc

fzf-history-append-widget() {
  selected=( $(fc -il -30000 | eval "$(__fzfcmd) +e --tac +m -n2..,.. --tiebreak=index --preview='echo {}' --preview-window=up:30%:wrap --toggle-sort=ctrl-r $FZF_CTRL_R_OPTS") )
  num=$selected[1]
  command=$(fc -nl "${num}" | head -n 1)
  LBUFFER="${LBUFFER}${command} "
  local ret=$?
  zle redisplay
  typeset -f zle-line-init >/dev/null && zle zle-line-init
  return $ret
}

zle     -N   fzf-history-append-widget
bindkey '^O' fzf-history-append-widget

```

jlescher:
fzf-history-widget() {
  local selected num lbuffer
  lbuffer=${LBUFFER}
  setopt localoptions noglobsubst noposixbuiltins pipefail no_aliases 2> /dev/null
  selected=( $(fc -rl 1 | perl -ne 'print if !$seen{(/^\s*[0-9]+\**\s+(.*)/, $1)}++' |
    FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} $FZF_DEFAULT_OPTS -n2..,.. --tiebreak=index --bind=ctrl-r:toggle-sort,ctrl-z:ignore $FZF_CTRL_R_OPTS --query=${(qqq)} +m" $(__fzfcmd)) )
[...]

PR #1161, opened 2017-11-24
[zsh] Update fzf-history-widget for multi-select

ericbn:
Allow fzf-history-widget in zsh to accept multi-select. If one entry is selected, then vi-fetch-history is invoked, as in previous code. If multiple entries are selected, then they're joined with ` && `.

### DIFF
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -68,15 +68,15 @@ bindkey '\ec' fzf-cd-widget
 # CTRL-R - Paste the selected command from history into the command line
 fzf-history-widget() {
   local selected num
-  setopt localoptions noglobsubst noposixbuiltins pipefail 2> /dev/null
-  selected=( $(fc -rl 1 |
-    FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} $FZF_DEFAULT_OPTS -n2..,.. --tiebreak=index --bind=ctrl-r:toggle-sort $FZF_CTRL_R_OPTS --query=${(qqq)LBUFFER} +m" $(__fzfcmd)) )
+  setopt localoptions noglobsubst noposixbuiltins pipefail extendedglob 2> /dev/null
+  selected=( "${(@f)$(fc -rl 1 |
+    FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} $FZF_DEFAULT_OPTS -n2..,.. --tiebreak=index --bind=ctrl-r:toggle-sort $FZF_CTRL_R_OPTS --query=${(qqq)LBUFFER} -m" $(__fzfcmd))}" )
   local ret=$?
-  if [ -n "$selected" ]; then
-    num=$selected[1]
-    if [ -n "$num" ]; then
-      zle vi-fetch-history -n $num
-    fi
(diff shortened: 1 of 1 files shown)


ericbn:
I was "intuitively" expecting for this to be available...  :- )

@junegunn, if you do approve the ideia, I can try to do the same for bash and fish (although I'm zero familiar with fish).

shell/key-bindings.zsh
```diff
@@ -68,15 +68,15 @@ bindkey '\ec' fzf-cd-widget
...
+  selected=( "${(@f)$(fc -rl 1 |
+    FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} $FZF_DEFAULT_OPTS -n2..,.. --tiebreak=index --bind=ctrl-r:toggle-sort $FZF_CTRL_R_OPTS --query=${(qqq)LBUFFER} -m" $(__fzfcmd))}" )
   local ret=$?
-  if [ -n "$selected" ]; then
-    num=$selected[1]
-    if [ -n "$num" ]; then
-      zle vi-fetch-history -n $num
-    fi
+  [[ -n $selected ]] && if [[ ${#selected} -eq 1 ]]; then
+    num=${${selected[1]## #}%%[^0-9]*}
```
ericbn: The extended globs here mean:

`${foo## #}` -> remove spaces at the left of foo
`${foo%%[^0-9]*}` -> remove everything from the right of foo, expect for the initial digits

shell/key-bindings.zsh
```diff
@@ -68,15 +68,15 @@ bindkey '\ec' fzf-cd-widget
...
+    [[ -n $num ]] && zle vi-fetch-history -n $num
+  else
+    LBUFFER=${(j: && :)${selected## #[^ ]# #}}
```
ericbn: The extended glob here means: `${foo## #[^ ]# #}` -> remove ` *[^ ]* *` (zero-or-more spaces, zero-or-more non-spaces, zero-or-more spaces) from the left of foo

junegunn:
Thanks, it's an interesting idea. However, for now, I'd rather not extend the functionality of the key bindings this repository provides to keep the code manageable in the long run. It's been pretty stressful for me to make sure that they work consistently across three different shells on different platforms, when I'm just a bash user.

There have been other suggestions regarding CTRL-R binding, e.g. key to append to the command line instead of replacing it (#1145), extra key bindings to extract positional arguments from commands (#1163). They can be nice enhancements, but when combined, they will make the implementation unnecessarily complex for most users who only need the default functionality, and the code will likely have more bugs and I'll have to put more of my time maintaining the code that I don't even use.

I'd like to see more advanced versions of the bindings maintained in separate, third-party repositories (`somebody/fzf-advanced-ctrl-r-for-zsh`, `another-user/fzf-better-ctrl-t-for-fish`, etc). Users looking for extra features can optionally install those "plugins", and the others can still the default implementation this repository provides, and don't have to learn about the features they don't care about.

I'm closing this, but thanks again for your interest in the project.

ericbn:
Maybe this can go to the [Examples](https://github.com/junegunn/fzf/wiki/Examples) wiki page? I just added the code to my `~/.zshrc` file after it sources `~/.fzf.zsh`.

Issue #449, opened 2015-12-15
Adding --line-number ?


netei:
Grep has this option : 

> -n, --line-number
       Prefix each line of output with the 1-based line number  within  its  input  file.   (-n  is
       specified by POSIX.)

This is useful when you would like to know the line the user selected, for example if multiple lines have the same content but the position in the file matters (which could be useful for a log searcher I am writing).

junegunn:
http://superuser.com/questions/10201/how-can-i-prepend-a-line-number-and-tab-to-each-line-of-a-text-file

Does this help?

netei:
Indeed, that does the trick combined with --delimiter and --nth

PR #1172, opened 2017-12-04
Made sure terminal closes on exit

ronmrdechai:
Resolves [issue #449 in fzf.vim](https://github.com/junegunn/fzf.vim/issues/449) for `vim8`.

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -699,7 +699,7 @@ function! s:execute_term(dict, command, temps) abort
     if has('nvim')
       call termopen(command, fzf)
     else
-      let t = term_start([&shell, &shellcmdflag, command], {'curwin': fzf.buf, 'exit_cb': function(fzf.on_exit)})
+      let t = term_start([&shell, &shellcmdflag, command], {'curwin': fzf.buf, 'exit_cb': function(fzf.on_exit), 'term_finish': 'close'})
       if !has('patch-8.0.1261') && !has('nvim') && !s:is_win
         call term_wait(t, 20)
       endif


junegunn:
Two separate issues are discussed in https://github.com/junegunn/fzf.vim/issues/449, so let me clarify what we're trying to fix here.

### Problem definition

When `let g:fzf_layout = { 'window': '10split enew' }` is set on Vim 8, `:FZF` opens the selected buffer on the temporary window, instead of opening it after closing the window.

The problem is not reproducible with `let g:fzf_layout = { 'window': '10split' }`. So we can presume that it has to do with the internal management of buffer ids.

### Cause

The bug is due to the inconsistent behavior of `term_start` of Vim 8 and `termopen` of nvim. Unlike `termopen`, `term_start` opens a new buffer with a new buffer id.

### Fix

I believe a proper fix would be as follows. We have to take the return value of `term_start`, a new buffer id, and use it in `on_exit` callback.

```diff
diff --git a/plugin/fzf.vim b/plugin/fzf.vim
index 8c3932f..27b5332 100644
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -699,9 +699,9 @@ function! s:execute_term(dict, command, temps) abort
     if has('nvim')
       call termopen(command, fzf)
     else
-      let t = term_start([&shell, &shellcmdflag, command], {'curwin': fzf.buf, 'exit_cb': function(fzf.on_exit)})
+      let fzf.buf = term_start([&shell, &shellcmdflag, command], {'curwin': 1, 'exit_cb': function(fzf.on_exit)})
       if !has('patch-8.0.1261') && !has('nvim') && !s:is_win
-        call term_wait(t, 20)
+        call term_wait(fzf.buf, 20)
       endif
     endif
   finally
```

What do you think?

tuyenpm9:
This works well for me. Title updated instantly.

tuyenpm9:
Sorry, It didn't work, just re-test. Redraw will work https://github.com/junegunn/fzf/pull/1173

PR #1173, opened 2017-12-05
[vim] Force title update on Windows terminal exits

janlazo:
This should fix the delayed update on window title for GVim (Windows)


junegunn:
Thanks, I was unable to reproduce the problem described in the thread yesterday, I'll check again.

By the way, is there any chance that https://github.com/junegunn/fzf/pull/1172#issuecomment-349227492 helps in the case?

tuyenpm9:
Thanks, by chance can we change title of terminal when fzf invoke it to something more clear?, for example: `fzf`

janlazo:
@junegunn It didn't make a difference for this issue.

@tuyenpm9 That's a separate issue but adding `'term_name': 'fzf'` to the Vim terminal job options should do the trick.

janlazo:
PR for the following Vim patch: https://github.com/vim/vim/pull/1969
Vim Patch: https://github.com/vim/vim/commit/62426e168ef29fb941d1bd91a112fa09445322a5

janlazo:
I can reproduce the delayed title update only after running `FZF` one after another. It's hard to reproduce it consistently on the first run of `FZF`. It doesn't have to be multiple terminal buffers open at once. It is more apparent in large directories.

tuyenpm9:
I see the delayed appears somehow but seem you found it. It's hard to reproduce because of inconsistency.

tuyenpm9:
Would we merge this PR?

PR #1177, opened 2017-12-11
[zsh] let key bindings be customized through zstyle

piec:
Hi,

This commits lets the user set its own keybinds using [zstyle](http://zsh.sourceforge.net/Doc/Release/Zsh-Modules.html#The-zsh_002fzutil-Module).
Example use case: keep the existing ^R widget and make fzf's history widget available with another binding.

For example in the standard `~/.fzf.zsh`
```shell
zstyle ':fzf:bindings' file '^T'
zstyle ':fzf:bindings' cd '\ec'
zstyle ':fzf:bindings' history '^R'
source …/key-bindings.zsh
```
Why [zstyle](http://zsh.sourceforge.net/Doc/Release/Zsh-Modules.html#The-zsh_002fzutil-Module)?  
I'm a [prezto](https://github.com/sorin-ionescu/prezto) user which uses zstyle for most user settings. I find it pretty clean because it doesn't pollute the variable environment. Also it's a standard zsh feature.

Would you prefer environment variables for this pull request?

---
A setting for the [Dedicated completion key](https://github.com/junegunn/fzf/wiki/Configuring-fuzzy-completion#dedicated-completion-key) could be handled the same way as well. Rather than having the original binding saved to `$fzf_default_completion` and ask the user to rebind it to `^I`

Cheers & thanks for fzf,
Pierre


nomasprime:
@junegunn I'm using the workarounds mentioned in existing issues but would be great if the bindings could be customised.

Looks like this PR's been sat around a long time. Any reason it shouldn't be merged?

piec:
I apply this patch after each fzf upgrade :p
Do not hesitate to tell me if you would like to see changes in this MR

piec:
Rebased on master :)

tjx666:
@piec Do you know how to configure the character sequence as zstyle key binding?

The context ref here: https://github.com/Aloxaf/fzf-tab/pull/148#issuecomment-1279920810

piec:
Hi @tjx666 no, but it looks like you found a way

PR #1185, opened 2017-12-25
Allow external config of fzf command

jflanglois:
This is mainly meant to support an environment like [NixOS](https://nixos.org/), where the binaries will not be in the same directory, and after-the-fact installation of binaries is either difficult or discouraged.

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -100,9 +100,9 @@ endfunction
 
 let s:default_layout = { 'down': '~40%' }
 let s:layout_keys = ['window', 'up', 'down', 'left', 'right']
-let s:fzf_go = s:base_dir.'/bin/fzf'
-let s:fzf_tmux = s:base_dir.'/bin/fzf-tmux'
-let s:install = s:base_dir.'/install'
+let s:fzf_go = get(g:, 'fzf_command', s:base_dir.'/bin/fzf')
+let s:fzf_tmux = get(g:, 'fzf_tmux_command', s:base_dir.'/bin/fzf-tmux')
+let s:install = get(g:, 'fzf_install_command', s:base_dir.'/install')
 let s:installed = 0
 
 let s:cpo_save = &cpo


junegunn:
Thanks but I don't like adding more configuration variables and I don't think this is necessary for the following reasons.

Given that you have installed fzf properly on your system,
1. the Vim plugin will work as expected using `fzf` binary found on `$PATH`
    - https://github.com/junegunn/fzf/blob/93aeae198511af76dc5b311601266bcaef36ba80/plugin/fzf.vim#L115-L116
2. `install` script is no longer needed
3. and `fzf-tmux` is not needed on modern Vim and Neovim

jflanglois:
@junegunn I understand that, but the whole point is to not have to pollute `PATH` just because it's used as a vim plugin. How about making just the fzf executable configurable?

junegunn:
Sorry but I don't understand your argument. How are we polluting `$PATH`? You can put fzf binary in any of the directories in your `$PATH` you prefer. Maybe you don't want to have fzf in your `$PATH`? In my opinion, fzf is more useful outside Vim and I don't see why anyone who's already using it in Vim would not want it to be available on their system.

If you really don't want to have fzf in your $PATH for some reason, you can manipulate your `$PATH` in your vim configuration file (`let $PATH .= ...`) so that it's only available inside Vim.

jflanglois:
Yeah, I don't want to have it in my path necessarily. One of the selling points of the nix package manager is that you can have isolated environments with all dependencies accounted for (and not coming from some global environment). In any case, I'll try out modifying the vim path as you suggest, but it seemed to me like there was a precedent among other vim plugins to specify binaries that they depend on.

junegunn:
Okay. I have over 2800 executables in my `$PATH` and I'm pretty sure I use maybe less than 10% of them, but I'm totally fine with it, and I don't mind if there's one more :)

By the way, like I said fzf is pretty useful outside Vim as well. There are many users who don't use it in Vim (or don't use Vim at all) but still use it regularly on command-line.

PR #1194, opened 2018-01-09
[vim] Detect fzf within wsl on Windows

### LINKED ISSUE #1191: Is it possible fzf on WSL work with GVim in Windows
- Category
    - [x] fzf binary
    - [ ] fzf-tmux script
    - [ ] Key bindings
    - [ ] Completion
    - [x] Vim
    - [ ] Neovim
    - [ ] Etc.
- OS
    - [ ] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [x] bash
    - [ ] zsh
    - [ ] fish

I have installed fzf on WSL and it works well, also have Vim 8.0 has terminal feature
Is that possible fzf on WSL work with gvim on Windows?

janlazo:
Close #1191

See the commit message for more info.
@rickywu Can you check if this works?

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -115,6 +115,12 @@ function! s:fzf_exec()
     elseif executable('fzf')
       let s:exec = 'fzf'
     elseif s:is_win && !has('win32unix')
+      if executable('wsl')
+        let wsl_fzf = get(split(system('wsl bash --login -c "command -v fzf"'), "\n"), 0, '')
+        if !v:shell_error && !empty(wsl_fzf)
+          return 'wsl.exe '.wsl_fzf
+        endif
+      endif
       call s:warn('fzf executable not found.')
       call s:warn('Download fzf binary for Windows from https://github.com/junegunn/fzf-bin/releases/')
       call s:warn('and place it as '.s:base_dir.'\bin\fzf.exe')
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -297,6 +297,8 @@ function! fzf#wrap(...)
     let opts.name = name
   end
 
+try
+  let [shell, shellslash, shellcmdflag, shellredir, shellxquote, shellquote] = s:use_sh()
   " Layout: g:fzf_layout (and deprecated g:fzf_height)
   if bang
     for key in s:layout_keys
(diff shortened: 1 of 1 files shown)


janlazo:
@rickywu Can you try again?

janlazo:
@rickywu Can you provide the full error log and any custom config you use for fzf? Also, can you test if you can run batchfiles (*.cmd, *.bat) within wsl.exe and whether it accepts `\r\n` or not?

janlazo:
@rickywu Do you have wslpath? See https://github.com/Microsoft/WSL/issues/1834#issuecomment-298984386

rickywu:
wsl.exe accept shell script with both dos/unix format

set shell=cmd.exe works well but set shell=wsl.exe not work and If I issue :FZF command just run vimrun and flash across, I only can capture error by print screen in time

I tried use BreakPts to debug but got some error message about not found script in current path

rickywu:
If you need Windows env, I'll try deploy an AWS windows if available

janlazo:
@rickywu If you don't set `shellredir` (use default value for cmd.exe), does it do anything?

Is this enough to run wsl bash using the default settings on GVim?
```
set shell=wsl.exe
set shellpipe=|
```

rickywu:
It's enough but still has problem,
I only set shell=wsl.exe also has problem.

I captured error msg, it's about unexpected EOF
fzf

janlazo:
Sorry, I forgot `shellcmdflag`. I'll edit my previous post. Mention anything else I missed as we go.

rickywu:
unexpected EOF while looking for matching `"'
syntax error: unexpected end of file

janlazo:
I guess it doesn't use `CommandLineToArgW` (for Neovim users, see `:h jobstart` for more info).
All arguments to wsl must be passed as one whole string.

Using bash.exe (msysgit), I get the following:
```dosbatch
> bash -c \"echo\" a b
/usr/bin/bash: -c: line 0: unexpected EOF while looking for matching `"'
/usr/bin/bash: -c: line 1: syntax error: unexpected end of file
> bash -c "\"echo\" a b"
a b
```

janlazo:
How about now?

rickywu:
the same problem, I jsut set shell=wsl.exe

rickywu:
applied your changes but the same problem

janlazo:
Minimal vimrc for gvim

```
set rtp+=~/.fzf
```

Assuming the user git-clone the fzf repo in both home directory (Windows, WSL) and fzf (wsl) is in user's bin/, this should be enough. Tested this on WSL.

rickywu:
I tested as I said it's running in cmd but not in wsl

my vimrc:
set rtp+=~/.fzf
set shell=wsl.exe

still the same problem, also test 
set rtp+=~/.fzf
set shell=wsl.exe
set shellxquote = "

I verified .fzf is under C:/Users/Ricky(Windows) and /home/ricky(WSL)

janlazo:
I haven't set `shellxquote` and `shellquote` for cmd.exe yet. I'll  still have to test with `set shell=wsl.exe`.

janlazo:
`fzf#wrap` temporarily sets the shell now so `fzf#shellescape` should work as expected. Setting the shell to something else caused it to use the default `shellescape` from Vim so none of the arguments were escaped properly if I set the shell to anything but cmd.exe prior to running `:FZF`.

rickywu:
nice, works fine, thank you.

janlazo:
@junegunn Can you review the changes?

rickywu:
someone can merge the wsl ag support for fzf.vim?

janlazo:
@rickywu That's a different issue and the changes you propose for `Ag` don't have to be in the fzf.vim repo. You have the autoload functions from autoload/fzf/vim.vim in the fzf.vim repo to define your new commands.

rickywu:
Works well

janlazo:
@junegunn I'm done with my changes. If you don't want to support running fzf (wsl) for GVim (Windows), then consider merging the patch for `fzf#wrap`.

janlazo:
Closing because of lack of interest.

PR #1195, opened 2018-01-10
Make default answer "y" in install script

coaxial:
The current install script hints that `y` will be the default answer to the various prompts, but if the user presses `Enter` without entering an answer, nothing happens.

This PR sets the default answer to `y` as hinted in the prompt `[y]/n` so that the user can keep pressing `Enter` to accept the default answer.

### DIFF
--- a/install
+++ b/install
@@ -69,6 +69,7 @@ fzf_base="$(pwd)"
 ask() {
   while true; do
     read -p "$1 ([y]/n) " -r
+    REPLY=${REPLY:-"y"}
     if [[ $REPLY =~ ^[Yy]$ ]]; then
       return 1
     elif [[ $REPLY =~ ^[Nn]$ ]]; then


coaxial:
The last two commits are shellcheck's findings. The build was failing for some reason so I thought it might be because of these, but it passes now `¯\_(ツ)_/¯`. Leaving them in anyway as they fix bash issues anyway but they're not strictly necessary.

install
```diff
@@ -351,8 +352,8 @@ fi
 echo
 for shell in $shells; do
   [[ "$shell" = fish ]] && continue
-  [ $shell = zsh ] && dest=${ZDOTDIR:-~}/.zshrc || dest=~/.bashrc
-  append_line $update_config "[ -f ~/.fzf.${shell} ] && source ~/.fzf.${shell}" "$dest" "~/.fzf.${shell}"
+  [ "$shell" = zsh ] && dest=${ZDOTDIR:-~}/.zshrc || dest=~/.bashrc
+  append_line $update_config "[ -f $HOME/fzf.${shell} ] && source $HOME/fzf.${shell}" "$dest" "$HOME/fzf.${shell}"
```
junegunn: We want to add literal `~` to the file (e.g. `[ -f ~/.fzf.bash ] && source ~/.fzf.bash`), so this was intentional.

coaxial: I see I dropped the `.` in `.fzf` my mistake too. I'll push an update today.

install
```diff
@@ -69,6 +69,7 @@ fzf_base="$(pwd)"
 ask() {
   while true; do
     read -p "$1 ([y]/n) " -r
+    REPLY=${REPLY:="y"}
```
junegunn: Although this works, the assignment is technically redundant. So we can instead write 1. `: ${REPLY:="y"}` which is more concise, 2. or `REPLY=${REPLY:-"y"}` to remove implicit assignment. What do you think?

coaxial: I usually do the second version.

junegunn: Me too, it clearly shows the intention.

junegunn:
Merged, thanks!

piotr-yuxuan:
I just installed version 0.17.3 and `/usr/local/opt/fzf/install` doesn't include this change. When would you release it?

``` diff
 ask() {
   while true; do
     read -p "$1 ([y]/n) " -r
+    REPLY=${REPLY:-"y"}
     if [[ $REPLY =~ ^[Yy]$ ]]; then
       return 1
     elif [[ $REPLY =~ ^[Nn]$ ]]; then
       return 0
     fi
   done
 }
```

junegunn:
Soon, probably in April.

PR #1197, opened 2018-01-11
Add vi command mode like mode for move up/down

t-msn:
Hello, 

Sometimes i'd like to move up/down without ctrl key and this commit will do that (just switching a keymap). Apparent problem of current implementation is keymap for vi-mode cannot be changed.

If you don't think this is good, please ignore this.
Thanks.

----
This commit adds vi command line like mode which can be enabled by ctrl-o or specifying --vi-mode at start time.

In this mode, key bindings will be changed as follows:
   i       ... back to default mode
   j       ... move down
   k       ... move up
   g       ... move top
   G       ... move bottom
   o       ... return
   q       ... quit
   f,space ... move half page
   b       ... move back half page

Also, action 'toDefaultMode' and 'toViMode' are added.

### DIFF
--- a/man/man1/fzf.1
+++ b/man/man1/fzf.1
@@ -146,6 +146,19 @@ actions are affected:
 .TP
 .BI "--jump-labels=" "CHARS"
 Label characters for \fBjump\fR and \fBjump-accept\fR
+.TP
+.BI "--vi-mode"
+Enable vi command line like mode by default (otherwise can be entered this mode by ctrl-O).
+In this mode, key bindings will be changed as follows:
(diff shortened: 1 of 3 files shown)


junegunn:
Thanks for the suggestion, it looks interesting, but it's not something I want to add to fzf. Since fzf is quite stable and I'm not planning to make significant changes to it, I suggest that you use your fork if you find it useful.

fsc0:
It's quite possible now with **--no-input** mode
Check https://github.com/junegunn/fzf/discussions/4215#discussion-7896938

PR #1208, opened 2018-01-31
Do not append an extra space when completing

lanker:
E.g., git commands.


junegunn:
Thanks, but I don't think this can be a solution. `nospace` is not the default mode of the generic bash completion. e.g. `ls SOME_FILE_NA<tab>` is supposed to append a space at the end. Though we can make our own completion function append a space after the selected entry, but it falls back to the default bash completion when completion trigger `**` is not given, and in that case `nospace` is not what we want.

PR #1212, opened 2018-02-07
[zsh] Make default ignore pattern to be configurable

mexus:
Since both `FZF_CTRL_T_COMMAND` and `FZF_ALT_C_COMMAND` share the same
ignore pattern for the `file` command, it makes sense to move the
pattern to a distinct variable and let the user to optionally provide it
via a global variable.

It is useful to let the user to *conceal* some commonly ignore directories (a `build` directory for example).

What do you think?

### DIFF
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -2,9 +2,17 @@
 # ------------
 if [[ $- == *i* ]]; then
 
+local default_ignore="${FZF_DEFAULT_IGNORE:-\
+    -path '*/\\.*' \
+    -o -fstype 'sysfs' \
+    -o -fstype 'devfs' \
+    -o -fstype 'devtmpfs' \
+    -o -fstype 'proc' \
+}"
+
 # CTRL-T - Paste the selected file path(s) into the command line
 __fsel() {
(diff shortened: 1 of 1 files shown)


mexus:
Actually I've come up with an alternative solution: what if we implement something like 'ctrl-g' command to go through only the git files? It will cover my use case at least :)

shell/key-bindings.zsh
```diff
@@ -2,9 +2,17 @@
 # ------------
 if [[ $- == *i* ]]; then
 
+local default_ignore="${FZF_DEFAULT_IGNORE:-\
```
mexus: This one should be probably documented :)

junegunn:
Thanks, but I don't think we need another variable that the users have to be aware of.

Because `find` is not good enough for most users these days as it does not respect `.gitignore`, and it's trivial to install a third-party tool that can completely replace `find` in this use case. I recommend https://github.com/sharkdp/fd.

```sh
export FZF_DEFAULT_COMMAND='fd --type f --hidden --follow --exclude .git'
export FZF_ALT_C_COMMAND='fd --type d --hidden --follow --exclude .git'
export FZF_CTRL_T_COMMAND='fd --type f --type d --hidden --follow --exclude .git'
```

mexus:
@junegunn okay, got it :) thanks for the tip about `fd`! It seems like it totally suits my needs.

PR #1218, opened 2018-02-17
Switch to dep vendoring tool

y0ssar1an:
- Replace `glide` vendoring tool with `dep`.
  - `dep` is the "official Go vendoring experiment"
  - state-of-the-art dependency manager
  - guarantees a flat dependency tree
  - makes it easy update dependencies
  - all Go projects are encouraged to switch to `dep`
  - [`glide` is in maintenance mode](https://github.com/masterminds/glide#golang-dep)
- Check in all dependencies to `vendor/`.
  - [traditional arguments against vendoring](https://peter.bourgon.org/go-best-practices-2016/#dependency-management) no longer apply in the age of `dep`
    - `fzf` is a program, not a library (It has `func main()`)
    - fetching deps at build-time can fail. prevent build failures by checking in everything.
  - now you can just clone and build
  - adds minimal bloat
    - `dep` guarantees a flat dependency tree. this means there's only one copy of every package in `vendor/` and there are no nested `vendor/` directories.
    - configuration in `Gopkg.toml` tells `dep` to aggressively prune away unnecessary files in `vendor/`
    - remaining files are just text, which `git` compresses very well
- Remove glide references in the `Makefile`.
- Remove `vendor` target from `Makefile`. We don't need to fetch dependencies anymore, since they're all present in `vendor/`.

### dep crash course
`Gopkg.lock` contains the commit hashes of every vendored dependency. It is the _output_ of the dependency solver. Do not edit this file.
`Gopkg.toml` contains the version and branch constraints for each package. It is the _input_ to the dependency solver. This is the file that you edit.

```sh
# start a new project. dep will read all the import paths in your code and create a Gopkg.toml manifest, containing all the branch/version constraints.
dep init

# fetch all dependencies. "ensure" that the vendor/ directory matches the constraints specified in Gopkg.toml
dep ensure

# update dependencies
# -v (verbose) recommended so you can see what dep is doing
dep ensure -v -update

# add a new dependency. just start using the package in your code, then run...
dep ensure

# print a nice summary of all the dependency versions
dep status
```

That's it. `init`, `ensure`, `ensure -update`, and `status` are the only commands to learn.


junegunn:
Thanks, I can consider switching to dep (though I don't see a good reason to do so now, nothing's broken) but I don't want to include vendor directory. Many users including myself install fzf by [cloning this repository](https://github.com/junegunn/fzf#using-git) and they don't need to download extra dependencies.

y0ssar1an:
The goal is for builds to be reproducible and highly reliable. You've achieved reproducible builds by including the `glide.lock` file (👍). With vendoring, you can make your builds a little more reliable by eliminating these failure cases:
1. failing to fetch `glide`. with vendoring, you don't need to fetch a vendoring tool because the deps are already present
2. failing to build `glide` because of problem upstream. unlikely given that `glide` is in maintenance mode, but it could happen. `make` is always `go get`-ting the latest `master`.
3. failing to fetch dependencies with `glide install`
4. failing to build `fzf` because of a mistake in `glide.lock`. with vendoring, it's still bad to have an incorrect `Gopkg.lock`, but it won't cause builds to fail

These are low risk failures, but they can be eliminated entirely. With vendoring, you can guarantee the build will succeed so long as the clone succeeds. The only downside is that you add a bit of bloat to the repo. That "bloat" is just text files, which `git` compresses very well. Still, it will take a few extra seconds to clone the repo. This is less time than it takes to clone an unvendored repo and fetch deps with `glide`.

Now here are all the benefits you get from vendoring:
1. `fzf` can be fetched and built with `go get`
2. simplified `Makefile`
3. contributors can run tests immediately after cloning
4. contributors and users don't have to fetch deps at build time

And here is the benefit you get from switching to `dep`:

5. much easier to update dependencies
6. you can vendor static analysis tools and linters, not just dependencies

### tl;dr
It's better to fetch deps at clone time instead of build time. `fzf` will be easier to build and maintain by vendoring with `dep`.

junegunn:
Yeah, I can see the benefits of the approach, but like I said above, most users of fzf simply download the official binaries using the install script rather than to build it from the source. For them, adding `vendor` directory will only make cloning slower. fzf has been pretty stable for quite some time now. It is not expecting much development effort in the near future and is not actively seeking more contributions, so I can't say that this is my priority.

Having said that, I do agree that it's nice to modernize the tool chain, so switching to `go dep` (without vendoring) is something I can consider.

y0ssar1an:
Ok. No worries. I'll close this. Thanks for considering.

PR #1219, opened 2018-02-17
go vet fixes, make tests pass on Go 1.10

y0ssar1an:
In Go 1.10, [`go test` runs `go vet`](https://golang.org/doc/go1.10#test). This is a problem because `fzf` currently doesn't pass `go vet`. So `fzf` doesn't pass tests with Go 1.10.

I based this off #1218, which vendors all the dependencies. I couldn't run `go vet ./...` or `go test ./...` on Go 1.10 because of missing packages in `vendor/` (for some reason `make` wasn't fetching them all). I recommend reviewing #1218 first. That will remove a lot of noise from this PR.

### go vet fixes
- add `String()` methods to `Token`, `term`, `Delimeter`, and `*util.Chars` so they can be printed with `%s`
- replace `%s` with `%v` where the default string representation is good enough


src/ansi_test.go
```diff
@@ -26,7 +26,7 @@ func TestExtractColor(t *testing.T) {
 		output, ansiOffsets, newState := extractColor(src, state, nil)
 		state = newState
 		if output != "hello world" {
-			t.Errorf("Invalid output: %s %s", output, []rune(output))
+			t.Errorf("Invalid output: %s %v", output, []rune(output))
```
y0ssar1an: `%s` doesn't work with `[]rune`.
`%v` will print `[]rune("hello")` like this: `[]rune{'h', 'e', 'l', 'l', 'o'}`.

y0ssar1an: Apparently, you can't use `%s` with a rune slice.
`%v` will print `[]rune("hello")` like `[]rune{'h', 'e', 'l', 'l', 'o'}`.

src/options_test.go
```diff
@@ -50,7 +50,7 @@ func TestDelimiterRegexString(t *testing.T) {
 		tokens[2].text.ToString() != "---*" ||
 		tokens[3].text.ToString() != "*" ||
 		tokens[4].text.ToString() != "---" {
-		t.Errorf("%s %s %d", delim, tokens, len(tokens))
+		t.Errorf("%s %v %d", delim, tokens, len(tokens))
```
y0ssar1an:
```go
tokens := []Token{
	Token{text: &Chars{}},
}
fmt.Printf("%v", tokens) // prints [Token{text: Chars{slice: []byte(""), inBytes: false, trimLengthKnown: false, trimLength: 0, Index: 0}, prefixLength: 0}]
```

src/options_test.go
```diff
@@ -71,7 +71,7 @@ func TestSplitNth(t *testing.T) {
 		if len(ranges) != 1 ||
 			ranges[0].begin != rangeEllipsis ||
 			ranges[0].end != rangeEllipsis {
-			t.Errorf("%s", ranges)
+			t.Errorf("%v", ranges)
```
y0ssar1an:
```go
r := []Range{
	{1, 2},
	{3, 4},
}
fmt.Printf("%v", r) // prints [{1 2} {3 4}]
```

src/pattern.go
```diff
@@ -34,6 +35,11 @@ type term struct {
 	caseSensitive bool
 }
 
+// String returns the string representation of a term.
+func (t term) String() string {
+	return fmt.Sprintf("term{typ: %d, inv: %v, text: []rune(%q), caseSensitive: %v}", t.typ, t.inv, string(t.text), t.caseSensitive)
+}
```
y0ssar1an: String representations are kind of an art. I tried to follow the Go syntax representation here. Please let me know if you prefer another format. I'm happy to change it.

junegunn:
Cherry-picked the commit that fixes go vet issue. Thanks.

PR #1224, opened 2018-02-23
Document inverse prefix exact match search syntax

JesseLeite:
Just documenting inverse prefix exact match search syntax.  Maybe there was a reason for not including this?  Or maybe it was just forgotten?  😊


junegunn:
Thanks! Sorry for the delay, by the way, I've been busy lately.

JesseLeite:
No worries!

PR #1253, opened 2018-04-04
remove final trailing delimiters under `--with-nth`

rolandwalker:
fzf retains the trailing delimiter on each field as divided for `--nth` or `--with-nth`.  This is useful for later joining fields, as the delimiters may be defined by regular expressions and thus may vary arbitrarily.

However, it is confusing/distracting for the final trailing delimiter to appear, particularly when control-character delimiters are displayed as `?` (which is true for me after e3e7b3360).

This PR continues to retain trailing delimiters internally, but right-trims the final delimiter just prior to display.  This also affects the `replace-query` action in a consistent way.  When `--nth` is used, it remains possible to query against the delimiter text, which is arguably a bug, but a separate one.

Example of bug, using visible delimiters:
```bash
$ printf '%s' $'one_1_I\ntwo_2_II\nthree_3_III\nfour_4_IV\n' | FZF_DEFAULT_OPTS='' fzf --height=6 --with-nth=1,2 --delimiter='_'
  four_4_
  three_3_
  two_2_
> one_1_    <--- final trailing delimiter
  4/4
> _
```

With the patch:

```bash
$ printf '%s' $'one_1_I\ntwo_2_II\nthree_3_III\nfour_4_IV\n' | FZF_DEFAULT_OPTS='' fzf --height=6 --with-nth=1,2 --delimiter='_'
  four_4
  three_3
  two_2
> one_1
  4/4
> _
```

This first draft of the patch likely has some problems.  I am unfamiliar with Go, and did not look into what `prefixLength` does.

### DIFF
--- a/src/core.go
+++ b/src/core.go
@@ -100,7 +100,7 @@ func Run(opts *Options, revision string) {
 	} else {
 		chunkList = NewChunkList(func(item *Item, data []byte) bool {
 			tokens := Tokenize(string(data), opts.Delimiter)
-			trans := Transform(tokens, opts.WithNth)
+			trans := Transform(tokens, opts.WithNth, opts.Delimiter)
 			transformed := joinTokens(trans)
 			if len(header) < opts.HeaderLines {
 				header = append(header, transformed)
--- a/src/pattern.go
+++ b/src/pattern.go
@@ -386,7 +386,7 @@ func (p *Pattern) transformInput(item *Item) []Token {
 	}
 
 	tokens := Tokenize(item.text.ToString(), p.delimiter)
-	ret := Transform(tokens, p.nth)
+	ret := Transform(tokens, p.nth, p.delimiter)
 	item.transformed = &ret
 	return ret
 }
--- a/src/pattern_test.go
+++ b/src/pattern_test.go
@@ -128,7 +128,7 @@ func TestCaseSensitivity(t *testing.T) {
 func TestOrigTextAndTransformed(t *testing.T) {
 	pattern := BuildPattern(true, algo.FuzzyMatchV2, true, CaseSmart, false, true, true, []Range{}, Delimiter{}, []rune("jg"))
 	tokens := Tokenize("junegunn", Delimiter{})
-	trans := Transform(tokens, []Range{Range{1, 1}})
+	trans := Transform(tokens, []Range{Range{1, 1}}, Delimiter{})
 
 	origBytes := []byte("junegunn.choi")
 	for _, extended := range []bool{false, true} {
--- a/src/terminal.go
+++ b/src/terminal.go
@@ -1220,7 +1220,7 @@ func replacePlaceholder(template string, stripAnsi bool, delimiter Delimiter, fo
 
 		for idx, item := range items {
 			tokens := Tokenize(item.AsString(stripAnsi), delimiter)
-			trans := Transform(tokens, ranges)
+			trans := Transform(tokens, ranges, delimiter)
 			str := string(joinTokens(trans))
 			if delimiter.str != nil {
 				str = strings.TrimSuffix(str, *delimiter.str)
--- a/src/tokenizer.go
+++ b/src/tokenizer.go
@@ -178,7 +178,7 @@ func joinTokens(tokens []Token) string {
 }
 
 // Transform is used to transform the input when --with-nth option is given
-func Transform(tokens []Token, withNth []Range) []Token {
+func Transform(tokens []Token, withNth []Range, delimiter Delimiter) []Token {
 	transTokens := make([]Token, len(withNth))
 	numTokens := len(tokens)
 	for idx, r := range withNth {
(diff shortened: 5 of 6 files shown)


junegunn:
Thanks for the suggestion, but trimming the rightmost delimiter is not necessarily desirable. For example,

```sh
fzf --reverse --delimiter '\|' --with-nth 1,4,2 << "EOF"
| Token     | Match type                 | Description                          |
| --------- | -------------------------- | ------------------------------------ |
| `sbtrkt`  | fuzzy-match                | Items that match `sbtrkt`            |
| `'wild`   | exact-match (quoted)       | Items that include `wild`            |
| `^music`  | prefix-exact-match         | Items that start with `music`        |
| `.mp3$`   | suffix-exact-match         | Items that end with `.mp3`           |
| `!fire`   | inverse-exact-match        | Items that do not include `fire`     |
| `!^music` | inverse-prefix-exact-match | Items that do not start with `music` |
| `!.mp3$`  | inverse-suffix-exact-match | Items that do not end with `.mp3`    |
EOF
```

I'm not sure what your exact use case is, but regarding control character delimiters, I use non-breaking spaces as delimiters when I want to hide them but ordinary spaces aren't suitable.

```sh
printf '%s' $'one_1_I\ntwo_2_II\nthree_3_III\nfour_4_IV\n' |
    sed $'s/_/\xc2\xa0/g' |
    FZF_DEFAULT_OPTS='' fzf --height=6 --with-nth=1,2 --delimiter=$'\xc2\xa0'
```

rolandwalker:
Excellent point; your example is very pretty.

However, I do think that precedents such as TSV and CSV are strong in favor of a different default.  Perhaps both forms could be accommodated within the `--with-nth` value by some syntax such as trailing-comma-requests-trailing-delimiter.  (`--with-nth 1,4,2,`)

My real usecase is ASCII NUL, of course!

```bash
$ printf '%s' $'one\0001\000I\ntwo\0002\000II\nthree\0003\000III\nfour\0004\000IV\n' | FZF_DEFAULT_OPTS='' fzf --height=6 --with-nth=1 --delimiter='\000'
  four?
  three?
  two?
> one?    <---- question marks start after commit e3e7b33
  4/4
>
```

Or when NUL is not available, `\034` ASCII FIELD SEPARATOR, because I am actually that old.

junegunn:
Would it help if fzf does not print those characters? (instead of printing `?`)

```sh
$ ruby -e 'puts (0..31).map { |n| "#{n}=(#{n.chr})" }.join'

0=()1=()2=()3=()4=()5=()6=()7=()8=)9=(  )10=(
)11=(
     )12=(
)14=()15=()16=()17=()18=()19=()20=()21=()22=()23=()24=()25=()26=()27=(8=()29=()30=()31=()
```

We can surely do better with these control characters.

https://github.com/junegunn/fzf/blob/851fa382519edc10a887d893abf54af3588c8de2/src/tui/light.go#L45-L60

junegunn:
```diff
diff --git a/src/tui/light.go b/src/tui/light.go
index 9b1489f..0d0d113 100644
--- a/src/tui/light.go
+++ b/src/tui/light.go
@@ -48,11 +48,13 @@ func (r *LightRenderer) stderrInternal(str string, allowNLCR bool) {
 	runes := []rune{}
 	for len(bytes) > 0 {
 		r, sz := utf8.DecodeRune(bytes)
-		if r == utf8.RuneError || r < 32 &&
-			r != '\x1b' && (!allowNLCR || r != '\n' && r != '\r') {
-			runes = append(runes, '?')
-		} else {
-			runes = append(runes, r)
+		nlcr := r == '\n' || r == '\r'
+		if r >= 32 || r == '\x1b' || nlcr {
+			if r == utf8.RuneError || nlcr && !allowNLCR {
+				runes = append(runes, '?')
+			} else {
+				runes = append(runes, r)
+			}
 		}
 		bytes = bytes[sz:]
 	}
@@ -789,7 +791,7 @@ func (w *LightWindow) Print(text string) {
 }
 
 func cleanse(str string) string {
-	return strings.Replace(str, "\x1b", "?", -1)
+	return strings.Replace(str, "\x1b", "", -1)
 }
 
 func (w *LightWindow) CPrint(pair ColorPair, attr Attr, text string) {
diff --git a/src/util/util.go b/src/util/util.go
index 867935a..95c4e1b 100644
--- a/src/util/util.go
+++ b/src/util/util.go
@@ -17,11 +17,12 @@ func RuneWidth(r rune, prefixWidth int, tabstop int) int {
 		return tabstop - prefixWidth%tabstop
 	} else if w, found := _runeWidths[r]; found {
 		return w
-	} else {
-		w := Max(runewidth.RuneWidth(r), 1)
-		_runeWidths[r] = w
-		return w
+	} else if r == '\n' || r == '\r' {
+		return 1
 	}
+	w := runewidth.RuneWidth(r)
+	_runeWidths[r] = w
+	return w
 }
 
 // Max returns the largest integer
```

rolandwalker:
Yes, thanks!  The above patch solves my usecase.

Issue #490, opened 2016-02-12
Bash: Multiple line PS1 prompt doubles up when redrawing


hughdavenport:
This one may be a bug with readline/bash itself, but I noticed after using fzf keybindings, which make use of redraw-current-line, that my prompt doubled up a lot.

You can reproduce this without fzf, with just bash by doing this:

```
bind '"\er": redraw-current-line'
PS1="first\nsecond"
```

Your prompt should now show

```
first
second
```

After pushing ESC-r a few times (or meta-r), then it shows

```
first
first
first
second
```

This is showing that the first line is overlapping with the last line. I say last, as you could set it to PS1="first\nsecond\nthird" and you get

```
first
second
first
second
first
second
third
```

Let me know if I should just file this with bash/readline directly, or whether you want it here to find a way around it?

Cheers,

Hugh

junegunn:
Interesting. Thanks for the heads up. Do you think we can find a workaround?

hughdavenport:
Haha, I'm googling this as we speak. I don't really want to go back to a single line prompt after getting used to my multi line one for so long... Though the extra new lines (my first line of my PS1 is just a blank) do take up a lot of vertical space.

Will let you know if I find a workaround, otherwise probably need to send this to bash/readline.

Not sure if similar behaviour happens on the other shells.

Cheers,

Hugh

hughdavenport:
Sent this to bash:

```
From: Hugh Davenport <scrubbed>
To: bug-bash AT gnu.org,bash AT packages.debian.org
Subject: redraw-current-line fails with multiline prompts

Configuration Information [Automatically generated, do not change]:
Machine: x86_64
OS: linux-gnu
Compiler: gcc
Compilation CFLAGS:  -DPROGRAM='bash' -DCONF_HOSTTYPE='x86_64' -DCONF_OSTYPE='linux-gnu' -DCONF_MACHTYPE='x86_64-pc-linux-gnu' -DCONF_VENDOR='pc' -DLOCALEDIR='/usr/share/locale' -DPACKAGE='bash' -DSHELL -DHAVE_CONFIG_H   -I.  -I../. -I.././include -I.././lib  -D_FORTIFY_SOURCE=2 -g -O2 -fstack-protector-strong -Wformat -Werror=format-security -Wall
uname output: Linux laptop 3.16.0-4-amd64 #1 SMP Debian 3.16.7-ckt20-1+deb8u2 (2016-01-02) x86_64 GNU/Linux
[...]

PR #1256, opened 2018-04-05
Fix multiline ps1 on zsh

ivanalejandro0:
These changes make fzf work perfect on a multiline ps1 prompt.
Without these changes you got lines "eaten" after canceling (and on some cases accepting) the fzf action triggered by one of these helper actions.

This fixes the behavior observed on #490, only for zsh.

### DIFF
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -29,7 +29,7 @@ __fzfcmd() {
 fzf-file-widget() {
   LBUFFER="${LBUFFER}$(__fsel)"
   local ret=$?
-  zle redisplay
+  zle -R
   typeset -f zle-line-init >/dev/null && zle zle-line-init
   return $ret
 }
@@ -43,6 +43,7 @@ fzf-redraw-prompt() {
     $precmd
   done
   zle reset-prompt
+  zle -R
 }
 zle -N fzf-redraw-prompt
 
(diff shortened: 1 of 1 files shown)


ivanalejandro0:
I see that the checks are failing but looking at the results I can't quite tell why. Any advice is welcome.
I'll get back to this to make the tests pass when I have some spare time again, I wanted to share the fix that worked for me in case this is being a pain for someone else.

junegunn:
Here's an easy way to reproduce the case:

```sh
mkdir -p /tmp/fzf-zsh
cp ~/.fzf.zsh /tmp/fzf-zsh/.zshrc
PS1= PROMPT_COMMAND= HISTSIZE=100 ZDOTDIR=/tmp/fzf-zsh zsh

# Now try CTRL-T
echo <CTRL-T>
```

When you make a selection, "echo" becomes invisible. This doesn't happen with `zle redisplay`.

Actually, I'm not a zsh user and do not know the exact difference between `zle redisplay` and `zle -R`. Can you explain?

Also, this seems related: https://github.com/sorin-ionescu/prezto/issues/1245

ivanalejandro0:
Oh, thanks for the code, it was useful. I'm able to reproduce the problem and see the `echo` disappear.

I'm not really familiar with zsh scripting. I gave a quick read to `zsh` and `zle` docs to get the basics. It seems that `zle redisplay` is a "widget" and `-R` is a switch of the `zle` command, they look very alike and I couldn't figure out yet why they behave differently and when you should use either of them.
I'll keep digging into this to see if I can figure out a fix that works on a multi-line prompt and doesn't break single line prompts (or the tests).

The ticket you linked looks like the same kind of problem `zle redisplay` vs `zle -R`, maybe we can fix them both. (I'm also a user of sorin-ionescu/prezto).

PR #1259, opened 2018-04-11
Add a new special event `init`

knu:
This new special event that is triggered right after initialization.  It comes in handy, for example, when you want to start fzf with everything selected by default: `fzf --multi --bind init:select-all`


knu:
Thanks for the detailed explanation!

Issue #447, opened 2015-12-11
Fatal Error: Failed to open /dev/tty


sbrl:
I just installed this on my server via ssh and got this error when I pressed `CTRL+R`:

```
$(__fzf_history__)panic: Failed to open /dev/tty

goroutine 8 [running]:
github.com/junegunn/fzf/src/curses.Init(0xc82000e180, 0x470100)
    /go/src/github.com/junegunn/fzf/src/curses/curses.go:261 +0xc9
github.com/junegunn/fzf/src.NewTerminal.func1()
    /go/src/github.com/junegunn/fzf/src/terminal.go:233 +0x36
github.com/junegunn/fzf/src.(*Terminal).Loop(0xc820058b60)
    /go/src/github.com/junegunn/fzf/src/terminal.go:744 +0x368
created by github.com/junegunn/fzf/src.Run
    /go/src/github.com/junegunn/fzf/src/core.go:202 +0xe84

goroutine 1 [semacquire]:
sync.runtime_Syncsemacquire(0xc820018350)
    /go1.5/src/runtime/sema.go:237 +0x201
sync.(*Cond).Wait(0xc820018340)
    /go1.5/src/sync/cond.go:62 +0x9b
github.com/junegunn/fzf/src/util.(*EventBox).Wait(0xc82000e520, 0xc82004bef0)
    /go/src/github.com/junegunn/fzf/src/util/eventbox.go:32 +0xb5
github.com/junegunn/fzf/src.Run(0xc82000c2c0)
    /go/src/github.com/junegunn/fzf/src/core.go:275 +0xfb2
main.main()
    /go/src/github.com/junegunn/fzf/src/fzf/main.go:6 +0x25

goroutine 17 [syscall, locked to thread]:
runtime.goexit()
    /go1.5/src/runtime/asm_amd64.s:1696 +0x1

goroutine 5 [syscall]:
os/signal.loop()
    /go1.5/src/os/signal/signal_unix.go:22 +0x18
created by os/signal.init.1
    /go1.5/src/os/signal/signal_unix.go:28 +0x37

goroutine 11 [runnable]:
github.com/junegunn/fzf/src.(*Terminal).Loop.func2(0xc82001c540, 0xc820058b60)
    /go/src/github.com/junegunn/fzf/src/terminal.go:736
created by github.com/junegunn/fzf/src.(*Terminal).Loop
    /go/src/github.com/junegunn/fzf/src/terminal.go:741 +0x330

[...]

PR #1266, opened 2018-04-18
Add an option: --tty=DEVICE

knu:
I have a similar problem to #447, where tty is somehow inaccessible via /dev/tty but $TTY still points to the working tty.  In this case, you could say `fzf --tty=$TTY` to force attaching to the tty.

In my case, I tried to invoke fzf from within rb-readline on Rails console run under spring, but I got the `Failed to open /dev/tty` error.  I guess there could be something wrong with the handling of tty in any of pry/spring/rb-readline, but I wasn't able to identify which and how.

### DIFF
--- a/man/man1/fzf.1
+++ b/man/man1/fzf.1
@@ -365,6 +365,9 @@ ncurses finder only after the input stream is complete.
 e.g. \fBfzf --multi | fzf --sync\fR
 .RE
 .TP
+.BI "--tty=" "DEVICE"
+Device name of tty (default: \fB/dev/tty\fR)
+.TP
 .B "--version"
 Display version information and exit
 
--- a/src/options.go
+++ b/src/options.go
@@ -86,6 +86,7 @@ const usage = `usage: fzf [options]
     --read0               Read input delimited by ASCII NUL characters
     --print0              Print output delimited by ASCII NUL characters
     --sync                Synchronous search for multi-staged filtering
+    --tty=DEVICE          Use DEVICE as tty (default: /dev/tty)
     --version             Display version information and exit
 
   Environment variables
@@ -188,6 +189,7 @@ type Options struct {
 	Bordered    bool
 	Tabstop     int
 	ClearOnExit bool
+	TtyDevice   string
 	Version     bool
 }
 
@@ -237,6 +239,7 @@ func defaultOptions() *Options {
 		Margin:      defaultMargin(),
 		Tabstop:     8,
 		ClearOnExit: true,
+		TtyDevice:   "/dev/tty",
 		Version:     false}
 }
 
@@ -1133,6 +1136,8 @@ func parseOptions(opts *Options, allArgs []string) {
 			opts.ClearOnExit = true
 		case "--no-clear":
 			opts.ClearOnExit = false
+		case "--tty":
+			opts.TtyDevice = nextString(allArgs, &i, "device name required")
 		case "--version":
 			opts.Version = true
 		default:
@@ -1188,6 +1193,8 @@ func parseOptions(opts *Options, allArgs []string) {
 				opts.HscrollOff = atoi(value)
 			} else if match, value := optString(arg, "--jump-labels="); match {
 				opts.JumpLabels = value
+			} else if match, value := optString(arg, "--tty="); match {
+				opts.TtyDevice = value
 			} else {
 				errorExit("unknown option: " + arg)
 			}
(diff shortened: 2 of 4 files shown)


junegunn:
Can you provide the repro steps? I'd like to make sure that the problem is reliably reproducible and there's absolutely no workaround for the issue.

knu:
Sure, here it is.  In the following environment I have:

- macOS 10.13.4 (High Sierra; the version would not matter)
- Ruby 2.4.4 installed via rbenv (Any version >= 2.2 will do)

Put the following in `~/.pryrc`:

```ruby
begin
  require 'rb-readline'
  require 'readline'
rescue LoadError
else
  def RbReadline.rl_reverse_search_history(sign, key)
    rl_insert_text `fzf --no-sort --tac < ~/.pry_history`.chomp
  end
end
```

Then create a Rails project and add Pry and Spring to it:

```
% gem install rails -v 5.2.0

% rails new foo

% cd foo
% echo "gem 'pry-rails', group: :development" >> Gemfile
% echo "gem 'rb-readline', group: :development" >> Gemfile
% bundle
```

Finally, run rails condole under Spring and invoke fzf via <key>^R</key> in the Pry prompt:

```
% spring rails condole
Running via Spring preloader in process 52196
Loading development environment (Rails 5.2.0)
[1] pry(main)> <^R>Failed to open /dev/tty
```

Passing `--tty=$TTY` to the patched fzf, to get it to open `$TTY` instead of `/dev/tty`, fixes the issue.

junegunn:
Thanks for the repro, I was able to reproduce the problem on `rails console`. By the way, `$TTY` is not set on my shell, do you set it manually?

Since fzf requires that STDERR is a tty device, we can find the name of the actual device via `ttyname(2)`. But fzf no longer uses `cgo` and we have to implement our own version of the function. I can confirm that the following patch fixes the issue on macOS and Linux, though I haven't tested it on other platforms such as FreeBSD or OpenBSD yet.

```diff
diff --git a/src/tui/light.go b/src/tui/light.go
index a96a844..e223f4a 100644
--- a/src/tui/light.go
+++ b/src/tui/light.go
@@ -32,6 +32,12 @@ var offsetRegexp *regexp.Regexp = regexp.MustCompile("\x1b\\[([0-9]+);([0-9]+)R"
 func openTtyIn() *os.File {
 	in, err := os.OpenFile(consoleDevice, syscall.O_RDONLY, 0)
 	if err != nil {
+		tty := ttyname()
+		if len(tty) > 0 {
+			if in, err := os.OpenFile(tty, syscall.O_RDONLY, 0); err == nil {
+				return in
+			}
+		}
 		fmt.Fprintln(os.Stderr, "Failed to open "+consoleDevice)
 		os.Exit(2)
 	}
diff --git a/src/tui/ttyname_unix.go b/src/tui/ttyname_unix.go
new file mode 100644
index 0000000..69bdfa5
--- /dev/null
+++ b/src/tui/ttyname_unix.go
@@ -0,0 +1,31 @@
+// +build !windows
+
+package tui
+
+import (
+	"io/ioutil"
+	"syscall"
+)
+
+var devPrefixes = [...]string{"/dev/pts/", "/dev/"}
+
+func ttyname() string {
+	var stderr syscall.Stat_t
+	if syscall.Fstat(2, &stderr) != nil {
+		return ""
+	}
+
+	for _, prefix := range devPrefixes {
+		files, err := ioutil.ReadDir(prefix)
+		if err != nil {
+			continue
+		}
+
+		for _, file := range files {
+			if stat, ok := file.Sys().(*syscall.Stat_t); ok && stat.Rdev == stderr.Rdev {
+				return prefix + file.Name()
+			}
+		}
+	}
+	return ""
+}
diff --git a/src/tui/ttyname_windows.go b/src/tui/ttyname_windows.go
new file mode 100644
index 0000000..5354b74
--- /dev/null
+++ b/src/tui/ttyname_windows.go
@@ -0,0 +1,7 @@
+// +build windows
+
+package tui
+
+func ttyname() string {
+	return ""
+}
```

junegunn:
I was a bit concerned about the performance of the`ttyname` implementation, but I can't notice any delay from the code, so I'm going to apply the patch instead of adding `--tty`. Let me know if you see any issues. Thanks.

knu:
Ah, it might look a little bit tricky but would work for most well-known unix-like systems including FreeBSD and OpenBSD.

junegunn:
Yeah, and the code is only for exceptional cases, so I think it's enough.

Issue #692, opened 2016-10-04
Custom selected character


wayneashleyberry:
- Category
    - [x] fzf binary
    - [ ] fzf-tmux script
    - [ ] Key bindings
    - [ ] Completion
    - [ ] Vim
    - [ ] Neovim
    - [ ] Etc.
- OS
    - [x] Linux
    - [x] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [ ] bash
    - [x] zsh
    - [ ] fish

There's an option to customise the prompt character, but the same character is used to highlight the selected result. It would be great to have an option to customise that character as well, to use something like `→` or `⇢` etc.

Here's a little example:

```sh
  github.com/foo/bar
⇢ github.com/baz
  github.com
  3/3
$ github
```

I think this is where it's hardcoded:

https://github.com/junegunn/fzf/blob/master/src/terminal.go#L629
https://github.com/junegunn/fzf/blob/master/src/terminal.go#L636

If this would be a valid option to add then I could try and send a pull request?

junegunn:
The idea was suggested in the past, but I declined. And I'm still not sure if I want this.

The problem is not that it's hard to implement – it's quite the opposite as you might have guessed – but that I do not want fzf to end up with a bunch of options for tweaking minor UI elements. Imagine, `--inline-info-separator`, `--info-format`, `--gutter-width`, `--border-style`, `--preview-padding`, the list goes on, and there's no reason that we should offer one but not the other. I don't think I want to go that direction.

One can argue that we already have `--prompt` in that family, but the way I see it is that the main purpose of `--prompt` is to indicate the context of the search (e.g. `fzf --prompt $(pwd)/`). So it's not just for changing the style of an element.

I may revisit the idea later, but for now I'll close this as wontfix. Thanks for the suggestion.

wayneashleyberry:
Fair point, I completely understand your viewpoint @junegunn - thank you for explaining and thank you for an incredible tool!

junegunn:
No problem, thanks for understanding.

PR #1267, opened 2018-04-18
Provide an option to reverse items only

knu:
I'm quite used to have a prompt at the bottom of the screen (not to mention vim & emacs) while I prefer a list of items shown from top down.  I just don't want to be distracted by what I type but just like to concentrate on what to pick.

So, I've extended the `--reverse` option so it optionally takes `all`, `items` or `none` so you can choose which area to reverse.

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,8 +217,8 @@ cursor with `--height` option.
 vim $(fzf --height 40%)
 ```
 
-Also check out `--reverse` option if you prefer "top-down" layout instead of
-the default "bottom-up" layout.
+Also check out `--reverse` and `--layout` options if you prefer
+"top-down" layout instead of the default "bottom-up" layout.
 
 ```sh
 vim $(fzf --height 40% --reverse)
@@ -228,7 +228,7 @@ You can add these options to `$FZF_DEFAULT_OPTS` so that they're applied by
 default. For example,
 
 ```sh
-export FZF_DEFAULT_OPTS='--height 40% --reverse --border'
+export FZF_DEFAULT_OPTS='--height 40% --layout=reverse-list --border'
 ```
 
 #### Search syntax
@@ -266,7 +266,7 @@ or `py`.
     - e.g. `export FZF_DEFAULT_COMMAND='fd --type f'`
 - `FZF_DEFAULT_OPTS`
     - Default options
-    - e.g. `export FZF_DEFAULT_OPTS="--reverse --inline-info"`
+    - e.g. `export FZF_DEFAULT_OPTS="--layout=reverse-list --inline-info"`
 
 #### Options
 
--- a/README.md
+++ b/README.md
@@ -228,7 +228,7 @@ You can add these options to `$FZF_DEFAULT_OPTS` so that they're applied by
 default. For example,
 
 ```sh
-export FZF_DEFAULT_OPTS='--height 40% --layout=reverse-list --border'
+export FZF_DEFAULT_OPTS='--height 40% --layout=reverse --border'
 ```
 
 #### Search syntax
@@ -266,7 +266,7 @@ or `py`.
     - e.g. `export FZF_DEFAULT_COMMAND='fd --type f'`
 - `FZF_DEFAULT_OPTS`
     - Default options
-    - e.g. `export FZF_DEFAULT_OPTS="--layout=reverse-list --inline-info"`
+    - e.g. `export FZF_DEFAULT_OPTS="--layout=reverse --inline-info"`
 
 #### Options
 
--- a/man/man1/fzf.1
+++ b/man/man1/fzf.1
@@ -157,19 +157,19 @@ Minimum height when \fB--height\fR is given in percent (default: 10).
 Ignored when \fB--height\fR is not specified.
 .TP
 .BI "--layout=" "LAYOUT"
-Choose the layout (default: bottom-up)
+Choose the layout (default: default)
 
 .br
-.BR bottom-up "     Show everything from bottom up"
+.BR default "     Show everything from bottom up"
 .br
-.BR top-down "      Show everything from top down"
+.BR reverse "     Show everything from top down"
 .br
-.BR top-down-below "Show items from top down, prompt and header at the bottom"
+.BR reverse-list "Show items from top down, prompt and header at the bottom"
 .br
 
 .TP
 .B "--reverse"
-Deprecated synonym for \fB--layout=top-down\fB
+A synonym for \fB--layout=reverse\fB
 
 .TP
 .B "--border"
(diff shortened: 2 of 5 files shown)


junegunn:
I know people have different preferences and tastes, but I've tried to refrain from adding more options for tweaking UI elements for quite some time because as the maintainer of the project I have to draw the line somewhere to prevent fzf from becoming too complex. You can find my previous comment on that matter here: https://github.com/junegunn/fzf/issues/692#issuecomment-251436739

Having said that, I'm not particularly against the idea, but I must say that I'm not convinced either. But anyway, if I were to add the option, I would consider supporting all 4 possible layouts, with new `--layout=LAYOUT` option that supersedes `--reverse` like so:

- `--layout=bottom-up` (default)
- `--layout=top-down` (equivalent to `--reverse`)
- `--layout=bottom-up-below` (bottom-up layout, but prompt at the top. Not sure if this is a straightforward name for the option.)
- `--layout=top-down-above` (prompt at the bottom)

By the way, you might want to check if mouse click positions are properly calculated with your patch.

knu:
Thanks for the feedback, I've fixed the mouse support.

Actually, I also considered the possibility of a --layout option, but I didn't feel the need for the fourth option: prompt at the top + items bottom up.  I believe the bottom-up list style is pretty uncommon; it only makes sense when the prompt is down below.

junegunn:
Mouse pointer still isn't working on the prompt line.

Maybe you're right, `bottom-up-below` is probably not going to be the most popular option, then we can consider not implementing that particular layout. However, I still think it would be more straightforward to the users that we provide `--layout` option than to extend `--reverse` as suggested (we should keep `--reverse` for backward compatibility though). `--reverse` seemed like a good name 4 years ago when fzf was simple and small, but now fzf has so many options (including `--tac`) `--reverse` no longer feels like a straightforward name, and I'm concerned that making it do more things may bring extra confusion.

knu:
I've addressed the mouse click issue and dropped the extension for `--reverse` in favor of a new `--layout` option.  What do you think?

man/man1/fzf.1
```diff
@@ -155,9 +155,22 @@ the full screen.
...
+.TP
+.BI "--layout=" "LAYOUT"
+Choose the layout (default: bottom-up)
+
+.br
+.BR bottom-up "     Show everything from bottom up"
+.br
+.BR top-down "      Show items from top down, prompt and header at the bottom"
+.br
+.BR top-down-above "Show everything from top down"
```
junegunn: Well, I wanted `top-down` to be the new name for `--reverse`, and `top-down-above` to be your new layout. I mean "top-down-above (the prompt and header)". But I can definitely see that the term can be confusing, do you have a better idea?

src/options.go
```diff
@@ -53,7 +53,9 @@ const usage = `usage: fzf [options]
                           height instead of using fullscreen
     --min-height=HEIGHT   Minimum height when --height is given in percent
                           (default: 10)
-    --reverse             Reverse orientation
+    --layout=LAYOUT       Choose layout: [bottom-up|top-down|top-down-above]
+                          (default: bottom-up)
+    --reverse             Deprecated synonym for --layout=top-down-above
```
junegunn: We can remove `--reverse` here as we still mention it on the man page.

src/options.go
```diff
@@ -1037,10 +1062,13 @@ func parseOptions(opts *Options, allArgs []string) {
 			opts.Bold = true
 		case "--no-bold":
 			opts.Bold = false
+		case "--layout":
+			opts.Layout = parseLayout(
+				nextString(allArgs, &i, "layout required (bottom-up / top-down / top-down-above)"))
 		case "--reverse":
-			opts.Reverse = true
+			opts.Layout = layoutTopDownAbove
```
junegunn: So this becomes `layoutTopDown`.

man/man1/fzf.1
knu: Ah, that's fine, I'll do that.

junegunn:
Can you add a basic test case for your new layout? The testing code is a bit messy, but you'll get the hang of it.

https://github.com/junegunn/fzf/blob/master/test/test_go.rb

knu:
@junegunn I think I've added just enough test cases, and while working on it I found and fixed a couple of bugs. 😆

junegunn:
Cool, thanks! I'll test the feature and merge it in a few days.

junegunn:
Looks like we had a miscommunication here, I commented above that I wanted `top-down-above`, as in "items are displayed in **top-down** layout, **above** the prompt". But I guess it equally makes sense to think "items are displayed in **top-down** layout, the prompt is shown **below** the list", so no, it's not really a good name 😕 

I'm open to suggestions. Some ideas:

- `upward`, `downward`, `downward/bottom` or `downward2`
- `asc`, `desc`, `desc/bottom`
- `1`, `2`, `3` (seriously)

knu:
I see...  What about the following names, then?

- `upward` (default, aka `--no-reverse`)
- `downward-top` (new)
- `downward-bottom` (aka `--reverse`)

junegunn:
I've thought about the options, but none of them really clicked with me. So I'd like to go with:

- `--layout=default`
- `--layout=reverse`
- `--layout=reverse-list`

These are quite similar to your initial proposal, but I still think `--layout=reverse` is more clear than just `--reverse=something`. And can you remove "deprecation" message for `--reverse` from the man page? Let's just leave `--reverse` as a shortcut option for `--layout=reverse`. Thanks.

knu:
OK, I'll do that.

README.md
```diff
@@ -228,7 +228,7 @@ You can add these options to `$FZF_DEFAULT_OPTS` so that they're applied by
 default. For example,
 
 ```sh
-export FZF_DEFAULT_OPTS='--height 40% --reverse --border'
+export FZF_DEFAULT_OPTS='--height 40% --layout=reverse-list --border'
```
junegunn: I want to present `--layout=reverse` as an example instead of the new `reverse-list`.

README.md
```diff
@@ -266,7 +266,7 @@ or `py`.
     - e.g. `export FZF_DEFAULT_COMMAND='fd --type f'`
 - `FZF_DEFAULT_OPTS`
     - Default options
-    - e.g. `export FZF_DEFAULT_OPTS="--reverse --inline-info"`
+    - e.g. `export FZF_DEFAULT_OPTS="--layout=reverse-list --inline-info"`
```
junegunn: Same here

src/options.go
```diff
@@ -90,7 +91,8 @@ const usage = `usage: fzf [options]
 
   Environment variables
     FZF_DEFAULT_COMMAND   Default command to use when input is tty
-    FZF_DEFAULT_OPTS      Default options (e.g. '--reverse --inline-info')
+    FZF_DEFAULT_OPTS      Default options
+                          (e.g. '--layout=reverse-list --inline-info')
```
junegunn: Here too.

knu:
Done!

junegunn:
Thanks, everything looks good now. I'll merge this just before I release 0.17.4, so that there's no confusion from the documentation on master branch vs. 0.17.3 binary.

knu:
Great, thanks!

junegunn:
Merged, thanks!

PR #1272, opened 2018-04-28
Restore tmux window options

heindsight:
The `fzf-tmux` script disables the tmux `synchronize-panes` and `remain-on-exit` window options without restoring the original values. This PR fixes that by saving the original values of these options and restoring them on exit.

Note: This fix suffers from a race condition which can result in incorrect and unexpected behaviour if `fzf-tmux` is run concurrently in more than one pane of the same tmux window. I don't think it's possible to avoid that without maintaining some global state for each tmux window, which I think is undesirable and unnecessarily complicated given that this is quite unlikely to occur anyway.

### DIFF
--- a/bin/fzf-tmux
+++ b/bin/fzf-tmux
@@ -136,6 +136,11 @@ fifo3="${TMPDIR:-/tmp}/fzf-fifo3-$id"
 cleanup() {
   \rm -f $argsf $fifo1 $fifo2 $fifo3
 
+  # Reset remain-on-exit window option
+  if [[ -n "$tmux_remain_on_exit" ]]; then
+    tmux set-window-option $tmux_remain_on_exit
+  fi
+
   # Remove temp window if we were zoomed
   if [[ -n "$zoomed" ]]; then
     tmux display-message -p "#{window_id}" > /dev/null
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks for the patch. It's a little unfortunate though that the additional commands make fzf-tmux slightly slower.

screen shot 2018-05-03 at 12 41 43 pm

The difference is not significant, so I would just merge this if there isn't a simple way to reduce the overhead.

junegunn:
What do you think about this version?

```diff
diff --git a/bin/fzf-tmux b/bin/fzf-tmux
index 79407e6..f99e68a 100755
--- a/bin/fzf-tmux
+++ b/bin/fzf-tmux
@@ -136,6 +136,11 @@ fifo3="${TMPDIR:-/tmp}/fzf-fifo3-$id"
 cleanup() {
   \rm -f $argsf $fifo1 $fifo2 $fifo3
 
+  # Reset remain-on-exit and synchronize-panes window option
+  if [[ "${#tmux_opts[@]}" -gt 0 ]]; then
+    eval "tmux ${tmux_opts[@]}"
+  fi
+
   # Remove temp window if we were zoomed
   if [[ -n "$zoomed" ]]; then
     tmux display-message -p "#{window_id}" > /dev/null
@@ -174,6 +179,8 @@ pppid=$$
 echo -n "trap 'kill -SIGUSR1 -$pppid' EXIT SIGINT SIGTERM;" > $argsf
 close="; trap - EXIT SIGINT SIGTERM $close"
 
+tmux_opts=( $(tmux show-window-options remain-on-exit \; show-window-options synchronize-panes | sed 's/^/set-window-option /; s/$/ \\;/') )
+
 if [[ -n "$term" ]] || [[ -t 0 ]]; then
   cat <<< "\"$fzf\" $opts > $fifo2; echo \$? > $fifo3 $close" >> $argsf
   TMUX=$(echo $TMUX | cut -d , -f 1,2) tmux set-window-option synchronize-panes off \;\
```

The same command (`echo 0 | time fzf-tmux -0 -q 1`) takes 0.10 ~ 0.12 seconds.

junegunn:
Merged, thanks!

heindsight:
Thanks!

Issue #600, opened 2016-06-15
Avoid duplicate entry when fzf-ing shell history


benoittgt:
- Category
    - [X] fzf binary
    - [ ] fzf-tmux script
    - [ ] Key bindings
    - [ ] Completion
    - [ ] Vim
    - [ ] Neovim
    - [ ] Etc.
- Shell
    - [ ] bash
    - [X] zsh
    - [ ] fish

I would like to know if it's possible to limit the duplication in fzf specially in the shell history command? Or maybe I'm do it it wrong?

![capture d ecran 2016-06-15 a 09 40 39](https://cloud.githubusercontent.com/assets/8417720/16072095/58fe0392-32df-11e6-8f68-3cf9acb4f720.png)

junegunn:
Hi, this is one of the recurring questions. Please see #49, #88, #492. Thanks.

benoittgt:
Thanks @junegunn and sorry for the duplicate

edi9999:
For other people encoutering this issue, here is the changes I've made to remove duplicates : 

```
__fzf_history__() {
  local line
  countskip="$(history | tail -n 1 | grep -E '^[0-9]+' -o | wc -c)"
  countskip="$(( countskip + 1 ))"
  echo "$countskip"
  line=$(
    HISTTIMEFORMAT= history |
    grep '.\{1,79\}' |
    sed 's/ *$//g' |
    tac |
    nauniq --skip-chars="$countskip" |
    tac |
    $(__fzfcmd) +s --tac +m -n2..,.. --tiebreak=index --toggle-sort=ctrl-r |
    \grep '^ *[0-9]') && sed 's/ *\([0-9]*\)\** .*/!\1/' <<< "$line"
}```

It requires nanuniq https://metacpan.org/pod/distribution/App-nauniq/bin/nauniq

chew-z:
`perl -ne 'print if $seen{$_}++'` should do.

chew-z:
`perl -ne 'print if !$seen{$_}++'` should do.

GCorbel:
Using `setopt HIST_IGNORE_ALL_DUPS` on zsh did the trick in my case.

benjaminbauer:
There are rare, desperate, cases, where I appreciate to browse through the history, in order to reconstruct something. 

On ZSH one might use `HIST_FIND_NO_DUPS` to suppress duplicates in the search. Dups still get written to history, as opposed to `HIST_IGNORE_ALL_DUPS`.

[...]

Issue #492, opened 2016-02-14
Option to remove duplicates


blueyed:
From https://github.com/junegunn/fzf.vim/issues/33#issuecomment-167243756:

It might be good to have an option for `fzf` to ignore duplicates (without the list required to be sorted)?!

While the issue itself has been fixed (by not having dupes in the input anymore, which was caused by another plugin), I still think that such an option would be useful to have with `fzf` (e.g. for  https://github.com/junegunn/fzf/issues/49 and whenever you cannot influence the input really).

(There's no default tool for that (`sort -u` changes the order, `uniq` needs it to be sorted), but you could use a awk script like this: https://www.gnu.org/software/gawk/manual/html_node/History-Sorting.html.)

The option should be able to either only keep the first or last occurrence of a duplicated entry.

junegunn:
Thanks for the suggestion but I don't think we should as fzf tries to follow the unix philosophy. I know there's no default unix tool for that – which is unfortunate but quite understandable at the same time considering the memory requirement – but it's easy to preprocess the input. The awk example you mentioned should work and this one is in Ruby:

```ruby
require 'set'
set = Set.new
while line = $stdin.gets
  puts line if set.add?(line)
end
```

> whenever you cannot influence the input really

There really aren't many cases where you can't preprocess the input. The code for `CTRL-R` is just about ten lines or so, so it's feasible to have a modified version of it in one's config file.

> The option should be able to either only keep the first or last occurrence of a duplicated entry.

We already have two options to provide, keep first and keep last. And in case of `CTRL-R`, fzf should ignore the numeric index of each entry so we need yet another option (`--filter-dupe-with-nth`?). Builtin processor will never be as flexible as the external ones, and I'd like fzf to focus on something we can't do with pre- and post-processors.

netei:
> Builtin processor will never be as flexible as the external ones, and I'd like fzf to focus on something we can't do with pre- and post-processors.

+1 

I personally use (nauniq)[https://github.com/perlancar/perl-App-nauniq/blob/master/bin/nauniq] for removing duplicates

aaronjensen:
@netei @junegunn is there an example of how to remove duplicates for ctrl-r with fzf? I've seen a couple open issues about it and it'd be a nice thing to have.

Issue #88, opened 2014-08-10
ctrl+r history results should be unique by command


bnovc:
I normally have used ctrl+r and hit it a few times to get the right command pre-fzf, which worked fine, and now with fzf, I can use my arrow keys to scroll the results, but I have _many_ more to scroll than before. I don't see any value in presenting numerous duplicate results here, and zsh's default ctrl+r behavior doesn't do this.

cbrwizard:
@junegunn could you please provide an example on how to enable this option in `fish`? It doesn't have `setopt`

vpshastry:
I want my history file to contain the duplicates, because I want to see my history. But I don't want them showing up in multiple entries when I search them through fzf. Can't we do that?

junegunn:
It's not simple as it sounds so it's not supported. If you know some shell scripting, you should be able to implement your own version of the binding.

Issue #49, opened 2014-05-29
Remove duplicates?


elado:
Command history contains plenty of duplicates usually, and these are just noise. Would be nice to have them removed from the list.

PR #1287, opened 2018-05-10
Avoid duplicates when filter history(#49, #88, #492, #600)

wfxr:
When hit `Ctrl - R`, what we want is to find some commands we can quickly reuse. Duplicates is meaningless and sometimes annoying in this situation.

Config the shell to not record repeated commands is a solution. But history is not only used for `Ctrl - R`. We want to recall what happened by looking up the history when something goes wrong. Information will be lost if we discard the repeated commands. Moreover, this solution doesn't solve the duplicates problem on `Ctrl + R` perfectly. Eg, we enter `ls` many times per day. When we want to filter a history command contains 'ls', there still be many duplicates `ls` appeared. Shell config like `HIST_IGNORE_ALL_DUPS` should not work because they are separated by other commands.

I think this commit solved the problem. No outer dependencies and influences. We can keep recording commands in history, but have unique entries when call fzf by `Ctrl - R`.

#49 #88 #492 #600

### DIFF
--- a/shell/key-bindings.bash
+++ b/shell/key-bindings.bash
@@ -55,7 +55,7 @@ __fzf_history__() (
   local line
   shopt -u nocaseglob nocasematch
   line=$(
-    HISTTIMEFORMAT= history |
+    HISTTIMEFORMAT= history | sort -uk2,1000 | sort -nr |
     FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} $FZF_DEFAULT_OPTS --tac --sync -n2..,.. --tiebreak=index --bind=ctrl-r:toggle-sort $FZF_CTRL_R_OPTS +m" $(__fzfcmd) |
     command grep '^ *[0-9]') &&
     if [[ $- =~ H ]]; then
--- a/shell/key-bindings.fish
+++ b/shell/key-bindings.fish
@@ -50,7 +50,7 @@ function fzf_key_bindings
         history -z | eval (__fzfcmd) --read0 -q '(commandline)' | perl -pe 'chomp if eof' | read -lz result
         and commandline -- $result
       else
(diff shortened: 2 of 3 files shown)


SerialVelocity:
I found this was quite slow on my machine. Instead I use:
```sh
fc -rl 1 | awk '{ FIRST=$1;$1="";if (!x[$0]++) {$1=FIRST; print $0 }}'
```

(my awk is pretty non-existent though so I'm pretty sure there's a better way)

What might be good is adding something like `FZF_CTRL_R_COMMAND` so we can set this to whatever we want

junegunn:
The delay is quite noticeable to me as well (I currently have 60K+ entries in command history), so I won't merge this as it is. Also, it seems that you haven't tested the patch with bash and fish.

Adding `$FZF_CTRL_R_COMMAND` can be non-trivial as the output of the command should conform to a certain format, which should be explicitly documented (but unfortunately from my experience, most users will conveniently ignore the part anyway and spam my inbox), otherwise the subsequent processing will fail. And because there are inconsistencies among the three shells fzf supports. For example, we can't use the same `$FZF_CTRL_R_COMMAND` when switching between shells. `fc` in zsh, `history` in bash and fish. `history` of bash lists commands in chronological order, while `fc -l 1` of zsh and `history` of fish list in reverse chronological order. bash and zsh uses the numeric index from the list to restore the command, but we don't have the equivalent in fish. At this point, I think it's much easier to just fully override the implemention on shell configuration file. It's just a few lines of code.

fgervais:
I tried the commit in bash and I don't quite like it because it keeps the first/oldest entry of all the duplicates. Also it reverse sorts it for some reasons?

To me, this gives a better result:

`HISTTIMEFORMAT= history | sort -nr | sort -uk2 | sort -n |`

Although all this sorting looks pretty crazy.

PR #1290, opened 2018-05-14
remove -y flag from Arch Linux installation

DanielFGray:
per https://wiki.archlinux.org/index.php/Partial_upgrades#Partial_upgrades_are_unsupported you should never `pacman -Sy <pkg>`, Arch users are expected to keep their system already up-to-date before installing anything.


junegunn:
Thanks, according to the documentation:

> Always upgrade (with pacman -Syu) before installing a package.

should we prepend `pacman -Syu` to the instruction?

PR #1297, opened 2018-05-20
Add pledge(2) support (OpenBSD only) via a 'protector' package.

qbit:
pledge(2) is a way to restrict operations an application can make use of on
OpenBSD. By defining a set of "pledges" the kernel knows to kill a program
that makes system calls outside of those "pledges".

More info on pledge: http://man.openbsd.org/pledge

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,30 +1,22 @@
+<!-- ISSUES NOT FOLLOWING THIS TEMPLATE WILL BE CLOSED AND DELETED -->
+
 <!-- Check all that apply [x] -->
-- Category
-    - [ ] fzf binary
-    - [ ] fzf-tmux script
-    - [ ] Key bindings
-    - [ ] Completion
-    - [ ] Vim
-    - [ ] Neovim
-    - [ ] Etc.
+
+- [ ] I have read through the manual page (`man fzf`)
+- [ ] I have the latest version of fzf
+- [ ] I have searched through the existing issues
+
+## Info
+
 - OS
     - [ ] Linux
     - [ ] Mac OS X
     - [ ] Windows
-    - [ ] Windows Subsystem for Linux
     - [ ] Etc.
 - Shell
     - [ ] bash
     - [ ] zsh
     - [ ] fish
-
-<!--
-### Before submitting
-
-- Make sure that you have the latest version of fzf
-- If you use tmux, make sure $TERM starts with "screen"
-- For more Vim stuff, check out https://github.com/junegunn/fzf.vim
-
-Describe your problem or suggestion from here ...
--->
+    
+## Problem / Steps to reproduce
 
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ Gemfile.lock
 doc/tags
 vendor
 gopath
+*.zwc
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,11 +16,11 @@ env:
 jobs:
   include:
     - stage: unittest
-      go: "1.11.x"
+      go: "1.13.x"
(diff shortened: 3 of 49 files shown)


rumpelsepp:
Aren't there any cross platform libraries available for go which implement these kind of things? The arguments you pass to `protector.Protect()` will not match any platforms but OpenBSD.

qbit:
There is a seccomp package from elasticsearch that offers similar functionality on linux but because the security techniques are vastly different, there is no consistent api for cross OS usage.

rumpelsepp:
Then IMO it makes more sense to put the parameters entirely in the platform specific implementation. I think the goal is simply "set up a sandbox for fzf"; we don't need to reinvent the wheel here. So, in `main()` just call `Protect()` without parameters and set up the parameters in the platform specific implementation. Does that make sense for you?

qbit:
Yep!

rumpelsepp:
@junegunn What's your opinion on sandboxing fzf? From a security standpoint it makes perfectly sense.

qbit:
Updated!

junegunn:
Can you ELI5 what pledge is and why I should care? I don't use BSD, so please don't tell me I have to do my own research :)

junegunn:
fzf allows users to start any program via `execute` action and `preview` command. Does this patch affect what they can or cannot do? In other words, is there any change that this change might break someone's workflow?

qbit:
> Does this patch affect what they can or cannot do? In other words, is there any change that this change might break someone's workflow?

Nope. If you look at the “promises” made you can see `exec` is one of them. From the man page:


**exec**
Allows a process to call execve(2). Coupled with the proc promise, this allows a process to fork and execute another program. If execpromises has been previously set the new program begins with those promises, unless setuid/setgid bits are set in which case execution is blocked with EACCESS. Otherwise the new program starts running without pledge active, and hopefully makes a new pledge soon.

junegunn:
Thanks for the explanation. Can you also explain how you built the list?

```go
unix.PledgePromises("stdio rpath wpath cpath tty proc exec inet")
```

Does fzf need `inet`, `cpath`, or `wpath`?

qbit:
> Can you also explain how you built the list?

Typically I use a wrapper tool to narrow it down (calls `execvp` with the app in question)

> Does fzf need inet, cpath, or wpath?

It doesn't seem to! Not sure how I missed that! I will update the PR to remove those.

qbit:
Ok, reduced set of promises now in place.

vext01:
Hi,

I'm about to commit an fzf update to the OpenBSD ports tree. It'd be really neat to have these changes upstreamed, because as it stands we have a local patch set.

> Can you give me a reproducible example where not applying this patch can be lead to a problem?

Pledge limits the impact of a successful exploit; it's a mitigation technique in the same sense ASLR is.

Pledge lets your application promise what it definitely will not need at startup. For example, if you program doesn't need the filesystem, then you can use pledge to promise that. Then if an attacker somehow persuades your program to behave in an unintended fashion (e.g. gains arbitrary code execution via a vulnerability), then the impact of such a compromise is limited: they can't access the filesystem because the kernel will kill the process if the attacker tries.

I think it would be a nice addition to fzf and it would make our life easier to not have to maintain the local patch set.

Thanks for fzf! Many of us OpenBSD folks use it, both on the shell and in vim/neovim :)

> Typically I use a wrapper tool to narrow it down (calls execvp with the app in question)

@qbit what wrapper is that? Sounds useful!

qbit:
Wrapper is [here](https://code.bolddaemon.com/qbit/utils/raw/branch/master/pledge/pledge.c).

junegunn:
Sorry for the delay, merged to master. Thanks.

vext01:
Excellent! Many thanks.

PR #1299, opened 2018-05-21
zsh completion: Add support for per-command completion triggers.

rickyz:
Allow users to configure custom completion triggers on a per-command
basis. For example, if FZF_COMPLETION_TRIGGER='**' and
FZF_PER_CMD_COMPLETION_TRIGGERS=(ssh ''), then:

ssh<space><tab> triggers fzf completion.

but

ls<space><tab> will not.

Previously, completion for the kill command was special cased to not
require a completion trigger. This patch reimplements kill in terms of
FZF_PER_CMD_COMPLETION_TRIGGERS as well. Doing this also required
modifying _fzf_complete to respect quotes when doing word splitting on
its fzf_opts param.

This is a minor behavior change for kill completion. Previously, kill's
hardcoded fzf invocation enabled the following options by default:

--height ${FZF_TMUX_HEIGHT:-50%}
--min-height 15

and also allowed any options to be overriden by FZF_COMPLETION_OPTS.
When reusing _fzf_complete, this will no longer happen (consistent with
all of the other built-in completion functions).

### DIFF
--- a/shell/completion.zsh
+++ b/shell/completion.zsh
@@ -4,12 +4,13 @@
 #  / __/ / /_/ __/
 # /_/   /___/_/ completion.zsh
 #
-# - $FZF_TMUX                 (default: 0)
-# - $FZF_TMUX_OPTS            (default: empty)
-# - $FZF_COMPLETION_TRIGGER   (default: '**')
-# - $FZF_COMPLETION_OPTS      (default: empty)
-# - $FZF_COMPLETION_PATH_OPTS (default: empty)
-# - $FZF_COMPLETION_DIR_OPTS  (default: empty)
+# - $FZF_TMUX                        (default: 0)
+# - $FZF_TMUX_OPTS                   (default: empty)
+# - $FZF_COMPLETION_TRIGGER          (default: '**')
+# - $FZF_PER_CMD_COMPLETION_TRIGGERS (default: none)
+# - $FZF_COMPLETION_OPTS             (default: empty)
+# - $FZF_COMPLETION_PATH_OPTS        (default: empty)
+# - $FZF_COMPLETION_DIR_OPTS         (default: empty)
 
 
 # Both branches of the following `if` do the same thing -- define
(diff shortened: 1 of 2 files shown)


junegunn:
Thanks for your interest in the project. Actually, I regret making kill completion an exception, and I've considered many times if I should rectify the decision at some point, to always require **, even though it means breaking backward compatibility.

Making the completion system more configurable can benefit some users, but it will inevitably bring more complexity which I, the maintainer of this project and a mere bash user, cannot afford to handle at this point. Also note that when I add a new feature, I have to make sure that it works both on bash and zsh in a consistent manner and there are test cases for regular, and exceptional use cases. And once the new feature is added, I need to take extra caution not to break backward compatibility of the feature and I'll have to deal with questions, issue reports, and feature requests around it. In short, any decision cannot be made lightly.

I'll probably pass on this pull request, but I'll leave the pull request open for now, so that other users can notice the issue. Thanks.

avamsi:
@junegunn can this be reconsidered?

rickyz:
For what it's worth, I would still be interested in merging this if the maintainers are willing. There have been some improvements made to fzf's zsh code since the original pull request, so the patch is now very simple compared to before. As far as I know, none of the caveats in the original pull request description apply to the latest version anymore.

rickyz:
@aur3l14no That line says to set trigger to the first non-null value out of $FZF_PER_CMD_COMPLETION_TRIGGERS[$cmd], $FZF_COMPLETION_TRIGGER, or '**'. See https://zsh.sourceforge.io/Doc/Release/Expansion.html#Parameter-Expansion for a description of this zsh syntax.

Can you provide a minimal zshrc that demonstrates your problem? I'm afraid I don't have enough information to explain why this isn't working for you.

rickyz:
Now that 52594355bfa63ce7d579c7961f4f2fb30b486101 has removed the special case for kill, this has effectively become a one-line change. I also added a simple test case.

It wouldn't be difficult to implement the same feature in bash - the main reasons I have not done so are:
 - Bash 3 (sadly still the version on a default macOS install) does not support associative arrays, so the configuration variable would need to be different.
 - There is already a working method to override the completion trigger on a per-command basis in Bash: https://github.com/junegunn/fzf/wiki/Examples-(completion)#bash-custom-trigger-less-completion

That said, I would be happy to implement bash support if that would help - we would just need to agree on the mechanism for specifying custom triggers.

avamsi:
(Coming from #2026 again) @junegunn are you still not interested in this (now very small) PR? Is diverging support (which already seems to be the case, looking at https://github.com/junegunn/fzf/wiki/Examples-(completion)) for Bash and Zsh the concern?

TaiSHiNet:
Upvoting, this does seem like a very useful feature and the changeset is small enough

PR #1302, opened 2018-05-25
Store current position in the history.

mrbiggfoot:
Address https://github.com/junegunn/fzf.vim/issues/290 by storing current position along with pattern in json format in the history file. Warning: breaks backward compatibility with the existing history files.

### DIFF
--- a/src/terminal.go
+++ b/src/terminal.go
@@ -3,6 +3,7 @@ package fzf
 import (
 	"bufio"
 	"bytes"
+	"encoding/json"
 	"fmt"
 	"io"
 	"os"
@@ -412,6 +413,39 @@ func NewTerminal(opts *Options, eventBox *util.EventBox) *Terminal {
 		initFunc:   func() { renderer.Init() }}
 	t.prompt, t.promptLen = t.processTabs([]rune(opts.Prompt), 0)
 	return &t
+}
+
+// Persistent state of the Terminal struct
+type TerminalState struct {
+	Cy     int
+	Offset int
+	Input  string
+}
+
+func (t *Terminal) Marshal() string {
+	ts := TerminalState{
+		Cy:			t.cy,
+		Offset:		t.offset,
+		Input:		string(t.input),
+	}
+	b, err := json.Marshal(ts)
+	if err != nil {
+		panic(err)
+	}
+	return string(b)
+}
+
+func (t *Terminal) Unmarshal(str string) {
+	var ts TerminalState
+	err := json.Unmarshal([]byte(str), &ts)
+	if err != nil {
+		fmt.Println(err)
+		return
+	}
+	t.input = trimQuery(ts.Input)
+	t.cx = len(t.input)
+	t.cy = ts.Cy
+	t.offset = ts.Offset
 }
 
 // Input returns current query string
@@ -1423,7 +1457,7 @@ func (t *Terminal) Loop() {
 		t.tui.Close()
 		code := getCode()
 		if code <= exitNoMatch && t.history != nil {
-			t.history.append(string(t.input))
+			t.history.append(t.Marshal())
 		}
 		// prof.Stop()
 		os.Exit(code)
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks for the suggestion, but I don't like the idea. fzf is a general purpose text filter that can be used with any type of lists and it consumes the input asynchronously (e.g. `for i in $(seq 10); do echo $i; sleep 1; done | fzf`), so there is no guarantee at all that the list will be exactly the same at a given point. This will only cause unnecessary confusion in most cases where that is not the case.

mrbiggfoot:
Ok, it will stay in my fork then.

maximbaz:
@junegunn you are right, it can be used with any type of lists and in general case there is no order guarantee.

However, there are certain specific cases when the order does persist - e.g. multiple invocations of `:Ag` with `fzf-vim` over a short period of time. The scenario I have in mind is the following: run `:Ag`, type filter, press `<Up>, <Up>, <Up>`, select it with `<Enter>`, realize that this is not what you were looking for, run `:Ag` again, bring the last history entry back, and expect that you are already focused on the 3rd entry so that you can continue searching upwards.

Would you consider a less radical approach that implements this? For example, I would suggest making `fzf` _write_ position and selection pattern to history, but _not read_ it by default (so that the default behavior doesn't change). And then introduce a new CLI argument `--use-history-selection` for `fzf` that will additionally restore the position and selection pattern, so that people can make their `fzf-vim` instance use this additional flag if they wish so. The new flag could even do nothing if the history file does not contain position information, in order to avoid breaking backwards compatibility with the old history files.

junegunn:
Sorry, I'm not interested in adding features for niche use cases, fzf is a bit too convoluted already.

For `:Ag`, sending the selected entries to the quickfix window as suggested in the comments solves the issue. Type in a query, select multiple entries with `tab` / `shift-tab`, or `alt-a`, and press enter to populate the quickfix window. Then you can navigate back and forth through the list using native Vim key bindings such as follows:

```vim
nnoremap ]q :cnext<cr>zz
nnoremap [q :cprev<cr>zz
```

I know it's not ideal, as it's a two-phase process. If you're unhappy with the workflow, I suggest you try other native Vim plugins that provides better integration with Vim.

mrbiggfoot:
@maximbaz 
I'm working on [a plugin called "neoview"](https://github.com/mrbiggfoot/neoview) that integrates `fzf` with neovim ([this PR](https://github.com/neovim/neovim/pull/8325) is needed, or just use https://github.com/mrbiggfoot/neovim which btw supports fuzzy search of completion candidates):

There are still some minor quirks and it's not documented, but it works already. I use it myself, e.g. https://github.com/mrbiggfoot/nvimp/blob/master/vimrc#L433. The main features are:
* Open preview as a neovim window
* Search and preview window positions can be specified using standard vim commands (with extended syntax for height/width percentage)
* Search window height is dynamically adjusted based on the number of candidates.
* It will have a "resume last search" feature, since I use it a lot, but `fzf` will have to be taken from me: https://github.com/mrbiggfoot/fzf in order to support that.

Basically, I want it to be a fast version of "unite.vim" and it's not very far already. I don't know when it will be finished completely since I do it mainly for myself and the main incentive to implement a feature is my own need of it.

mrbiggfoot:
I made a `fzf` fork with binary releases. IMO this is how `--history` should work.
https://github.com/mrbiggfoot/fzf

maximbaz:
@mrbiggfoot allow creating issues in your fork, for future discussions 😉 

I tried it out (took the binary for linux-amd64), the selection position works, but not the selection pattern (unless I'm doing something wrong).

I tried:

```
$ ls -al | fzf -m
<Shift+Tab>, <Shift+Tab>, <Shift+Tab>, <Enter>
$ ls -al | fzf -m
<Ctrl+p> (get previous entry from the history)
```

The cursor moves to the correct position, but the 3 entries that I selected with `<Shift+Tab>` are not selected again. I would expect them to be selected.

mrbiggfoot:
@maximbaz issues are now allowed.
Yeah, `-m` is not supported at the moment, just the cursor position. I will look into it.

mrbiggfoot:
@maximbaz check out the new version. Do `rm -Rf ~/.fzf` before updating via `vim-plug`, since I modified the commit history.

maximbaz:
Now this is very cool, nice work!

mrbiggfoot:
@junegunn I will implement "one line up/down" actions at some point. The idea is to have the behavior similar to vim's `<C-y>` and `<C-e>`. Will you be interested in merging it? Otherwise, I don't want to waste time on a PR.

mrbiggfoot:
The train has left the station:
https://github.com/mrbiggfoot/fzf/commit/0a8b48ff578b46a75f53e82878aecc8f51d39409#diff-f755b71f11d73dbdb3f3cbde6d068f65
All my work on `fzf` from now on will be in my fork.
Cheers!

PR #1303, opened 2018-05-25
[bash, zsh] Fix missing fuzzy completions

### LINKED ISSUE #1301: **<TAB> does not list all matches
- Category
    - [x] Completion
- OS
    - [x] Mac OS X
- Shell
    - [x] bash

Given the following hierarchy
```
├── foo
│   ├── baz
│   └── qux
└── foobar
```
`cat foo**<TAB>` only shows `foo`, `foo/baz` and `foo/qux`. I would expect `foobar` to be included as well.

If I instead do `cat fo**<TAB>`, all four matches are listed.

I guess the problem has something to do with `foo` being identified as the target directory in the first case, hence `foobar` being excluded?

ptzz:
`cat foo**<TAB>` did not display the file `foobar` if there was a directory
named `foo`.

Fixes #1301

### DIFF
--- a/shell/completion.bash
+++ b/shell/completion.bash
@@ -148,7 +148,7 @@ __fzf_generic_path_completion() {
     base=${cur:0:${#cur}-${#trigger}}
     eval "base=$base"
 
-    dir="$base"
+    [[ $base = *"/"* ]] && dir="$base"
     while true; do
       if [ -z "$dir" ] || [ -d "$dir" ]; then
         leftover=${base/#"$dir"}
--- a/shell/completion.zsh
+++ b/shell/completion.zsh
@@ -36,8 +36,7 @@ __fzfcmd_complete() {
 
 __fzf_generic_path_completion() {
   local base lbuf compgen fzf_opts suffix tail fzf dir leftover matches
-  # (Q) flag removes a quoting level: "foo\ bar" => "foo bar"
-  base=${(Q)1}
+  base=$1
   lbuf=$2
   compgen=$3
   fzf_opts=$4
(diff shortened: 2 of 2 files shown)


junegunn:
Make sense, thanks. zsh completion also seems to have the same issue, can you fix it as well?

junegunn:
Zsh test case is not passing as the new code doesn't support expanding `ls ~username**`.

A separate, related issue I noticed is that `ls $HOME**` currently expands on bash but not on zsh, just like `~username`. Ideally, I'd like to see if it's possible to make the behavior consistent across bash and zsh

ptzz:
> A separate, related issue I noticed is that ls `$HOME**` currently expands on bash but not on zsh, just like `~username`.

It should expand for zsh as well with the last fix.

junegunn:
Thanks, but I noticed that it has an unwanted side effect when the prefix contains spaces:

```sh
> mkdir -p /tmp/foo\ ls
> cat /tmp/foo\ ls/**<TAB>(eval):1: command not found: ls/

> eval "a=b c"
zsh: command not found: c
```

`base=${~base}` instead of `eval` handles `~username`, but not `$HOME`. Any idea? If there isn't a simple solution, we should just go with `base=${~base}`.

ptzz:
No idea, sorry, not a zsh user. Spamming the CI at the moment, wasn't able to run the test suite locally unfortunately.

Another inconsistency I noted was that `cat foo/**<TAB>` shows the dir `foo` in zsh but not in bash. I guess bash's result is the expected one.

ptzz:
So with the last commit, tests pass and also your case with prefix containing spaces should work. Do you think it's acceptable? The issue with `$HOME` still remains.

junegunn:
Using`eval` is probably not a good practice, but I realized that we already use it in bash completion, and that's why `$HOME` or `~user` are expanded on bash (but it also means that `ls $(sleep 5)**` are evaluated too, which is probably not a good thing as the command may have side effects).

I applied the same approach to the zsh version and it seems to work as expected. Can you review the commit I pushed to your branch?

ptzz:
Looks good to me!

junegunn:
Merged, thanks!

PR #1326, opened 2018-06-29
Support for ppc64le

ymanton:
These patches add ppc64le to the build system and update the crypto lib to pull in a fix [1] for syscalls that's needed to get fzf working. These changes and building with Go 1.10 allows fzf to work on ppc64le (tested on Ubuntu 16.04). Building with Go 1.6-1.9 does not produce a working fzf, presumably due to bugs in Go.

[1] https://go.googlesource.com/crypto/+/558b6879de74bc843225cde5686419267ff707ca

Note: I'm not a Go user myself so I have no idea if the generated `glide.lock` file is correct, it's what was emitted when I did `glide update`.


ymanton:
Fixed.

junegunn:
Merged, thanks!

PR #1341, opened 2018-07-30
Initial version of height option for Windows

kelleyma49:
--height works under Windows versions that support vt100 (https://twitter.com/shanselman/status/695738082209853440). 
Mouse support doesn't work.

### DIFF
--- a/src/tui/light.go
+++ b/src/tui/light.go
@@ -565,7 +565,6 @@ func (r *LightRenderer) Refresh() {
 }
 
 func (r *LightRenderer) Close() {
-	r.closePlatform()
 
 	// r.csi("u")
 	if r.clearOnExit {
@@ -585,6 +584,7 @@ func (r *LightRenderer) Close() {
 		r.csi("?1000l")
 	}
 	r.flush()
+	r.closePlatform()
 	r.restoreTerminal()
 }
 
--- a/src/tui/light.go
+++ b/src/tui/light.go
@@ -7,7 +7,6 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
-	"syscall"
 	"time"
 	"unicode/utf8"
 
@@ -28,21 +27,6 @@ const (
 const consoleDevice string = "/dev/tty"
 
 var offsetRegexp *regexp.Regexp = regexp.MustCompile("(.*)\x1b\\[([0-9]+);([0-9]+)R")
-
-func openTtyIn() *os.File {
-	in, err := os.OpenFile(consoleDevice, syscall.O_RDONLY, 0)
-	if err != nil {
-		tty := ttyname()
-		if len(tty) > 0 {
-			if in, err := os.OpenFile(tty, syscall.O_RDONLY, 0); err == nil {
-				return in
-			}
-		}
-		fmt.Fprintln(os.Stderr, "Failed to open "+consoleDevice)
-		os.Exit(2)
-	}
-	return in
-}
 
 func (r *LightRenderer) stderr(str string) {
 	r.stderrInternal(str, true)
@@ -80,26 +64,31 @@ func (r *LightRenderer) flush() {
 
 // Light renderer
 type LightRenderer struct {
-	theme         *ColorTheme
-	mouse         bool
-	forceBlack    bool
-	clearOnExit   bool
-	prevDownTime  time.Time
-	clickY        []int
-	ttyin         *os.File
-	buffer        []byte
-	origState     *terminal.State
-	width         int
-	height        int
-	yoffset       int
-	tabstop       int
-	escDelay      int
-	fullscreen    bool
-	upOneLine     bool
-	queued        string
-	y             int
-	x             int
-	maxHeightFunc func(int) int
+	theme           *ColorTheme
+	mouse           bool
+	forceBlack      bool
+	clearOnExit     bool
+	prevDownTime    time.Time
+	clickY          []int
+	ttyin           *os.File
+	ttyinChannel    chan byte
+	inHandle        uintptr // Windows only
+	outHandle       uintptr // Windows only
+	buffer          []byte
+	origState       *terminal.State
+	origStateInput  uint32 // Windows only
+	origStateOutput uint32 // Windows only
+	width           int
+	height          int
+	yoffset         int
+	tabstop         int
+	escDelay        int
+	fullscreen      bool
+	upOneLine       bool
+	queued          string
+	y               int
+	x               int
+	maxHeightFunc   func(int) int
 }
 
 type LightWindow struct {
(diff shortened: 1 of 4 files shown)


kelleyma49:
Hmm, found an issue with the arrow keys. I need to look into why it's exiting if you bang on the up/down arrow keys.

kelleyma49:
I found another issue with calling `fzf --height` from within a PsReadline key handler. So I still have some work to do.

@junegunn , at least look at how I separated out the light renderer to see if you have any issues with how I implemented the separation.

junegunn:
Thanks for working on this. The code separation looks good to me.

/cc @janlazo

src/tui/light_windows.go
```diff
@@ -0,0 +1,131 @@
...
+	if err := windows.GetConsoleMode(windows.Stderr, &oldState); err != nil {
+		return false
+	}
+	canSetVt100 := windows.SetConsoleMode(windows.Stderr, windows.ENABLE_VIRTUAL_TERMINAL_PROCESSING) == nil
+	windows.SetConsoleMode(windows.Stderr, oldState)
+	return canSetVt100
+}
+
+func (r *LightRenderer) initPlatform() error {
+	IsLightRendererSupported()
```
janlazo: Return value of this function isn't checked. Intentional?

src/tui/light_windows.go
```diff
@@ -0,0 +1,131 @@
...
+			return err
+		}
+	}
+
+	inHandle := windows.Stdin
+	//inHandle, _ := syscall.Open("CONIN$", syscall.O_RDWR, 0)
+	if err := windows.GetConsoleMode(windows.Handle(inHandle), &oldStateInput); err != nil {
+		return err
+	}
+	var requestedInModes uint32 = windows.ENABLE_VIRTUAL_TERMINAL_INPUT | windows.ENABLE_PROCESSED_INPUT | windows.ENABLE_WINDOW_INPUT | windows.ENABLE_MOUSE_INPUT | windows.ENABLE_EXTENDED_FLAGS
```
janlazo: Does this depend on Windows 10?

src/tui/light.go
```diff
@@ -132,11 +121,10 @@ func NewLightRenderer(theme *ColorTheme, forceBlack bool, mouse bool, tabstop in
 	return &r
 }
 
-func (r *LightRenderer) fd() int {
-	return int(r.ttyin.Fd())
-}
-
 func (r *LightRenderer) defaultTheme() *ColorTheme {
+	if util.IsWindows() && os.Getenv("ConEmuANSI") == "ON" {
```
janlazo: winpty doesn't unset `ConEmuANSI` (#106) so this can break Vim `:terminal` on Windows.

kelleyma49: I'm fine with removing this check unless you have another suggestion.

janlazo:
Can we use Travis to test this on [Windows](https://docs.travis-ci.com/user/reference/windows/)?

src/tui/light_windows.go
kelleyma49: This was addressed in a later commit.

kelleyma49:
@janlazo: please take a look at the latest version when you get a chance.

janlazo:
@kelleyma49 Sure. Does this feature support Windows 10 only? I don't have my own Windows machine for Windows 7,8 so I can test it only in the VM.

kelleyma49:
> @kelleyma49 Sure. Does this feature support Windows 10 only? I don't have my own Windows machine for Windows 7,8 so I can test it only in the VM.

It's only Windows 10, Anniversary Update and newer (https://devblogs.microsoft.com/commandline/new-experimental-console-features/). There's code in my commit to detect if vt100 sequences are supported.

src/tui/light_windows.go
```diff
@@ -0,0 +1,132 @@
+//+build windows
+
+package tui
+
+import (
+	"os"
+	"syscall"
+
+	"github.com/junegunn/fzf/src/util"
+	"golang.org/x/sys/windows"
```
janlazo: `go build` fails because this is missing from `go.mod`. `go get 'golang.org/x/sys/windows'` should be enough. I can run fzf with `--height` on Windows 10 after this is resolved. Tested on powershell default terminal.

kelleyma49: @janlazo - Should I submit a separate pull request with golang/sys updated to the latest version?

janlazo: Sure.

src/tui/light_unix.go
```diff
@@ -0,0 +1,100 @@
...
+
+func (r *LightRenderer) fd() int {
+	return int(r.ttyin.Fd())
+}
+
+func (r *LightRenderer) initPlatform() error {
+	fd := r.fd()
+	origState, err := terminal.GetState(fd)
+	if err != nil {
+		errorExit(err.Error())
```
junegunn: We can just return `err` here.

src/tui/light.go
janlazo: It's ok to keep it now. There's a way to modify or unset environment variables for (neo)vim terminal without relying on the shell.

wjrogers:
I am testing this patch locally after merging with master @ d9c6a0305b8c7f83a21fa583ae662660b940b1b6. My primary use case is CTRL+R command history using [PSFzf](https://github.com/kelleyma49/PSFzf) (see also https://github.com/kelleyma49/PSFzf/issues/30, https://github.com/junegunn/fzf/issues/1766).

- PowerShell 7.0
- Windows Terminal (Preview) 0.9.433.0
- go 1.14
- `$Env:FZF_DEFAULT_OPTS="--ansi --layout=reverse --height=40%"`

I had no problem building or running `fzf`. At first glance, the height option works correctly. I will keep using this binary and report back if I encounter anything unexpected.

--

I had some hope that these changes (and updating the sys module) would also fix #1766, but they do not. Any portion of the terminal buffer erased by fzf is not restored. I also observed some additional weird interactions w/ PSFzf, which I mention here because the PR author is also the author of PSFzf:
- The prompt still gets erased after using CTRL+R, even when `height` is set
- When there are fewer blank lines remaining in the terminal viewport than the value of `height`, the selected command appears at the bottom of the terminal after invoking CTRL+R

KindDragon:
@wjrogers, which terminal are you using?

kelleyma49:
Thanks for testing this! Hopefully this will expedite the merging of this request. :) 

> 
> * The prompt still gets erased after using CTRL+R, even when `height` is set
> * When there are fewer blank lines remaining in the terminal viewport than the value of `height`, the selected command appears at the bottom of the terminal after invoking CTRL+R

I just submitted a workaround for the missing prompt issue in PSFzf. You can download a Prerelease version here: https://www.powershellgallery.com/packages/PSFzf/1.1.36-alpha. 

It works around the issue by telling PsReadLine to redraw the prompt. This could be an issue if you have a slow prompt function.

Note that if you call fzf.exe directly, the screen buffer is correctly displayed after you fzf.exe, so the issues you're seeing seem to be only when fzf is executed through PSReadLine.

junegunn:
I'm very sorry for keeping this open for so long. I finally had a chance to test this on Windows Server 2019 and it works really well. Rebased and merged to master. Thank you very much for your contribution! 🎉 

I'll be releasing a new version pretty soon.

kelleyma49:
> Rebased and merged to master.

Woo hoo! 🎉

> I'm very sorry for keeping this open for so long

No need to apologize. Thank you for creating such an awesome tool! 👏

PR #1346, opened 2018-07-31
add --reverse to __fzf_history__() to be consistent with the rest

gertcuykens:


junegunn:
Thanks, but I prefer the way it is now, as the command history is displayed in chronological order. We used this setup for almost five years by now and I don't want to change it all of a sudden. I believe you already know this, but you can customize the options via `FZF_CTRL_R_OPTS`.

PR #1355, opened 2018-08-11
Fuzzy completion for Make

micha-k:
Add fuzzy completion for Make command by looking up targets inside a makefile

### DIFF
--- a/shell/completion.bash
+++ b/shell/completion.bash
@@ -266,6 +266,12 @@ _fzf_complete_unalias() {
   )
 }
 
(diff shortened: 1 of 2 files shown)


junegunn:
Thanks for the suggestion, but I decided not to maintain more custom completion functions in this repository. You might want to add your example to https://github.com/junegunn/fzf/wiki/Examples-(completion) instead.

By the way, one issue I noticed with the implementation is the grep pattern picks up false positives, such as `FOO := BAR`.

Issue #270, opened 2015-06-18
fzf needs a -u/--unique option


stephentalley:
It would be nice if fzf had a way to omit repeated lines.

I typically run fzf with this pattern:

    (cmd1; cmd2) | fzf

where:

* cmd1 and cmd2 return the same results, mostly.
* cmd1 is fast but unreliable (results may be out of date).
* cmd2 is slow but reliable (results are always current).

This allows me to get fast matches that are probably good, while slower ones that are definitely good trickle in.

Unfortunately this leaves me with a lot of duplicated lines in fzf.

For example, I'd like for:

    (echo 'hello'; sleep 3; echo 'hello') | fzf

...to only display "hello" once.

A -u/--unique option would be nice.

Apologies if there's already a way to do this and I've overlooked it.

Thanks!

junegunn:
Thanks for the suggestion, but it's not the responsibility of fzf and should be done by a separate filter. Trying to implement such behavior in fzf is non-trivial and it's not in accordance to the unix philosophy.

We both know `uniq` doesn't work here as it is for sorted input, and I don't know if there's a pre-existing unix filter for the purpose, but it's trivial to write one using any scripting language. For example, with Ruby,

```ruby
#!/usr/bin/env ruby

require 'set'
found = Set.new
$stdout.sync = true
ARGF.each do |line|
  unless found.include? line
    found << line
    puts line
  end
end
```

```sh
(echo 'hello'; sleep 3; echo 'hello') | nodup | fzf
```

stephentalley:
Thanks, that's a fair assessment.

edi9999:
Great to see that you don't accept too many features so that the thing doesn't get too complicated to maintain (I guess it's already quite big)

rafaeln:
I understand why you can't use `uniq` to solve the original problem, but I solved a similar problem using `uniq` and thought I could share my solution with you people. I use `fzf`'s `ctrl-r` command a lot to, for instance, obsessively issue the command to update my vim plugins, as you can see in the image below.

![image](https://cloud.githubusercontent.com/assets/2940362/23773386/fb1ed450-04ec-11e7-941f-c42ce2a0070e.png)

I didn't want to see repeated commands anymore, and changing the definition of `ctrl-r` in `key-bindings.bash` solves this issue. The only change is on the line `HISTTIMEFORMAT= history |`, which becomes `HISTTIMEFORMAT= history | sort -k 2 | uniq -f 1 |`. I copy the new version below:

[...]

PR #1363, opened 2018-08-28
[bash,zsh] unique shell history (optional through FZF_UNIQUE_HISTORY)

jerdna-regeiz:
Hey,

I like to have a deduplicated shell history when using CTRL-R.

Yes, I am aware that it is a recurring question and I have seen them (#749, #270, #49, #88, #492, #600).
And thus it occurred to me, that it might bother some more people than myself.

The solutions so far did not solve my desire, as most of them would omit the duplicated entries from the history entirely and I want to keep it for histories sake.
Others propose nanuniq, which is yet an other not standard tool (not even in my software repositories), and thus I did not want to rely on it. Further I guess I have an easier approach with much less code counting in nauniq.
Using an alternative CTRL-R command would be able to solve the problem as well, but that was declined (with reasonable arguments) and I feel it would be a to complicated feature for a simple task.

Therefore I provided an option ( FZF_UNIQUE_HISTORY=1 ) to enable a unique history without changing the default as it is (therefore breaking no tests). I am sad that I can not provide further tests for the new "feature" as I have experience in go yet (maybe it's time to start).
I only rely on perl which is installed on most systems per default (perl was only choosen as I guess it's the most compact and simple way to express the duplicate filtering).
Lines are filtered for uniqueness ignoring numbers and spaces at the beginning.
Only the first match is included. It could easily be extended to include only the last match, but it yould require two tail -r. I'd like to open that up for discussion if it is worth it (haven't taken any performance measures).
I included and tested bash and zsh. I am unfamiliar with fish, but per default it does filter out duplicates in history on my system - therefore I didn't dig any deeper.

I really hope this might be included or may at least help some others looking for a simple solution without extra scripts.

Cheers and thanks for this awesome project,
 Andrej

### DIFF
--- a/shell/key-bindings.bash
+++ b/shell/key-bindings.bash
@@ -51,11 +51,18 @@ __fzf_cd__() {
   dir=$(eval "$cmd" | FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} --reverse $FZF_DEFAULT_OPTS $FZF_ALT_C_OPTS" $(__fzfcmd) +m) && printf 'cd %q' "$dir"
 }
 
+__fzf_history_dedup__() {
+  if [ -z ${FZF_UNIQUE_HISTORY+x} ]; then
+    cat
+  else
+    perl -ne 'print if !$seen{($_ =~ s/^[0-9\s]*//r)}++'
+  fi
+}
 __fzf_history__() (
   local line
   shopt -u nocaseglob nocasematch
   line=$(
-    HISTTIMEFORMAT= history |
+    HISTTIMEFORMAT= history | __fzf_history_dedup__ |
     FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} $FZF_DEFAULT_OPTS --tac --sync -n2..,.. --tiebreak=index --bind=ctrl-r:toggle-sort $FZF_CTRL_R_OPTS +m" $(__fzfcmd) |
     command grep '^ *[0-9]') &&
     if [[ $- =~ H ]]; then
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -65,11 +65,19 @@ fzf-cd-widget() {
 zle     -N    fzf-cd-widget
 bindkey '\ec' fzf-cd-widget
 
+fzf-history-dedup() {
+  if [ -z ${FZF_UNIQUE_HISTORY+x} ]; then
+    cat
+  else
+    perl -ne 'print if !$seen{($_ =~ s/^[0-9\s]*//r)}++'
+  fi
+}
+
 # CTRL-R - Paste the selected command from history into the command line
 fzf-history-widget() {
   local selected num
   setopt localoptions noglobsubst noposixbuiltins pipefail 2> /dev/null
-  selected=( $(fc -rl 1 |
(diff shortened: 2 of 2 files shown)


jerdna-regeiz:
Don't really want to argue, as your mind seems to be set, but it really does NOT have that much of an impact.

I already mentioned, that one might reverse it twice. I uncovered tail -r is not that widespread, but tac is (reverse of cat).

Thus I ran some performance test with a large history (enlarged my own...):
History file:  69993 ./myhist entries
100 rounds each, remember to divide results by 100 for a single run!
Zsh
Time of history duplicated (cat | fzf | wc -l)
zsh ./hist_dupl.sh  9.08s user 1.06s system 182% cpu 5.550 total
Time of history deduplicated
zsh ./hist_dedup.sh  11.26s user 0.76s system 109% cpu 10.965 total
Time of history reverse, dedup, reverse
zsh ./hist_dedup_double_reverse.sh  11.79s user 1.48s system 110% cpu 11.969 total
Bash
Time of history duplicated (cat | fzf | wc -l)
bash ./hist_dupl.sh  9.16s user 0.99s system 177% cpu 5.728 total
Time of history deduplicated
bash ./hist_dedup.sh  11.68s user 0.88s system 107% cpu 11.640 total
Time of history reverse, dedup, reverse
bash ./hist_dedup_double_reverse.sh  12.06s user 1.31s system 106% cpu 12.510 total

Meaning a delay for zsh from doing nothing but cat into fzf to reverse, deduplicate, reverse into fzf of 0.027s per call. I guess you test was with sort, but we really don't need to sort.
Test it four yourself, I attached the crude scripts I used.

[hist_dedup.txt](https://github.com/junegunn/fzf/files/2472428/hist_dedup.txt)
[hist_dedup_double_reverse.txt](https://github.com/junegunn/fzf/files/2472429/hist_dedup_double_reverse.txt)
[hist_dupl.txt](https://github.com/junegunn/fzf/files/2472430/hist_dupl.txt)
[time.txt](https://github.com/junegunn/fzf/files/2472431/time.txt)

junegunn:
> as it will most likely break in the future, when fzf receives updates

No, I can assure you that you don't have to worry about it.

breuerfelix:
awesome feature! thats actually the only solution that is working like a charm for me :)  
i use it in conjunction with the `setopt HIST_FIND_NO_DUPS` from zsh.

Here is what you have to insert into your `.bashrc` or `.zshrc` to enable it with this opt
```bash
source ~/.fzf.zsh

fzf-history-dedup() {
  if [ $options[HIST_FIND_NO_DUPS] = on ]; then
    perl -ne 'print if !$seen{($_ =~ s/^[0-9\s]*//r)}++'
  else
    cat
  fi
}

fzf-history-widget() {
  local selected num
  selected=( $(fc -rl 1 | fzf-history-dedup |
    FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} $FZF_DEFAULT_OPTS -n2..,.. --tiebreak=index --bind=ctrl-r:toggle-sort $FZF_CTRL_R_OPTS --query=${(qqq)LBUFFER} +m" $(__fzfcmd)) )
  local ret=$?
  if [ -n "$selected" ]; then
    num=$selected[1]
    if [ -n "$num" ]; then
      zle vi-fetch-history -n $num
    fi
  fi
  zle reset-prompt
  return $ret
}
```

PR #1392, opened 2018-09-26
Don't drop buffered input data in findOffset()

timbertson:
Sometimes when `fzf` takes a while to startup, I start typing but my input gets dropped once the UI actually renders, and it's slightly annoying. It's rare (because `fzf` is usually quick), but noticeable particularly on first run after booting.

To reproduce, you can simulate a slow load with:

```
git ls-files | (sleep 1; fzf)
```

Currently anything you type in the first second will be dropped, with this change it gets consumed as search input once fzf loads.

I don't know if it's a bit weird to modify `r.buffer` during `findOffsets()`. It could alternatively return `(skipped, row, col)` and leave it up to the caller to add `skipped` to `r.buffer`, but that might get awkward.

I also don't know if this is the only place which needs this fix - is there a dark TUI as well as a light one?

### DIFF
--- a/src/tui/light.go
+++ b/src/tui/light.go
@@ -27,7 +27,7 @@ const (
 
 const consoleDevice string = "/dev/tty"
 
-var offsetRegexp *regexp.Regexp = regexp.MustCompile("\x1b\\[([0-9]+);([0-9]+)R")
+var offsetRegexp *regexp.Regexp = regexp.MustCompile("(.*)\x1b\\[([0-9]+);([0-9]+)R")
 
 func openTtyIn() *os.File {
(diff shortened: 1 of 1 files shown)


junegunn:
Although I can't reproduce the issue with the command above, I do experience drops when using CTRL-R binding for bash and your patch fixes it.

`findOffset` is not a pure function already, so I can live with it appending leftover keystrokes to the internal buffer. I'll just merge the patch, thanks!

> is there a dark TUI as well as a light one?

Ah, no. LightRenderer was named "light" just because it did not depend on ncurses unlike the old [FullscreenRenderer](https://github.com/junegunn/fzf/commit/e3e7b3360cd83c20a109ccc550bd0508fc5b1cba) :)

Issue #867, opened 2017-03-02
FZF clears previous line of prompt


paymog:
- Category
    - [x] fzf binary
    - [x] fzf-tmux script
    - [ ] Key bindings
    - [x] Completion
    - [ ] Vim
    - [ ] Neovim
    - [ ] Etc.
- OS
    - [ ] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Windows Subsystem for Linux
    - [x] CentOS7
- Shell
    - [ ] bash
    - [x] zsh
    - [ ] fish

I have a multiline prompt in zsh and recently updated fzf (I don't know the previous version I was on unforutnately). I've found that now when I hit CTRL-T in my prompt to bring up the search window then ESC out of it, the previous line in my prompt gets erased. See https://imgur.com/a/TJyep

I just tested making my prompt a single line and the issue doesn't persist. See https://imgur.com/a/9UDtj

junegunn:
Can you share the source code for your prompt so I can test with it?

paymog:
Here's a copy of my prompt: http://pastebin.com/9EG9aCWx

It's a shittily modified version of https://github.com/sorin-ionescu/prezto/blob/master/modules/prompt/functions/prompt_sorin_setup

Thanks for following up on this!

junegunn:
Well, I tried your config (which required me to install prezto as well), but it seems like it has issues not only with fzf but also with other command-line completions. For example, simply doing `vim <tab>` or `ls <tab>` causes the problem you described. I think you should fix the issue first. Let me know if you still have problems with fzf bindings even after that.

paymog:
Interesting, that doesn't happen for me.
![ls not broken](https://cloud.githubusercontent.com/assets/1986950/23563699/a3986a60-fffc-11e6-9449-6730f68c23ae.gif)

paymog:
Oh, nevermind. I was able to repro with `ls` too.

paymog:
@junegunn I can't repro this issue with `ls <tab> <ctrl-c>` or `vim <tab> <ctrl-c>` anymore but now `<ctrl-t> <down arrow> <enter>`, `<ctrl-t> <esc>` and `<ctrl-t> <ctrl-c>` all cause the previous line to be eaten.

If you don't mind looking into this again, here's my latest prompt: https://github.com/paymog/prezto/blob/4a5a78b744c111f506623754a91bebb2925bbbb0/modules/prompt/functions/prompt_sorin_setup

[...]

PR #1397, opened 2018-10-03
fix multiline prompt issue with 'zle reset-prompt'

pfrybar:
Builds upon PR https://github.com/junegunn/fzf/pull/1256, fixes the zsh multiline PS1 issue explained in https://github.com/junegunn/fzf/issues/867.

Tried this with both single and multi-line PS1 prompts and it works for both. This was inspired by a fix to a similar problem with pure (https://github.com/sindresorhus/pure) described here - https://github.com/sindresorhus/pure/issues/143, with the relevant fix here - https://github.com/sindresorhus/pure/pull/273/files#diff-ca464c5c9e47e8a26c7e6ac0ed786406R61.

I don't think this breaks anything, but it would be good to have someone else try this to make sure.

### DIFF
--- a/shell/completion.zsh
+++ b/shell/completion.zsh
@@ -60,8 +60,7 @@ __fzf_generic_path_completion() {
       if [ -n "$matches" ]; then
         LBUFFER="$lbuf$matches$tail"
       fi
-      zle redisplay
-      typeset -f zle-line-init >/dev/null && zle zle-line-init
+      zle reset-prompt
       break
     fi
     dir=$(dirname "$dir")
@@ -100,8 +99,7 @@ _fzf_complete() {
   if [ -n "$matches" ]; then
     LBUFFER="$lbuf$matches"
   fi
-  zle redisplay
-  typeset -f zle-line-init >/dev/null && zle zle-line-init
+  zle reset-prompt
   command rm -f "$fifo"
 }
 
@@ -165,8 +163,7 @@ fzf-completion() {
     if [ -n "$matches" ]; then
       LBUFFER="$LBUFFER$matches"
     fi
-    zle redisplay
-    typeset -f zle-line-init >/dev/null && zle zle-line-init
+    zle reset-prompt
   # Trigger sequence given
   elif [ ${#tokens} -gt 1 -a "$tail" = "$trigger" ]; then
     d_cmds=(${=FZF_COMPLETION_DIR_COMMANDS:-cd pushd rmdir})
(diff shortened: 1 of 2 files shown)


junegunn:
Can you explain why I can't reproduce the issue described in #867, with a simple multi-line prompt: `PS1=$'foo\nbar> '` ?

Also, I noticed that that you removed `zle zle-line-init` which was added in #610 for fixing #279. Can you confirm that it's no longer needed?

/cc @eigengrau

junegunn:
We also have `zle redisplay` in completion.zsh, should we change them as well?

eigengrau:
> Also, I noticed that that you removed zle zle-line-init which was added in #610 for fixing #279. Can you confirm that it's no longer needed?

So, I tested one fzf-widget without calling `zle-line-init`, and #279 did no longer trigger. It’s odd though… The only reason the call would not be needed any longer would be if the main fzf application (or any terminal-handling libraries it uses) no longer puts the terminal into raw mode.

junegunn:
Thanks for the comment. Maybe the issue was fixed from the zsh side.

https://github.com/sindresorhus/pure/issues/319#issuecomment-321004414

If that's the case, and the issue still persists on earlier versions of zsh, we may need to keep those lines for the users with old zsh.

eigengrau:
> Maybe the issue was fixed from the zsh side.

I wondered that, too, but doubt it, since Zsh never really touched the terminal mode anywhere and the FAQ still tell folks to roll their own `zle-line-{init,finish}` if they want to use terminfo-escapes.

pfrybar:
Hm, so I use [prezto](https://github.com/sorin-ionescu/prezto) and the pure theme which is what causes issues I guess. When I try a simple multi-line prompt I don't see the problem (like you said).

Here's a screencast of the issue: https://imgur.com/a/Zp9ZCio. This no longer happens with my patch.

As for `zle-line-init` it was trial-and-error. I figured since we are resetting the prompt, we wouldn't need to explicitly call this init function. When I removed it, it looked like everything was still working. I'm really not sure if it's still needed or not.

pfrybar:
I just pushed up a new commit which makes the changes in `completion.zsh`. If you think we still need the `zle-line-init`, I can add those back.

junegunn:
I was able to reproduce the issue with `steeef` theme of pretzo and confirmed that this patch fixes it. Realized that #279 was a very old issue before fzf rewrote its renderer and dropped ncurses dependency, so I guess we don't have to worry about it.

Merging, thanks!

juanibiapina:
This doesn't fix it on `zsh 5.7.1 (x86_64-apple-darwin17.7.0)`

`zle -R` fixes it, though.

PR #1401, opened 2018-10-11
Support completion for tig

sharils:
This change is [Reviewable](https://reviewable.io/reviews/junegunn/fzf/1401)

### DIFF
--- a/shell/completion.bash
+++ b/shell/completion.bash
@@ -278,7 +278,7 @@ a_cmds="
   basename bunzip2 bzip2 chmod chown curl cp dirname du
   find git grep gunzip gzip hg jar
   ln ls mv open rm rsync scp
-  svn tar unzip zip"
+  svn tar tig unzip zip"
 x_cmds="kill ssh telnet unset unalias export"
 
 # Preserve existing completion


shell/completion.bash
```diff
@@ -278,7 +278,7 @@ a_cmds="
   basename bunzip2 bzip2 chmod chown curl cp dirname du
   find git grep gunzip gzip hg jar
   ln ls mv open rm rsync scp
-  svn tar unzip zip"
+  svn tar tig unzip zip"
```
ptzz: @junegunn Perhaps `a_cmd` could include an environment variable, e.g. `a_cmds=${FZF_COMPLETION_COMMANDS} awk cat ...`, so that it's easy extend any command with fzf completion?

My use case is I'd like fzf completion on `**<TAB>` for a custom defined alias that has existing tab completion.

junegunn: @ptzz Yeah, that seems like a better idea.

junegunn: @ptzz But then again, I'm not sure if we really need another variable when it's already possible to set up fuzzy completion for individual commands.

https://github.com/junegunn/fzf#supported-commands

ptzz: @junegunn IIRC original tab completions were not preserved when setting up fuzzy completion for new commands.

junegunn: Good point. I think a better approach would be to provide an API for setting up completion, but that needs more discussion.

junegunn:
I'm going to close this because you can set up additional fuzzy completion as shown in https://github.com/junegunn/fzf#supported-commands.

shell/completion.bash
ptzz: For anyone interested in a workaround/hack not requiring changes to `fzf`, I currently have this in my `.bash_profile`:

```
_fzf_orig_completion_tig=_tig
complete -F _fzf_path_completion -o default -o bashdefault tig
```

PR #1410, opened 2018-10-21
Support Ag command in mac os

eyaltrabelsi:
fix(Ag command doesn't work on mac with current installation)

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,6 +133,12 @@ Plug 'junegunn/fzf', { 'dir': '~/.fzf', 'do': './install --all' }
   " and you don't have to run install script if you use fzf only in Vim.
 ```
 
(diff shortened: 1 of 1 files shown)


i-am-eva:
The ":Ag" command is actually created by `https://github.com/junegunn/fzf.vim`, the "vim plugin" included with stock `fzf` only provides the `:FZF` command as well as the `fzf#run` and `fzf#wrap` functions. 

The readme for `fzf.vim` already includes links to `Ag` and `Rg` though specifically noting that those tools must be installed separately might not be a terrible idea.

Perhaps you could close this request and add some (non-platform specific) instructions to it's `Readme.md`? Probably at the end of "Installation".

PR #1420, opened 2018-11-06
Also lookup on ~/.ssh/config.d/* files when doing ssh host complete

rjcoelho:
Because I use 
`Include ~/.ssh/config.d/*`
on my ~/.ssh/config

Anyone knows how to put this into a FZF_SSH_CONFIG env var ?


junegunn:
Thanks, it's merged to `devel` branch for now.

PR #1439, opened 2018-11-27
Suppress the zsh autocomplete line number output

ysmood:
People normally don't use line number to filter the result when pressing CTRL-R. By default it's better to hide them if possible.

Before:

screen shot 2018-11-27 at 18 57 00

After:

screen shot 2018-11-27 at 18 59 00


junegunn:
Thanks. With this approach, new-line characters in a multi-line command are replaced to `\n`, which may or may not be desirable. You're probably right that most users don't care about the numbers, but I can't be 100% sure there's no one typing index number for picking up a command or looking at it to see if the selected command is a recent one. Please understand that, with the large number of users, changing a behavior is not an easy decision for me as it always breaks someone's workflow.

ysmood:
Understood. Then how about add an option for it?

ysmood:
So you mean we could just add some doc to the wiki?

PR #1449, opened 2018-12-01
Instead of running edit run vsplit for first file and split after

Harrian:
When opening files instead of using edit by default use vsplit if the current window number is less than two otherwise use split.


Harrian:
Closed, this change was meant for a personal fork of mine, not the original repository.

PR #1459, opened 2018-12-20
replace from ${TMPDIR:-/tmp}  to /tmp

nnao45:
Hi All.

I make PR is replace from ${TMPDIR:-/tmp}  to /tmp in some file.

I think need because, `${TMPDIR}` is **global** vars and popular name.
So...If we set `TMPDIR` vars in the another file(for example, .zshrc),
when using fzf-tmux, other software move unintentionally
(for exmaple, other software exec `rm -rf ${TMPDIR}`).

I think This `${TMPDIR}` is noneed set as vars.
And use `${TMPDIR}` in place of `/tmp` (default value).

Do you think?

Thanks.

### DIFF
--- a/bin/fzf-tmux
+++ b/bin/fzf-tmux
@@ -129,10 +129,10 @@ set -e
 
 # Clean up named pipes on exit
 id=$RANDOM
-argsf="${TMPDIR:-/tmp}/fzf-args-$id"
-fifo1="${TMPDIR:-/tmp}/fzf-fifo1-$id"
-fifo2="${TMPDIR:-/tmp}/fzf-fifo2-$id"
-fifo3="${TMPDIR:-/tmp}/fzf-fifo3-$id"
+argsf="/tmp/fzf-args-$id"
+fifo1="/tmp/fzf-fifo1-$id"
+fifo2="/tmp/fzf-fifo2-$id"
+fifo3="/tmp/fzf-fifo3-$id"
 cleanup() {
   \rm -f $argsf $fifo1 $fifo2 $fifo3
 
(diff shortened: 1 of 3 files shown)


rumpelsepp:
Why not just use `mktemp`?

junegunn:
It was three years ago so I don't quite remember why, but I intentionally changed the code to use `$TMPDIR` in 659f49a09a0e77efb60f4adcb092575c8aa58d87. Maybe there was a system, or VM or a docker container, without `/tmp`? 🤷‍♂️ 

Anyway, I found an SO answer that says we should respect `TMPDIR`. You might want to take a look.
- https://unix.stackexchange.com/a/362107/62171

> (for exmaple, other software exec rm -rf ${TMPDIR})

fzf or fzf-tmux process is usually short-lived, so I wouldn't worry too much about such scenario. And I think any program shouldn't really do `rm -rf ${TMPDIR}`. Can you describe the problem you're having in more detail?

> Why not just use mktemp?

`mktemp` itself uses `$TMPDIR` so it won't really help if another program runs `rm -rf ${TMPDIR}`.

QuLogic:
I don't see anything wrong with using `TMPDIR` for `install`; for named pipes, how about [`XDG_RUNTIME_DIR`](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html#variables)?

> `$XDG_RUNTIME_DIR` defines the base directory relative to which user-specific non-essential runtime files and other file objects (such as sockets, named pipes, ...) should be stored.

nnao45:
okay, i will

PR #1466, opened 2018-12-30
Two fixes and one cosmetics for install script

step-:
Fix: line 94: it is possible for `fzf --version` to fail with output
that satisfies awk, therefore $? is 0 in spite of the error. Fixed by
adding pipefail. Seen in attached install log. Due to previous error in
the same log.

Change: line 113 (cosmetics). Reverse `->` link notation to match that
of `ls -l`.

Fix: line 114: It is possible for the bin directory to not exist. Fixed
by adding `mkdir -p`. Seen in attached install log. Execution continued
to expose the error on line 94 (fixed above).

Attached: install log file [install-log-2018-12-30.txt](https://github.com/junegunn/fzf/files/2717556/install-log-2018-12-30.txt)

### DIFF
--- a/install
+++ b/install
@@ -91,7 +91,7 @@ ask() {
 check_binary() {
   echo -n "  - Checking fzf executable ... "
   local output
-  output=$("$fzf_base"/bin/fzf --version 2>&1 | awk '{print $1}')
+  output=$(set -o pipefail; "$fzf_base"/bin/fzf --version 2>&1 | awk '{print $1}')
   if [ $? -ne 0 ]; then
     echo "Error: $output"
     binary_error="Invalid binary"
@@ -110,8 +110,8 @@ check_binary() {
 link_fzf_in_path() {
   if which_fzf="$(command -v fzf)"; then
     echo "  - Found in \$PATH"
-    echo "  - Creating symlink: $which_fzf -> bin/fzf"
-    (cd "$fzf_base"/bin && rm -f fzf && ln -sf "$which_fzf" fzf)
+    echo "  - Creating symlink: bin/fzf ->  $which_fzf"
+    (mkdir -p "$fzf_base"/bin && cd "$fzf_base"/bin && rm -f fzf && ln -sf "$which_fzf" fzf)
     check_binary && return
   fi
   return 1


junegunn:
Thanks.

If the absence of awk causes the comparison to fail, I think we should better warn about it or, even better, remove awk dependency of the script.

> Change: line 113 (cosmetics). Reverse `->` link notation to match that
of `ls -l`.

Agreed.

> It is possible for the bin directory to not exist.

`mkdir -p` seems harmless, but I don't quite understand how it can be missing. The install script assumes that it's on this specific directory structure of the project. Can you elaborate?

step-:
Hi,

> If the absence of awk causes the comparison to fail, I think we should better warn about it or, even better, remove awk dependency of the script.

I think the absence of awk isn't the issue.  In that [line](https://github.com/junegunn/fzf/blob/315e568de006e80138f79c77d5508c7e4853e6b2/install#L94
) if `fzf --version` fails `awk` will print nothing and exit 0.  Why would `fzf --version` fail?  I don't know but the log I attached shows that it did.

> The install script assumes that it's on this specific directory structure of the project. Can you explain why this happened?

I ran `bash -x /root/bin/../share/app/install-fzf --bin` when `/root/share/app/install-fzf` existed but `/root/share/app/bin` didn't. It was part of the installation script for another app that uses fzf. I have since then changed the app installer to install the full fzf package of my distribution instead of the fzf script. So the app doesn't call `install-fzf` anymore.

Thanks

junegunn:
> I think the absence of awk isn't the issue.

Right, I think I misinterpreted your writing.

> Why would `fzf --version` fail?

When you have a corrupt binary or when you have an invalid binary for the system, e.g. linux binary on macOS, etc.

The install script will run as expected regardless of `pipefail`, because even if `$?` is zero, the subsequent version comparison will fail. But `pipefail` will yield a better error message here, so it's definitely an improvement.

> So the app doesn't call `install-fzf` anymore.

Should we drop `mkdir` part then? It's harmless, but `link_fzf_in_path` doesn't make much sense if the directory doesn't exist. The whole install script was written with the assumption of the standard directory layout, and using it out of the context doesn't seem like a great idea.

step-:
> Should we drop mkdir part then? It's harmless, but link_fzf_in_path doesn't make much sense if the directory doesn't exist. The whole install script was written with the assumption of the standard directory layout, and using it out of the context doesn't seem like a great idea.

Before I can answer, I need to apologize for having given you some misleading information - I didn't remember the details exactly myself. So this time I looked at the code I wrote months ago to make sure what it does. Indeed it calls a frozen version of `install_fzf` that is distributed with my application. That frozen (at fzf version 1.17.5) is patched with the changes that we're discussing here. You can see on [line 114](https://github.com/step-/find-n-run/blob/3235b42fd15c9b60989d522944770fe28195564d/usr/share/findnrun/install-fzf#L114) that `mkdir -p` is included because it is needed only in that particular application context. For [your install script](https://github.com/junegunn/fzf/blob/315e568de006e80138f79c77d5508c7e4853e6b2/install#L114) `mkdir -p` isn't needed, and you're free to decide what's best for your target.

My application deploys fzf on demand when the user accepts a dialog. On some distributions it installs fzf using the package manager, therefore without running `install_fzf`. On other distributions, where `fzf` isn't available as a package, my application calls the included, frozen `install_fzf` to get fzf from this GitHub repo.

PR #1474, opened 2019-01-09
Fix "bind: warning: line editing not enabled"

dyno:
Scp a file to my machine got warning like:
> .fzf/shell/completion.bash: line 33: bind: warning: line editing not enabled

### DIFF
--- a/shell/completion.bash
+++ b/shell/completion.bash
@@ -29,8 +29,8 @@ fi
 
 ###########################################################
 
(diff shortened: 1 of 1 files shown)


junegunn:
We already have the guard statement in the generated fzf.bash script.

https://github.com/junegunn/fzf/blob/master/install#L258

But this looks harmless. Do we really need `|| true` at the end?

PR #1477, opened 2019-01-14
[Bash-complete] Add a compatibility/pure fzf mode

rockandska:
Hi @junegunn ,

First, thanks for fzf, i love it.

But even if i love it, i would love it more with a better completion and that's why i spend some time to try to achieve what i've called "Compatibility Mode" and "Pure FZF mode" after reading one of your [comments](https://github.com/junegunn/fzf/pull/1128#issuecomment-342372679) in the discussion with @barlik a few days ago.

First, a simple example : 
<a href="https://asciinema.org/a/54WAaYgPgoO7K77iTB4aG5cx5?speed=2&autoplay=1"></a>

This PR remove the trigger capability but could be re-implemented if needed.

Except that, this PR add the possibility to : 
- surcharge the core autocomplete functions with fzf easily
- override core function if requested
- filter autocomplete function override by fzf ( in case there is a bug ) and keep with core function
- customize the trap function for specific functions ( _longopt is an example because it always return code 1 )
- let the users add/replace the fzf autocomplete function by their own or surcharge the defaults one
- let the users to choose the maxpdeth option used by find command when in pure fzf mode
- allow custom options for path / dir lookup

If i'm correct it also close those PR: 
- #1115 
- #1128
- #1149
- #1420 

I didn't take the time to fix the travis test for now before knowing if you would mind to see those function merged

Any comments welcome

Best regards,


rockandska:
closing in favor of https://github.com/rockandska/fzf-obc

junegunn:
Thanks for sharing, I'll take a look at it when I get some time. I think it's great that you decided to provide and maintain it as a separate project that users can optionally install. I added it to [the wiki page](https://github.com/junegunn/fzf/wiki/Related-projects).

rockandska:
thanks, any advices / comments are welcome

Issue #477, opened 2016-01-25
Ctrl-r history direct output


Zaplanincan:
Firs congratulations for make it to ArchLinux Community repo and second, I know I see that answer, but I did not make a mark (so mad at myself). How to make **_ctrl-r**_ commandline output without option to edit command and just directly start command.
Thank you and all the best, Boris

junegunn:
Haha thanks but I have no idea what's going on with ArchLinux stuff.

Anyway the feature you asked is not implemented. It would be ideal if we could "choose" if we want to edit the line or not, like pressing CTRL-X instead of Enter will fire the command immediately. This should be possible in zsh using `--expect ctrl-x` option, but I don't think the same can be easily done on bash.

naphthalene:
I would love this feature as well - if ctrl-x isn't bound to anything i also think that's an appropriate shortcut. I'll try my hand at adding this.

junegunn:
For zsh, this is what we can do:

```diff
diff --git a/shell/key-bindings.zsh b/shell/key-bindings.zsh
index ea65c0c..f1fb720 100644
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -47,12 +47,18 @@ bindkey '\ec' fzf-cd-widget
 fzf-history-widget() {
   local selected num
   setopt localoptions noglobsubst pipefail
-  selected=( $(fc -l 1 | eval "$(__fzfcmd) +s --tac +m -n2..,.. --tiebreak=index --toggle-sort=ctrl-r $FZF_CTRL_R_OPTS -q ${(q)LBUFFER}") )
+  selected=( $(fc -l 1 | eval "$(__fzfcmd) +s --tac +m -n2..,.. --tiebreak=index --toggle-sort=ctrl-r --expect=ctrl-x $FZF_CTRL_R_OPTS -q ${(q)LBUFFER}") )
   local ret=$?
   if [ -n "$selected" ]; then
+    local accept=0
+    if [[ $selected[1] = ctrl-x ]]; then
+      accept=1
+      shift selected
+    fi
     num=$selected[1]
     if [ -n "$num" ]; then
       zle vi-fetch-history -n $num
+      [[ $accept = 1 ]] && zle accept-line
     fi
   fi
   zle redisplay
```

but I couldn't find a solution for bash.

bananatranada:
[...]

PR #1492, opened 2019-02-05
[bash] Fire command directly with CTRL-X in CTRL-R

4z3:
This PR adds the ability to choose between direct execution and prior editing in CTRL-R for bash.

I'm not sure whether this implementation is acceptable as it is quite hacky and requires two dedicated readline bindings, but as demand for this feature seemed quite high, I've created a PR anyway =)

The trick used is to bind a command (to CTRL-O) that inspects the readline buffer and rebinds another binding (CTRL-P) to either accept a line or do nothing. The rebinding happens if the readline buffer contains a "magic suffix". `__fzf_history__`, in turn, adds the suffix based on the key used to make a selection. By default Enter allows the user to edit the selection, and CTRL-X will execute it immediately, i.e the same way [it was proposed for zsh](https://github.com/junegunn/fzf/issues/477#issuecomment-230338992). As [some users](https://github.com/junegunn/fzf/issues/477#issuecomment-444053054) (me included) prefer to execute by default instead, the keys to choose between execution and editing have been made configurable, e.g.:
```bash
FZF_CTRL_R_EDIT_KEY=ctrl-e
FZF_CTRL_R_EXEC_KEY=enter
```

Related #477

### DIFF
--- a/shell/key-bindings.bash
+++ b/shell/key-bindings.bash
@@ -52,18 +52,52 @@ __fzf_cd__() {
 }
 
 __fzf_history__() (
-  local line
   shopt -u nocaseglob nocasematch
-  line=$(
+  edit_key=${FZF_CTRL_R_EDIT_KEY:-enter}
+  exec_key=${FZF_CTRL_R_EXEC_KEY:-ctrl-x}
+  if selected=$(
     HISTTIMEFORMAT= history |
-    FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} $FZF_DEFAULT_OPTS --tac --sync -n2..,.. --tiebreak=index --bind=ctrl-r:toggle-sort $FZF_CTRL_R_OPTS +m" $(__fzfcmd) |
-    command grep '^ *[0-9]') &&
-    if [[ $- =~ H ]]; then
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks for your interest in the project.

Due to increasing maintenance burden, I decided not to add more options to the key bindings this repository provides. Ensuring consistent, stable user experience across three different shells – bash, zsh, and fish – turned out to be quite painful and has taken up much of my time.

So instead of having this merged to this repository, I suggest that you create a separate "plugin" with extended features and options, so that the users can optionally install your bindings if they want.

```sh
[ -f ~/.fzf.bash ] && source ~/.fzf.bash

# Load your plugin to override the bindings
```

markmcd:
If you turned this into a plugin, I'd love to use it. Could you share a link here so other Google-searchers can find it?

4z3:
This PR's code has been copied to [4z3/fzf-plugins](https://github.com/4z3/fzf-plugins).

nhooyr:
This was quite easy in zsh.

See https://github.com/nhooyr/dotfiles/blob/92623b7a3e9fa421d191f6c2bb77de4ce60aa367/zsh/fzf.zsh#L48-L61

Issue #1486, opened 2019-01-29
[Ubuntu] --preview error : Failed to read /dev/tty


erch:
- Category
    - [x ] fzf binary
    - [ ] fzf-tmux script
    - [ ] Key bindings
    - [ ] Completion
    - [ ] Vim
    - [ ] Neovim
    - [ ] Etc.
- OS
    - [x ] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Windows Subsystem for Linux
    - [ ] Etc.
- Shell
    - [ x] bash
    - [ ] zsh
    - [ ] fish

When I use --preview I got the error message: Failed to read /dev/tty
```bash
$ echo "hello" | fzf
hello
$ echo "hello" | fzf --preview 'echo {}'
Failed to read /dev/tty
$ echo $TERM
xterm-256color
$ tty
/dev/pts/6
$ fzf --version
0.17.5 (b46227d)
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 18.04.1 LTS
Release:	18.04
Codename:	bionic
```

winvicta:
I have just run into the same problem. The problem occurred just after a system update. System is ubuntu 18.04 LTS. Before the system update, everything worked flawlessly.
Symptoms: 
1. When entering ALT-C (default binding), system responded by printing the message "Failed to read /dev/tty".
2. When entering /dev/tty, system responded by printing "bash: /dev/tty: Permission denied"
3. When entering tty, system responded by printing "/dev/pts/1"

"sudo chmod 777 /dev/tty" won't make a difference. uninstalling/reinstalling won't make a difference.

The other commands CTRL-R and CTRL-T work without any problems.

Any help will be appreciated. Thanks.

bartels:
I'm seeing this too. It appears to have started after an Ubuntu 18.04 kernel update to `4.15.0-44`. Booting back into the previous `4.15.0-43` kernel and the error does not happen.

ianchanning:
* fzf v0.17.4 
* Ubuntu 18.04
* `uname -a`: `Linux ICT0115123 4.15.0-44-generic #47-Ubuntu SMP Mon Jan 14 11:26:59 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux`

I'm hitting this problem too, it breaks the fzf.vim plugin. I get the following error in Vim:

```
[...]

PR #1496, opened 2019-02-07
Handle EAGAIN properly

ethomas2:


junegunn:
Thanks. I believe you're running into #1486. It's a Kernel regression so I'd rather not a merge temporary fix. The patch is not needed because we should not see EAGAIN on blocking read. `Failed to read /dev/tty` error only occurs during blocking read.

- https://github.com/junegunn/fzf/blob/0.17.5/src/tui/light.go#L155
- https://github.com/junegunn/fzf/blob/0.17.5/src/tui/light.go#L291-L294

ethomas2:
Oh sorry, I didn't even mean to make this pull request against this repo yet. I thought I made a pr from a branch in my fork to master in my fork. I wanted to include a pr description before making a pull request, and maybe clean this up a little. 

You're correct that #1486 describes my issue. Thanks for the fast response.

PR #1510, opened 2019-02-27
Document CTRL-D shortcut to exit

nsauzede:


junegunn:
Thank you, but CTRL-D acts a little differently – it deletes a character on the cursor, or close fzf if the prompt is empty, just like it does on your shell and any other programs that support standard readline key bindings. So mentioning it with the other keys can be misleading. You can read more about the default key bindings is `man fzf`.

nsauzede:
Ah, OK, nevermind..

PR #1528, opened 2019-03-13
[vim] Respect user settings to disable mouse

### LINKED ISSUE #1527: fzf.vim does not respect user settings to disable mouse
- Category
    - [ ] fzf binary
    - [ ] fzf-tmux script
    - [ ] Key bindings
    - [ ] Completion
    - [x] Vim
    - [ ] Neovim
    - [ ] Etc.
- OS
    - [ ] Linux
    - [x] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [x] bash
    - [ ] zsh
    - [ ] fish

I disable my mouse everywhere:
1. Vim `mouse` option is the default empty string
2. `~/.bash_profile` sets `export FZF_DEFAULT_OPTS='--no-mouse'`
3. `~/.tmux.conf` has `set -g mouse off`

When I called `:Files` from Vim, I was surprised that fzf was capturing my mouse. I believe that it should pass the `--no-mouse` option if numbers 1 or 2 above are true.

I've installed fzf via Homebrew. My `.vimrc` contains:
```vim
Plug '/usr/local/opt/fzf' | Plug 'junegunn/fzf.vim'
  nnoremap <C-p> :Files<CR>
```

m1foley:
Fix #1527

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -405,6 +405,11 @@ try
   elseif use_term
     let optstr .= ' --no-height'
   endif
+
+  if &mouse !=# 'a' || stridx($FZF_DEFAULT_OPTS, '--no-mouse') != -1
+    let optstr .= ' --no-mouse'
+  endif
+
   let command = prefix.(use_tmux ? s:fzf_tmux(dict) : fzf_exec).' '.optstr.' > '.temps.result
 
   if use_term


junegunn:
Thanks, but I think you're confused somehow. fzf always respects FZF_DEFAULT_OPTS, so having `--no-mouse` in it should do the trick. Try this:

```sh
FZF_DEFAULT_OPTS='--no-mouse' vim +Files
```

Forcing `--no-mouse` when mouse is disabled on Vim may or may not be desirable depending on the use case. Mouse on fzf can be quite useful for scrolling the preview window, so you may still want it. And if you don't, you can still disable it by setting FZF_DEFAULT_OPTS, right? So this is a bit more flexible.

m1foley:
You're right about `FZF_DEFAULT_OPTS` already working -- adding that to my `.bash_profile` fixed it. I must have forgotten to reload my terminal earlier.

I disagree re: Vim `mouse` setting, but it's not a big deal because there's an easy workaround like you said. Thanks for being patient and kind, and for maintaining this great tool! 🙏

Issue #577, opened 2016-05-23
Long commands aren't horizontally scrollable


BBCMarkMcDonnell:
- Category
    - [x] fzf binary
    - [ ] fzf-tmux script
    - [ ] Key bindings
    - [x] Completion
    - [ ] Vim
    - [ ] Neovim
    - [ ] Etc.
- OS
    - [ ] Linux
    - [x] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [x] bash
    - [ ] zsh
    - [ ] fish

<img width="1436" alt="Long commands aren't horizontally scrollable" src="https://cloud.githubusercontent.com/assets/15904246/15468445/323b4e0a-20dc-11e6-9991-74becdc6d1b9.png">

I can't seem to be able to scroll horizontally along any results that are longer that my terminal screen. The example image above shows many different `curl` requests that _look_ the same, but in fact point to different URLs so it's impossible to know which one I want.

edi9999:
If you start matching on the right, (in your example, by typing "component"), you can view more (but there is no way to bind a key to scroll to the right for you)

junegunn:
@edi9999 Thanks for chiming in.

> If you start matching on the right, (in your example, by typing "component"), you can view more

Yes that's what I would do. A large `--hscroll-off` value in `$FZF_CTRL_R_OPTS` can definitely help (the default is 10).

But anyway I don't have a great answer to this. Adding support for manual horizontal scrolling is the last thing I want to do as getting it right and implementing a clean, straightforward interface is not going to be trivial (manual scroll vs. auto scroll, short item vs. long item)

When the above trick doesn't work you, you might want to fall back to the original "reverse-search-history" that can display the command in multiple lines by binding it to a different key sequence (e.g. `bind '"\C-x\C-r": reverse-search-history'`), or simply use <kbd>CTRL-P</kbd> or <kbd>CTRL-N</kbd> to browse the recent entries on bash.

Closing as "wontfix" (at least for the moment). Thanks.

Integralist:
No problem. Thanks for the feedback everyone :-)

junegunn:
With the latest fzf, we have these options:

```sh
# always-visible preview window that shows full command
export FZF_CTRL_R_OPTS="--preview 'echo {}' --preview-window down:2"

# hidden by default, made visible on `?`
export FZF_CTRL_R_OPTS="--preview 'echo {}' --preview-window down:3:hidden --bind ?:toggle-preview"
```

[...]

PR #1551, opened 2019-04-08
key-bindings.bash: better support multiline commands

telotortium:
This makes it possible to see multiple lines of a multi-line command with the C-r keybinding when using `FZF_CTRL_R_OPTS` to provide a multi-line preview window, as done in https://github.com/junegunn/fzf/issues/577#issuecomment-473241837.

### DIFF
--- a/shell/key-bindings.bash
+++ b/shell/key-bindings.bash
@@ -55,21 +55,44 @@ __fzf_history__() (
   local line
   shopt -u nocaseglob nocasematch
   line=$(
-    # Use a random number as a unique marker in the output of `history` below in
-    # order to distinguish continuations of multiline commands.
-    rnd=$RANDOM
-    HISTTIMEFORMAT="_${rnd}_" history | gawk '
+    HISTTIMEFORMAT="" history | awk '
     # Terminate history entries with ASCII NUL instead of newline. This
     # requires parsing the output of `history` to determine which lines
     # are continuations of multiline commands.
-    BEGIN { line = ""; re = "^( *[0-9]+ *)_'"${rnd}"'_(.*)$"; }
+    BEGIN {
+        cmd = "";   # Content of command currently being read from history
+        idx = 1;  # Index of next history item to be read from the input
+        # Format of lines containing the beginning of commands in history;
+        # commands continued from the previous line will not start with this.
+        # Matches everything before the command.
+        #
+        # Bash always outputs 2 spaces between the date and the command.
+        re = "^ *[0-9]+  ";  # Yes, two spaces at the end is guaranteed.
+    }
     {
+        # Check that the current line starts with the correct format.
         if (match($0, re) != 0) {
-            if (NR != 1) { printf("%s\0", line); }
-            line = gensub(re, "\\1\\2", "g");
-        } else {
-            line = line "\n" $0;
+            # Extract the index from the history line
+            prefix = substr($0, RSTART, RLENGTH)
+            match(prefix, "[0-9]+");
+            idx_read = substr(prefix, RSTART, RLENGTH);
+
+            # Compare index we are expecting to read to index read from line.
+            if (idx == idx_read) {
+                # Update next index to read from input.
+                idx++;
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks, but there are a few points against this approach.

1. We can't assume that the system has `gawk` installed. Mine didn't, and I had to install it using `brew`.
    ```
    `__fzf_history__`-bash: gawk: command not found
    ```
2. We could add a conditional branch, yes, but I'd really like to keep the implementation simple and minimal so that the users can take a look at it and quickly understand how it works.
3. For some reason, setting `HISTTIMEFORMAT` makes `history` much slower, that almost makes `CTRL-R` unusable for me. Tested on bash 3.2 and 5.0.
    ```sh
    > echo $BASH_VERSION
    5.0.3(1)-release
    
    > time HISTTIMEFORMAT="" history | wc -l
       96513
    
    real    0m0.220s
    user    0m0.129s
    sys     0m0.201s
    
    > time HISTTIMEFORMAT="foo" history | wc -l
       96514
    
    real    0m16.960s
    user    0m2.827s
    sys     0m14.310s
    ```
    This could be a bug of bash, but whatever the reason may be, we can't ship this version.

telotortium:
Hi, I've patched the code to address your concerns. PTAL.

junegunn:
Thanks for the update, but I'm still not sure if I want it because of the additional delay.

```sh
jg@mbp2018:/private/tmp> time HISTTIMEFORMAT="" history > /dev/null

real    0m0.128s
user    0m0.090s
sys     0m0.038s
jg@mbp2018:/private/tmp> time HISTTIMEFORMAT="" history | awk -f proc.awk > /dev/null
awk: proc.awk:16: (FILENAME=- FNR=88613) warning: Invalid multibyte data detected. There may be a mismatch between your data and your locale.

real    0m0.338s
user    0m0.417s
sys     0m0.050s
```

The difference is quite noticeable to me. Since I have very few multi-line commands in my command history, I do not like that I have to wait another 100ms every time I hit CTRL-R for no real benefit. So this comes down to personal preference. If you use multi-line commands frequently, I can imagine that you would want proper support for multi-line commands even if you have to wait a little longer, but that's not the case for me.

Also, I'm seeing some issues.

```
jg@mbp2018:~/.vim/plugged/fzf/shell (master)> echo "foo
> bar"
foo
bar
jg@mbp2018:~/.vim/plugged/fzf/shell (master)> echo "foo
> bar
> baz"
foo
bar
baz
jg@mbp2018:~/.vim/plugged/fzf/shell (master)> `__fzf_history__`awk: cmd. line:17: (FILENAME=- FNR=88613) warning: Invalid multibyte data detected. There may be a mismatch between your data and your locale.
─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
  97342  ### ###.###
  97343  #### ####
  97344  #### #######
  97345  ####
  97346  #### #######
  97347  #### #######
  97348  ### ######.###
  97349  ## #########/
  97350  ## ..
  97351  ### ######.###
  97352  ### ######.###
  97353  ## -## ######/
  97354  ## ###
  97355  ..
  97356  ####### ######/######.###
  97357  ## ######/
  97358  ##
  97359  #### .
  97360  #### __###_#######__
  97361  echo "foo bar"
>
  Press CTRL-Y to copy command into clipboard
  97362/97362 +S
>
─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
```

Having said that I'm leaning towards not merging it to the main repository and publishing the code as a plugin/extension in a separate repository (see my previous comment on that matter: https://github.com/junegunn/fzf/pull/1161#issuecomment-347002445).

telotortium:
Yes, that seems like a good idea to me. In any case, I just found out that macOS awk doesn't allow you to print out NUL at all, so I'd have to either use gawk anyway or implement some more complicated escaping scheme. Is there a particular format for the "plugins" that I should follow?

junegunn:
Fixed in #1837

PR #1562, opened 2019-04-29
Add FZF_COMPLETE_{DIR,PATH,PID}_FOR

### LINKED ISSUE #1560: Provide way for users to add commands for autocomplete
Hmm this is effectively a dup of #536 but since that is closed and I can't reopen a closed issue in GitHub (unlike Bugzilla) I am filing this separately.

- Category
    - [x] Completion
- OS
    - [x] Linux
- Shell
    - [x] bash

Hello. I'm new to `fzf` and tried to use `**` to trigger it while trying at the terminal to open a file using `kate` which is my favourite editor. However it did not work.

I was able to dig in and find that only commands that are given under `a_cmds` in `fzf/shell/completion.bash` are completed this way. So I added `kate` and a few other (like `nano`, `clang` since I prefer it to `gcc` which is included, `python3` apart from existing `python` since I use only Py 3) to this variable and got the completion working.

However, this is not optimal since if there is a new `fzf` release and I re-pull the repo, the changes would get overwritten or I have to maintain a separate branch.

It would be better if there is some way like an environment variable (which can be set under `.bashrc`) or config file where one can add one's own custom commands for which one would like `fzf` `**` completion.

In #536 I see you suggesting to add a `complete` line to `.bashrc` but this method (preferably envvar) would be much more easier. Something like `FZF_COMPLETE_FOR='clang clang++ kate nano'` which would then get included at the end of `a_cmds`.

BTW I think you should at least include `nano` and `python3` in the basic list since they are so ubiquitous and used by users of all desktop environments.

jamadagni:
Provide variables FZF_COMPLETE_DIR_FOR, FZF_COMPLETE_PATH_FOR and FZF_COMPLETE_PID_FOR for user convenience. Fixes #1560.

### DIFF
--- a/README.md
+++ b/README.md
@@ -434,15 +434,16 @@ _fzf_compgen_dir() {
 }
 ```
 
-#### Supported commands
+#### Commands supported on bash
 
 On bash, fuzzy completion is enabled only for a predefined set of commands
-(`complete | grep _fzf` to see the list). But you can enable it for other
-commands as well as follows.
+(do `complete | grep _fzf` to see the list). But you can enable it for other
+commands as well by editing the following variables in `~/.fzf.bash`:
 
 ```sh
-complete -F _fzf_path_completion -o default -o bashdefault ag
-complete -F _fzf_dir_completion -o default -o bashdefault tree
+FZF_COMPLETE_PATH_FOR='clang clang++ ag'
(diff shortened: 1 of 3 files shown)


junegunn:
Related: https://github.com/junegunn/fzf/commit/9d842630c931ae08f6f16f902d7b0ab42b10fda2

Issue #1170, opened 2017-12-02
Fuzzy completion for custom commands is broken.


gregorias:
- Category
    - [ ] fzf binary
    - [ ] fzf-tmux script
    - [ ] Key bindings
    - [x] Completion
    - [ ] Vim
    - [ ] Neovim
    - [ ] Etc.
- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Windows Subsystem for Linux
    - [ ] Etc.
- Shell
    - [x] bash
    - [ ] zsh
    - [ ] fish

**Steps to reproduce and behaviour**

1. Run `complete -F _fzf_path_completion -o default -o bashdefault fake_cmd`
2. Type `fake_cmd any_dir<TAB><TAB>`

Now typing `fake_cmd **<TAB><TAB>` stops working and returns default tab completion.

**Root cause**

In step 2. the completion function for fake_cmd is changed to bash' default - `_minimal`.

Before step 2., `complete | grep fake_cmd` outputs `complete -F _fzf_path_completion -o default -o bashdefault fake_cmd`, but after step 2., the command outputs `complete -F _minimal fake_cmd`.

This change happens, when `_completion_loader` is run at: https://github.com/junegunn/fzf/blob/9615c4edf13b0cedbc8c9b1c22bac905e12de8af/shell/completion.bash#L125
L125 affects all other commands, however the original completion function is restored later - https://github.com/junegunn/fzf/blob/9615c4edf13b0cedbc8c9b1c22bac905e12de8af/shell/completion.bash#L127-L128 
This restoration doesn't affect fake_cmd, because it is not in `a_cmds`: https://github.com/junegunn/fzf/blob/9615c4edf13b0cedbc8c9b1c22bac905e12de8af/shell/completion.bash#L265

Note that this if the `fake_cmd` is inside `/etc/bash_completion.d/`, then `_completion_loader` doesn't fail, and the switch doesn't happen.

junegunn:
Thanks for the detailed report. Let me see if I can find a way to fix it.

gregorias:
I made `a_cmds` variable configurable in https://github.com/gregorias/fzf/commit/3fc31ac3b6a15a1c18c292dae841954216de4ebe and use it in my personal config: I presume that you might want something different for a wider audience.

junegunn:
I believe 620f0649d4e01766115a994bad7cf381d5942134 fixes the issue (it's not in the master branch).

[...]

PR #1564, opened 2019-04-30
[bash-completion] Fix custom completion for non-standard command names

jbriales:
Related to [previous issue](https://github.com/junegunn/fzf/issues/1170).
[Previous solution](https://github.com/junegunn/fzf/commit/d6588fc835aff6d952d86b1bdc5ed7378a846ab9#diff-c16a25d47a578fa3d03aa4328a4f59b0)
wouldn't work for commands with non-standard names, where `$cmd` and `$orig_cmd` differ,
e.g. `s.foo` -> `s_foo`


junegunn:
Looks good to me, thanks!

PR #1571, opened 2019-05-06
Remove allow_legacy variable in install script

Dom-R:
Just removing an unused variable that I detected when analyzing the install script.


junegunn:
Thanks!

PR #1579, opened 2019-05-15
Make fzf conditionally respect old zsh key bindings

arturoescaip:
I'm annoyed that fzf overrides some key-bindings unconditionally. I find the default bindings of `C-r`, `C-t` and `M-c` very useful. I also think fzf is very useful. This pull request mixes both behaviors. I've been using this for a while and I'm very happy with it.

Specifically, these changes make fzf respect old zsh key-bindings (if any) in certain situations:
- `C-t`, `M-c`: If the cursor is not surrounded by whitespace, then use old binding.
- `C-r`: only use fzf if the command line is empty or only whitespace

In all cases, if old bindings were not set, fzf is used.

### DIFF
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -103,9 +103,11 @@ bindkey '\ec' fzf-cd-widget
 # CTRL-R - Paste the selected command from history into the command line
 fzf-history-widget() {
   local oldbinding=$__fzf_old_binding[^R]
+  setopt local_options extended_glob
 
-  # If the cursor is not at the start of the line, then fall back to the old key binding.
-  if [[ -n $oldbinding ]] && (( CURSOR )); then
+  # If the command line is empty or only whitespace, use fzf. Otherwise fall
+  # back to old binding.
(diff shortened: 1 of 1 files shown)


arturoescaip:
Any chance this anyone can look at this?

junegunn:
I would suggest the following behaviors instead:

- `C-t`: Use old binding only when the cursor is surrounded by *alphanumeric characters*
    - Because I can imagine a few cases where you want to hit CTRL-T to paste the selection after non-alphanumeric characters
        - `vim "▒`
        - `echo $(pwd)/▒`
- `C-r`: Keep it unchanged. We currently pass the current command line as `--query`. If we hit `C-r` after typing `git log`, fzf starts with `git log` as the initial query.
- I'm not sure what would be the best strategy for `M-c`. The original binding works even when the cursor is on a whitespace.
    - `foo   ▒   bar`
        - `M-c` capitalizes `bar`
    - We should note that we're actually breaking backward compatibility for those who expect `M-c` to clear whatever's on the command line and start directory selector.

PR #1604, opened 2019-06-12
Update fzf-tmux

jiahut:
In my case I have defined  `alias rm=~/bin/safe-rm` which move it into trash, `fzf-tmux` generate temporary file which can safe be deleted immediate. so add hook `FZF_CLEANUP_CMD` which default is  `/bin/rm`


junegunn:
Notice `\` prefix of `\rm`. It makes bash ignore any aliases, so I think you're confused somehow.

jiahut:
thinks @junegunn , I found the issue  that  there is an other `rm` cli  at  `$PATH`  before `/bin/rm`

PR #1611, opened 2019-06-20
Enable PIDs completion to lsof, strace and gdb

tonyseek:


junegunn:
Thanks, I've given thought about this, but it's not clear if we should launch PID completion or some other completion in different circumstances. So I think a better option would be to create a dedicated key binding (e.g. CTRL-G + CTRL-P) for PID completion like so:

```sh
# bash
if [[ $- =~ i ]]; then
  gp() {
    ps -ef | fzf --height 50% --border --header-lines 1 --info inline --layout reverse --multi |
      awk '{print $2}'
  }

  bind '"\er": redraw-current-line'
  bind '"\C-g\C-p": "$(gp)\e\C-e\er"'
fi
```

PR #1616, opened 2019-06-22
Use the vim-specific shell instead of the environment variable

AdminXVII:
This will allow to use a specific shell for vim, without needing to change the $SHELL environment variable too.

The shell option is set by default to $SHELL, so it should not matter for users with standard setup.

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -367,7 +367,7 @@ try
   if !has_key(dict, 'source') && !empty($FZF_DEFAULT_COMMAND) && !s:is_win
     let temps.source = s:fzf_tempname()
     call writefile(s:wrap_cmds(split($FZF_DEFAULT_COMMAND, "\n")), temps.source)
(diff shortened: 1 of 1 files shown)


plugin/fzf.vim
```diff
@@ -367,7 +367,7 @@ try
   if !has_key(dict, 'source') && !empty($FZF_DEFAULT_COMMAND) && !s:is_win
     let temps.source = s:fzf_tempname()
     call writefile(s:wrap_cmds(split($FZF_DEFAULT_COMMAND, "\n")), temps.source)
-    let dict.source = (empty($SHELL) ? &shell : $SHELL).' '.fzf#shellescape(temps.source)
+    let dict.source = &shell.' '.fzf#shellescape(temps.source)
```
janlazo: `&shell` is set to `sh` so there's no user customization here. You could set/unset `$SHELL` instead.

PR #595, opened 2016-06-11
optimize fzf_default_completion binding

aykamko:
I spent the past few hours doing a bit of yak shaving on speeding up my `zsh` startup time and discovered that one of the biggest bottlenecks was this chunk of code in `shell/completion.zsh`! 😮 

The slowdown is a result of launching the `grep` and `awk` programs to set the `fzf_default_completion` environment variable. I rewrote the code to use only zsh built-ins and got a reasonable speedup! Here are the results:
```
All times in microseconds (us), 100 samples each
=====
Control       -- Mean: 9035.12, Stddev: 326.39
Experiment    -- Mean: 5979.47, Stddev: 372.748
```

[Here is a gist of the script I used](https://gist.github.com/anonymous/3ecef48effacc72f205ed40a58b3112d) to profile the control (i.e. `junegunn master`) vs the experiment (i.e. this PR).

FYI: here are the 25 slowest lines evaluated via `source ~/.fzf.zsh` on `junegunn master` in descending order. Times are in milliseconds here, not microseconds.
```
04.17709351: /Users/Aleks/.fzf/shell/completion.zsh:185> fzf_default_completion=expand-or-complete-with-indicator 
00.76699257: /Users/Aleks/.fzf/shell/completion.zsh:23> declare -f _fzf_compgen_dir
00.33903122: /Users/Aleks/.fzf.zsh:19> source /Users/Aleks/.fzf/shell/key-bindings.zsh
00.22387505: /Users/Aleks/.fzf.zsh:15> source /Users/Aleks/.fzf/shell/completion.zsh
00.18501282: /Users/Aleks/.zshrc:311> source /Users/Aleks/.fzf.zsh
[...]

PR #1621, opened 2019-06-25
[vim] Use keepjump

foonathan:
I use FZF with 'window': 'enew'. When FZF switches back, the current position in the FZF buffer will be added to the jumplist. When I then want to jump back to the position I was in before I started FZF, it is a buffer that no longer exists, so I end up at the beginning of the current file for some reason.

By adding 'keepjump' when switching back, the position is not added anymore. I believe this also fixes #595, at least for me.

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -664,7 +664,7 @@ function! s:execute_term(dict, command, temps) abort
   function! fzf.switch_back(inplace)
     if a:inplace && bufnr('') == self.buf
       if bufexists(self.pbuf)
-        execute 'keepalt b' self.pbuf
+        execute 'keepalt keepjump b' self.pbuf
       endif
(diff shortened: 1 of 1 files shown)


foonathan:
Any update on this?

junegunn:
I can't reproduce the problem.

```sh
# Start Vim with a minimal vimrc
vim -Nu <(curl https://gist.githubusercontent.com/junegunn/6936bf79fedd3a079aeb1dd2f3c81ef5/raw)
```

Then with,

```vim
let g:fzf_layout = { 'window': 'enew' }
```

`:FZF` and selected a file, but no such problem.

foonathan:
Oh, you're right, the problem was a `keepjump` command in my custom file sink...

Sorry for bothering you :D

PR #1634, opened 2019-07-09
Fix name-based colours for gvim/8.0 without the in-built terminal

charlton1:
Name-based colour specifications for guibg and guifg end up trying to call xterm with spurious backslashes. While this works for hex colours, it causes issues with the name-based colours. This removes the backslash in the format string and, instead, escapes the pound character as needed (i.e. if it is present).


junegunn:
Thanks!

charlton1:
Thanks! Thanks for making and maintaining this plugin.

PR #1635, opened 2019-07-13
Remove perl from fish key bindings.

ssjhv:
Perl was used to remove the trailing newline character, but fzf already
has --print0 to use null character as terminators, and fish read -z is
expecting null character as terminators. There is no reason to depend on
perl if --print0 is passed to fzf invocation.


junegunn:
Good point, merged, thanks!

PR #1640, opened 2019-07-19
Simplified code in various places

muesli:
- Replaced time.Now().Sub() with time.Since()
- Replaced unnecessary string/byte slice conversions
- Removed obsolete return and value assignment in range loop


junegunn:
Looks good, thanks!

muesli:
No problem, thanks for merging!

PR #1654, opened 2019-07-31
Add the 'shell' option to the run function

ian-howell:
This allows a user to specify which shell to use for the command that
gets piped into fzf. This is valuable if the user wants to use non-POSIX
compliant features such as process redirection.

Possibly related to #1616

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -333,24 +333,24 @@ function! fzf#wrap(...)
   return opts
 endfunction
 
-function! s:use_sh()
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks for the pull request, but I don't think we need it as it's already possible to use a different shell by specifying it in the command.

```vim
call fzf#run({'source': 'zsh -c '.shellescape('echo $ZSH_VERSION')})
```

Let me know if there is a case where this is absolutely necessary, otherwise, I'd like to keep the feature set small.

PR #1667, opened 2019-08-15
$(...) calls should be quoted: "$(...)"

vandenoever:
`vim $(fzf)` will not open files that have a space in the name. To fix that, use double quotes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ to install fzf.
 brew install fzf
 
 # To install useful key bindings and fuzzy completion:
-$(brew --prefix)/opt/fzf/install
+"$(brew --prefix)/opt/fzf/install"
 ```
 
 fzf is also available [via MacPorts][portfile]: `sudo port install fzf`
@@ -244,7 +244,7 @@ files excluding hidden ones. (You can override the default command with
 `FZF_DEFAULT_COMMAND`)
 
 ```sh
-vim $(fzf)
+vim "$(fzf)"
 ```
 
 #### Using the finder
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -89,7 +89,7 @@ to install fzf.
 brew install fzf
 
 # To install useful key bindings and fuzzy completion:
-$(brew --prefix)/opt/fzf/install
+"$(brew --prefix)/opt/fzf/install"
```
LuRsT: I'm unsure about this one :thinking:

leonklingele: The brew prefix path might include spaces, so strictly speaking it _must_ be quoted (no one ever uses a non-default brew prefix, so yeah..)

LuRsT: Makes sense!

vandenoever:
Looks good

junegunn:
I assumed that most of the users already know the implication of omitting quotes – files paths with spaces are quite rare on Unix systems, so we casually omit quotes when typing commands, and put them only when they are required. The commands in the README page are just simple examples (except for `brew` one), so I decided to omit quotes to have less visual clutter. But I don't have a strong opinion on how we should present them.

vandenoever:
On my system they are not rare. I do not think I'm an exception.

```bash
$ find ~ -name '* *' | wc -l
13637
```

Since omitting quotes can lead to bugs, it's always good to quote. shellcheck is a handy tool to remind of that.

LuRsT:
How about a happy medium, where we consistently don't use quotes, but have a note somewhere stating that using quotes is necessary for files with spaces?

PR #1668, opened 2019-08-15
Added ed, Gnu ed and Open ed Support

bjoernwibben:
Just added Autocompletion when using ed.


junegunn:
Thanks, but it has become clear to me that I can't keep adding all the commands people use to the list. Instead, fzf now provides a helper function, undocumented yet, that lets you easily set up fuzzy completion for any command.

Since 50958992455708f4a9ec82f745f7496fbca243c7, you can do something like this:

```sh
_fzf_setup_completion path ed sed ged oed
```

bjoernwibben:
Cool, thanks for the hint. :-)

Issue #963, opened 2017-07-06
Character Set Not Supported


Chinggis6:
- Category
    - [x] fzf binary
    - [ ] fzf-tmux script
    - [ ] Key bindings
    - [ ] Completion
    - [ ] Vim
    - [ ] Neovim
    - [ ] Etc.
- OS
    - [ ] Linux
    - [ ] Mac OS X
    - [x] Windows
    - [ ] Windows Subsystem for Linux
    - [ ] Etc.
- Shell
    - [ ] bash
    - [ ] zsh
    - [ ] fish
    - [x] cmd.exe
    - [x] powershell.exe
- Terminal
    - [x] hyper

apart from the above

   character set not supported

I installed `fzf` through [chocolatey](https://github.com/chocolatey) on Windows 10.
Issuing the following failed:

    chcp 65001

as I found [here](https://stackoverflow.com/questions/388490/unicode-characters-in-windows-command-line-how).

janlazo:
Unset TERM before running fzf.

```dosbatch
> echo %TERM%
xterm-256color
> fzf
character set not supported
> set "TERM="

> fzf
```

janlazo:
``` dosbatch
> echo %TERM_PROGRAM%
Hyper
```

cmd.exe, powershell.exe, sh.exe, and mintty.exe don't use it and none of them set it to Hyper so fzf can internally unset TERM like in cygwin.

Chinggis6:
It didn't run on Cygwin after unsetting TERM hanging forever, though I could run it on Git Bash after setting `TERM=`.

I think plain `cmd.exe` is the place where most people would run it.

janlazo:
`xterm-256color` is hardcoded in [app/session.js](https://github.com/zeit/hyper/blob/aa7e79a0399fddefe67c248a0d5a9b13bb27659b/app/session.js#L26-L31)

```javascript
const baseEnv = Object.assign({}, process.env, {
      LANG: app.getLocale().replace('-', '_') + '.UTF-8',
      TERM: 'xterm-256color',
      TERM_PROGRAM: productName,
      TERM_PROGRAM_VERSION: version
}, envFromConfig);
```

Chinggis6:
Uncommenting (prepend `//`) the 28th line in `%localappdata%\hyper\app-1.3.3\resources\app\session.js` on Windows solved the issue. Thanks for the help.

janlazo:
[...]

PR #1677, opened 2019-08-31
Add MSYS2 support as a vim plugin

msr1k:
Add &shellcmdflag and TERM environment variable treatment.

- Make &shellcmdflag `/C` when &shell turns into `cmd.exe`
- Delete %TERM% environment variable before fzf execution

On MSYS2 environment, vim is typically configured as follows:

- &shell: bash
- &shellcmdflag: -c
- $TERM(%TERM%): xterm or something

Since this plugin does not change &shellcmdflag while it replaces &shell with cmd.exe,
vim on MSYS2 environment cannot start fzf properly.
(It try to do `cmd.exe -c something...` but it should be `cmd.exe /C something...`.)

And at the time of starting cmd.exe %TERM% is no longer needed.
So delete this environment variable in the batch file to be executed.

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -50,7 +50,9 @@ if s:is_win
   " Use utf-8 for fzf.vim commands
   " Return array of shell commands for cmd.exe
   function! s:wrap_cmds(cmds)
-    return map(['@echo off', 'set TERM= > nul', 'for /f "tokens=4" %%a in (''chcp'') do set origchcp=%%a', 'chcp 65001 > nul'] +
+    return map(['@echo off'] +
+          \ (has('gui_running') ? ['set TERM= > nul'] : []) + 
+          \ ['for /f "tokens=4" %%a in (''chcp'') do set origchcp=%%a', 'chcp 65001 > nul'] +
           \ (type(a:cmds) == type([]) ? a:cmds : [a:cmds]) +
           \ ['chcp %origchcp% > nul'], 'v:val."\r"')
   endfunction
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -77,7 +77,7 @@ function! s:shellesc_cmd(arg)
 endfunction
 
 function! fzf#shellescape(arg, ...)
-  let shell = get(a:000, 0, s:is_win ? 'cmd.exe' : &shell)
+  let shell = get(a:000, 0, s:is_win ? 'cmd.exe' : 'sh')
   if shell =~# 'cmd.exe$'
     return s:shellesc_cmd(a:arg)
   endif
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -50,7 +50,7 @@ if s:is_win
   " Use utf-8 for fzf.vim commands
   " Return array of shell commands for cmd.exe
   function! s:wrap_cmds(cmds)
-    return map(['@echo off', 'for /f "tokens=4" %%a in (''chcp'') do set origchcp=%%a', 'chcp 65001 > nul'] +
+    return map(['@echo off', 'set TERM= > nul', 'for /f "tokens=4" %%a in (''chcp'') do set origchcp=%%a', 'chcp 65001 > nul'] +
           \ (type(a:cmds) == type([]) ? a:cmds : [a:cmds]) +
           \ ['chcp %origchcp% > nul'], 'v:val."\r"')
   endfunction
@@ -334,19 +334,20 @@ function! fzf#wrap(...)
 endfunction
 
 function! s:use_sh()
-  let [shell, shellslash] = [&shell, &shellslash]
+  let [shell, shellslash, shellcmdflag] = [&shell, &shellslash, &shellcmdflag]
   if s:is_win
     set shell=cmd.exe
+    set shellcmdflag=/c
     set noshellslash
   else
     set shell=sh
   endif
-  return [shell, shellslash]
+  return [shell, shellslash, shellcmdflag]
 endfunction
 
 function! fzf#run(...) abort
 try
-  let [shell, shellslash] = s:use_sh()
+  let [shell, shellslash, shellcmdflag] = s:use_sh()
 
   let dict   = exists('a:1') ? s:upgrade(a:1) : {}
   let temps  = { 'result': s:fzf_tempname() }
@@ -416,7 +417,7 @@ try
   call s:callback(dict, lines)
   return lines
 finally
-  let [&shell, &shellslash] = [shell, shellslash]
+  let [&shell, &shellslash, &shellcmdflag] = [shell, shellslash, shellcmdflag]
 endtry
 endfunction
 
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -75,7 +75,7 @@ function! s:shellesc_cmd(arg)
 endfunction
 
 function! fzf#shellescape(arg, ...)
-  let shell = get(a:000, 0, &shell)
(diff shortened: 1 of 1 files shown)


msr1k:
junegunn/fzf.vim uses fzf#shellescape function.

But when it calls this function, &shell is not changed to `cmd.exe`
and shellescaping is executed not for `cmd.exe` through it is eventually passed to `cmd.exe`.

So I changed shell default value &shell to `cmd.exe` when s:is_win is true at c95d52fa0cbdacecda4dc8c0aaa7d0515f5e18ad.

junegunn:
@janlazo Hi, can you take a look at this?

plugin/fzf.vim
```diff
@@ -50,7 +50,7 @@ if s:is_win
   " Use utf-8 for fzf.vim commands
   " Return array of shell commands for cmd.exe
   function! s:wrap_cmds(cmds)
-    return map(['@echo off', 'for /f "tokens=4" %%a in (''chcp'') do set origchcp=%%a', 'chcp 65001 > nul'] +
+    return map(['@echo off', 'set TERM= > nul', 'for /f "tokens=4" %%a in (''chcp'') do set origchcp=%%a', 'chcp 65001 > nul'] +
```
janlazo: fzf unsets `TERM` when `TERM=cygwin` to support msys2 (without mintty). fzf failing on mintty is intended behavior because of tcell. The exception is https://github.com/junegunn/fzf/issues/963#issuecomment-525186950 because of ssh from Unix-like OS to Windows.

What's the usecase for `TERM=xterm-256colors` on a bash shell without mintty?

plugin/fzf.vim
```diff
@@ -75,7 +75,7 @@ function! s:shellesc_cmd(arg)
 endfunction
 
 function! fzf#shellescape(arg, ...)
-  let shell = get(a:000, 0, &shell)
+  let shell = get(a:000, 0, s:is_win ? 'cmd.exe' : &shell)
```
janlazo: Is this for `fzf#wrap` only? `fzf#run` calls `s:use_sh` to set the shell.

janlazo:
This is ok for a first step since you set `shellcmdflag` but to support custom shells, https://github.com/junegunn/vim-plug/pull/856 and https://github.com/junegunn/vim-plug/pull/860 should be ported over.

plugin/fzf.vim
msr1k: It is for external program such as [junegunn/fzf.vim's vim script](https://github.com/junegunn/fzf.vim/blob/master/autoload/fzf/vim.vim) which seemingly calls this function before `fzf#run` execution.

plugin/fzf.vim
msr1k: In my case, `gvim.exe` which started from mintty inherits `TERM` and it prevents running fzf.

janlazo: What if you unset `TERM` only if `has('gui_running')` is true? fzf does not work in mintty no matter what value `TERM` is set to.

plugin/fzf.vim
janlazo: If you're doing this, go all the way. Change all invocations of `fzf#shellescape` within the context of `fzf#run` to pass `&shell`. Then

```
let shell = get(a:000, 0, s:is_win ? 'cmd.exe' : 'sh')
```

plugin/fzf.vim
```diff
@@ -334,19 +334,20 @@ function! fzf#wrap(...)
 endfunction
 
 function! s:use_sh()
-  let [shell, shellslash] = [&shell, &shellslash]
+  let [shell, shellslash, shellcmdflag] = [&shell, &shellslash, &shellcmdflag]
   if s:is_win
     set shell=cmd.exe
+    set shellcmdflag=/c
```
janlazo: Setting `shell` options have too many edge cases on Windows because there are 3 incompatible shells to work with and fzf's Vim plugin supports neovim. Like vim-plug, don't set any shell options on Windows.

plugin/fzf.vim
msr1k: I'm sorry but I can't get what exactly means to add `&shell` for all invocations of `fzf#shellescape`.
Could you tell me what aim is laid here, or what problems exist if I don't do that.

In my opinion, since `s:use_sh` have already replaces `&shell` to `sh` or `cmd.exe`,
there are not so much differences regardless of whether I do that or not.

janlazo: Idea is to not set the shell in `s:use_sh` on Windows in the future, similar to what vim-plug does, because there are many shell option combinations to consider depending on the OS and editor. vim-plug passes `&shell` to `fzf#shellescape` when it needs to run a shell command with the user's `&shell`.

Perhaps, it's too much to do it all in this PR. fzf supported cmd.exe on Windows so doing it halfway (ie. hardcode the default fallback shell to cmd.exe for `fzf#shellescape`) is fine for now.

plugin/fzf.vim
janlazo: Not setting shell options here will take a lot of work so consider Neovim's default shell options here.

```vim
let &shellcmdflag = has('nvim') ? '/s /c' : '/c'
```

msr1k: Thank you for your advice.
I applied it at abde45180f947f5b6f1d3320973d9a7025fc8575.

plugin/fzf.vim
msr1k: I partially introduced your suggestion at ca770ec63940fa665a283afc30c92bd3c9ffd9a4 for now.

plugin/fzf.vim
msr1k: Thank you. I did it at 502432f7a6764362fa03093c0b7491140f8f3ffc.

plugin/fzf.vim
```diff
@@ -334,19 +336,20 @@ function! fzf#wrap(...)
 endfunction
 
 function! s:use_sh()
-  let [shell, shellslash] = [&shell, &shellslash]
+  let [shell, shellslash, shellcmdflag] = [&shell, &shellslash, &shellcmdflag]
   if s:is_win
     set shell=cmd.exe
     set noshellslash
+    let &shellcmdflag = has('nvim') ? '/s /c' : '/c'
```
janlazo: `/s` requires setting `shellxquote`. Use the default values for Windows: `"` for nvim, `(` for vim.

janlazo:
Wait for https://github.com/junegunn/vim-plug/pull/913 to be merged to vim-plug and then be ported to fzf. Then, rebase your branch so that it can be merged.

msr1k:
I got it. Thank you for your review.

junegunn:
https://github.com/junegunn/vim-plug/pull/913 is backported now, and I've resolved the conflict. Is everything okay?

plugin/fzf.vim
```diff
@@ -54,6 +54,7 @@ if s:is_win
     return map([
       \ '@echo off',
       \ 'setlocal enabledelayedexpansion']
+    \ + (has('gui_running') ? ['set TERM= > nul'] : [])
```
janlazo: Should port this over to vim-plug.

janlazo:
Yes. It's safe to merge but we can wait for @msr1k to verify. Concern in https://github.com/junegunn/vim-plug/pull/913 is unicode and msys2/mingw/cygwin allows unicode.

junegunn:
I'm going to merge this as I want to push out a new release with it. Thanks @msr1k and @janlazo

janlazo:
Tested it on a Windows VM today. It works with my config before and after switching the shell to bash. It's good to go, at least for English users.

PR #1697, opened 2019-09-26
[zsh] fix completion from specified path prefix when set FZF_COMPLEON_TRIGGER=;

zyf0330:


junegunn:
Merged as https://github.com/junegunn/fzf/commit/b911af200c4ed4085a76d3e0a12458ccba2d3b3f.
Thanks!

Issue #1709, opened 2019-10-05
After processing command for fzf-vim plugin


NightWolf007:
- Category
    - [ ] fzf binary
    - [ ] fzf-tmux script
    - [ ] Key bindings
    - [ ] Completion
    - [x] Vim
    - [x] Neovim
    - [ ] Etc.
- OS
    - [ ] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Windows Subsystem for Linux
    - [ ] Etc.
- Shell
    - [ ] bash
    - [ ] zsh
    - [ ] fish

I've created a simple application for adding icons to files:
https://github.com/NightWolf007/izer

To use it with fzf from shell i just need to run this:
```shell
export FZF_DEFAULT_PREVIEW="fd --type file | izer iconize -f=nerd -c"
fzf --preview 'bat $(izer deiconize {})' | izer deiconize
```

And it's ok, but things get complicated in fzf-vim plugin.

It would be nice to have an option in fzf-vim plugin to provide a post fzf command,
like `izer deiconize`. That will substitute this command in pipe after running fzf.

If there are any other ideas to implement feature, let's discuss.

junegunn:
We already have `sink` (or `sink*`) for post-processing the output.

```vim
call fzf#run(fzf#wrap({
  \ 'source': 'ls | sed "s/^/>> /"',
  \ 'sink': {line -> execute('e '. line[3:])
\ }}))
```

And you can override the `:Files` command to use your program by redefining it with a custom sink function like so in a similar manner:

```vim
function! MySink(lines)
  " Trim the prefix except the first line which is the name of the key pressed (--expect)
  let a:lines[1:] = map(a:lines[1:], {_, line -> line[3:]})

  " Then reuse the default sink function fzf#wrap generates
  call fzf#wrap()['sink*'](a:lines)
endfunction

command! -bang -nargs=? -complete=dir Files
  \ call fzf#vim#files(<q-args>, {'source': 'ls | sed "s/^/>> /"', 'sink*': function('MySink')}, <bang>0)
```

Note that I explicitly specified `source` option instead of setting FZF_DEFAULT_COMMAND (there's no such thing called FZF_DEFAULT_PREVIEW by the way).

I hope this answers your question.

tbrodbeck:
Hey guys! I just tried to reproduce this workflow. The only thing that does not work for me is the color information. The icons in fuzzy find look like this: `[38;5;255m[0m iconful/iconful`
[...]

PR #1712, opened 2019-10-13
#1709 Add after processing command

NightWolf007:


junegunn:
Thanks, but we don't really need this. See my comment https://github.com/junegunn/fzf/issues/1709#issuecomment-541414749

PR #1730, opened 2019-10-23
add support for a limit when using --multi

### LINKED ISSUE #1718: max limit for multi select
- Category
    - [x] fzf binary
    - [ ] fzf-tmux script
    - [ ] Key bindings
    - [ ] Completion
    - [ ] Vim
    - [ ] Neovim
    - [ ] Etc.
- OS
    - [x] Linux
    - [x] Mac OS X
    - [x] Windows
    - [ ] Etc.
- Shell
    - [ ] bash
    - [ ] zsh
    - [ ] fish

In certain situations it would be helpful, if there was a max limit for how many items one can select.

Would you @junegunn accept a PR that allows people to specify a max limit for multi select?
To keep backwards compatibility, maybe a new argument like `--multi-limit`.

What do you think?

tessus:
I have implemented an option --multi-limit=LIMIT which allows to set a max number of selections one can make.

closes #1718

### DIFF
--- a/src/options.go
+++ b/src/options.go
@@ -39,6 +39,7 @@ const usage = `usage: fzf [options]
 
   Interface
     -m, --multi           Enable multi-select with tab/shift-tab
+    --multi-limit=LIMIT   Max limit for multi-select (default: no limit)
     --no-mouse            Disable mouse
     --bind=KEYBINDS       Custom key bindings. Refer to the man page.
     --cycle               Enable cyclic scroll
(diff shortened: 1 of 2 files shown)


junegunn:
Thanks, but there are a few more things to be handled, such as `select-all`, `deselect-all`, and `toggle-all`. I'm working on my own version, I'll let you know when it's done.

tessus:
I thought about those edge cases but then came to the conclusion that nobody in their right mind would use those when limiting a selection. But then I thought, you'd just tell me to implement them, if necessary.
Anyway, no matter, I'm glad that you thought it was a useful feature. 👍

PR #1731, opened 2019-10-23
Nix install instructions

mgttlinger:
fzf is also available via the nix package manager


junegunn:
Added NixOS section in https://github.com/junegunn/fzf/commit/ef80bd401fdab3979a72696dac585f1cbd463efc. Thanks.

mgttlinger:
Note that installing via nix is not limited to the NixOS operating system but works well on other Linux derivations, OS X, and afaik nix even works on Windows in the Unix subsystem.

Issue #751, opened 2016-11-30
[Feature request] invoke `ag` interactively


lotabout:
- Category
    - [X] fzf binary
- OS
    - [X] Linux
- Shell
    - [X] bash

I've used `fzf` and `fzf.vim` and they are great! `Ag` command in `fzf.vim` will allows us to:

1. `:Ag <pattern>` will return the search result and pipe it to `fzf`
2. `fzf` will allows us to narrow down the above result to get what we want.

However, in `ag`'s case, more often, we may want to change the `<pattern>` dynamically instead of changing the query to narrow down the result of a specific `<pattern>`. I'll give an image as example of this:

![interactive mode example](https://cloud.githubusercontent.com/assets/1527040/16977634/e03c9484-4e89-11e6-8255-69394964cb90.gif)

[skim](https://github.com/lotabout/skim) is a rust version of fzf, which provides an "interactive" mode to dynamically invoke the commands.

I understand the role of `fzf` is a general purpose "filter", and the above feature may be far from what `fzf` tries to be. Though as an end-user, I'd really like `fzf` to have this feature. Please consider about it, thanks!

junegunn:
I don't have enough time these days, so I can make any promises, but I'll think about it.

Anyway, here are some concerns that come to my mind.

- Rebuilding the input list from scratch can feel clunky. The list will be cleared, cursor will always be brought down to the bottom of the screen, tab-selected entries will be lost. And depending on the command used, there can be noticeable initial delay which will make the user wait until the first entry from the new command pops up. 
- So restarting ag command every time you press a key can be an overkill, especially when the selectivity of the initial query is good. Isn't it silly to restart ag command when there are only a handful of matches?. It may suffice to rebuild the list only when some key is pressed such as `CTRL-R` (r for refresh). Note that it is possible to implement the idea externally using `--expect` and `--print-query` options, though it requires closing the current fzf window and opening a new one. So the experience will not be as nice, but in that way, we can keep the simplicity of fzf.
- Search syntax of fzf is not compatible with the regex used by ag command. Passing fzf query directly to ag command may or may not make sense. (e.g. `!test 'run`)

junegunn:
Since I don't have answers to the above questions, I'll close the issue for now.

[...]

PR #1736, opened 2019-11-03
Adding initial support for running interactive commands

mateuszr:
I looked at feature request #751 and I it seems to me that this is a recurring topic. I would give it a try. This is a straightforward implementation of running interactive command in fzf:
- we specify the command using -c param
- query is fed directly to the command
- the command is run every time the query changes (is rerun as-you-type)
- the list is fed with the new result of the command (and any current selection is gone)
The speed is acceptable, even though we run a command multiple times. 
I've put some more observations in #751.

### DIFF
--- a/README.md
+++ b/README.md
@@ -570,6 +570,19 @@ For more advanced examples, see [Key bindings for git with fzf][fzf-git]
 
 [fzf-git]: https://junegunn.kr/2016/07/fzf-git/
 
+### Using fzf in interactive mode
+
+You can also use fzf in "interactive command" mode. This mode changes the "default"
+way how fzf retrieves input for filtering. Instead of reading from stdin, fzf
+executes command specified by `--cmd` option. Every change to the filter query
+results in the command being reevaulated and the result is fed back to fzf.
+The query is passed to the command through `{}`.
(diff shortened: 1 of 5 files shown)


andreyorst:
Thanks on working on this!

It would be nice if you could make args compatible with those that `skim` uses. As maintainer of [fzf.kak](https://github.com/andreyorst/fzf.kak) I try to keep plugin generic and because `skim` uses pretty much the same args as `fzf` it was easy to support both tools with the same code.

mateuszr:
hey!
Actually they are the same: `-c` and alternatively `--cmd`, are exactly the same that `skim` uses. Or I missed something?

andreyorst:
> hey!
> Actually they are the same: `-c` and alternatively `--cmd`, are exactly the same that `skim` uses. Or I missed something?

There's also `-i` in `skim`. https://github.com/lotabout/skim#interactive-mode

Maybe it's not necessary though.

junegunn:
I'm taking a different approach in #1750, which I believe is more extensible and generic. It doesn't introduce a new "mode", instead, it builds upon the existing features.

Issue #1727, opened 2019-10-22
fzf works but exits with SIGINT(2)


phatsk:
- Category
    - [x] fzf binary
    - [ ] fzf-tmux script
    - [ ] Key bindings
    - [ ] Completion
    - [ ] Vim
    - [x] Neovim
    - [ ] Etc.
- OS
    - [ ] Linux
    - [x] Mac OS X
    - [ ] Windows
    - [ ] Windows Subsystem for Linux
    - [ ] Etc.
- Shell
    - [ ] bash
    - [x] zsh
    - [ ] fish

I noticed the following when `:q` quitting `fzf` in NeoVim - note that fzf works fine, and this only happens when an option is not selected:

```sh
Error running ( ag --nogroup --column --color  '^(?=.)' )|'/Users/phatsk/.fzf/bin/fzf' '--color=bg+:#343d46,bg:#1b2b34,spinner:#fac863,hl:#65737e,fg:#c0c5ce,header:...selec
t-all' '--color' 'hl:4,hl+:12' --delimiter : --nth 4.. --expect=ctrl-v,ctrl-s,ctrl-t --no-height > /var/folders/1p/2y8_8lvx3131z80ktzz6ftcr0000gn/T/nvimBvoevO/2
```

After digging a bit, I tried updating with `~/.fzf/install` which seemed to go fine. I restarted my terminal session and tried again to no avail. 

Next, I ran `fzf` directly in the terminal. It runs fine, but upon exit it leaves with `SIGINT(2)`. No errors are reported to the screen. My guess is that this exit code is what is causing it to throw the error in nvim.

junegunn:
> when :q quitting fzf in NeoVim

Why do you close fzf with `:q`? You can just hit ESC or CTRL-C to close it and if you do that, the exit code is handled gracefully.

phatsk:
Interesting. It's just habit for me for quitting windows. ESC does not quit my `fzf` window, however `^C` does. Previously to updating NeoVim, `:q` would quit the window gracefully.

junegunn:
You must have remapped ESC to something else using `tnoremap` (see `:tmap`).

To override:

```vim
autocmd! FileType fzf tnoremap <buffer> <esc> <c-c>
```

Issue #1700, opened 2019-09-26
error message when closing fzf buffer


frescoraja:
- Category
    - [X] Neovim

- OS
    - [X] Mac OS X

- Shell
    - [X] zsh

I do not remember this happening always - seems to be a fairly recent thing .. but if i close the fzf search buffer (by hitting `<Esc>` to get out of insert mode, then `:q<Enter>`, i get an error message like the following:

`Error running ( /usr/local/bin/zsh '/var/folders/gv/qdygj2ss0_jgn8xdmjqt6sdr0000gp/T/nvimxuvGwv/3' )|'/Users/username/.fzf/bin/fzf'  '-m' '--prompt' '~/D/F/c/c/s/g/servicecode/' '--preview-window' 'right:50%:hidden' '--preview' ''\'
'/Users/username/.vim/vimplugs/fzf.vim/bin/preview.sh'\'' {}' '--bind' '?:toggle-preview' --expect=ctrl-v,ctrl-x,ctrl-t --no-height > /var/folders/gv/qdygj2ss0_jgn8xdmjqt6sdr0000gp/T/nvimxuvGwv/2`

is this intended behavior? is it a bug?
thanks for looking into it

junegunn:
It's not a bug. Pressing ESC should normally close the fzf window; you probably mapped it to exiting insert mode using `tnoremap`. In that case, you can just press CTRL-C instead to close fzf. Or you can override the mapping only on fzf windows.

```vim
autocmd! FileType fzf tnoremap <buffer> <esc> <c-c>
```

frescoraja:
thanks, you're right. 👍

PR #1749, opened 2019-11-11
Nvim: handle SIGHUP in exit handler

mhinz:
In recent Nvim versions, an "Error running ..." message is shown even for normal
use cases, such as:

    :Files
    <c-\><c-n>
    :close

Closing the window will :bwipeout! the terminal buffer, because fzf sets
bufhiden=wipe.

When deleting the terminal buffer while fzf is still running, Nvim sends SIGHUP.
This happens for quite some time already, but the bug only manifests since this
commit:

  https://github.com/neovim/neovim/commit/939d9053b

It's The Right Thing to do when the application exited due to a signal.

Before that commit, no "Error running ..." message was shown, because 1 (instead
of 128 + 1 == SIGHUP) was returned which the exit handler in fzf.vim treats as
"NO MATCH".

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -527,6 +527,10 @@ endif
 function! s:exit_handler(code, command, ...)
   if a:code == 130
     return 0
+  elseif has('nvim') && a:code == 129
+    " When deleting the terminal buffer while fzf is still running,
+    " Nvim sends SIGHUP.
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks. For reference, previous issues regarding the problem:
- #1700 
- #1727

PR #1758, opened 2019-11-16
[install] Support busybox uname on Windows

janlazo:
uname (busybox) outputs `Windows_NT` on Windows 10 so `install` script could not download the binary. This PR should fix it.


junegunn:
Thanks!

PR #1779, opened 2019-12-03
Update README.MD

kozachynskyi:
It was long until I understood that the /path/to/file/ has to have /plugin/fzf.vim inside. That's my proposal to make it clearer for user, not sure, though, if it's understandable.


kozachynskyi:
Same changes should can done on 'junegunn/fzf.vim' repo

junegunn:
Thanks, but I think the topic is already covered on the page. Hopefully, a59e846 makes it more clear.

PR #1781, opened 2019-12-04
Open Files in New Tab

asommer70:
Created g:fzf_new_tab config option for opening files in a new tab if option is true.

I like opening files in a new tab with the super handy Ctrl+p.  Hope it's useful for others.

Great plugin and utility, thanks for creating it.


junegunn:
It is already possible via `g:fzf_action`.

```vim
let g:fzf_action = { 'enter': 'tab split' }
```

It is described in https://github.com/junegunn/fzf/blob/master/README-VIM.md

PR #1784, opened 2019-12-05
ssh completion: handle multiple hosts per line

henrebotha:
This commit allows the ssh completion to handle the case where multiple
hostnames are declared on one line, such as in this example that
declares three hosts:

	Host foo.com bar.dev baz

It does not correcly handle quoted hostnames that contain whitespace,
but this was the case before this commit.


junegunn:
Thanks, but I think you're confused somehow. It already works. See #1062.

henrebotha:
Ah, I see the problem then. Let me update this real quick.

PR #1788, opened 2019-12-06
ssh completion: skip only aliases matching *

henrebotha:
This commit fixes a bug where lines that declare multiple hostnames get
omitted from completion entirely if one of the hostnames matches *. For
example:

	Host foo.com bar.dev baz.*


henrebotha:
This is what I was trying to "fix" in #1784.

junegunn:
LGTM, thanks.

PR #1791, opened 2019-12-12
add Ripgrep as source of files for FZF

johntellsall:


junegunn:
Thanks, but the section was written as a step-by-step illustration of how one can set up any program as the default source of fzf, and I don't think showing more examples is really necessary. And IMO, fd is better suited for the purpose than rg.

PR #1793, opened 2019-12-15
[vim] use cterm colors on Windows

janlazo:
Truecolor does not work on default Windows terminal.
It is a problem in neovim GUIs.

https://github.com/sainnhe/edge/issues/5#issuecomment-565748240


junegunn:
Makes sense, thanks!

PR #1802, opened 2019-12-24
[zsh completion] Expand aliases recursively

sei40kr:
## Description

What this PR does:
- Expand aliases recursively to find proper completer

e.g.
```
$ alias s=ssh

$ s ** # press tab here to complete hostnames
```


NemesisRE:
I just started to work with kubectl for some time default alias everyone suggesting is k which is convinient cause the rest of the commands can get pretty long. If that pr works as expect it would make things so much easier.

PR #1809, opened 2020-01-04
Add FZF_PATH_COMPLETION_OPTS to allow preview for path completion

jawang35:
It would be nice to allow enabling previews to path completion. Adding an `FZF_PATH_COMPLETION_OPTS` environment variable would allow users to set `--preview` for path completions while not breaking other completions for which a preview doesn't make sense.

### DIFF
--- a/README.md
+++ b/README.md
@@ -411,6 +411,9 @@ export FZF_COMPLETION_TRIGGER='~~'
 # Options to fzf command
 export FZF_COMPLETION_OPTS='+c -x'
 
+# Preview path completion list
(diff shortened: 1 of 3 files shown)


jannes:
could you also add an analogous FZF_DIR_COMPLETION_OPTS?

junegunn:
Thanks, but in this way, we would have to add a new option variable for each type of completion, and for each command, and I'd like to avoid doing that. We need a more generic solution.

Also see https://github.com/junegunn/fzf/issues/1850#issuecomment-586615919

PR #1810, opened 2020-01-05
fix issue #1637 show error message when preview command fails

### LINKED ISSUE #1637: fzf --preview is empty if $SHELL points to missing file
- Category
    - [X] fzf binary
    - [ ] fzf-tmux script
    - [ ] Key bindings
    - [ ] Completion
    - [ ] Vim
    - [ ] Neovim
    - [ ] Etc.
- OS
    - [ ] Linux
    - [X] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [ ] bash
    - [ ] zsh
    - [X] fish

If my `SHELL` variable points to a file that doesn't exist, the preview window shown with `fzf --preview` is completely empty, not even showing an error.

I didn't even know the preview command was executed using my current shell. I don't see any documentation to this effect. Apparently the preview works by executing the equivalent of `${SHELL:-sh} -c "$previewCommand"`, so if `SHELL` points at a missing file, this fails. But in this case, the failure isn't printed to the preview window, it's just silently swallowed.

I discovered this when using `fzf` from a `fish` shell after having reinstalled fish in a new location and uninstalling the old (and then running `exec fish -l` from my active terminal window, which preserved existing env vars including `SHELL`).

Due to this, I think two things need to be done:

1. The fact that `--preview` uses `$SHELL` should be documented somewhere, ideally both in the manpage and in the repo README.
2. If the execution of `$SHELL` itself fails, fzf should print an error to the preview window indicating that. It should probably only do so if the execution result indicates that the command couldn't be found or executed at all, rather than on any non-zero result from the shell, as there are many commands that return an expected non-zero status code (such as `grep`)

tonymet:
fix #1637 

## STEPS
```
SHELL=kaka go run .  --preview 'cat {}'
```

## BEFORE:
[preview window is empty]

## AFTER
Preview window:
```
 exec: "kaka": executable file not found in $PATH
```

## Other Testing
test suite pass
```
 make test > /dev/null && echo $?                                   
0
```

### Test existing behavior with return code ~= 0
```
  138   go run .  --preview 'cat {} |grep fzf'
  149   go run .  --preview 'cat {} |grep -v fzf'
  152   go run .  --preview 'cat {} |grep -v fzf'
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ doc/tags
 vendor
 gopath
 *.zwc
+fzf
--- a/README.md
+++ b/README.md
@@ -484,7 +484,8 @@ See *KEY BINDINGS* section of the man page for details.
 ### Preview window
 
 When `--preview` option is set, fzf automatically starts an external process with
-the current line as the argument and shows the result in the split window.
+the current line as the argument and shows the result in the split window. Your
+`$SHELL` is used to execute the command with `$SHELL -c COMMAND`
 
 ```bash
 # {} is replaced to the single-quoted string of the focused line
--- a/src/terminal.go
+++ b/src/terminal.go
@@ -1572,7 +1572,10 @@ func (t *Terminal) Loop() {
 					var out bytes.Buffer
 					cmd.Stdout = &out
 					cmd.Stderr = &out
-					cmd.Start()
+					err := cmd.Start()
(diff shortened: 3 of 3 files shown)


junegunn:
Thanks!

PR #1815, opened 2020-01-07
[bash-completion] Populate emacs and vi keymaps

### LINKED ISSUE #39: Doc that vi-mode requires "set -o vi"

jablko:
Enables the right bindings when switching between editing modes.

Fixes #39

### DIFF
--- a/README.md
+++ b/README.md
@@ -340,10 +340,6 @@ If you're on a tmux session, you can start fzf in a split pane by setting
 `FZF_TMUX` to 1, and change the height of the pane with `FZF_TMUX_HEIGHT`
 (e.g. `20`, `50%`).
 
-If you use vi mode on bash, you need to add `set -o vi` *before* `source
-~/.fzf.bash` in your .bashrc, so that it correctly sets up key bindings for vi
-mode.
-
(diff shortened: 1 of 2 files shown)


junegunn:
Very nice. Thanks!
How come nobody has ever suggested this before!

PR #1817, opened 2020-01-09
[bash] Multiline C-r without histexpand

### LINKED ISSUE #1370: fzf doesn't handle multiline history
- Category
    - [x] fzf binary

- OS
    - [x] Linux

- Shell
    - [x] bash

Describe your problem or suggestion from here ...
1. While I'm in tmux and do ctrl - r then multiline commands will be split into lines
2. When I select a line in the multiline it won't put the line or the command

Hopefully this can be fixed as it would be awesome to go back in my sql command history :100: 
I saw some other https://github.com/junegunn/fzf/issues/146, but I couldn't make it work...

jablko:
```Shell
> echo 'foo
bar'
```

Previously, if you <kbd>C-r</kbd> and select the `bar'` line, your command line remains empty: `__fzf_history__()` fails because the the selected line doesn't [start with a history number](https://github.com/junegunn/fzf/blob/82bf8c138daaabf28faf7819b2f5a2b03af37d43/shell/key-bindings.bash#L60).

If you select the `echo 'foo` line, it pastes both lines [only if `histexpand` is enabled](https://github.com/junegunn/fzf/commit/392da53f5380907fdc7c21f398612be82f7dc3e0).

If you search for `foo bar`, neither line is found.

Fixes #1370

### DIFF
--- a/shell/key-bindings.bash
+++ b/shell/key-bindings.bash
@@ -52,22 +52,13 @@ __fzf_cd__() {
 }
 
 __fzf_history__() (
-  local line
-  shopt -u nocaseglob nocasematch
-  line=$(
-    HISTTIMEFORMAT= builtin history |
-    FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} $FZF_DEFAULT_OPTS --tac --sync -n2..,.. --tiebreak=index --bind=ctrl-r:toggle-sort $FZF_CTRL_R_OPTS +m" $(__fzfcmd) |
-    command grep '^ *[0-9]') &&
-    if [[ $- =~ H ]]; then
-      sed 's/^ *\([0-9]*\)\** .*/!\1/' <<< "$line"
-    else
-      sed 's/^ *\([0-9]*\)\** *//' <<< "$line"
-    fi
+  HISTTIMEFORMAT='' builtin history |
+    sed -E -e :a -e 'N;$!ba' -e 's/(^|\n) *[1-9][0-9]*[ *] /\x00/g' |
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks for taking a stab at the issue, but I won't merge this for the following reasons:

1. The `sed` command used doesn't work on macOS unless you manually installed GNU sed
2. I prefer having index numbers. They help me grok the relative ordering of the commands, and I sometimes rely on them to pinpoint the exact command I want. Also, I don't want to break backward compatibility.
3.  The additional `sed` process makes getting the entire list slower.
    ```
    > HISTTIMEFORMAT= builtin history | wc -l
      122501
    > time HISTTIMEFORMAT= builtin history > /dev/null

    real    0m0.095s
    user    0m0.068s
    sys     0m0.028s
    > time HISTTIMEFORMAT= builtin history | gsed -E -e :a -e 'N;$!ba' -e 's/(^|\n) *[1-9][0-9]*[ *] /\x00/g' > /dev/null

    real    0m0.450s
    user    0m0.427s
    sys     0m0.145s
    ```
    Any delay is noticeable because we use `--tac --sync`. Having no delay is much more important to me because I face the delay every time I press CTRL-R which is very often, while I rarely have to fetch a multi-line command from the middle of it. 

I also have to mention that I'm not interested in adding more options for configuring the behavior because maintaining these scripts for three different shells has been really painful to me and I don't want to increase the complexity anymore. Users can easily override these functions in their shell configuration files. And I believe more advanced implementations should be external, third-party projects.

Anyway, thanks again for your interest in the project.

jablko:
@junegunn I've updated this PR to work on macOS, and replaced `--tac --sync` with `fc -r`. Would you consider giving it another review?

PR #1820, opened 2020-01-12
[zsh-completion] Fix quoting/splitting issues

chitoku-k:
This PR fixes issues derived from quoting and word-splitting in Zsh completion.

## Quoting issue

Since `fzf-completion` calls `eval` without carefully quoting `$lbuf`, it broke in the following situation.

First, type:
```zsh
telnet -l "User that is quoted" **<TAB>
```
and hit <kbd>enter</kbd> to select `localhost` resulted in:
```zsh
telnet -l Userlocalhost
```
whereas the correct behaviour should be:
```zsh
telnet -l "User that is quoted" localhost
```

Simply quote the parameter using `${(q)lbuf}` solves this issue.

## Word-splitting issue

`_fzf_complete` splits its parameter (`$1`) in order to pass them to `fzf` in shell manner; however, the way it does this is not exactly the same way as Zsh. For instance, let's say we want to pass a parameter that contains spaces as follows:

```zsh
_fzf_complete '--tiebreak index --preview "echo {}"' < <(find . -type f)
```

`${=fzf_opts}` splits the `$fzf_opts` in this way:
- `--tiebreak`
- `index`
- `--preview`
- `"echo`
- `{}"`

but the expected way should look like:
- `--tiebreak`
- `index`
- `--preview`
- `echo {}`

Using [`Z:`*opts*`:`](http://zsh.sourceforge.net/Doc/Release/Expansion.html#index-substitution_002c-parameter_002c-flags) and `Q` together enables this shell parsing not word-splitting.

Thanks.

### DIFF
--- a/shell/completion.zsh
+++ b/shell/completion.zsh
@@ -55,7 +55,7 @@ __fzf_generic_path_completion() {
       leftover=${leftover/#\/}
       [ -z "$dir" ] && dir='.'
       [ "$dir" != "/" ] && dir="${dir/%\//}"
-      matches=$(eval "$compgen $(printf %q "$dir")" | FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} --reverse $FZF_DEFAULT_OPTS $FZF_COMPLETION_OPTS" ${=fzf} ${=fzf_opts} -q "$leftover" | while read item; do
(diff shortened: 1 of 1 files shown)


junegunn:
Superb, thank you very much!

Issue #1497, opened 2019-02-07
Provide a snap package


soullivaneuh:
Snap is a package manager aim to work on multiple OS.

Apparently, anybody can create a snap.

Maybe you can provide your own official snap?

Two already exist: https://snapcraft.io/search?category=&q=fzf

But the first one is not up to date from a time and the second one does not seem to work after the installation.

What do you think?

Regards

tetienne:
Will be a great and nice alternative to Linuxbrew!

pckilgore:
Checkout the branch above for a start at this, but for those who want to be able to install a fully-functional v0.20 right now, you should be able to:
`sudo snap install --edge fzf-unofficial-18-04 --devmode`

To use it you need to invoke `fzf-unofficial-18-04.fzf` so you'll probably want to setup an alias :upside_down_face: 

Obviously, you should not be installing random snaps will full permissions like this built by internet strangers, so make your own decision whether me and this package is trustworthy.

But for reference, here's the basic recipe I used for this edge package:

```
name: fzf-unofficial-18-04
version: git
summary: A general-purpose command-line fuzzy finder.
description: |
  An unofficial snap build of fzf for Ubuntu 18.04.  The fzf package provides an
  interactive Unix filter for command-line that can be used with any list;
  files, command history, processes, hostnames, bookmarks, git commits, etc.
base: core18
grade: stable
confinement: devmode

parts:
  fzf:
    plugin: go
    go-importpath: github.com/junegunn/fzf
    source: ..
    source-type: git

apps:
  fzf:
    command: bin/fzf
```

Note this assumes that your `snapcraft.yml` file sits in the /snap directory of the master branch if you're paranoid (you should be) and want to build and install it yourself (you should, use --dangerous).

PR #1824, opened 2020-01-15
Adds initial snapcraft build.

pckilgore:
This is a first pass at #1497, and will require some additional ideas/help (I am willing to provide).

The idea here is to allow targeting 18.04 for conservative farts like me. The snap model is also pretty neat.

Right now this is published under an intentionally `unofficial` name while I continue to sort out permissions and plugins locally. 

But there is another option available.  We can [request to publish this snap under classic confinement](https://forum.snapcraft.io/t/process-for-reviewing-classic-confinement-snaps/1460).  This is a much easier thing to do than piece together the necessary plugs one at a time and asking for auto-connect permissions but it isn't necessarily clear to me if `fzf` would qualify given the link above.  But right now `ripgrep` has `--classic` so I'm thinking it is achievable for `fzf` given a well-reasoned case.

Left to do (assuming there is no consensus around just going the `--classic` route):
- [ ] Identify additional plugs/permissions for `fzf` to work as intended
- [ ] have @junegunn reserve the `fzf` name on snapcraft
- [ ] publish edge release under official snapcraft name
- [ ] Request automatic connections to identified plugs (ideal user experience, imo no point to publication if users must create manual connections *after* installation).
- [ ] (probably) update CI to automatically publish snaps on release
- [ ] Publish first stable snap!

Thanks for an awesome tool @junegunn and contributors!

### DIFF
--- a/snap/.gitignore
+++ b/snap/.gitignore
@@ -1,0 +1,3 @@
+*
+!snapcraft.yaml
+!.gitignore
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -7,15 +7,18 @@ description: |
   files, command history, processes, hostnames, bookmarks, git commits, etc.
 base: core18
 grade: stable
-confinement: classic
+confinement: strict
 
 parts:
   fzf:
     plugin: go
     go-importpath: github.com/junegunn/fzf
-    source: ..
+    source: .
     source-type: git
 
 apps:
   fzf:
     command: bin/fzf
+    plugs:
+      - home
+      - mount-observe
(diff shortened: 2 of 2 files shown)


pckilgore:
Right now I can use `fzf` for some basics (navigating the file tree in $HOME) but it still isn't doing everything the fully-permissioned package would do).

### Installation instructions

assuming snapcraft and snap are installed and the snap daemon (`snapd`) is started
```console
checkout add-snap
cd snap
snapcraft
sudo snap install fzf-unofficial-18-04_0.20.0_amd64.snap --dangerous
```

### To run

`fzf-unofficial-18-04.fzf` 

FYI namespacing will not be necessary if `fzf` is published under the name `fzf` instead of the the current name on this pr (I didn't want to squat it).

anna1980-6:
Sell

pckilgore:
I've run into enough trouble with snaps at this point in other packages I'm going to abandon this for now.

Issue #1309, opened 2018-06-07
Duplicates in history search and auto-Enter upon selections


gerases:
- Category
    - [ ] fzf binary
    - [ ] fzf-tmux script
    - [x] Key bindings
    - [ ] Completion
    - [ ] Vim
    - [ ] Neovim
    - [ ] Etc.
- OS
    - [x] Linux
    - [x ] Mac OS X
    - [ ] Windows
    - [ ] Windows Subsystem for Linux
    - [ ] Etc.
- Shell
    - [x] bash
    - [ ] zsh
    - [ ] fish

I have a couple of suggestions:

- When searching through the history, provide an option to remove the duplicates in the history. I know that bash has options to remove duplicates, but I don't want to remove them since they provide important info for debugging and forensics.

- How do I make fzf behave for CTRL-R/CTRL-T the same way it behaves for ALT-C. That is, I would like Enter to pressed automatically upon me selecting an entry?

edi9999:
1) For removing duplicates, use nauniq or similar to remove the duplicates : 

Personally, I have written __fzf_history__() 

by adding : 

```
		tac |
		nauniq --skip-chars="$countskip" |
		tac |
		$(__fzfcmd) +s --tac +m -n2..,.. --tiebreak=index --toggle-sort=ctrl-r |
```

gerases:
Hi @edi9999,

Maybe it should be an option to fzf? It seems useful enough for me.

Thanks,
  Sergei

edi9999:
If you want an option about uniqueness, this won't be part of fzf, because we want to have tools that do one thing and one thing well, and they should be able to be combined (using stdin/stdout and pipes).

sverhagen:
I'm interested in the "unique" feature, and I'm perfectly happy adding this to `key-bindings.bash` (which is where I found a `__fzf_history__` function, which I assume you're talking about). But I'm unclear how to exactly put your changes in there. And `nauniq`, is that in reference to the Perl function? Thanks for your help.

edi9999:
My current __fzf_history function is the following : 

```
	__fzf_history__() {
		local line
		countskip="$(history | tail -n 1 | grep -E '^ *[0-9]+' -o | wc -c)"
		countskip="$(( countskip + 1 ))"
		line=$(
		HISTTIMEFORMAT= history |
		grep '^.\{1,130\}$' --text |
		sed 's/ *$//g' |
[...]

Issue #626, opened 2016-07-20
zsh Ctrl+R must use option HIST_FIND_NO_DUPS


vaxXxa:
- Category
  - [ ] fzf binary
  - [ ] fzf-tmux script
  - [x] Key bindings
  - [ ] Completion
  - [ ] Vim
  - [ ] Neovim
  - [ ] Etc.
- OS
  - [ ] Linux
  - [x] Mac OS X
  - [ ] Windows
  - [ ] Etc.
- Shell
  - [ ] bash
  - [x] zsh
  - [ ] fish

I set `setopt HIST_FIND_NO_DUPS` and expect, that my Ctrl+R search history filtered without dups. I don't want to use option HIST_IGNORE_ALL_DUPS cause my history must be full and with dups, and only in search term I don't want to see duplicates.

junegunn:
Any idea how we can implement it?

Related: https://github.com/zsh-users/zaw/issues/71

edi9999:
I have replaced __fzf_history in .fzf-key-bindings.bash with the following : 

```
__fzf_history__() {
  local line
  countskip="$(history | tail -n 1 | grep -E '^[0-9]+' -o | wc -c)"
  countskip="$(( countskip + 1 ))"
  line=$(
    HISTTIMEFORMAT= history |
    grep '.\{1,79\}' |
    sed 's/ *$//g' |
    tac |
    nauniq --skip-chars="$countskip" |
    tac |
    $(__fzfcmd) +s --tac +m -n2..,.. --tiebreak=index --toggle-sort=ctrl-r |
    \grep '^ *[0-9]') && sed 's/ *\([0-9]*\)\** .*/!\1/' <<< "$line"
}
```

You will need nauniq (which you can find here : https://metacpan.org/release/App-nauniq)

edi9999:
But I wouldn't put this in the main repository

balta2ar:
@vaxXxa @junegunn @edi9999 Is this helpful http://unix.stackexchange.com/a/84838/44493?

The idea is to add numbers as the first column, sort & uniq by the second column (actual history lines), sort again by the first column (original indices) and remove indices. Only standard tools are used.

junegunn:
@balta2ar Thanks. But since the command history can be huge depending on the configuration I'm a bit concerned about the performance. Since fzf command for ctrl-r binding uses `--tac` option, the input stream should finish very quickly. Any delay will be noticeable and hurt the usability.

edi9999:
[...]

PR #1836, opened 2020-01-25
shell/bash/history-search: filter out duplicates

grantzvolsky:
This solves #1309. #626 should be solvable in a similar fashion.

Credit for the cleverness in the awk associative array approach goes to
@iridakos's [remove-duplicate-lines-preserving-order-linux](https://iridakos.com/programming/2019/05/16/remove-duplicate-lines-preserving-order-linux).

### DIFF
--- a/shell/key-bindings.bash
+++ b/shell/key-bindings.bash
@@ -55,7 +55,7 @@ __fzf_history__() (
   local line
   shopt -u nocaseglob nocasematch
   line=$(
-    HISTTIMEFORMAT= builtin history |
+    HISTTIMEFORMAT= builtin history | awk '!visited[substr($0, index($0, $2))]++' |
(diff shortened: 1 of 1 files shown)


grantzvolsky:
@junegunn Hello, in response to your comment [here](https://github.com/junegunn/fzf/pull/1817#issuecomment-572892825).

ad 1. OS X compatibility is a valid concern. I'm using gawk so equivalence check with nawk and possibly other awk flavours is needed.
ad 3. It is possible that with sufficiently long history, filtering out duplicates saves more cycles than it introduces.

junegunn:
The problem with the suggested approach is that it prints the first occurrence of a command when it's more desirable to print the last one, the most recent one. Let's say you just typed in a series of commands and you naturally expect to see them at the bottom of the list, but some commands are missing because you run the same commands a few years ago. It's going to be quite confusing.

Having said that, @jablko is working on #1837, which uses `fc -rl` instead of `history`. With `fc -rl`, fzf doesn't have to wait until the end, so the latency becomes less of an issue. Leaving the last occurrence of a command while filtering out the earlier duplicates becomes more viable. I'll revisit the issue once it's merged though I'm still not sure if I want to introduce a new option for duplicate filtering. Thanks.

PR #1837, opened 2020-01-25
[bash] Multiline C-r without histexpand

jablko:
I updated #1817 to work on macOS, and replaced --tac --sync with fc -r. Would you consider giving it another review?

### DIFF
--- a/shell/key-bindings.bash
+++ b/shell/key-bindings.bash
@@ -53,21 +53,16 @@ __fzf_cd__() {
 
 __fzf_history__() (
   local line
-  shopt -u nocaseglob nocasematch
   line=$(
-    HISTTIMEFORMAT= builtin history |
-    FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} $FZF_DEFAULT_OPTS --tac --sync -n2..,.. --tiebreak=index --bind=ctrl-r:toggle-sort $FZF_CTRL_R_OPTS +m" $(__fzfcmd) |
-    command grep '^ *[0-9]') &&
-    if [[ $- =~ H ]]; then
-      sed 's/^ *\([0-9]*\)\** .*/!\1/' <<< "$line"
-    else
-      sed 's/^ *\([0-9]*\)\** *//' <<< "$line"
-    fi
+    builtin fc -lnr -2147483648 |
+      perl -p -l0 -e 'BEGIN { getc; $/ = "\n\t" } s/^[ *]//; $_ = $ENV{HISTCMD} - $. . "\t$_"' |
+      FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} $FZF_DEFAULT_OPTS --tiebreak=index --bind=ctrl-r:toggle-sort $FZF_CTRL_R_OPTS +m --read0" $(__fzfcmd)
+  )
+  echo "${line#*$'\t'}"
 )
 
 # Required to refresh the prompt after fzf
 bind -m emacs-standard '"\er": redraw-current-line'
-bind -m emacs-standard '"\e^": history-expand-line'
 
 # CTRL-T - Paste the selected file path into the command line
 if [ $BASH_VERSINFO -gt 3 ]; then
@@ -79,7 +74,7 @@ else
 fi
 
 # CTRL-R - Paste the selected command from history into the command line
-bind -m emacs-standard '"\C-r": " \C-e\C-u\C-y\ey\C-u`__fzf_history__`\e\C-e\er\e^"'
+bind -m emacs-standard '"\C-r": " \C-e\C-u\C-y\ey\C-uHISTCMD=$HISTCMD\e\C-e __fzf_history__`\"\C-a\"`\C-e\e\C-e\er"'
 
 # ALT-C - cd into the selected directory
 bind -m emacs-standard '"\ec": " \C-e\C-u`__fzf_cd__`\e\C-e\er\C-m"'
--- a/shell/key-bindings.bash
+++ b/shell/key-bindings.bash
@@ -55,7 +55,7 @@ __fzf_history__() (
   local line
   line=$(
     builtin fc -lnr -2147483648 |
-      perl -p -l0 -e 'BEGIN { getc; $/ = "\n\t" } s/^[ *]//; $_ = $ENV{HISTCMD} - $. . "\t$_"' |
+      perl -p -l0 -e 'BEGIN { getc; $/ = "\n\t" } s/^[ *]//; $_ = '"$1"' - $. . "\t$_"' |
       FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} $FZF_DEFAULT_OPTS --tiebreak=index --bind=ctrl-r:toggle-sort $FZF_CTRL_R_OPTS +m --read0" $(__fzfcmd)
   )
   echo "${line#*$'\t'}"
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks for the update. Using `fc -r` is a nice idea, but like I mentioned in [my previous comment]( https://github.com/junegunn/fzf/pull/1817#issuecomment-572892825), I do not wish to remove index numbers as I find them quite useful.

I believe we can still display index numbers of the commands while using this approach. What do you think?

jablko:
:+1: I've updated this PR and restored the index numbers.

shell/key-bindings.bash
```diff
@@ -53,21 +53,16 @@ __fzf_cd__() {
...
-    HISTTIMEFORMAT= builtin history |
-    FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} $FZF_DEFAULT_OPTS --tac --sync -n2..,.. --tiebreak=index --bind=ctrl-r:toggle-sort $FZF_CTRL_R_OPTS +m" $(__fzfcmd) |
-    command grep '^ *[0-9]') &&
-    if [[ $- =~ H ]]; then
-      sed 's/^ *\([0-9]*\)\** .*/!\1/' <<< "$line"
-    else
-      sed 's/^ *\([0-9]*\)\** *//' <<< "$line"
-    fi
+    builtin fc -lnr -2147483648 |
+      perl -p -l0 -e 'BEGIN { getc; $/ = "\n\t" } s/^[ *]//; $_ = $ENV{HISTCMD} - $. . "\t$_"' |
```
junegunn: Any reason you manually put index numbers instead of using `fc` without `-n`?

shell/key-bindings.bash
```diff
@@ -53,21 +53,16 @@ __fzf_cd__() {
...
   line=$(
-    HISTTIMEFORMAT= builtin history |
-    FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} $FZF_DEFAULT_OPTS --tac --sync -n2..,.. --tiebreak=index --bind=ctrl-r:toggle-sort $FZF_CTRL_R_OPTS +m" $(__fzfcmd) |
-    command grep '^ *[0-9]') &&
-    if [[ $- =~ H ]]; then
-      sed 's/^ *\([0-9]*\)\** .*/!\1/' <<< "$line"
-    else
-      sed 's/^ *\([0-9]*\)\** *//' <<< "$line"
-    fi
+    builtin fc -lnr -2147483648 |
```
junegunn: I'm actually not really familiar with `fc` command of bash. But it seems to me that `1` instead of `-2147483648` suffices. Am I missing something?

shell/key-bindings.bash
jablko: With `-n` you don't need to parse the index numbers with a regular expression. You can split the history list with the input record separator, which to me seemed clean. Also it makes the item format independent of `fc`'s output, in case something different is preferred.

shell/key-bindings.bash
jablko: The problem is it doesn't work if the history list doesn't start at `1`, i.e. if you overflow `$HISTSIZE`:
```Shell
> HISTSIZE=7
> for i in {1..9}; do history -s $i; done
> fc -lnr 1
bash: fc: history specification out of range
```

junegunn:
Sorry for the delay. I've been busy lately.
Have you tested this on Linux? It works as expected on mac, but I'm seeing a weird behavior on some of my Linux machines.

```sh
# Build docker image with your change
make docker

# Inside docker container
ls -al
# Press CTRL-R and select ls -al, and the cursor position is not at the end of the line
-bash-5.0# ls▒-al
```

jablko:
Thanks for reviewing this, I've updated it and fixed the cursor position.

shell/key-bindings.bash
```diff
@@ -79,7 +74,7 @@ else
 fi
 
 # CTRL-R - Paste the selected command from history into the command line
-bind -m emacs-standard '"\C-r": " \C-e\C-u\C-y\ey\C-u`__fzf_history__`\e\C-e\er\e^"'
+bind -m emacs-standard '"\C-r": "\C-k\C-uHISTCMD=$HISTCMD\e\C-e __fzf_history__`\"\C-a\"`\e\C-e\C-e\er"'
```
junegunn: The reason we have `<space>` at the beginning of the sequence is to suppress beep/bell from the terminal. It beeps when `<ctrl-u>` is pressed on an empty command line. You can test it on tmux by enabling visual bell like so:

```sh
make docker

tmux set -g visual-bell on

# <CTRL-R>
# Visual bell rings
```

- ` `: Paste empty space
- `\C-e`: Move the cursor to the end
- `\C-u`: Delete the whole command line. We can be sure that the command line is not empty, so no beep.
- `\C-y\ey`: Remove the new entry (possibly just an empty space) in the "kill ring"

jablko: :man_facepalming: Whoops! Thank you for explaining, now I get it and I've undone that change. Thanks!

junegunn:
Thanks, one thing I'm not quite happy with the change is how the prompt looks when fzf is running.

Before:

```sh
$ `__fzf_history__`
...
```

After:

```sh
$ "`
...
```

Would it be possible to improve the situation?

jablko:
Absolutely: Done.

junegunn:
Thanks, it's better.

```
$ "`HISTCMD=128226 __fzf_history__`"
```

But can we just pass `$HISTCMD` as the argument to `__fzf__history`?

```diff
diff --git a/shell/key-bindings.bash b/shell/key-bindings.bash
index af23720..9bf1eb8 100644
--- a/shell/key-bindings.bash
+++ b/shell/key-bindings.bash
@@ -55,7 +55,7 @@ __fzf_history__() (
   local line
   line=$(
     builtin fc -lnr -2147483648 |
-      perl -p -l0 -e 'BEGIN { getc; $/ = "\n\t" } s/^[ *]//; $_ = $ENV{HISTCMD} - $. . "\t$_"' |
+      perl -p -l0 -e 'BEGIN { getc; $/ = "\n\t" } s/^[ *]//; $_ = '$1' - $. . "\t$_"' |
       FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} $FZF_DEFAULT_OPTS --tiebreak=index --bind=ctrl-r:toggle-sort --read0 $FZF_CTRL_R_OPTS +m" $(__fzfcmd)
   )
   echo "${line#*$'\t'}"
@@ -74,7 +74,7 @@ else
 fi
 
 # CTRL-R - Paste the selected command from history into the command line
-bind -m emacs-standard '"\C-r": " \C-e\C-u\C-y\ey\C-uHISTCMD=$HISTCMD\e\C-e __fzf_history__`\"\C-a\"`\C-e\e\C-e\er"'
+bind -m emacs-standard '"\C-r": " \C-e\C-u\C-y\ey\C-u__fzf_history__ $HISTCMD\e\C-e`\C-a`\C-e\e\C-e\er"'
 
 # ALT-C - cd into the selected directory
 bind -m emacs-standard '"\ec": " \C-e\C-u`__fzf_cd__`\e\C-e\er\C-m"'
```

which gives:

```
`__fzf_history__ 128230`
```

junegunn:
Merged, thanks for your patience!

leonklingele:
Thank you _so_ much for this change. It decreases the TTRIP (time-to-reverse-i-prompt) from ~0.6 seconds to basically 0. ❤️

jan-warchol:
Thanks @junegunn and @jablko!! This has been bothering me for a long time, now I see it's already fixed and my workaround is no longer needed! <3

Issue #1769, opened 2019-11-24
[feature request] customizable line continuation characters


herrsimon:
- [x ] I have read through the manual page (`man fzf`)
- [x ] I have the latest version of fzf
- [x] I have searched through the existing issues

## Info

- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [ ] bash
    - [x] zsh
    - [ ] fish
    
## Problem / Steps to reproduce

Could you please add an option to specify the line continuation characters (currently '.. ') which are inserted by fzf when --no-hscroll is used and a line exceeds the terminal width? In particular, please allow to specify the empty string so that no continuation characters are printed at all.

In my use case (which is rather typical), I'm selecting some id by its human-readable description by outputting the description with terminal width and then adding the id to the right. The two dots and the space inserted by fzf are not very useful in this case.

Thank you very much!

junegunn:
When it comes to the formatting of the input, it's almost impossible to come up with a general solution that meets all kinds of different requirements. So fzf leaves it up to the users. Instead of using `--no-hscroll`, you can pre-process the input to fit in screen before passing it to fzf.

```sh
input-generating-program | formatting-program | fzf
```

herrsimon:
I completely agree to what your saying but my problem does not fit your solution. Let me elaborate a bit more: Imagine you have a text file with lines of the form

``UUID       DESCRIPTION``

where only DESCRIPTION is meaningful for the user. Then, when you would like to select a UUID by its DESCRIPTION, your ``formatting-program`` has no choice as to display the full (and rather long) UUID somewhere on each line, unnecessarily using a lot of screen space without adding any additional information for the user. In such a situation, it is very useful to use --no-hscroll and hide the UUID on the right of the viewable screen area, and use something like 

``input-generating-program | formatting-program | fzf --no-hscroll | post-process-program``,

where post-process-program extracts the UUID. I think such a use case is not that exotic (I currently use it for a textfile-based personal database). 

Another point would be to just replace the '.. ' line continuation characters by '>' or any other symbol, in order to save some screen space. 

Of course we can now raise the question whether it is worthwile to discuss about three columns of screen space, but as fzf is perfect and fully customizable in every other aspect it would be sad not to have such a feature.

[...]

PR #1844, opened 2020-01-29
Make pointer and selected sign customizable

relastle:
Thank you for developing and maintaining this state-of-the-art CLI.

I opened this PR because as well as prompt, pointer sign and selected signs (in multi-select mode) should also be customizable with command line arguments.

fzf_demo

Such feature, for example, would be useful in a case, where UI-provider really want users to know "WHAT" will happen ( `multi-delete` :fire: or `multi-star` ⭐️ ) with emoji.

### DIFF
--- a/man/man1/fzf.1
+++ b/man/man1/fzf.1
@@ -231,10 +231,10 @@ A synonym for \fB--info=hidden\fB
 .BI "--prompt=" "STR"
 Input prompt (default: '> ')
 .TP
-.BI "--pointer-sign=" "STR"
+.BI "--pointer=" "STR"
 Pointer sign of cursor line (default: '>')
 .TP
-.BI "--selected-sign=" "STR"
+.BI "--marker=" "STR"
 Signs of selected ling in multi-mode  (default: '>')
 .TP
 .BI "--header=" "STR"
--- a/man/man1/fzf.1
+++ b/man/man1/fzf.1
@@ -231,6 +231,12 @@ A synonym for \fB--info=hidden\fB
 .BI "--prompt=" "STR"
 Input prompt (default: '> ')
 .TP
+.BI "--pointer-sign=" "STR"
+Pointer sign of cursor line (default: '>')
+.TP
+.BI "--selected-sign=" "STR"
+Signs of selected ling in multi-mode  (default: '>')
+.TP
 .BI "--header=" "STR"
 The given string will be printed as the sticky header. The lines are displayed
 in the given order from top to bottom regardless of \fB--layout\fR option, and
--- a/man/man1/fzf.1
+++ b/man/man1/fzf.1
@@ -232,10 +232,10 @@ A synonym for \fB--info=hidden\fB
 Input prompt (default: '> ')
 .TP
 .BI "--pointer=" "STR"
-Pointer sign of cursor line (default: '>')
+Pointer to the current line (default: '>')
 .TP
 .BI "--marker=" "STR"
-Signs of selected ling in multi-mode  (default: '>')
+Multi-select marker (default: '>')
 .TP
 .BI "--header=" "STR"
 The given string will be printed as the sticky header. The lines are displayed
--- a/shell/completion.bash
+++ b/shell/completion.bash
@@ -72,6 +72,8 @@ _fzf_opts_completion() {
     --margin
     --inline-info
     --prompt
+    --pointer-sign
+    --selected-sign
     --header
     --header-lines
     --ansi
--- a/shell/completion.bash
+++ b/shell/completion.bash
@@ -73,7 +73,7 @@ _fzf_opts_completion() {
     --inline-info
     --prompt
     --pointer
-    --selected
+    --marker
     --header
     --header-lines
     --ansi
--- a/shell/completion.bash
+++ b/shell/completion.bash
@@ -72,8 +72,8 @@ _fzf_opts_completion() {
     --margin
     --inline-info
     --prompt
-    --pointer-sign
-    --selected-sign
+    --pointer
+    --selected
     --header
     --header-lines
     --ansi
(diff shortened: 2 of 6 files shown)


relastle:
Sorry, this PR is currently behind the upstream HEAD, i will fix it now.

junegunn:
Thanks for your interest in the project. This was suggested a few times in the past, but I rejected the idea because I felt that adding it would eventually lead to feature creep. You can read my previous comments on that matter:

- https://github.com/junegunn/fzf/issues/692#issuecomment-251436739
- https://github.com/junegunn/fzf/issues/1769#issuecomment-557946235

Fast forward a few years, I'm not sure if I still feel the same. But you made a good point that using different marker signs can provide additional context to the users, so that this is not purely cosmetic.

There are a few things I'd do differently:

- We already use the terms "pointer" and "marker" (see `man fzf`), so I would go for more concise option names: `--pointer` and `--marker`
- I'd like to limit the width of the signs to 2 (Emojis or CJK wide characters are allowed), and make sure that they don't contain `\n` or `\t` so we have less to worry about screen overflow issues: (`fzf --pointer "$(seq 100)"`)
- Make it compatible with `jump` feature.
- Test cases would be nice.

relastle:
@junegunn 

Thank you for the insightful review comments.

>We already use the terms "pointer" and "marker" (see man fzf), so I would go for more concise option names: --pointer and --marker

I'm sorry for not following the existing naming concistency.
It was fixed in (https://github.com/junegunn/fzf/pull/1844/commits/2aff65ccafcd306946e439c1793f9d0301dd1e6d)

> I'd like to limit the width of the signs to 2 (Emojis or CJK wide characters are allowed), and make sure that they don't contain \n or \t so we have less to worry about screen overflow issues: (say no to fzf --pointer "$(seq 100)")

That's really reasonable. I added option parameter validation and some table-driven test-cases (https://github.com/junegunn/fzf/pull/1844/commits/c54dbc30d13a7636abc422efecdf74d6d8eda521) 

> Make it compatible with jump feature.

I forgot to test new functions working with that feature.
It was fixed in (https://github.com/junegunn/fzf/pull/1844/commits/df6b1441f29ffdf32c79c75100a6d2a169c5d81f)

-- -

Finally, I personally don't think that `pointer` and `marker` display length should be absolutely less than or equal `2`. (i.g. Someone may want to use `>>>` as pointer 🤔 )
So I implemented the logics so that only few changes should be done when the limit of display length would be longer (than 2) in the future.

relastle:
fzf_jump_demo

Here is the demonstration of multi-display-width marker and pointer working with the jump feature.

man/man1/fzf.1
```diff
@@ -231,6 +231,12 @@ A synonym for \fB--info=hidden\fB
 .BI "--prompt=" "STR"
 Input prompt (default: '> ')
 .TP
+.BI "--pointer=" "STR"
+Pointer sign of cursor line (default: '>')
+.TP
+.BI "--marker=" "STR"
+Signs of selected ling in multi-mode  (default: '>')
```
junegunn: Let's just use the same expressions in `--color` section of the man page.

```
pointer    Pointer to the current line
marker     Multi-select marker
```

shell/completion.bash
```diff
@@ -72,6 +72,8 @@ _fzf_opts_completion() {
     --margin
     --inline-info
     --prompt
+    --pointer
+    --selected
```
junegunn: `--marker`

src/options.go
```diff
@@ -1333,6 +1354,33 @@ func parseOptions(opts *Options, allArgs []string) {
...
+
+	if validateMarker {
+		if err := validateSign(opts.Marker, "marker"); err != nil {
+			errorExit(err.Error())
+		}
+	}
+}
+
+func validateSign(sign string, signOptName string) error {
+	if strings.Contains(sign, "\n") || strings.Contains(sign, "\t") {
```
junegunn: It would be more precise to use https://golang.org/pkg/unicode/#IsGraphic
Error message can be `invalid character in %v`

src/options.go
```diff
@@ -59,6 +60,8 @@ const usage = `usage: fzf [options]
     --margin=MARGIN       Screen margin (TRBL / TB,RL / T,RL,B / T,R,B,L)
     --info=STYLE          Finder info style [default|inline|hidden]
     --prompt=STR          Input prompt (default: '> ')
+    --pointer=STR         Cursor line pointer sign (default: '>')
+    --marker=STR          Selected sign in multi-select mode (default: '>')
```
junegunn: Also, the same expressions:

```
Pointer to the current line (default: '>')
Multi-select marker (default: '>')
```

relastle:
@junegunn 

Thank you for your reviews!

I amended the description of the options and added three integration test-cases.

src/options.go
```diff
@@ -1333,6 +1355,32 @@ func parseOptions(opts *Options, allArgs []string) {
...
+	widthSum := 0
+	for _, r := range sign {
+		if !unicode.IsGraphic(r) {
+			return fmt.Errorf("invalid character in %v", signOptName)
+		}
+		widthSum += runewidth.RuneWidth(r)
+	}
+	if widthSum > 2 {
+		return fmt.Errorf("%v display width should be up to 2", signOptName)
+	}
```
junegunn: We can put `if` statement inside `for` loop for early termination.

relastle: Thank you.

fixed in (ddc7435)

relastle:
> We should disallow empty pointer or marker.

Exactly. Validation for empty string and its test-case was added in (5b8bca2)

junegunn:
Merged, thanks a lot!

relastle:
Thank you, too!

PR #1847, opened 2020-02-03
Use native walker since output of DOS command is not UTF-8 encoded (Makes fzf 300x faster on Windows)

mattn:
Current implementation of walker on Windows use MSDOS command. But the output of the command is not UTF-8 encoded. Also heavily slow.

image

This pull-request delete it and add implementation of native walker. This works 300 times faster on Windows on my environment.


mattn:
old
terminal7

new
terminal7

mattn:
This implementation ignore hidden files. @junegunn If you don't like, I fix this to show all.

src/reader.go
```diff
@@ -144,6 +150,25 @@ func (r *Reader) readFromStdin() bool {
...
+			if r.pusher([]byte(path)) {
+				atomic.StoreInt32(&r.event, int32(EvtReadNew))
+			}
+		}
+		return nil
+	}
+	cb := walker.WithErrorCallback(func(pathname string, err error) error {
+		return nil
+	})
+	return walker.Walk(".", fn, cb) == nil
```
junegunn: Thanks @mattn!

Looks like we can just omit error callback here. i.e. `return walker.Walk(".", fn) == nil`

mattn: Default behavior of walker stop walking when error found. For example, the directory not permitted.

junegunn: Thanks for the clarification.

junegunn:
Would it be possible to make the loop terminate early on `reload` action? Will it make it much slower?

```sh
FZF_DEFAULT_COMMAND= fzf --bind 'space:reload:echo foo'
```

mattn:
Added canceling

junegunn:
Thanks, do you think it's okay to read `r.killed` without locking the mutex? Can we ignore complaints from `GORACE` in this context?

mattn:
Done

junegunn:
Do you notice any slowdown because of the locking?

mattn:
Should I modify title of this issue to "250x times faster" ?

junegunn:
😲 

So not much difference? I was just wondering how much performance impact it would make, though I was pretty sure that it's not going to dwarf the initial performance improvement. Anyway, thanks, I'll merge this now.

mattn:
Thank you.

janlazo:
Does the native file walker work on directory symlinks and junctions?

mattn:
No but this seems to be intended.

https://github.com/saracen/walker/blob/324a081bae7e580aa0bf3afe8164acb16634afca/walker.go#L85

Issue #1856, opened 2020-02-06
set ambiwidth=double breaks popup border layouts


kyoh86:
- [x] I have read through the manual page (`man fzf`)
- [x] I have the latest version of fzf
  - v0.20.0
- [x] I have searched through the existing issues

## Info

- OS
    - [ ] Linux
    - [x] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [x] bash
    - [x] zsh
    - [ ] fish
    
## Problem / Steps to reproduce

```
:let g:fzf_layout = { 'window': { 'width': 0.6, 'height': 0.6, 'rounded': v:false } }
:set ambiwidth=double
:Files
```

<img width="954" alt="068a6322433b1a7acf5ced08a92ea04b" src="https://user-images.githubusercontent.com/5582459/73912806-d1f7f900-48f8-11ea-97ff-e17cbf8d645a.png">

kyoh86:
it will be fixed by:

```diff
diff --git a/plugin/fzf.vim b/plugin/fzf.vim
index 7e3046a..bf3c987 100644
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -844,6 +844,9 @@ endif
 function! s:popup(opts) abort
   " Size and position
   let width = min([max([0, float2nr(&columns * a:opts.width)]), &columns])
+  if &ambiwidth == 'double'
+    let width += width % 2
+  endif
   let height = min([max([0, float2nr(&lines * a:opts.height)]), &lines - has('nvim')])
   let row = float2nr(get(a:opts, 'yoffset', 0.5) * (&lines - height))
   let col = float2nr(get(a:opts, 'xoffset', 0.5) * (&columns - width))
@@ -860,16 +863,21 @@ function! s:popup(opts) abort
     let style = 'sharp'
   endif
 
+  let ambidouble = 1
+  if &ambiwidth == 'double'
+    let ambidouble = 2
+  endif
+
   if style == 'horizontal'
-    let hor = repeat('─', width)
+    let hor = repeat('─', width / ambidouble)
     let mid = repeat(' ', width)
     let border = [hor] + repeat([mid], height - 2) + [hor]
     let margin = 0
   else
     let edges = style == 'sharp' ? ['┌', '┐', '└', '┘'] : ['╭', '╮', '╰', '╯']
-    let bar = repeat('─', width - 2)
+    let bar = repeat('─', width / ambidouble - 2)
[...]

PR #1857, opened 2020-02-06
add "ascii" option and focus "ambiwidth" option

kyoh86:
fix: #1856


kyoh86:
it's alternative for #1861 .

junegunn said: 
> let's keep it simple (until someone actually complains)

agree. so I close this one.

PR #1861, opened 2020-02-10
consider ambiwidth for border

### LINKED ISSUE #1856: set ambiwidth=double breaks popup border layouts

kyoh86:
fix: #1856

Both of this and #1857 will fix #1856,
but this one won't create new option "ascii".

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -848,7 +848,6 @@ function! s:popup(opts) abort
   " Size and position
   let width = min([max([0, float2nr(&columns * a:opts.width)]), &columns])
   let width += width % ambidouble
-
   let height = min([max([0, float2nr(&lines * a:opts.height)]), &lines - has('nvim')])
   let row = float2nr(get(a:opts, 'yoffset', 0.5) * (&lines - height))
   let col = float2nr(get(a:opts, 'xoffset', 0.5) * (&columns - width))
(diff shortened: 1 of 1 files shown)


plugin/fzf.vim
```diff
@@ -860,16 +863,21 @@ function! s:popup(opts) abort
     let style = 'sharp'
   endif
 
+  let ambidouble = 1
+  if &ambiwidth == 'double'
+    let ambidouble = 2
+  endif
```
junegunn: I would suggest simplifying this using ternary operator:

```vim
let ambidouble = &ambiwidth == 'double' ? 2 : 1
```

plugin/fzf.vim
```diff
@@ -844,6 +844,9 @@ endif
 function! s:popup(opts) abort
   " Size and position
   let width = min([max([0, float2nr(&columns * a:opts.width)]), &columns])
+  if &ambiwidth == 'double'
```
junegunn: Let's calculate `ambidouble` before this line and use it (because `x % 1` is always 0)

```vim
let width += width % ambidouble
```

kyoh86: thanks

kyoh86:
@junegunn thank you for your review!
I fixed.

junegunn:
Merged, thanks!

PR #1864, opened 2020-02-10
terminal.go: fix double-lock by adding missing unlock in loop

BurtonQin:
t.mutex.Lock() is on Line 1710 in a loop.
The loop starts from Line 1703 to 2140. 
When the loop continue without unlock on Line 2096 and 2103,
double lock happens.

### DIFF
--- a/src/terminal.go
+++ b/src/terminal.go
@@ -2093,6 +2093,7 @@ func (t *Terminal) Loop() {
 				}
 			}
 			if !doActions(actions, mapkey) {
+				t.mutex.Unlock()
 				continue
 			}
 			t.truncateQuery()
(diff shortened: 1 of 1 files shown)


junegunn:
Interesting, some kind of static analysis?
But in this case, `doActions` always unlocks the mutex when before it returns false: 
- https://github.com/junegunn/fzf/blob/02c5e62efeda372ff918eb72fec3c1d2829275b3/src/terminal.go#L2007-L2008
- https://github.com/junegunn/fzf/blob/02c5e62efeda372ff918eb72fec3c1d2829275b3/src/terminal.go#L1766-L1767

So this change will introduce "double unlocking". Looks like we need some code cleanup nevertheless. Thanks.

PR #1867, opened 2020-02-11
add {r}aw flag to disable quoting in templates

wrboyce:


wrboyce:
@junegunn anything I can do to assist in getting this PR merged? Cheers

junegunn:
I don't see why this is needed. Any reason you can't use the existing `f` flag to get unquoted items? Injecting unquoted items to the command string is not safe and it can cause unexpected results.

DanSM-5:
Is this {r}aw flag not going to be implemented?

I found this PR while looking for a way to get unquoted strings for the preview. I wanted to get an `ls -A` on some directories but the quoted tilda `~` in some of the paths make it imposible to work with ls. E.g. `ls -A '~/files'` results in `ls: cannot access '~/files': No such file or directory`.

I'm doing a workaround by replacing the `~` with the value of `$HOME` but I can see value on the proposed `{r}`.

wrboyce:
Well, [31fd207b](https://github.com/junegunn/fzf/commit/31fd207b) feels a bit like poor form really but whatever at least the feature has finally landed.

PR #1876, opened 2020-02-16
[bash] Put C-t items at point in vi mode

jablko:
Be consistent with emacs mode and put the items at the point vs. the end
of the command line.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,27 +1,22 @@
 language: go
-dist: xenial
+env: GO111MODULE=on
+os:
+  - linux
+  - osx
+dist: bionic # For fish >= 2.3.0 string builtin
 addons:
   apt:
-    sources:
-      - sourceline: "ppa:pi-rho/dev"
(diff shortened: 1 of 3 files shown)


junegunn:
I really like this approach, thanks. I'm a bit concerned with these new bindings though. 

```sh
bind -m vi-command '"\C-e": emacs-editing-mode'
bind -m vi-insert '"\C-e": emacs-editing-mode'
```

`CTRL-E` is such an "accessible", easy key chord, that I can imagine people already have their own bindings on it. How about if we choose some esoteric key sequence that is less likely to conflict? We've been using `CTRL-X CTRL-E`, so it's a good candidate. What do you think?

jablko:
How about <kbd>C-z</kbd> like in [Viper](https://www.gnu.org/software/emacs/manual/html_mono/viper#States-in-Viper) or [Evil](https://evil.readthedocs.io/en/latest/settings.html#elispobj-evil-toggle-key)?

junegunn:
Superb, thanks!

PR #1878, opened 2020-02-18
Add explanation for the `g:fzf_color` vim setting

tartansandal:
Getting my vim colors to work with  fzf was quite a challenge. After some experimentation and "code diving" I think I've made sense of it. Adding some notes on this to the README-VIM file that may help others facing similar issues.

### DIFF
--- a/README-VIM.md
+++ b/README-VIM.md
@@ -159,31 +159,34 @@ let g:fzf_colors =
 let g:fzf_history_dir = '~/.local/share/fzf-history'
 ```
 
-The above `g:fzf_colors` setting is a little tricky.
+##### Explanation of `g:fzf_colors`
 
-This variable is a dictionary mapping fzf elements to a color specification
+`g:fzf_colors` is a dictionary mapping fzf elements to a color specification
 list:
 
-    element: [ layer, group1 (, group2, ...) ]
+    element: [ component, group1 [, group2, ...] ]
 
-- `element` is the fzf element to apply a color to:
+- `element` is an fzf element to apply a color to:
 
-  - **fg  / bg  / hl** : non-selected lines (foreground / background / highlight)
-  - **fg+ / bg+ / hl+** : selected lines  (foreground / background / highlight)
-  - **hl  / hl+** : strings matching the search (normal / selected)
-  - **gutter** : the background of the first 2 columns
-  - **pointer / marker** : the '>' indicator in the gutter
-  - **border** : the border around terminal window
-  - **header** : the line containing the current file
-  - **info** : match counters
-  - **spinner** : the char at column 1 before the match counters
-  - **prompt** : the prompt before input strings
+  | Element               | Description                                           |
+  | ---                   | ---                                                   |
+  | `fg`  / `bg`  / `hl`  | Item (foreground / background / highlight)            |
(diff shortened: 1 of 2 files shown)


junegunn:
Thanks.

> For the **fg+**, **bg+**, **hl+**, **prompt**, **pointer**, and **marker**
> elements there is an extra condition: the color only matches if the highlight
> group also has `term=bold` set.  This last bit can cause a lot of confusion.

I'm not sure about this. There is no such restriction, and if it works that way, it's probably a bug.

tartansandal:
> I'm not sure about this. There is no such restriction, and if it works that way, it's probably a bug.

Glad to here it. It seemed a very strange restriction to me. 

I've had a go at finding minimal test case for a possible bug. Sadly it seems the issue is not as simple as I described. There is merely a confluence of the groups that behaved as I expected and those with term=bold set, which is probably a peculiarity of my colorscheme.  I'll dig a deeper still if I have a chance.

README-VIM.md
```diff
@@ -159,6 +159,47 @@ let g:fzf_colors =
 let g:fzf_history_dir = '~/.local/share/fzf-history'
 ```
 
+The above `g:fzf_colors` setting is a little tricky.
```
LuRsT: I'd say this is unnecessary to add, let people judge how tricky it is on their own :)

junegunn:
Merged, thanks. Let me know if you find out what's going on with `term=bold` thing.
Also note that `fzf#wrap()` will generate 24-bit colors if `termguicolors` is used.

```vim
:set termguicolors
:echo fzf#wrap()
{'options': '''--color=bg+:#3F3F3F,bg:#4B4B4B,border:#6B6B6B,spinner:#98BC99,hl:#719872,fg:#D9D9D9,header:#719872,info:#BDBB72,pointer:#E12672,marker:#E17899,fg+:#D9D9D9,prompt:#98BEDE,hl+:#98BC99''  --expect
=ctrl-v,ctrl-x,ctrl-t', 'sink*': function('64'), 'window': {'width': 0.9, 'height': 0.6}, '_action': {'ctrl-v': 'vsplit', 'ctrl-x': 'split', 'ctrl-t': 'tab split'}}

:set notermguicolors
:echo fzf#wrap()
{'options': '''--color=bg+:236,bg:237,border:240,spinner:108,hl:65,fg:252,header:65,info:143,pointer:161,marker:168,fg+:252,prompt:110,hl+:108''  --expect=ctrl-v,ctrl-x,ctrl-t', 'sink*': function('65'), 'wind
ow': {'width': 0.9, 'height': 0.6}, '_action': {'ctrl-v': 'vsplit', 'ctrl-x': 'split', 'ctrl-t': 'tab split'}}
```

tartansandal:
Thanks for sorting out the merge @junegunn! Sorry that I haven't had a chance to look at this, and might not for a while. I've just started a new job a couple of weeks ago and its been frantic. I'm pretty sure the issue I was seeing had something to do with, gui, `&termguicolors`, a weird darcula colorscheme, `g:terminal_ansi_colors`, and possibly `t_Co=256`. Hope to get a chance to figure it out in a couple of weeks. :-)

junegunn:
No hurries, thanks.

PR #1881, opened 2020-02-21
[bash] Restore insertion point pre Bash 4

jablko:
Make <kbd>C-t</kbd> more consistent pre and post Bash 4. It already kills the
command line separately before and after the insertion point. Add
[`set-mark`](https://www.gnu.org/software/bash/manual/bash#index-set_002dmark-_0028C_002d_0040_0029) and [`exchange-point-and-mark`](https://www.gnu.org/software/bash/manual/bash#index-exchange_002dpoint_002dand_002dmark-_0028C_002dx-C_002dx_0029) to restore the insertion point
after yanking back and apply the same behavior to <kbd>M-c</kbd>.


junegunn:
Thanks, I merged this with a minor tweak.

jablko:
:+1: Thanks!

PR #1886, opened 2020-02-23
[bash] Start C-r search with current command line

### LINKED ISSUE #432: Ctr-r shuld use the current command as query as default Ctr-r do in bash

### LINKED ISSUE #731: Start the history search with the word you typed at the prompt in bash

jablko:
Retain the current command line when aborting <kbd>C-r</kbd>. Add `--query
"$READLINE_LINE"` and fall back to the current behavior pre Bash 4.

Fixes #432
Fixes #731

### DIFF
--- a/test/test_go.rb
+++ b/test/test_go.rb
@@ -63,8 +63,6 @@ class Shell
 end
 
 class Tmux
-  TEMPNAME = '/tmp/fzf-test.txt'
-
   attr_reader :win
 
   def initialize(shell = :bash)
@@ -119,12 +117,7 @@ class Tmux
   end
 
   def capture(pane = 0)
-    File.unlink TEMPNAME while File.exist? TEMPNAME
-    wait do
-      go("capture-pane -t #{win}.#{pane} \\; save-buffer #{TEMPNAME} 2> /dev/null")
-      $CHILD_STATUS.exitstatus.zero?
-    end
-    File.read(TEMPNAME).split($INPUT_RECORD_SEPARATOR)[0, @lines].reverse.drop_while(&:empty?).reverse
+    go("capture-pane -p -t #{win}.#{pane}")
   end
 
   def until(refresh = false, pane = 0)
@@ -1714,7 +1707,10 @@ end
 
 module TestShell
   def setup
-    super
+    @tmux = Tmux.new shell
+
+    # Why is this necessary? To wait until the shell is ready?
+    tmux.prepare
   end
 
   def teardown
@@ -1829,6 +1825,16 @@ module TestShell
     tmux.until { |lines| lines[-1] == 'echo 3rd' }
     tmux.send_keys :Enter
     tmux.until { |lines| lines[-1] == '3rd' }
+  end
+
+  def test_ctrl_r_abort
+    skip "doesn't restore the original line when search is aborted pre Bash 4" if shell == :bash && /(?<= version )\d+/.match(`#{Shell.bash} --version`).to_s.to_i < 4
+    tmux.send_keys 'foo'
+    tmux.until { |lines| lines[-1].start_with? 'foo' }
+    tmux.send_keys 'C-r'
+    tmux.until { |lines| lines[-1].start_with? '>' }
+    tmux.send_keys 'C-g'
+    tmux.until { |lines| lines[-1].start_with? 'foo' }
   end
 
   def retries(times = 3)
@@ -2020,15 +2026,14 @@ class TestBash < TestBase
   include TestShell
   include CompletionTest
 
+  def shell
+    :bash
+  end
+
   def new_shell
     tmux.prepare
     tmux.send_keys "FZF_TMUX=1 #{Shell.bash}", :Enter
     tmux.prepare
-  end
-
-  def setup
-    super
-    @tmux = Tmux.new :bash
   end
 
   def test_dynamic_completion_loader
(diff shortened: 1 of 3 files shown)


junegunn:
I pushed a commit to your branch, please take a look and tell me what you think.

jablko:
Thanks! Looks good, except that in my tests the command line was already restored when fzf is interrupted?

I incorporated your changes to the Bash 3 command line (although testing `$READLINE_POINTER` to see if was called by `bind -x`). Thanks! Please let me know if it doesn't actually restore the command line when fzf is interrupted, or restore your changes.

junegunn:
Maybe we have different definitions of "restored"? The command-line is reset if you terminate fzf because you modified `$READLINE_LINE`.

```sh
# On bash 4
$ git add<CTRL-R>
# fzf pops up
# CTRL-C to stop fzf
$
  # We are left with an empty command-line.
  # Would be better if it restores "git add", it's how CTRL-R of zsh and fish work
```

jablko:
No, same definition ... On my machine (also Bash 4) if I interrupt fzf (with <kbd>C-c</kbd>) I get `git add` back at the command line, same as if I abort with <kbd>C-g</kbd> ... I'll continue to investigate.

junegunn:
```sh
FOO=$(seq 10 | fzf); echo "Foo is $FOO"
  # Enter: Foo is 1
  # CTRL-C: Foo is 
```
Similarly `READLINE_LINE` is set to an empty string when you interrupt fzf with CTRL-C. And that's why I assigned the value to a different variable first and suspended the execution with `|| return` when fzf is interrupted.

You can easily see what's going on with `set -x`.

```sh
++ __fzf_history__
+++ builtin fc -lnr -2147483648
++++ set +o history
++++ fc -l 0
++++ __fzfcmd
++++ __fzf_use_tmux__
++++ '[' -n %81 ']'
++++ '[' 0 '!=' 0 ']'
++++ echo fzf
+++ last_hist='87837     unset FZF_DEFAULT_OPTS'
+++ perl -p -l0 -e 'BEGIN { getc; $/ = "\n\t"; $HISTCMD = $ENV{last_hist} + 1 } s/^[ *]//; $_ = $HISTCMD - $. . "\t$_"'
++++ printf %q 'git add'
+++ FZF_DEFAULT_OPTS='--height 40%  --tiebreak=index --bind=ctrl-r:toggle-sort  +m --query git\ add --read0'
+++ fzf
++ READLINE_LINE=
++ READLINE_LINE=
++ '[' -z 7 ']'
++ READLINE_POINT=0x7fffffff
```

test/test_go.rb
```diff
@@ -1831,6 +1824,20 @@ def test_ctrl_r
     tmux.until { |lines| lines[-1] == '3rd' }
   end
 
+  def test_ctrl_r_abort
+    skip "doesn't restore the original line when search is aborted pre Bash 4" if self.is_a?(TestBash) && /(?<= version )\d+/.match(`#{Shell.bash} --version`).to_s.to_i < 4
+    # Why is this necessary? To wait until the shell is ready?
+    tmux.send_keys 'echo hello', :Enter
+    tmux.until { |lines| lines[-1] == 'hello' }
```
junegunn: Yeah, I guess. You probably should use `tmux.prepare` instead, which also retries for a few time. It's used all over the test cases.

jablko: Thanks! I'm skeptical about that method in this context: It might add enough delay for the test to succeed, but is keystrokes echoed to the terminal confirmation that the shell is ready? Then again I don't 100% understand why this test doesn't succeed straight up. And if the line subsequently starts with `foo`, that's confirmation that <kbd>C-u</kbd> was handled by the shell, so it's ready. I'll give it a try! :+1:

junegunn: > is keystrokes echoed to the terminal confirmation that the shell is ready?

Maybe it's just better than nothing? But I agree that it's not sufficient. `git blame` on the method shows that we used to send `Right` key as well (https://github.com/junegunn/fzf/commit/400e443a0a397e559250fa090b5c437e44d01bf8), but for some reason I don't quite remember I dropped it. It makes more sense to have it because if the shell is not ready, the command line will print something like `foo^[[C`, so the subsequent `end_with?` test will fail.

So I just pushed https://github.com/junegunn/fzf/commit/2ff7db1b36e4f976e4e24e56e996b9128f131141, which hopefully works better.

junegunn: > that's confirmation that C-u was handled by the shell, so it's ready.

Yep, that was the intention. But it looks like CTRL-U can be handled even when the shell is not available, i.e. `sleep 60` and `foo<CTRL-U>` just works. I don't know if something has changed over the years. But anyway, cursor keys still leaves `^[[D` or `^[[C` on the screen, so I believe 2ff7db1 should definitely help.

test/test_go.rb
```diff
@@ -1714,7 +1707,10 @@ def test_clear_selection
 
 module TestShell
   def setup
-    super
+    @tmux = Tmux.new shell
+
+    # Why is this necessary? To wait until the shell is ready?
```
junegunn: We can remove this comment.

junegunn:
Merged, thanks!

Issue #1887, opened 2020-02-24
_fzf_complete function can not be interrupted


doubleloop:
- [x ] I have read through the manual page (`man fzf`)
- [x ] I have the latest version of fzf
- [x ] I have searched through the existing issues

## Info

- OS
    - [ x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [ ] bash
    - [x ] zsh
    - [ ] fish
    
## Problem / Steps to reproduce

Define custom completin:
```zsh
_foo() {
    for i in {1..5}; do
        echo "completion$i"
    done
}

_bar() {
    # this is some slower command, or command that may hang
    sleep 10
    for i in {6..10}; do
        echo -e "bar completion$i"
    done
}

_fzf_complete_foobar() {
    _fzf_complete "--multi --ansi -n 2" "$@" < <(
        _foo
        _bar
    )
}
```

This completion, when triggered, will allow to select foo completions instantly, but the shell is not responding until `_bar` finishes.

junegunn:
Thanks, I've done some research and here's what I've found.

### bash

bash has the same problem, but it is trivial to fix it. All I need to do is to remove unnecessary `cat`

```diff
diff --git a/shell/completion.bash b/shell/completion.bash
index b953cc8..bd94c51 100644
--- a/shell/completion.bash
+++ b/shell/completion.bash
@@ -200,7 +200,7 @@ _fzf_complete() {
   if [[ "$cur" == *"$trigger" ]]; then
     cur=${cur:0:${#cur}-${#trigger}}
 
-    selected=$(cat | FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} --reverse $FZF_DEFAULT_OPTS $FZF_COMPLETION_OPTS $1" __fzf_comprun "$2" -q "$cur" | $post | tr '\n' ' ')
+    selected=$(FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} --reverse $FZF_DEFAULT_OPTS $FZF_COMPLETION_OPTS $1" __fzf_comprun "$2" -q "$cur" | $post | tr '\n' ' ')
     selected=${selected% } # Strip trailing space not to repeat "-o nospace"
     if [ -n "$selected" ]; then
       COMPREPLY=("$selected")
```

And we can even kill the running processing with its process ID bash reports, `$!`.

```sh
_fzf_complete_f() {
[...]

PR #1888, opened 2020-02-24
Extended custom completions API

doubleloop:
Fizes #1887. This is an alternative approach to #1895 PR.

With this PR completion causing issue:
```zsh
_fzf_complete_foobar() {
    _fzf_complete "--multi --ansi -n 2" "$@" < <(
        echo foo
        sleep 10
        echo bar
    )
}
```

can be transformed to:
```zsh
_fzf_complete_foobar_func() {
    echo foo
    sleep 10
    echo bar
}

_fzf_complete_foobar() {
    _fzf_complete_async _fzf_complete_foobar_func "--multi --ansi -n 2" "$@"
}
```
which can be interrupted (ctrl-c) or foo can be selected.

Wandering if `_fzf_complete_async` is the best name, any suggestions for better name?
If this approach is OK, I can add bash/fish version.

### DIFF
--- a/shell/completion.zsh
+++ b/shell/completion.zsh
@@ -124,6 +124,30 @@ _fzf_complete() {
   command rm -f "$fifo"
 }
 
+_fzf_complete_async() {
+  setopt localoptions nomonitor
+  local fifo fzf_opts lbuf cmd matches post func pid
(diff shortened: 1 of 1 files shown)


junegunn:
https://github.com/junegunn/fzf/issues/1887#issuecomment-596057258

PR #1891, opened 2020-02-25
feat: Add action for backspace if query empty

### LINKED ISSUE #1797: Abort on <BS> if prompt is empty

## Info

- OS
    - [ ] Linux
    - [x] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [ ] bash
    - [x] zsh
    - [ ] fish
    
## Feature Request
It would be really convenient to add functionality to abort FZF if current search line is empty and backspace is pressed. Howerver, it should not abort if backspace was held to actually delete characters as it would be hard to unpress arrow button right before it reaches the start of line to avoid accidental exiting. Is it possible to implement?

jrwrigh:
Action 'backward-delete-char-non-empty' will either abort if query if empty or delete one character backwards.

This is more geared towards something like [zsh-interactive-cd](https://github.com/changyuheng/zsh-interactive-cd), where pressing backspace on an empty query is the (at least my) natural tendency when you want to exit out of fzf.

This would partially resolve changyuheng/zsh-interactive-cd#5.

Also partially resolves #1797 . Doesn't implement the long-press repetition request though.

### DIFF
--- a/src/terminal.go
+++ b/src/terminal.go
@@ -187,6 +187,7 @@ const (
 	actAcceptNonEmpty
 	actBackwardChar
 	actBackwardDeleteChar
+    actBackwardDeleteCharNonEmpty
 	actBackwardWord
 	actCancel
 	actClearScreen
@@ -1845,6 +1846,13 @@ func (t *Terminal) Loop() {
 					t.yanked = t.input
 					t.input = []rune{}
 					t.cx = 0
+				}
+			case actBackwardDeleteCharNonEmpty:
+				if len(t.input) == 0 {
+					req(reqQuit)
+				} else {
+					t.input = append(t.input[:t.cx-1], t.input[t.cx:]...)
+					t.cx--
 				}
 			case actForwardChar:
 				if t.cx < len(t.input) {
--- a/src/terminal.go
+++ b/src/terminal.go
@@ -187,7 +187,7 @@ const (
 	actAcceptNonEmpty
 	actBackwardChar
 	actBackwardDeleteChar
-    actBackwardDeleteCharNonEmpty
+    actBackwardDeleteCharEOF
 	actBackwardWord
 	actCancel
 	actClearScreen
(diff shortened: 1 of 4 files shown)


jrwrigh:
I think what would be a bit more flexible would be to add an event that basically says any "deletion" event on an empty query should abort fzf, where "deletion" would be backwards-char, backwards-word, clear, etc. I just wasn't sure how to do an event for something like that. 

I can open a separate issue if desired for a more general event.

junegunn:
- It should be named `backward-delete-char/eof` to be consistent with the similar existing acttion (https://github.com/junegunn/fzf/commit/13e040baee77d75736f9fbb53f269d5ddc97efb9).
- You should check if `t.cx` is positive before decrementing it.
- Please update the man page and add a test case to `test/test_go.rb`.

jrwrigh:
I think I'm good to go. Only thing I'm not sure of if my implementation of the test function. I pretty much copied after what `test_cancel` used as far as testing the abort capabilities of the action.

Also, I added a snippet to the `delete-char/eof` action on the man page since it wasn't really documented before. I can remove that commit if you want.

Lastly, if we're all good, I can squash the first two commits together since the second is just a fix of the first. (assuming you're not going to squash merge the whole PR).

src/terminal.go
```diff
@@ -1846,6 +1847,13 @@ func (t *Terminal) Loop() {
 					t.input = []rune{}
 					t.cx = 0
 				}
+			case actBackwardDeleteCharEOF:
+				if len(t.input) == 0 {
+					req(reqQuit)
+				} else if t.cx > 0 {
+                    t.input = append(t.input[:t.cx-1], t.input[t.cx:]...)
+                    t.cx--
```
junegunn: Go uses tab instead of spaces for indentation.  Please `gofmt` the file.

test/test_go.rb
```diff
@@ -1710,6 +1710,20 @@ def test_clear_selection
     tmux.until { |lines| lines.match_count.zero? }
     tmux.until { |lines| !lines[-2].include?('(1)') }
   end
+
+  def test_backward_delete_char_eof
+    tmux.send_keys "seq 1000 | #{fzf "--multi --jump-labels 12345 --bind 'bs:backward-delete-char/eof'"}", :Enter
```
junegunn: `jump-labels` and `multi` are irrelevant to the test, so let's remove them.

junegunn:
Looks good now. Thanks!

PR #1892, opened 2020-02-25
Test multiline C-r

jablko:


test/test_go.rb
```diff
@@ -88,30 +88,19 @@ def initialize(shell = :bash)
...
-      if pair.first == :Escape
-        arg = pair.compact.map { |key| %("#{key}") }.join(' ')
-        go(%(send-keys -t #{target} #{arg}))
-        enum.next if pair.last
-      else
-        go(%(send-keys -t #{target} "#{pair.first}"))
-      end
-      break unless pair.last
-    end
+    go(%W[send-keys -t #{target}] + args.map { |key| key.to_s })
```
junegunn: So you're reverting https://github.com/junegunn/fzf/commit/2b98fee13602a90bcfaf76a0d75388b76692a95f here. It was an attempt to mitigate intermittent Travis CI failures. It seemed at that time that sending keys one by one sequentially instead of sending them all at once, especially in the presence of escape sequences, improved the situation. Do you think we no longer need it?

jablko: Admittedly I have a lot less context. I have seen (current) intermittent failures but there doesn't appear to be any more or less with more or less calls to `send-keys` ...

My (baseless) suspicion is that more calls to `send-keys` added enough pauses to pave over some intermittent failures and the better resolution is to `tmux.until` a reaction before sending more keys, in the appropriate places? Sorry to revisit an old issue if that was already tried!

Would you be opposed to trying it out? I offer to check out any failures and restore the extra `send-keys` if dropping them was my mistake.

junegunn: I wish we could just get rid of all these hacks. Since the tests are currently passing on Travis CI and on my machine, and we just made some progress in stabilizing the test setup, I think I'm just gonna merge this and see what happens.

FYI, this is running on my machine right now:

```sh
make docker
while true; do /fzf/test/test_go.rb || break; done
```

junegunn:
Merged, thank you.

PR #1895, opened 2020-02-27
Extended custom completions API (alternative approach)

### LINKED ISSUE #1887: _fzf_complete function can not be interrupted

doubleloop:
Fixes #1887. This is an alternative approach to #1888 PR.

Instead of introducing extra function, reuse existing one.

With this PR completion causing issue:
```
_fzf_complete_foobar() {
    _fzf_complete "--multi --ansi -n 2" "$@" < <(
        echo foo
        sleep 10
        echo bar
    )
}
```
can be transformed to:

```
_fzf_complete_foobar_func() {
    echo foo
    sleep 10
    echo bar
}
_fzf_complete_foobar() {
    _fzf_complete "--multi --ansi -n 2" "$@"
}
```
which can be interrupted (`ctrl-c`) or `foo` can be selected

Wandering if `_func` postfix is OK (callback, source ?)

If this approach is OK, I can add bash/fish version.

### DIFF
--- a/shell/completion.zsh
+++ b/shell/completion.zsh
@@ -100,27 +100,29 @@ _fzf_dir_completion() {
     "" "/" ""
 }
 
-_fzf_feed_fifo() (
-  command rm -f "$1"
-  mkfifo "$1"
-  cat <&0 > "$1" &
-)
-
 _fzf_complete() {
-  local fifo fzf_opts lbuf cmd matches post
+  setopt localoptions nomonitor
+  local fifo fzf_opts lbuf cmd matches post func pid
   fifo="${TMPDIR:-/tmp}/fzf-complete-fifo-$$"
   fzf_opts=$1
   lbuf=$2
   cmd=$(__fzf_extract_command "$lbuf")
+  func="${funcstack[2]}_func"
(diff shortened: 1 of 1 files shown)


junegunn:
https://github.com/junegunn/fzf/issues/1887#issuecomment-596057258

PR #1897, opened 2020-02-28
Added auto select only filtered option

MaanooAk:
So basically what `--select-1` does but the check (for only one item) is performed every time the user types a character (interactive mode).

I added a new option instead of changing the `--select-1` behavior because this is more like a user choice than a "script" choice.

The name `--select-only` is just what came first to mind.

(I'm not very familiar with the code base...)

### DIFF
--- a/man/man1/fzf.1
+++ b/man/man1/fzf.1
@@ -152,6 +152,9 @@ actions are affected:
 .TP
 .BI "--jump-labels=" "CHARS"
 Label characters for \fBjump\fR and \fBjump-accept\fR
+.TP
+.B "--select-only"
+Automatically select the only filtered match
 .SS Layout
 .TP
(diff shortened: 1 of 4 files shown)


junegunn:
Thanks for the suggestion but I don't want to add this to fzf, because I don't need the feature myself and I believe it can be confusing in a number of scenarios. e.g. `(echo 1; sleep 1; echo 2) | fzf` or when you type in a search term with `!` prefix; the number of matches increases as you type. Please consider building fzf from your fork if this is a crucial feature for you.

MaanooAk:
I have been using it for a while myself, this is the reason a made the pr. I will not insist, its your project, but both the cases can be easily handled (wait for the input to end before auto select, and just do nothing is there is a `!`)

FYI I have been using it for fast terminal cd loop and single character select menu.

PR #1900, opened 2020-03-02
Prompt generally signals that the shell is ready

jablko:
What do you think of this (hopefully) simplified `tmux.prepare`? (relying on the prompt to signal that the shell is ready)


junegunn:
I don't think repeating `tmux.until { |lines| lines[-1] == '$' }` instead of `tmux.prepare` is an improvement. We can change the implementation of `tmux.prepare` method to rely on `$` prompt (which is probably a more robust way to check) but removing the layer of abstraction is not something we should do.

And it looks like you did a bunch of additional refactorings here, which is nice, but it's a different concern, so it should be in a separate commit (or in a separate pull request).

jablko:
:+1: Will do: I've replaced `tmux.until { |lines| lines[-1] == '$' }` with `tmux.prepare` and split apart changes that stand on their own.

test/test_go.rb
```diff
@@ -1748,120 +1725,111 @@ def test_ctrl_t
...
     skip "doesn't restore the original line when search is aborted pre Bash 4" if shell == :bash && /(?<= version )\d+/.match(`#{Shell.bash} --version`).to_s.to_i < 4
     %w[foo ' "].each do |query|
       tmux.prepare
       tmux.send_keys(query)
-      tmux.until { |lines| lines[-1].start_with? query }
+      tmux.until { |lines| lines[-1].start_with? "$ #{query}" }
+      # Keys are possibly echoed before the (fish) shell is ready for
+      # C-r? In this case there's no other feedback to check for so add
+      # an arbitrary delay.
+      sleep 0.5 if shell == :fish
```
junegunn: I'd like to avoid this arbitrary delay. This seems to the case where the old Left, Right hack should work well. What do you think?

```sh
tmux.send_keys(query, :Left, :Right)
# $ foo^[[D^[[C (not ready)
# $ foo (ready)
tmux.until { |lines| lines[-1] == "$ #{query}" }
```

test/test_go.rb
```diff
@@ -1935,61 +1903,54 @@ def test_dir_completion
...
-    lines = tmux.until { |lines| lines[-1].start_with? '[1]' }
-    pid = lines[-1].split.last
     tmux.prepare
-    tmux.send_keys 'C-L'
+    tmux.send_keys 'sleep 12345 &', :Enter
+    pid = nil
+    tmux.until { |lines| pid = /(?<= )\d+/.match lines[-2] }
     tmux.send_keys 'kill ', :Tab
-    tmux.until { |lines| lines.match_count.positive? }
+    tmux.until { |lines| lines.item_count.positive? }
```
junegunn: Can you explain why you changed `match_count` to `item_count`? Both will work here.

test/test_go.rb
```diff
@@ -1748,120 +1725,111 @@ def test_ctrl_t
...
-    lines = retries do
-      tmux.prepare
-      tmux.send_keys :Escape, :c
-      tmux.until { |lines| lines.match_count.positive? }
-    end
-    expected = lines.reverse.select { |l| l.start_with? '>' }.first[2..-1]
+    Dir.mkdir 'foo'
+    tmux.prepare
+    tmux.send_keys 'M-c'
+    tmux.until { |lines| lines.item_count.positive? }
```
junegunn: Same question: any reason you prefer `item_count` over `match_count`?

test/test_go.rb
```diff
@@ -1935,61 +1903,54 @@ def test_dir_completion
...
-    tmux.send_keys 'C-L'
+    tmux.send_keys 'sleep 12345 &', :Enter
+    pid = nil
+    tmux.until { |lines| pid = /(?<= )\d+/.match lines[-2] }
     tmux.send_keys 'kill ', :Tab
-    tmux.until { |lines| lines.match_count.positive? }
+    tmux.until { |lines| lines.item_count.positive? }
     tmux.send_keys 'sleep12345'
-    tmux.until { |lines| lines.any_include? 'sleep 12345' }
+    tmux.until { |lines| lines.match_count == 1 }
```
junegunn: How about changing this to `lines.match_count >= 1`? We can't be 100% sure if our process is the only match on the system.

test/test_go.rb
```diff
@@ -23,7 +23,6 @@
 
 FILE = File.expand_path(__FILE__)
 BASE = File.expand_path('..', __dir__)
-Dir.chdir(BASE)
```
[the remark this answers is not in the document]
jablko: I was mistaken about the current working directory inside tmux, but I've now replaced it with `Dir.chdir(Dir.mktmpdir)`.

test/test_go.rb
```diff
@@ -1748,120 +1725,111 @@ def test_ctrl_t
...
 
     tmux.send_keys '2'
-    tmux.until { |lines| lines.match_count == 1 }
+    tmux.until { |lines| lines[-1] == '> fzf-unicode 테스트2' }
     tmux.send_keys :Tab
     tmux.until { |lines| lines.select_count == 2 }
 
     tmux.send_keys :Enter
-    tmux.until { |lines| lines.join.match(/echo.*fzf-unicode.*1.*fzf-unicode.*2/) }
+    tmux.until { |lines| /echo .*fzf-unicode.*1.* .*fzf-unicode.*2/.match lines.join }
```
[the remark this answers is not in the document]
jablko: Yes, purely for symmetry with `test_file_completion_unicode`.

test/test_go.rb
```diff
@@ -1748,120 +1725,111 @@ def test_ctrl_t
...
   end
 
   def test_alt_c
-    lines = retries do
-      tmux.prepare
-      tmux.send_keys :Escape, :c
-      tmux.until { |lines| lines.match_count.positive? }
-    end
-    expected = lines.reverse.select { |l| l.start_with? '>' }.first[2..-1]
+    Dir.mkdir 'foo'
```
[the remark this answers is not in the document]
jablko: :+1: Done!

test/test_go.rb
```diff
@@ -1748,120 +1725,111 @@ def test_ctrl_t
...
-      tmux.prepare
-      tmux.send_keys 'C-r'
-      tmux.until { |lines| lines.match_count.positive? }
-    end
+    tmux.prepare
     tmux.send_keys 'C-r'
+    tmux.until { |lines| lines[-1] == '>' }
     tmux.send_keys '3d'
-    tmux.until { |lines| lines[-3].end_with? 'echo 3rd' }
+    tmux.until { |lines| lines[-3].end_with? 'echo 3d' }
```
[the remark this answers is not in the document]
jablko: :heavy_check_mark: Will do! But for the time being I've undone the changes.

test/test_go.rb
```diff
@@ -1748,120 +1725,111 @@ def test_ctrl_t
...
       tmux.send_keys(query)
-      tmux.until { |lines| lines[-1].start_with? query }
+      tmux.until { |lines| lines[-1].start_with? "$ #{query}" }
+      # Keys are possibly echoed before the (fish) shell is ready for
+      # C-r? In this case there's no other feedback to check for so add
+      # an arbitrary delay.
+      sleep 0.5 if shell == :fish
       tmux.send_keys 'C-r'
-      tmux.until { |lines| lines[-1] == "> #{query}" }
+      tmux.until { |lines| lines[-1].start_with? '>' }
```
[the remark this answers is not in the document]
jablko: :heavy_check_mark: Done.

test/test_go.rb
jablko: I like it! But I tried it out and the intermittent failure returned: https://travis-ci.org/jablko/fzf/jobs/657850503#L534

test/test_go.rb
jablko: Yes, no difference, `item_count` just seemed more "correct" before entering any search terms?

test/test_go.rb
jablko: If there's more than one match can we be sure we'll the right one?

junegunn: No, but it will avoid false negatives. Any reason you decided to remove `any_include?` which should be more robust in that regard? If we would really like to avoid false positives, we could generate a large random integer instead of `12345` and search for that number.

test/test_go.rb
junegunn: Is the link correct?

image

jablko: Yes, the backtrace shows that it errored at ``test/test_go.rb:1842:in `block in test_ctrl_r_abort'``, which is: https://github.com/jablko/fzf/blob/28a39ddbc5a8761f151d8b05e3da4b92f98ad812/test/test_go.rb#L1842

So it sends `query` (in this case a double quote) followed by <kbd>:Left</kbd> and <kbd>:Right</kbd>. It finds `$ #{query}` (in this case `$ "`) so determines that the shell should be ready! It sends <kbd>C-r</kbd> but fish still stumbles and doesn't launch fzf :frowning_man:

test/test_go.rb
jablko: > No, but it will avoid false negatives.

Sorry, I didn't think carefully enough! You're right: I've changed it to `lines.match_count >= 1` :+1:

> Any reason you decided to remove `any_include?` which should be more robust in that regard?

Is it more robust? It checks that the process is among the matches but another match might still be selected?

junegunn: It is not perfect, but I'm sure we can agree that it is better than not checking at all.

test/test_go.rb
junegunn: That's strange. Let me see.

junegunn: You should use the equality condition (`lines[-1] == "$ #{query}"`) [instead of `start_with?`](https://github.com/jablko/fzf/blob/28a39ddbc5a8761f151d8b05e3da4b92f98ad812/test/test_go.rb#L1840) which should return true before fish processes left and right arrow keys.

junegunn: If this left-right dance turns out to work well, we probably should put it in `tmux.prepare`.

jablko: :man_facepalming: Sorry I missed `.start_with?`. Thanks for your patience! However there are still intermittent failures:
- - https://travis-ci.org/junegunn/fzf/jobs/658740657#L635
  - https://github.com/junegunn/fzf/blob/191c308a3dd00b83c2b088304521082a0c6b6d9a/test/test_go.rb#L1793
- - https://travis-ci.org/jablko/fzf/jobs/658775012#L527
  - https://github.com/jablko/fzf/blob/5cdaf389648d2415d915bd7408ff2c18ebf64356/test/test_go.rb#L1793
- - https://travis-ci.org/jablko/fzf/jobs/658781294#L524
  - https://github.com/jablko/fzf/blob/51fa9973434ebf990f9e9afd692d4283871e25a5/test/test_go.rb#L1793

I used `lines[-1] != "$ #{query}^[[D^[[C"` or `!lines[-1].start_with?("$ #{query}^")` instead of `lines[-1] == "$ #{query}"` because fish sometimes offers an autocompletion: https://travis-ci.org/jablko/fzf/jobs/658726395#L628

test/test_go.rb
jablko: I see your point, how about `lines[5].end_with?(' sleep 12345')`?

test/test_go.rb
junegunn: > because fish sometimes offers an autocompletion

Hmm, this is painful. One thing I noticed on my machine is that I append a space after double-quote, the suggestion text disappears. So,

```ruby
tmux.send_keys(query, :Space, :Left, :Right)
```

might simplify the check. But anyway, this trick doesn't seem to completely prevent fish from not responding on CTRL-R.

There are other test cases for CTRL-R, should we apply the same hack on those cases as well? 

We used to wrap these key bindings with `retries` method,

- https://github.com/junegunn/fzf/blob/d9c6a0305b8c7f83a21fa583ae662660b940b1b6/test/test_go.rb#L1758-L1762
- https://github.com/junegunn/fzf/blob/d9c6a0305b8c7f83a21fa583ae662660b940b1b6/test/test_go.rb#L1839-L1843
- https://github.com/junegunn/fzf/blob/d9c6a0305b8c7f83a21fa583ae662660b940b1b6/test/test_go.rb#L1820-L1824

but you removed them all in this pull request. Maybe we should keep them?

jablko: I think in those cases there's a prompt we can rely on instead?

junegunn:
I just triggered a rebuild on Travis CI, but it failed this time.

- https://travis-ci.org/junegunn/fzf/jobs/659950165#L693-L698

So aside from the refactorings, I'm not convinced that the change proposed here reduces the probability of intermittent failures. Whether we like it or not, we'll have to live with a bunch of retries and "prepares".

jablko:
> I'm not convinced that the change proposed here reduces the probability of intermittent failures.

I think I've fixed it: The loop below now runs for 50 minutes without failing (it exhausts the Travis CI job time limit): https://travis-ci.org/github/jablko/fzf/builds/670342136

Would you consider giving it another look?
```Shell
while ruby test/test_go.rb --verbose; do :; done
```

jablko:
I've rebased this change and confirmed that an indefinite loop still exhausts the job time limit without failing: https://travis-ci.org/github/jablko/fzf/builds/670732268

PR #1903, opened 2020-03-02
Abort history search, but keep already typed text

rashack:
When searching through history fails (nothing found), pressing Ctrl-G
aborts the search but retains already typed text. This prevents one from
having to type it again.

### DIFF
--- a/src/terminal.go
+++ b/src/terminal.go
@@ -267,7 +267,7 @@ func defaultKeymap() map[int][]action {
 	keymap[tui.CtrlA] = toActions(actBeginningOfLine)
 	keymap[tui.CtrlB] = toActions(actBackwardChar)
 	keymap[tui.CtrlC] = toActions(actAbort)
-	keymap[tui.CtrlG] = toActions(actAbort)
+	keymap[tui.CtrlG] = toActions(actPrintQuery)
 	keymap[tui.CtrlQ] = toActions(actAbort)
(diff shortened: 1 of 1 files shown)


junegunn:
The suggested change doesn't make much sense. I'm not sure what you exactly mean by "searching through history", but if you're referring to CTRL-R key binding we provide for bash, it now retains the previous command line since https://github.com/junegunn/fzf/commit/7c447bbdc7adb1fe0c848e171780f68cacf3e80e.

rashack:
Sorry if I was unclear. I meant something like "when I search through history using C-r, and there is no result, but I would still like to keep what I have type so far".
How to reproduce:
1. C-r
2. type 'foo' (or something that is not in the history)
3. C-g
Now I'm back to an empty prompt, but with my small addition, 'foo' is there. Did that make sense?

junegunn:
That's something you can already do with `--bind` option.

```sh
export FZF_CTRL_R_OPTS='--bind ctrl-g:print-query'
```

rashack:
Aha, ok. I couldn't find that. A colleague and I tried to find a solution, the closest we came was https://news.ycombinator.com/item?id=20362184 
Maybe it would be a good default? Seems like a shame to waste three keybindings on the same action ;).

junegunn:
No, we don't break backward compatibility.

PR #1912, opened 2020-03-07
[vim] Use native border for popup terminal window

lacygoill:
Currently, when an `fzf` command is invoked from Vim, and the user has configured the plugin to use Vim's popup terminal window, `fzf` creates 2 popup windows; one for the terminal, and one for the border.

I think it would be better to only create 1 popup window, and make Vim draw the border using the `border`, `borderchars`, `borderhighlight`, and `padding` keys (cf. `:h popup_create-arguments`).

It would allow the removal of the one-shot autocmd listening to `BufWipeOut`, and make the logic around the Vim's implementation a little simpler.

I've tried to make sure that the geometry of the window was identical as in Neovim.

Feel free to close the PR if you prefer the current implementation.

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -859,7 +859,6 @@ else
       \ borderchars: borderchars,
       \ borderhighlight: [a:hl],
       \ padding: padding,
-      \ zindex: 50,
       \ })
     let &g:wincolor = a:hl
   endfunction
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -860,7 +860,7 @@ else
       \ padding: padding,
       \ })
     let &g:wincolor = a:hl
-    execute 'autocmd BufWipeout * ++once bwipeout! '..buf
+    execute 'autocmd TerminalWinOpen * ++once autocmd BufWipeout <buffer> ++once bwipeout! '..buf
   endfunction
 endif
 
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -861,6 +861,7 @@ else
       \ padding: padding,
       \ })
     let &g:wincolor = a:hl
+    execute 'autocmd BufWipeout * ++once bwipeout! '..buf
   endfunction
 endif
 
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -857,7 +857,6 @@ else
       \ minheight: a:opts.height,
       \ border: [],
       \ borderchars: borderchars,
-      \ borderhighlight: [a:hl],
       \ padding: padding,
       \ })
     let &g:wincolor = a:hl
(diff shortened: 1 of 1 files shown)


lacygoill:
I was wrong; the `BufWipeout` is still needed.

lacygoill:
Actually, one of the autocmd is not needed anymore, the one which wipes the border; but the other is still required (the one wiping the fzf buffer).

lacygoill:
I've removed the `'borderhighlight'` key, because it's useless.
It sets the `'wincolor'` option but its value is not used here, because `fzf` runs `term_start()` which loads a new buffer in the popup window.  When that happens, Vim doesn't re-use the value of the option from the previous buffer; it initializes the local value from the global one, which is why I wrote this line:

    let &g:wincolor = a:hl

Btw, the global value of a window-local option only affects the current window; it does not affect the other windows.  IOW, it's not global to all buffers, but to all buffers in the current window.  All of this is explained at `:h local-options`.

You can check this with the following experiment:

    $ vim -Nu NONE +vs +'setg cc=12'

This starts Vim with 2 vertical splits, and sets the global value of `'cc'` to `12` in the first window, which is confirmed by the output of `:echo &g:cc`.  But if you run the same command in the second split, you should see that the output is an empty string (the default value in Vim).  Setting the global value in the first window had no effect on the second window.

lacygoill:
> Edit: I wonder whether there is a Vim bug which makes this autocmd necessary. When I'll have the time, I'll try to bisect the code to find a MWE where Vim fails to wipe the buffer without. If this is a Vim bug, I'll report it, and if it gets fixed, maybe the current implementation could be simplified further (i.e. the autocmd could be removed in the future).

Ok, so I've checked, and this is not due to a Vim bug.  It's just that fzf creates 2 buffers: one for the popup window via `popup_create()` and `term_start()`, then another one for the fzf process via `term_start()` alone.
When the fzf buffer is wiped out, the original buffer which was displayed in the popup window needs to also be wiped out; otherwise it would hang around.

That being said, `:bw` is run on the next `BufWipeout` regardless of the buffer which is wiped out.
Although there is no issue right now, who knows, maybe in the future the implementation could change, and a `BufWipeout` could be fired for another buffer before being fired for the fzf buffer; if that happened, maybe wiping the original buffer too early would cause an issue.

To fix this, I've pushed a commit which makes the autocmd more restrictive; now, `:bw` should only wipe out the original popup buffer when the fzf buffer is wiped out, and not when any other buffer is wiped out.

This is allowed by the the `<buffer>` pattern:

    execute 'autocmd TerminalWinOpen * ++once autocmd BufWipeout <buffer> ++once bwipeout! '..buf
                                                                 ^^^^^^^^

During my limited tests, it seemed to work as expected; i.e. the popup buffer is still correctly wiped out after the fzf process is terminated.

PR #1924, opened 2020-03-13
[zsh-completion] Use noksh_arrays in _fzf_complete

chitoku-k:
The changes produced in 50b7608f9d27c09093846e172f230d92b401f956 broke some behaviours with many other Zsh plugins like [zsh-users/zsh-syntax-highlighting](https://github.com/zsh-users/zsh-syntax-highlighting) that rely on ksh\_arrays not being enabled. The scope of `localoptions` is complicated in widget context as it may be hooked by others and this kind of setting should be kept default unless it cannot be avoided IMO.

The example line of the currently broken plugin:
https://github.com/zsh-users/zsh-syntax-highlighting/blob/b85e313/zsh-syntax-highlighting.zsh#L93

### DIFF
--- a/shell/completion.zsh
+++ b/shell/completion.zsh
@@ -107,13 +107,13 @@ _fzf_feed_fifo() (
 )
 
 _fzf_complete() {
-  setopt localoptions ksh_arrays
+  setopt localoptions noksh_arrays
   # Split arguments around --
   local args rest str_arg i sep
   args=("$@")
   sep=
   for i in {0..$#args}; do
-    if [[ "${args[$i]}" = -- ]]; then
+    if [[ "${args[i+1]}" = -- ]]; then
       sep=$i
       break
     fi
@@ -131,9 +131,9 @@ _fzf_complete() {
 
   local fifo lbuf cmd matches post
   fifo="${TMPDIR:-/tmp}/fzf-complete-fifo-$$"
-  lbuf=${rest[0]}
+  lbuf=${rest[1]}
   cmd=$(__fzf_extract_command "$lbuf")
-  post="${funcstack[1]}_post"
+  post="${funcstack[2]}_post"
   type $post > /dev/null 2>&1 || post=cat
 
   _fzf_feed_fifo "$fifo"


junegunn:
Thanks, I'm not a zsh user so I didn't expect this to happen.

But the way this plugin ecosystem works, where one plugin setting a local option breaking other plugins, sounds fundamentally broken. 😕 Should every plugin code manually protect against other plugin's local options like this? Maybe we should also update zsh-syntax-highlighting to set `noksh_arrays` to make it work with any plugins that might set any incompatible local options?

Please note that I'm not against what you're suggesting here, I'm just curious if this is the only way to do.

junegunn:
Seem related: https://github.com/zsh-users/zsh-syntax-highlighting/commit/8f19af6b319d

chitoku-k:
I understand what you are talking about. For this specific plugin causing the issue, I'll try to bump that in https://github.com/zsh-users/zsh-syntax-highlighting/pull/622.

Anyway, I definitely recommend `noksh_arrays` to be as-is!

junegunn:
Thanks. I intentionally chose to set `ksh_arrays` to use the same code for both bash and zsh.

https://github.com/junegunn/fzf/blob/14f90502a49b868b6c9fed4a86d413427052fb81/shell/completion.zsh#L111-L130
https://github.com/junegunn/fzf/blob/14f90502a49b868b6c9fed4a86d413427052fb81/shell/completion.bash#L193-L212

Are you aware of any other plugins that stopped working because of this change?

danielshahaf:
> Maybe we should also update zsh-syntax-highlighting to set `noksh_arrays` to make it work with any plugins that might set any incompatible local options?

z-sy-h already uses `emulate` precisely for this reason, but the KSH_ARRAYS option affects parsing, including the parsing that happens before the `emulate` is executed.

That can be discussed in the referenced z-sy-h issue.  As a workaround, however, try using z-sy-h's feature/redrawhook branch.

> But the way this plugin ecosystem works, where one plugin setting a local option breaking other plugins, sounds fundamentally broken. confused Should every plugin code manually protect against other plugin's local options like this?

Option settings are dynamically scoped, so yes, plugins need to guard against whatever options their callers might set.

danielshahaf:
https://github.com/zsh-users/zsh-syntax-highlighting/issues/688 might be another workaround.

junegunn:
@danielshahaf Thanks for the comment.

> Option settings are dynamically scoped, so yes, plugins need to guard against whatever options their callers might set.

True, so in that sense, fzf is locally setting `ksh_arrays` to protect its own code against the environment that may have `noksh_arrays` set. I just couldn't see how that could break some other code, that fzf is not even calling.

chitoku-k:
@junegunn
Actually this is the only plugin that I found it corrupted. It doesn't sound quite usual to make an 0-based arrays for Zsh scripts escpecially when one of then already contains 1-based arrays in another part. Feel free to choose which way to go as the aforementioned issue looks like it's going to get solved there.

danielshahaf:
@junegunn Presumably zsh-syntax-highlighting code gets to run because fzf invokes some widget (via `zle $foo` without a leading dot). zsh-syntax-highlighting wraps most widgets.

On the zsh-syntax-highlighting's end the fix is threefold: https://github.com/zsh-users/zsh-syntax-highlighting/issues/689 to fix the current instance, and https://github.com/zsh-users/zsh-syntax-highlighting/issues/688 and feature/redrawhook to prevent external parse-affecting setopts from affecting zsh-syntax-highlighting in the first place.

danielshahaf:
I just pushed to z-sy-h a commit that may address this.

junegunn:
@danielshahaf Thanks for the quick fix!

@chitoku-k I'll close this as I'd like to avoid maintaining two marginally different implementations. But please let me know if you run into the same problem again with some other plugin. I can reconsider.

PR #1925, opened 2020-03-13
Use $HISTFILE for Ctrl-R to find results from multiple terminals

pjio:
This PR allows to search in the history of multiple terminals in Bash while maintaining separate histories for each terminal.

To enable it one has to set:
```
export PROMPT_COMMAND="history -a"
```

While I find it desirable to not sync history between all terminals immediately, I think it is very useful to to be able to search for commands from other terminals.

With https://github.com/junegunn/fzf/pull/749 the same could be achieved, but this PR aims for a more specialized functionality. I'm curious if there is some interest for this feature :-)

### DIFF
--- a/shell/key-bindings.bash
+++ b/shell/key-bindings.bash
@@ -53,9 +53,17 @@ __fzf_cd__() {
 
 __fzf_history__() {
   local output
+  if [[ "$PROMPT_COMMAND" =~ (^|;)\ *history\ +-a\ *($|;) ]]; then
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks for the suggestion, but I'm not interested in adding more options to these shell extension scripts because maintaining these scripts for three different shells has been really painful and time-consuming that I don't want to increase the complexity anymore. I suggest that you override the function in your shell configuration file.

Issue #1916, opened 2020-03-09
Invoking fzf from Vim raises E994 when a terminal popup is already open and forces you to quit Vim entirely


lacygoill:
- [x] I have read through the manual page (`man fzf`)
- [x] I have the latest version of fzf
- [x] I have searched through the existing issues

## Info

- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [ ] bash
    - [x] zsh
    - [ ] fish

## Problem / Steps to reproduce

In Vim, open a popup terminal via `term_start()` and `popup_create()`.
Then invoke `fzf#run()` to filter some output.
fzf opens a second popup terminal.
After it's closed the focus gets back to the original popup, but:

   - these errors are raised:

         Error detected while processing function 2[5]..1:
         line    3:
         E994: Not allowed in a popup window
         Error detected while processing BufWipeout Autocommands for "*":
         E994: Not allowed in a popup window

   - the terminal buffer in the first popup is no longer displayed

   - the popup window can't be closed anymore (`:q!`, `bw!`, `:close!`, `:call popup_close(win_getid())`, `:call popup_clear()` ...), and the focus can't be changed; the only command which seems to work is `:qa!`

---

Here's a minimal example.  Write this in `/tmp/vimrc`:

    let g:loaded_netrwPlugin = 1
    let g:fzf_layout = {
        \ 'window': {
        \     'width': 0.9,
        \     'height': 0.6,
        \     'xoffset': 0.5,
        \     'yoffset': 0.5,
        \     'highlight': 'Title',
        \     'border': 'sharp',
        \ }}
    set rtp^=~/.fzf rtp-=~/.vim rtp-=~/.vim/after hidden lines=24 columns=80
    au VimEnter * call popup_create(term_start(&shell, #{hidden: v:true}), {'highlight': 'Normal', 'border': [], 'line': 6, 'borderhighlight': ['Title'], 'borderchars': ['─', '│', '─', '│', '┌', '┐', '┘', '└'], 'minheight': 12, 'minwidth': 68, 'col': 5, 'maxwidth': 68, 'zindex': 1, 'padding': [0, 1, 0, 1], 'maxheight': 12})
    nno cd :call fzf#run(fzf#wrap('test', {'source': ['foo', 'bar', 'baz'], 'sink': {line -> execute('echom '..string(line), '')}}))<cr>

Open an xterm terminal, and inside run this shell command:

[...]

PR #1927, opened 2020-03-14
fix issue #1916

lacygoill:
This PR tries to fix the issue #1916.

Here is the purpose of each change:

    -  return {'tab': tabpagenr(), 'win': winnr(), 'cnt': winnr('$'), 'tcnt': tabpagenr('$')}
    +  return {'tab': tabpagenr(), 'win': winnr(), 'winid': win_getid(), 'cnt': winnr('$'), 'tcnt': tabpagenr('$')}

Inside `fzf.on_exit()`, this test wrongly succeeds when fzf has been invoked from a popup terminal:

      if s:getpos() == self.ppos

This is because a popup window does not have a number and `winnr()` always evaluates to 0 in a popup terminal.  By including `win_getid()` in the dictionary, the test correctly fails when fzf is invoked from a popup terminal.

---

    +        if !has('nvim') | call term_sendkeys(bufnr('#'), "exit\<cr>") | endif

This line terminates the `sh` job bound to the `!sh` terminal buffer which [was originally created by fzf](https://github.com/junegunn/fzf/blob/c39113ee41f032c72a08e015b23d00f60d61a57c/plugin/fzf.vim#L840).

It's necessary so that Vim automatically wipes out the buffer when the window is closed.  There's already a one-shot autocmd with the same purpose, but imo, it's better to rely on Vim's builtin mechanism than create an ad-hoc one.

---

    -      execute 'tabnext' self.ppos.tab
    -      execute self.ppos.win.'wincmd w'
    +      silent! execute 'tabnext' self.ppos.tab
    +      silent! execute self.ppos.win.'wincmd w'

Here, `silent!` suppresses `E994` which is raised when we're still in a popup terminal; commands like `:tabnext` and `:wincmd` are forbidden then.

---

    -    let buf = is_frame ? '' : term_start(&shell, #{hidden: 1})
    +    let buf = is_frame ? '' : term_start(&shell, #{hidden: 1, term_finish: 'close'})

For some reason, `term_finish: 'close'` is necessary for the `!sh` buffer to be automatically wiped out by Vim.  Not sure why; it's not necessary to close the window, since the exit callback already executes `:close`; [it could be a bug](https://github.com/vim/vim/issues/5768).

---

    -    else
    -      execute 'autocmd BufWipeout * ++once bwipeout! '..buf

This one-shot autocmd is not necessary anymore; Vim automatically wipes out a terminal buffer when:

   - its associated job is finished (hence the previous `term_sendkeys(...exit...)`)
   - it's hidden
   - it's not modified

See `:h E947`.

> When the job has finished and no changes were made to the buffer: closing the
> window will wipe out the buffer.

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -723,9 +723,6 @@ function! s:execute_term(dict, command, temps) abort
       call self.switch_back(1)
     else
       if bufnr('') == self.buf
-        if exists('*win_gettype') && win_gettype() ==# 'popup'
-          call term_sendkeys(bufnr('#'), "exit\<CR>")
-        endif
         " We use close instead of bd! since Vim does not close the split when
         " there's no other listed buffer (nvim +'set nobuflisted')
         close
@@ -853,6 +850,8 @@ else
       call setwinvar(id, '&wincolor', a:hl)
       call setbufline(winbufnr(id), 1, a:opts.border)
       execute 'autocmd BufWipeout * ++once call popup_close('..id..')'
+    else
+      execute 'autocmd BufWipeout * ++once call term_sendkeys('..buf..', "exit\<CR>")'
     endif
     return winbufnr(id)
   endfunction
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -723,7 +723,9 @@ function! s:execute_term(dict, command, temps) abort
       call self.switch_back(1)
     else
       if bufnr('') == self.buf
-        if !has('nvim') | call term_sendkeys(bufnr('#'), "exit\<CR>") | endif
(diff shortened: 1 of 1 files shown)


lacygoill:
Ah, I think something is still wrong.  I need to check that the user is using Vim's popup windows before invoking `term_sendkeys()`.

lacygoill:
The last commit invokes `win_gettype()` to make sure that the user did configure fzf to use Vim popup windows.

lacygoill:
In the last commit, I found a way to get rid of `win_gettype()`.

junegunn:
Sorry for the delay, merged to master. Thanks!

Issue #1931, opened 2020-03-20
Remove perl dependency for bash


rickywu:
refer to this issue https://github.com/junegunn/fzf/pull/1635

Could it possible?

junegunn:
I don't know. Pull requests are welcome.

PR #1934, opened 2020-03-21
remove perl dependency for bash

rickywu:
refer to https://github.com/junegunn/fzf/issues/1931

juse use builtin histtory and sort by reverse order, and use sed and tr to replace perl

### DIFF
--- a/shell/completion.bash
+++ b/shell/completion.bash
@@ -290,7 +290,7 @@ d_cmds="${FZF_COMPLETION_DIR_COMMANDS:-cd pushd rmdir}"
 a_cmds="
   awk cat diff diff3
   emacs emacsclient ex file ftp g++ gcc gvim head hg java
-  javac ld less more mvim nvim patch perl python ruby
+  javac ld less more mvim nvim patch python ruby
   sed sftp sort source tail tee uniq vi view vim wc xdg-open
   basename bunzip2 bzip2 chmod chown curl cp dirname du
   find git grep gunzip gzip hg jar
--- a/shell/completion.bash
+++ b/shell/completion.bash
@@ -290,7 +290,7 @@ d_cmds="${FZF_COMPLETION_DIR_COMMANDS:-cd pushd rmdir}"
 a_cmds="
   awk cat diff diff3
   emacs emacsclient ex file ftp g++ gcc gvim head hg java
-  javac ld less more mvim nvim patch python ruby
+  javac ld less more mvim nvim patch perl python ruby
   sed sftp sort source tail tee uniq vi view vim wc xdg-open
   basename bunzip2 bzip2 chmod chown curl cp dirname du
   find git grep gunzip gzip hg jar
--- a/file
+++ b/file
@@ -1,4 +1,0 @@
-"one", 
-"three
-four",
-"seven"
--- a/file
+++ b/file
@@ -1,0 +1,4 @@
+"one", 
+"three
+four",
+"seven"
--- a/shell/key-bindings.bash
+++ b/shell/key-bindings.bash
@@ -55,7 +55,7 @@ __fzf_cd__() {
 __fzf_history__() {
   local output
   output=$(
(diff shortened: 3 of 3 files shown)


junegunn:
Please make sure that the existing tests all pass. (`make docker-test`)
Hmm, I don't think there's a good way to handle multi-line commands with sed as we do now.

rickywu:
Not sure sed could not, but awk can do this as perl:

HISTTIMEFORMAT='' fc -lr -2147483648 | awk '(NR-1)%2{$1=$1}1' RS=\" ORS=\" | sed -r 's/^([0-9]+\t) +/\1/g;' | tr '\n' '\000' |

image

I've tested in my shell but failed when do docker-test:
image

rickywu:
OK, both sed and awk can do this

`fc -lr -2147483648 | sed -r -e 's/^([0-9]+\t) +/\1/g;' | sed -z -r -e 's/\n^([0-9]+\t)/\x00/g;'`

or 

`fc -lr -2147483648 | awk 'BEGIN{RS="\n[0-9]+\t ";ORS="\x00";OFS=""}{print num,$0}{num=gensub(/\n([0-9]+\t) +/,"\\1","g",RT)}' | sed -r -e 's/^([0-9]+\t) +/\1/g;'`

Here are the two options test result:
> docker run -it fzf-arch
> [exited]
> Run options: --seed 37568
> 
> # Running:
> 
> ............................................................................................................................................
> 
> Finished in 68.682246s, 2.0384 runs/s, 1.7181 assertions/s.
> 
> 140 runs, 118 assertions, 0 failures, 0 errors, 0 skips

One problem is sed -z option is only available version 4.2.2+

junegunn:
> One problem is sed -z option is only available version 4.2.2+ which released at Sat, 22 Dec 2012

Then we can't use it, macOS Catalina ships with old BSD sed (2005).

```sh
$ sed -z
sed: illegal option -- z
```

shell/completion.bash
```diff
@@ -290,7 +290,7 @@ d_cmds="${FZF_COMPLETION_DIR_COMMANDS:-cd pushd rmdir}"
 a_cmds="
   awk cat diff diff3
   emacs emacsclient ex file ftp g++ gcc gvim head hg java
-  javac ld less more mvim nvim patch perl python ruby
+  javac ld less more mvim nvim patch python ruby
```
junegunn: This is unrelated.

rickywu:
gensub is not available for BSD awk on mac os, just hold on, I'll find a better solution

junegunn:
Yeah, thanks. Is there a reason you want to remove perl dependency? Perl is available on virtually every Unixy system nowadays.

rickywu:
I‘ve tested awk with sub function works well on mac OS and passed docker-test, you can review now.

The reason is I want to use fzf in docker container and 
most of bash distribution include sed and awk but except perl, such as alpine Linux distribution

junegunn:
> works well on mac OS

While it works with GNU awk, which can be installed using Homebrew (`brew install gawk` -> `/usr/local/bin/awk`), it doesn't work with the default awk from macOS (`/usr/bin/awk`).

rickywu:
Yes, you are right.
I've checked, for compatible BSD bash, the only option is use perl.
So I changed check if gnu awk available, otherwise use perl, is this solution acceptable?

junegunn:
Thanks, but I'm not a fan of this code branch. (FYI, we rely on Perl even more since https://github.com/junegunn/fzf/commit/18261fe31cf1270f9aa783f99bbdbca343f89479)

Minimizing dependency is nice, but I'd just use Perl in this case as I don't want to increase code complexity and I don't think it's too much to require Perl on interactive command-line environments. 

> The reason is I want to use fzf in docker container and
most of bash distribution include sed and awk but except perl, such as alpine Linux distribution

That's a valid point but such Docker images are mainly for building lightweight containers rather than interactive command-line environments. So I think it's forgivable that interactive fzf scripts don't work in those images out of the box without additional installation steps. Perl on Alpine is [quite small](https://pkgs.alpinelinux.org/package/edge/main/x86/perl), so I suggest that you just install Perl when you build your Docker containers.

PR #1939, opened 2020-03-25
10x faster AtomicBool implementation

brualan:
It's very straightforward to use _actual_ atomic instead of mutex. And it's paying off in nearly 10x margin in terms of performance
```
goos: linux
goarch: amd64
pkg: github.com/junegunn/fzf/src/util
BenchmarkMutex_Set-4            20000000                69.4 ns/op
BenchmarkMutex_Get-4            20000000                68.9 ns/op
BenchmarkAtomic_Set-4           200000000                7.66 ns/op
BenchmarkAtomic_Get-4           2000000000               0.42 ns/op
```

### DIFF
--- a/src/util/atomicbool.go
+++ b/src/util/atomicbool.go
@@ -1,32 +1,38 @@
 package util
 
-import "sync"
+import (
+	"sync/atomic"
+)
 
 // AtomicBool is a boxed-class that provides synchronized access to the
 // underlying boolean value
 type AtomicBool struct {
-	mutex sync.Mutex
-	state bool
+	state int32 // "1" is true, "0" is false
 }
 
 // NewAtomicBool returns a new AtomicBool
 func NewAtomicBool(initialState bool) *AtomicBool {
-	return &AtomicBool{
-		mutex: sync.Mutex{},
-		state: initialState}
+	var state int32 = 0
+	if initialState == true {
+		state = 1
+	}
+	return &AtomicBool{state: state}
 }
 
 // Get returns the current boolean value synchronously
 func (a *AtomicBool) Get() bool {
-	a.mutex.Lock()
-	defer a.mutex.Unlock()
-	return a.state
+	if atomic.LoadInt32(&a.state) != 0 {
+		return true
+	}
+	return false
 }
 
 // Set updates the boolean value synchronously
 func (a *AtomicBool) Set(newState bool) bool {
-	a.mutex.Lock()
-	defer a.mutex.Unlock()
-	a.state = newState
-	return a.state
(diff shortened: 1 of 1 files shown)


brualan:
Also, `AtomicBool` is used only here https://github.com/junegunn/fzf/blob/master/src/matcher.go#L150 for, basically, context cancellation. I plan later to use proper methods for that task. As explained here https://blog.golang.org/context, for example

src/util/atomicbool.go
```diff
@@ -1,32 +1,35 @@
...
-	state bool
+	state int32 // "1" is true, "0" is false
 }
 
 // NewAtomicBool returns a new AtomicBool
 func NewAtomicBool(initialState bool) *AtomicBool {
-	return &AtomicBool{
-		mutex: sync.Mutex{},
-		state: initialState}
+	var state int32 = 0
```
leonklingele: Can be simplified:

```go
var state int32
if initialState {
    state = 1
}
```

src/util/atomicbool.go
```diff
@@ -1,32 +1,35 @@
...
+	return atomic.LoadInt32(&a.state) == 1
 }
 
 // Set updates the boolean value synchronously
 func (a *AtomicBool) Set(newState bool) bool {
-	a.mutex.Lock()
-	defer a.mutex.Unlock()
-	a.state = newState
-	return a.state
+	var state int32 = 0
```
leonklingele: Can be simplified:

```go
var state in32
if newState {
    state = 1
}
```

src/util/atomicbool.go
brualan: Yes, we can simplify this. But in this particular case I prefer to explicitly set variable to `0`, so that it would be more obvious that `"1" is true` and `"0" is false`

src/util/atomicbool.go
brualan: Yes, we can simplify this. But in this particular case I prefer to explicitly set variable to `0`, so that it would be more obvious that `"1" is true` and `"0" is false`

junegunn: I see your point, but golint complains that we should remove `= 0`, so can you change the code as @leonklingele suggested?

```
src/util/atomicbool.go:15:20: should drop = 0 from declaration of var state; it is the zero value
src/util/atomicbool.go:29:20: should drop = 0 from declaration of var state; it is the zero value
```

brualan:
I'm think that code in latest commit is much better and shouldn't get any complains from linter. Please review

junegunn:
Merged, thanks.

PR #1940, opened 2020-03-27
shell: add support to deduplicate history command output

leonklingele:
This change allows to deduplicate commands from the shell history
when filtering with fzf.

To enable deduplication, set "FZF_HIST_FIND_NO_DUPS" to "on":

    export FZF_HIST_FIND_NO_DUPS=on

On zsh, fall back to reading the "HIST_FIND_NO_DUPS" shell option
which can be set as follows:

    setopt HIST_FIND_NO_DUPS

A big »Thanks« goes to Felix Breuer for his idea which this change is
based upon ([0]).

Updates [1] and [2].

[0]: https://github.com/junegunn/fzf/pull/1363#issuecomment-598228199
[1]: https://github.com/junegunn/fzf/pull/1287
[2]: https://github.com/junegunn/fzf/pull/1363

### DIFF
--- a/shell/key-bindings-common.sh
+++ b/shell/key-bindings-common.sh
@@ -1,0 +1,8 @@
+# Keep in sync with version for fish (in key-bindings.fish)
+__fzf_history_dedup__() {
+  if [ "$FZF_HIST_FIND_NO_DUPS" = on ]; then
+    perl -ne 'print if !$seen{($_ =~ s/^[0-9\s]*//r)}++'
+  else
+    cat
+  fi
+}
--- a/shell/key-bindings.bash
+++ b/shell/key-bindings.bash
@@ -13,6 +13,10 @@ __fzf_select__() {
 
 if [[ $- =~ i ]]; then
 
+# https://stackoverflow.com/a/246128
+CWD="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+source "$CWD/key-bindings-common.sh"
+
 __fzf_use_tmux__() {
   [ -n "$TMUX_PANE" ] && [ "${FZF_TMUX:-0}" != 0 ] && [ ${LINES:-40} -gt 15 ]
 }
@@ -54,7 +58,7 @@ __fzf_cd__() {
 __fzf_history__() {
   local output
   output=$(
-    builtin fc -lnr -2147483648 |
+    builtin fc -lnr -2147483648 | __fzf_history_dedup__ |
       last_hist=$(HISTTIMEFORMAT='' builtin history 1) perl -p -l0 -e 'BEGIN { getc; $/ = "\n\t"; $HISTCMD = $ENV{last_hist} + 1 } s/^[ *]//; $_ = $HISTCMD - $. . "\t$_"' |
       FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} $FZF_DEFAULT_OPTS -n2..,.. --tiebreak=index --bind=ctrl-r:toggle-sort $FZF_CTRL_R_OPTS +m --read0" $(__fzfcmd) --query "$READLINE_LINE"
   ) || return
--- a/shell/key-bindings.fish
+++ b/shell/key-bindings.fish
@@ -47,14 +47,23 @@ function fzf_key_bindings
       # history's -z flag was added in fish 2.4.0, so don't use it for versions
       # before 2.4.0.
       if [ "$FISH_MAJOR" -gt 2 -o \( "$FISH_MAJOR" -eq 2 -a "$FISH_MINOR" -ge 4 \) ];
-        history -z | eval (__fzfcmd) --read0 --print0 -q '(commandline)' | read -lz result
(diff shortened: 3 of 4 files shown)


junegunn:
- My biggest concern regarding the deduplication was the increased delay. This is much less of an issue because in bash we no longer use `history` with `fzf --tac --sync` and fzf starts immediately without waiting for the full list of commands.
- I don't like adding an option for this. If we think this is generally preferable to most users, we will just ship it as the default behavior. More options lead to more documentation, more edge cases, more test code, all in all, more maintenance burden. Users of fzf don't have to use the code we publish as it is, they can freely change it. We don't have to maintain options for every niche use case.
- Deduping seems to be the default behavior of fish (see https://fishshell.com/docs/current/#history-search), if that's the case, we don't have to do anything for fish.
- Don't like the idea of extracting "common" file as some users would prefer taking a single file for their shell instead of keeping the whole directory structure.

So here's what I'm suggesting:

```diff
diff --git a/shell/key-bindings.bash b/shell/key-bindings.bash
index 71cd999..16e1ec5 100644
--- a/shell/key-bindings.bash
+++ b/shell/key-bindings.bash
@@ -55,7 +55,7 @@ __fzf_history__() {
   local output
   output=$(
     builtin fc -lnr -2147483648 |
-      last_hist=$(HISTTIMEFORMAT='' builtin history 1) perl -p -l0 -e 'BEGIN { getc; $/ = "\n\t"; $HISTCMD = $ENV{last_hist} + 1 } s/^[ *]//; $_ = $HISTCMD - $. . "\t$_"' |
+      last_hist=$(HISTTIMEFORMAT='' builtin history 1) perl -n -l0 -e 'BEGIN { getc; $/ = "\n\t"; $HISTCMD = $ENV{last_hist} + 1 } s/^[ *]//; print $HISTCMD - $. . "\t$_" if !$seen{$_}++' |
       FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} $FZF_DEFAULT_OPTS -n2..,.. --tiebreak=index --bind=ctrl-r:toggle-sort $FZF_CTRL_R_OPTS +m --read0" $(__fzfcmd) --query "$READLINE_LINE"
   ) || return
   READLINE_LINE=${output#*$'\t'}
diff --git a/shell/key-bindings.zsh b/shell/key-bindings.zsh
index e291677..3408b08 100644
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -68,7 +68,7 @@ bindkey '\ec' fzf-cd-widget
 fzf-history-widget() {
   local selected num
   setopt localoptions noglobsubst noposixbuiltins pipefail no_aliases 2> /dev/null
-  selected=( $(fc -rl 1 |
+  selected=( $(fc -rl 1 | perl -ne 'print if !$seen{($_ =~ s/^[0-9\s]*//r)}++' |
     FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} $FZF_DEFAULT_OPTS -n2..,.. --tiebreak=index --bind=ctrl-r:toggle-sort $FZF_CTRL_R_OPTS --query=${(qqq)LBUFFER} +m" $(__fzfcmd)) )
   local ret=$?
   if [ -n "$selected" ]; then
diff --git a/test/test_go.rb b/test/test_go.rb
index 765108c..83d189f 100755
--- a/test/test_go.rb
+++ b/test/test_go.rb
@@ -1839,15 +1839,18 @@ module TestShell
     tmux.send_keys 'echo 1st', :Enter; tmux.prepare
     tmux.send_keys 'echo 2nd', :Enter; tmux.prepare
     tmux.send_keys 'echo 3d',  :Enter; tmux.prepare
-    tmux.send_keys 'echo 3rd', :Enter; tmux.prepare
+    3.times { tmux.send_keys 'echo 3rd', :Enter; tmux.prepare }
     tmux.send_keys 'echo 4th', :Enter
     retries do
       tmux.prepare
       tmux.send_keys 'C-r'
       tmux.until { |lines| lines.match_count.positive? }
     end
-    tmux.send_keys 'C-r'
     tmux.send_keys '3d'
+    # Duplicates removed: 3d (1) + 3rd (1) => 2 matches
+    tmux.until { |lines| lines.match_count == 2 }
+    tmux.until { |lines| lines[-3].end_with? 'echo 3d' }
+    tmux.send_keys 'C-r'
     tmux.until { |lines| lines[-3].end_with? 'echo 3rd' }
     tmux.send_keys :Enter
     tmux.until { |lines| lines[-1] == 'echo 3rd' }
```

junegunn:
Patch pushed to master. Please let me know if it doesn't work for you. Thanks.

leonklingele:
Hey.. 😂 I was just about to push an updated version of this. Thank you for your work @junegunn!

PaulCapestany:
Just wanted to say that I'd been using my own (zsh-only) hacked up solution to this for years and I had to check my sanity when I saw deduping working out-of-the-box for fzf after an upgrade. Thanks for this super pleasant surprise @leonklingele @junegunn :)

gaganis:
Just wanted to add that this has been a feature that I always wanted. I had cases where all my screen was filled with duplicates and I often needed to scroll to find what I wanted.

Thanks @leonklingele @junegunn  for moving forward and making this the default! Kudos!

PR #1944, opened 2020-03-29
[zsh] Ensure that fzf code always parses the same way

### LINKED ISSUE #1938: fzf-completion() breaks when global alias `tail` is defined

## Info

- OS
    - [ ] Linux
    - [x] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [ ] bash
    - [x] zsh
    - [ ] fish
    
When I have the following line in my `.zshrc`:
```shell
alias -g tail="multitail -Cs --follow-all"
```
then when I try to use `fzf-completion()`, it fails with the following error message:
```shell
fzf-completion:local:1: not valid in this context: -Cs
```

The solution is probably just to add `setopt no_aliases`, like you did here: https://github.com/junegunn/fzf/issues/1708#issuecomment-543000287

romkatv:
At the top of each zsh file options are set to their standard values (those marked with `<Z>` in `man zshoptions`) and `aliases` option is disabled.

At the bottom of the file the original options are restored.

Fixes #1938.

### DIFF
--- a/shell/completion.zsh
+++ b/shell/completion.zsh
@@ -9,7 +9,72 @@
 # - $FZF_COMPLETION_TRIGGER (default: '**')
 # - $FZF_COMPLETION_OPTS    (default: empty)
 
-if [[ $- =~ i ]]; then
+# Both branches of the following `if` do the same thing -- define
+# __fzf_completion_options such that `eval $__fzf_completion_options` sets
+# all options to the same values they currently have. We'll do just that at
+# the bottom of the file after changing options to what we prefer.
+#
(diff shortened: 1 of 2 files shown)


junegunn:
Thanks.

I'm not a zsh user and honestly, it's really hard to understand the code. For me and for those who want to understand the code, can you briefly explain how the code works? It would be extra helpful if you could post some documentation URLs that we can refer to.

romkatv:
I've added enough comments for a chapter in a book :wink: PTAL

junegunn:
Wow, the comment is super helpful, thanks!

romkatv:
Thanks for the quick merge!

PR #1961, opened 2020-04-07
Add completion for Makefiles

eds000n:
I added a completion for targets in `Makefile` and `makefile` for bash. What do you think?


junegunn:
Thanks, but I decided not to add more completion implementations here because it's just not possible for me to maintain an ever-increasing number of features for three different shells.

PR #1962, opened 2020-04-07
Improve error messages

jablko:
What do you think of adding minitest assertions, to improve the automated tests output?

### Before

```
  1) Error:
TestZsh#test_process_completion:
RuntimeError: timeout
    test/test_go.rb:47:in `wait'
    test/test_go.rb:118:in `until'
    test/test_go.rb:2025:in `test_process_completion'
```

### After

```
  1) Failure:
TestZsh#test_process_completion [test/test_go.rb:2047]:
Expected: "kill 16358"
  Actual: "kill 1"
```

PR #1964, opened 2020-04-09
Use tmux shell-command

jablko:
Since [`Tmux.new(shell)`](https://github.com/junegunn/fzf/blob/master/test/test_go.rb#L1736) ([`:bash`](https://github.com/junegunn/fzf/blob/master/test/test_go.rb#L2098), [`:zsh`](https://github.com/junegunn/fzf/blob/master/test/test_go.rb#L2130), [`:fish`](https://github.com/junegunn/fzf/blob/master/test/test_go.rb#L2143)) are just indexes to `Shell.send(shell)`,
https://github.com/junegunn/fzf/blob/master/test/test_go.rb#L90

what do you think about simply passing the command to `Tmux.new` instead?
https://github.com/jablko/fzf/blob/21afe1c09bd7e6775503195a1aed00eb3b40e0fc/test/test_go.rb#L57

This was originally part of #1900


junegunn:
Thanks but why don't we open a separate pull request for Rubocop changes you made first? They are not related to what you're proposing here. That way, it would be easier to discuss different concerns (#1900 #1962 #1964).

jablko:
:+1: Thank you for reviewing these changes. I've opened a PR for just the RuboCop changes: #1967

PR #1967, opened 2020-04-09
Run rubocop --auto-correct --disable-uncorrectable

jablko:
Separate these changes from the others in #1900

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -8,6 +8,7 @@ Style/MethodCallWithArgsParentheses:
   Enabled: true
   IgnoredMethods:
     - assert
+    - exit
     - paste
     - puts
     - raise
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -8,7 +8,9 @@ Style/MethodCallWithArgsParentheses:
   Enabled: true
   IgnoredMethods:
     - assert
+    - paste
     - puts
+    - raise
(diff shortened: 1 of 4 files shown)


junegunn:
Thanks.

I'm not a fan of enforcing `Style/MethodCallWithArgsParentheses` everywhere. For example, using parentheses on `require` statements looks very unconventional to a Ruby programmer's eyes.

In the same vein, I would not like to add parentheses to `tmux.send_keys` and `assert_*` because they read like domain-specific language, and adding parentheses doesn't really add readability in my opinion.

References:
- https://github.com/rubocop-hq/ruby-style-guide#methods-that-are-part-of-an-internal-dsl
- https://github.com/rubocop-hq/ruby-style-guide#non-declarative-methods-that-have-keyword-status-in-ruby
- https://github.com/rubocop-hq/rubocop/pull/3797#issuecomment-267887795

But other than that, parentheses are generally nice for consistency, so here's what I'm suggesting:

```yaml
Style/MethodCallWithArgsParentheses:
  Enabled: true
  IgnoredMethods:
    - require
    - puts
    - send_keys
    - assert_equal
```

jablko:
:heavy_check_mark: Sounds good: I've made the changes you suggest.

junegunn:
Merged, thanks!

Issue #1968, opened 2020-04-10
Add support for 'cd ..' on backspace on empty query


mitinarseny:
## Info

- OS
    - [x] Linux
    - [x] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [ ] bash
    - [x] zsh
    - [ ] fish
    
## Problem / Steps to reproduce
As for [this PR](https://github.com/junegunn/fzf/pull/1891) now it is possible to bind `bspace` to `backward-delete-char/eof`. It also will be useful to make it possible to bind `bspace` to action when fzf changes its working directory to the parent of current one and executes all subsequent commands from this directory.

There:
```sh
fzf --bind 'bspace:backward-delete-char+reload(cd .. && find .)'
```
But in this case it executes every time `bspace` is pressed, so even if you type `abc` in query and then press `bspace` it would perform search for `ab` in parent directory instead of current.

It seems like a solution would be to add `backward-delete-char/cd-parent` action which either deletes one char or takes you to the parent directory if query is empty.

PR #1969, opened 2020-04-10
Implement backward-delete-char/cd-parent

mitinarseny:
I realized that I can implement #1968 by myself, so here it is :)

### DIFF
--- a/man/man1/fzf.1
+++ b/man/man1/fzf.1
@@ -646,62 +646,63 @@ or any single character
 A key or an event can be bound to one or more of the following actions.
 
   \fBACTION:                   DEFAULT BINDINGS (NOTES):
-    \fBabort\fR                     \fIctrl-c  ctrl-g  ctrl-q  esc\fR
-    \fBaccept\fR                    \fIenter   double-click\fR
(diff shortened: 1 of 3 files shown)


junegunn:
Thanks, but I don't want to add this to fzf, because fzf is a general-purpose text filter and we should not start adding features for a specific type of input or use case even if they look trivial at this point, to avoid inevitable feature bloat. I can understand that you may not agree with my decision as the long-term maintainer of this project. So it's a win-win situation that fzf is an open-source project and you can freely change it to your needs.

In this case, I would just bind a special key to the reload action:

```sh
(pwd | tee /tmp/fzf-pwd; find *) |
  fzf --header-lines=1 \
    --bind 'ctrl-p:clear-query+reload:
      cd "$(cat /tmp/fzf-pwd)"/..
      pwd | tee /tmp/fzf-pwd
      find *'
```

NightMachinary:
@junegunn @mitinarseny Can fzf make it possible to bind an arbitrary action to backspace when the query is empty? It’s a very handy “shortcut” that let’s one easily “switch modes.”

`fzf -bind ‘bspace/eof:reload: ...’`

The current version where it aborts fzf is just one of many possible actions.

PR #1970, opened 2020-04-11
[shell] zsh completion for the Debian package manager (aptitude or `apt`)

sim590:
This is inspired by the efforts that led to writing the small snippet found at https://github.com/krickelkrakel/fzf-apt.

### DIFF
--- a/shell/completion.zsh
+++ b/shell/completion.zsh
@@ -249,6 +249,36 @@ _fzf_complete_unalias() {
   )
 }
 
+list_aptitude_packages() {
+  if [[ $@ == *' install'* ]] || [[ $@ == *' show'* ]] || [[ $@ == *' search'* ]]; then
+    command apt list --verbose
+  elif [[ $@ == *' remove'* ]] || [[ $@ == *' reinstall'* ]] || [[ $@ == *' purge'* ]]; then
+    command apt list --verbose --installed
+  fi
+}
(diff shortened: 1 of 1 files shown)


sim590:
After thinking about it. Maybe this contribution should be a good fit for the Debian package. I guess I could make a PR [there][deb]. What do the developers think? 

[deb]: https://salsa.debian.org/debian/fzf/-/tree/master/debian

sim590:
I have prepared [patch for the debian package][patch]. Therefore, I could close this PR if that is more appropriate to merge in the Debian package.

[patch]: https://salsa.debian.org/sim590-guest/fzf/-/commit/1c892d291baaa40ff3f0581a1a25a682c4a59ca4

junegunn:
Thanks, but I decided not to accept more completion implementations because it's just not possible for me to maintain ever-increasing number of features for three different shells.

sim590:
@junegunn Alright. This is fine, but what about the Debian package? Would you at least accept the patch for the Debian package as aptitude is fundamental element of Debian?

junegunn:
I'm not a maintainer of the Debian package.

sim590:
@junegunn sorry. You're right. I misread the page. Thanks for your answers.

junegunn:
No problem. Take care.

Issue #1973, opened 2020-04-12
completion.zsh:__fzf_extract_command fails with '\curl'


MenkeTechnologies:
- [x] I have read through the manual page (`man fzf`)
- [x] I have the latest version of fzf
- [x] I have searched through the existing issues

## Info

- OS
    - [x] Linux
    - [x] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [ ] bash
    - [x] zsh
    - [ ] fish
    
## Problem / Steps to reproduce

when using the fzf tab completion on command \curl <tab>
the following happens.
<img width="350" alt="Screen Shot 2020-04-12 at 1 11 00 AM" src="https://user-images.githubusercontent.com/18195247/79061084-7b8d9a80-7c5a-11ea-8ca6-e4fe479737be.png">

 __fzf_extract_command fails to return \curl as the command like it returns \ls when \ls is the command.  The first word needs to be escaped.  \curl is used in command position when an alias curl exists that should not be used.

PR #1974, opened 2020-04-12
escape token before extracting command

MenkeTechnologies:
this PR fixes https://github.com/junegunn/fzf/issues/1973


junegunn:
Thanks, this fixes the problem but I think we can do slightly better.

- `\curl **<tab>`: should trigger `curl` completion if exists
- `'curl' **<tab>`: should also trigger `curl` completion
- `\\curl **<tab>`: should not trigger `curl` completion & should not fail

Let me take this over.

PR #1978, opened 2020-04-12
Make flaky tests reliable

jablko:
This was originally part of #1900

### DIFF
--- a/test/test_go.rb
+++ b/test/test_go.rb
@@ -109,7 +109,7 @@ class Tmux
   end
 
   def capture
-    go(%W[capture-pane -p -t #{win}]).reverse.drop_while(&:empty?).reverse
+    go(%W[capture-pane -p -J -t #{win}]).map(&:rstrip).reverse.drop_while(&:empty?).reverse
   end
 
   def until(refresh = false)
@@ -254,7 +254,7 @@ class TestGoFZF < TestBase
 
   def test_fzf_default_command
     tmux.send_keys fzf.sub('FZF_DEFAULT_COMMAND=', "FZF_DEFAULT_COMMAND='echo hello'"), :Enter
-    tmux.until { |lines| lines.last =~ /^>/ }
+    tmux.until { |lines| lines[-3] == '> hello' }
 
     tmux.send_keys :Enter
     assert_equal 'hello', readonce.chomp
@@ -528,7 +528,7 @@ class TestGoFZF < TestBase
 
   def test_expect
     test = lambda do |key, feed, expected = key|
-      tmux.send_keys "seq 1 100 | #{fzf(:expect, key)}; sync", :Enter
+      tmux.send_keys "seq 1 100 | #{fzf(:expect, key)}", :Enter
       tmux.until { |lines| lines[-2].include?('100/100') }
       tmux.send_keys '55'
       tmux.until { |lines| lines[-2].include?('1/100') }
@@ -880,14 +880,14 @@ class TestGoFZF < TestBase
     opts = "--history=#{history_file} --history-size=4"
     input = %w[00 11 22 33 44].map { |e| e + $INPUT_RECORD_SEPARATOR }
     input.each do |keys|
+      tmux.prepare
       tmux.send_keys "seq 100 | #{fzf(opts)}", :Enter
       tmux.until { |lines| lines[-2].include?('100/100') }
       tmux.send_keys keys
       tmux.until { |lines| lines[-2].include?('1/100') }
       tmux.send_keys :Enter
-      readonce
     end
-    assert_equal input[1..-1], File.readlines(history_file)
+    wait { File.exist?(history_file) && File.readlines(history_file) == input[1..-1] }
 
     # Update history entries (not changed on disk)
     tmux.send_keys "seq 100 | #{fzf(opts)}", :Enter
@@ -907,8 +907,7 @@ class TestGoFZF < TestBase
     tmux.send_keys 0
     tmux.until { |lines| lines[-1].end_with?('> 310') }
     tmux.send_keys :Enter
-    readonce
-    assert_equal %w[22 33 44 310].map { |e| e + $INPUT_RECORD_SEPARATOR }, File.readlines(history_file)
+    wait { File.exist?(history_file) && File.readlines(history_file) == %w[22 33 44 310].map { |e| e + $INPUT_RECORD_SEPARATOR } }
 
     # Respect --bind option
     tmux.send_keys "seq 100 | #{fzf(opts + ' --bind ctrl-p:next-history,ctrl-n:previous-history')}", :Enter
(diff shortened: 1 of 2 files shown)


jablko:
@junegunn I've rebased this change: I think it's ready for review.

junegunn:
Can you explain what strategies are used here to reduce flakies? It's not clear to me how removing waits and retries in some places can make things less susceptible to timing errors. Isn't it the opposite? Yes, we should remove them if they are truly unnecessary, but I'm not convinced.

test/test_go.rb
```diff
@@ -109,7 +109,7 @@ def paste(str)
   end
 
   def capture
-    go(%W[capture-pane -p -t #{win}]).reverse.drop_while(&:empty?).reverse
+    go(%W[capture-pane -p -J -t #{win}]).map(&:rstrip).reverse.drop_while(&:empty?).reverse
```
junegunn, line 112: Big 👍to `-J`

test/test_go.rb
```diff
@@ -880,14 +880,14 @@ def test_history
...
+      tmux.prepare
       tmux.send_keys "seq 100 | #{fzf(opts)}", :Enter
       tmux.until { |lines| lines[-2].include?('100/100') }
       tmux.send_keys keys
       tmux.until { |lines| lines[-2].include?('1/100') }
       tmux.send_keys :Enter
-      readonce
     end
-    assert_equal input[1..-1], File.readlines(history_file)
+    wait { File.exist?(history_file) && File.readlines(history_file) == input[1..-1] }
```
junegunn, line 890: 👍

test/test_go.rb
```diff
@@ -923,34 +922,25 @@ def test_history
   def test_execute
     output = '/tmp/fzf-test-execute'
     opts = %[--bind "alt-a:execute(echo /{}/ >> #{output}),alt-b:execute[echo /{}{}/ >> #{output}],C:execute:echo /{}{}{}/ >> #{output}"]
-    wait = ->(exp) { tmux.until { |lines| lines[-2].include?(exp) } }
     writelines(tempname, %w[foo'bar foo"bar foo$bar])
-    tmux.send_keys "cat #{tempname} | #{fzf(opts)}; sync", :Enter
-    wait['3/3']
+    tmux.send_keys "cat #{tempname} | #{fzf(opts)}", :Enter
+    tmux.until { |lines| lines[-2] == '  3/3' }
```
junegunn, line 927: `execute` action switches to the alternate screen, so I was concerned that the keys we send may not register to fzf. But I could be wrong. Do you think it's fine without waits?

test/test_go.rb
```diff
@@ -923,34 +922,25 @@ def test_history
...
     tmux.send_keys :Escape, :b
-    wait['/3']
-    tmux.send_keys :Enter
-    readonce
-    assert_equal %w[/foo'bar/ /foo'bar/
-                    /foo"barfoo"bar/ /foo"barfoo"bar/
-                    /foo$barfoo$barfoo$bar/],
-                 File.readlines(output).map(&:chomp)
+    wait do
+      File.exist?(output) && File.readlines(output).map(&:chomp) == %w[/foo'bar/ /foo'bar/
```
junegunn, line 940: How about if we split the line for readability?

```ruby
wait do
  return unless File.exist?(output)

  lines = File.readlines(output).map(&:chomp)
  lines == ...
end
```

test/test_go.rb
```diff
@@ -1016,15 +1003,13 @@ def test_execute_plus_flag
...
-      %(foo bar/foo bar/bar/bar),
-      %(123 456/foo bar/456/bar),
-      %(123 456 foo bar/foo bar/456 bar/bar)
-    ], File.readlines(output).map(&:chomp)
+    wait do
+      File.exist?(output) && File.readlines(output).map(&:chomp) == [
+        %(foo bar/foo bar/bar/bar),
+        %(123 456/foo bar/456/bar),
+        %(123 456 foo bar/foo bar/456 bar/bar)
+      ]
```
junegunn, line 1011: Oh, this reads much better.

test/test_go.rb
```diff
@@ -1870,29 +1833,17 @@ def test_ctrl_r_multiline
   end
 
   def test_ctrl_r_abort
-    skip("doesn't restore the original line when search is aborted pre Bash 4") if shell == :bash && /(?<= version )\d+/.match(`#{Shell.bash} --version`).to_s.to_i < 4
+    skip("doesn't restore the original line when search is aborted pre Bash 4") if shell == :bash && `#{Shell.bash} --version`[/(?<= version )\d+/].to_i < 4
     %w[foo ' "].each do |query|
-      retries do
```
junegunn, line 1875: I had to add this `retries` due to Travis CI failure and I believe it helped.

https://github.com/junegunn/fzf/commit/7c40a424c0bf5a8967816d51ead6a71a334f30bb

Can you explain why you decided that retries are no longer needed here?

test/test_go.rb
```diff
@@ -1924,7 +1875,7 @@ def test_file_completion
     tmux.until { |lines| lines.count { |l| l.include?('.fzf-home') } > 1 }
     tmux.send_keys :Enter
     tmux.until(true) do |lines|
-      lines[-1].end_with?('.fzf-home')
+      lines[-1] =~ %r{cat .*/\.fzf-home}
```
junegunn, line 1878: Nice. Thanks to `-J`, right?

test/test_go.rb
jablko: Hmm, in theory I'm unsure whether, in this scenario, the keys might not register to fzf. However I've run the tests in an indefinite loop several times, for 50 minutes at a time, and not seen this happen, so I'm confident that, at least in practice, they're not needed.

test/test_go.rb
jablko: I think you mean the line breaks inside the brackets: I'll go ahead and add them to 940 and 966 :+1:

test/test_go.rb
jablko: I've seen those intermittent failures have been resolved in recent versions of the fish shell. I can now run the tests in an indefinite loop without failures, whereas before, I agree, you couldn't.

test/test_go.rb
jablko: Thanks! `lines[-1].end_with?('.fzf-home')` is supposed to confirm that fzf put the selected item on the shell's command line, but in fact it matches the last fzf result (when the query `'.fzf-home` has one result). The test is then liable to send subsequent keys (`cat ~such**` etc.) before the shell is ready, causing this intermittent failure: https://travis-ci.org/github/jablko/fzf/jobs/673032619#L928-L974

[Line 1934](https://github.com/jablko/fzf/blob/298eda3806811964cc31f6c381a4df3b8c3ddfe7/test/test_go.rb#L1932-L1934) sent `cat ~such**` and is waiting for fzf to start, but the shell never saw those keys and still shows `cat /home/travis/.fzf-home`.

`%r{cat .*/\.fzf-home}` fixes the test so it confirms what it should, which resolves the intermittent failure by not sending subsequent keys too quickly.

test/test_go.rb
junegunn: Got it, let's remove them. We can always restore them when we see errors again.

test/test_go.rb
junegunn: Nice to know! Good riddance then.

test/test_go.rb
junegunn: Ah, thanks for the explanation. Yep, that makes sense.

junegunn:
Let me merge this once the running Travis build succeeds. Thanks for your contribution!

jablko:
Awesome, thank you for reviewing! I'll update #1962: Would you consider giving it a look, when you get a chance?

Issue #385, opened 2015-10-18
Inconsistent `kill -9` completion behaviour


ELLIOTTCABLE:
Why are `unset`, `export`, `unalias`, `ssh`, `telnet`, and `cd` all configured to do special-case completion on the otherwise-general `**` suffix … but `kill` is configured to complete on a simple `⇥`, without any `**`? Is there a rationale behind this, or was it an accident?

junegunn:
Both, I guess. I initially thought that I was never going to use the default shell completion for kill command, so I made the decision which I kind of regret now. But I haven't tried to rectify the inconsistency not to break backward compatibility. Note that kill completion still uses the default completion if there's prefix (e.g. `kill -9 $(cat <TAB>`), so it's probably okay in most situations, but I should probably make it consistent with the other ones.

ELLIOTTCABLE:
Glad to hear you're aware. It doesn't bother me at all; if you need to keep it for backwards-compat, that's fine.

bag-man:
I was going to raise this as a question too. Luckily it is pretty trivial to get the same behaviour for ssh as it is with kill. It makes sense to me to not have to ad the ** to get the auto completion. I would definitely +1 for making this the default behaviour. 

```
_fzf_complete_ssh() {
  fzf="$(__fzfcmd_complete)"
  cat <(cat ~/.ssh/config /etc/ssh/ssh_config 2> /dev/null | command grep -i '^host' | command grep -v '*') \
      <(command grep -oE '^[a-z0-9.,-]+' ~/.ssh/known_hosts | tr ',' '\n' | awk '{ print $1 " " $1 }') \
      <(command grep -v '^\s*\(#\|$\)' /etc/hosts | command grep -Fv '0.0.0.0') |
      awk '{if (length($2) > 0) {print $2}}' | sort -u | $fzf
}
```

allixsenos:
this is as easy as adding this in your bashrc after `source ~/.fzf.bash`

```
_fzf_complete_ssh_notrigger() {
    FZF_COMPLETION_TRIGGER='' _fzf_complete_ssh
}

complete -o bashdefault -o default -F _fzf_complete_ssh_notrigger ssh
complete -o bashdefault -o default -F _fzf_complete_ssh_notrigger mosh
complete -o bashdefault -o default -F _fzf_complete_ssh_notrigger my_ssh_like_command
```

lanza:
Any idea how to get the notrigger variant to work in zsh?

[...]

PR #1988, opened 2020-04-17
Consistent kill behaviour

hellricer:
This addresses issue #385 .


junegunn:
Thanks, but it's a 6+-year-old behavior and we don't have a consensus if we should do this or not. I'll leave this open.

junegunn:
A nice middle ground would be to support both cases, i.e. `kill <tab>` and `kill foo**<tab>`.

RobSis:
Awesome! Thanks a lot.

hellricer:
Awesome! Thanks a lot.

PR #1995, opened 2020-04-20
Improve makefile

### LINKED ISSUE #1994: make command fials on LDFLAGS with spaces

## Info

- OS
    - [X] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [X] bash
    - [ ] zsh
    - [ ] fish
    
When I try to build fzf, by running make, I receive the following message:
```
GOARCH=amd64 go build -a -ldflags "-X main.revision=334a4fa -w -extldflags='-Wl,-O1 -Wl,--as-needed'" -tags "" -o target/fzf-linux_amd64
# github.com/junegunn/fzf
flag provided but not defined: -Wl,--as-needed'
usage: link [options] main.o
  -B note
    	add an ELF NT_GNU_BUILD_ID note when using ELF
  -E entry
    	set entry symbol name
  -H type
    	set header type
  -I linker
    	use linker as ELF dynamic linker
  -L directory
    	add specified directory to library path
  -R quantum
    	set address rounding quantum (default -1)
  -T address
    	set text segment address (default -1)
  -V	print version and exit
  -X definition
    	add string value definition of the form importpath.name=value
  -a	disassemble output
  -buildid id
    	record id as Go toolchain build id
  -buildmode mode
    	set build mode
  -c	dump call graph
  -compressdwarf
    	compress DWARF if possible (default true)
  -cpuprofile file
    	write cpu profile to file
  -d	disable dynamic executable
  -debugtramp int
    	debug trampolines
  -dumpdep
    	dump symbol dependency graph
  -extar string
    	archive program for buildmode=c-archive
  -extld linker
    	use linker when linking in external mode
  -extldflags flags
    	pass flags to external linker
  -f	ignore version mismatch
  -g	disable go package data checks
  -h	halt on error
  -importcfg file
    	read import configuration from file
  -installsuffix suffix
    	set package directory suffix
  -k symbol
    	set field tracking symbol
  -libgcc string
    	compiler support lib for internal linking; use "none" to disable
  -linkmode mode
    	set link mode
  -linkshared
    	link against installed Go shared libraries
  -memprofile file
    	write memory profile to file
  -memprofilerate rate
    	set runtime.MemProfileRate to rate
  -msan
    	enable MSan interface
  -n	dump symbol table
  -o file
    	write output to file
  -pluginpath string
    	full path name for plugin
  -r path
    	set the ELF dynamic linker search path to dir1:dir2:...
  -race
    	enable race detector
  -s	disable symbol table
  -strictdups int
    	sanity check duplicate symbol contents during object file reading (1=warn 2=err).
  -tmpdir directory
    	use directory for temporary files
  -u	reject unsafe packages
  -v	print link trace
  -w	disable DWARF generation
make: *** [Makefile:103: target/fzf-linux_amd64] Error 2
```

The issue is that my LDFLAGS variable contains a space. This, along with the fact that `-extldflags` is not quoted in the **makefile**, leads to the error.
Although I can change my LDFLAG to accomodate the `go build` command, I thought it would be much better if the make command can quote the `-extldflags` argument.

Something like this should do:
> BUILD_FLAGS    := -a -ldflags "-X main.revision=$(REVISION) -w '-extldflags=$(LDFLAGS)'" -tags "$(TAGS)"

rafspiny:
Making sure that `extldflags` is quoted so that **LDFLAGS ** containing spaces won't break the build command.
Close #1994

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ ROOT_DIR       := $(shell dirname $(MAKEFILE))
 SOURCES        := $(wildcard *.go src/*.go src/*/*.go) $(MAKEFILE)
 
 REVISION       := $(shell git log -n 1 --pretty=format:%h -- $(SOURCES))
-BUILD_FLAGS    := -a -ldflags "-X main.revision=$(REVISION) -w -extldflags=$(LDFLAGS)" -tags "$(TAGS)"
+BUILD_FLAGS    := -a -ldflags "-X main.revision=$(REVISION) -w '-extldflags=$(LDFLAGS)'" -tags "$(TAGS)"
 
 BINARY32       := fzf-$(GOOS)_386
 BINARY64       := fzf-$(GOOS)_amd64


junegunn:
Merged, thanks.

Issue #2001, opened 2020-04-23
Ctrl+R doesn't work but fc | fzf does


gvelchuru:
- [X] I have read through the manual page (`man fzf`)
- [X] I have the latest version of fzf
- [X] I have searched through the existing issues

## Info

- OS
    - [X] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [ ] bash
    - [X] zsh
    - [ ] fish
    
## Problem / Steps to reproduce
When I try fc | fzf, everything is fine. However, I wanted to use the fzf history widget, and when I press Ctrl+R, it does not work. I think this line fails: 
```
fc -rl 1 | perl -ne 'print if !$seen{($_ =~ s/^\s*[0-9]+\s+//r)}++' 
```
When I try this I get this error:
```
Bareword found where operator expected at -e line 1, near "s/^\s*[0-9]+\s+//r"
syntax error at -e line 1, near "s/^\s*[0-9]+\s+//r"
syntax error at -e line 1, at EOF
Execution of -e aborted due to compilation errors.
```

junegunn:
What's your `perl --version`?

And does this help?

```diff
diff --git a/shell/key-bindings.zsh b/shell/key-bindings.zsh
index 1960892..23ef8cf 100644
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -100,7 +100,7 @@ bindkey '\ec' fzf-cd-widget
 fzf-history-widget() {
   local selected num
   setopt localoptions noglobsubst noposixbuiltins pipefail no_aliases 2> /dev/null
-  selected=( $(fc -rl 1 | perl -ne 'print if !$seen{($_ =~ s/^\s*[0-9]+\s+//r)}++' |
+  selected=( $(fc -rl 1 | perl -ne '$o = $_; s/^\s*[0-9]+\s+//; print $o if !$seen{$_}++' |
     FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} $FZF_DEFAULT_OPTS -n2..,.. --tiebreak=index --bind=ctrl-r:toggle-sort $FZF_CTRL_R_OPTS --query=${(qqq)LBUFFER} +m" $(__fzfcmd)) )
   local ret=$?
   if [ -n "$selected" ]; then
```

PR #2003, opened 2020-04-24
compat with older perls

eserte:
s///r is only available since perl 5.14. The perl oneliner
was changed to work with older perls, possibly even with 5.000.


junegunn:
Thanks.

This should fix #2001.

PR #2014, opened 2020-05-03
Added a flag to the install script to skip download and build

xSetech:
The added --conf flag is the complement of --bin.

I added this flag since I wanted shell autocomplete without downloading or building fzf (already provided by my distribution at /usr/bin/fzf).

### DIFF
--- a/install
+++ b/install
@@ -7,6 +7,7 @@ auto_completion=
 key_bindings=
 update_config=2
 binary_arch=
+skip_download=0
 shells="bash zsh fish"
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks, but the installer already checks if fzf is in your $PATH, and skips downloading if the version of the binary is the latest, so I don't think we need this flag. Also, it's not a good idea to allow only installing the latest scripts without making sure that you have the latest binary because the scripts may require the new features from the latest version.

PR #2037, opened 2020-05-14
[zsh fzf-history-widget] allow editing and executing directly

commiyou:


junegunn:
Thanks for sharing your solution, however, I'm going to close this. See my previous comment here: https://github.com/junegunn/fzf/pull/1492#issuecomment-461670530

Issue #2041, opened 2020-05-15
Options altered in Vim window after using popup


lacygoill:
- [x] I have read through the manual page (`man fzf`)
- [x] I have the latest version of fzf
- [x] I have searched through the existing issues

## Info

- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [ ] bash
    - [x] zsh
    - [ ] fish

## Problem / Steps to reproduce

After using fzf in a Vim popup window, the options `'bufhidden'`, `'winfixwidth'`, `'winfixheight'` are not properly restored.  This can cause unexpected behaviors.  For example, the focused window is sometimes wrongly restored when quitting a window.

**To Reproduce**

Run this shell command:

    $ vim -Nu <(cat <<'EOF'
        set rtp^=~/.fzf rtp-=~/.vim rtp-=~/.vim/after hidden
        let g:fzf_layout = {'window': #{width: 0.9, height: 0.6}}
        au VimEnter * wincmd b
    EOF
    ) -o /tmp/file{1..2}

Run this Ex command:

    :call fzf#run(fzf#wrap(#{source: ['a'], sink: {-> 0}}))

Press Escape to quit the fzf popup, and run these Ex commands:

    :sp
    :q

Vim focuses the window 1.  This is wrong; it should focus the window 2.
I think that's because the option `'winfixheight'` has been wrongly set in the window 2.

Opening the fzf popup also causes `'winfixwidth'` to be set, and `'bufhidden'` to be set to `hide` in the current Vim window.

**Expected behavior**

The options `'winfixwidth'`, `'winfixheight'` and `'bufhidden'` are not altered in the current Vim window after using the fzf popup.

**Screenshots**

![gif](https://user-images.githubusercontent.com/8505073/82052645-48588400-96bc-11ea-9a34-5d487edb09ec.gif)

**Environment**

 - Vim version: 8.2 Included patches: 1-752
 - OS: Ubuntu 16.04.6 LTS
 - Terminal: XTerm(322)

**Additional context**

This is a regression introduced in https://github.com/junegunn/fzf/commit/d631c76e8d2dd18876a09d2c0919ece1a42a8721.

/cc @ichizok

lacygoill:
The issue comes from here: https://github.com/junegunn/fzf/blob/d631c76e8d2dd18876a09d2c0919ece1a42a8721/plugin/fzf.vim#L838-L844

[...]

PR #2042, opened 2020-05-17
[vim] Don't set wfw, wfh, bh options when opening popup

lacygoill:
This PR tries to fix the issue #2041.

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -680,7 +680,9 @@ function! s:split(dict)
     endif
     return [ppos, { '&l:wfw': &l:wfw, '&l:wfh': &l:wfh }, is_popup]
   finally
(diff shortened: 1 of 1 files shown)


junegunn:
Makes sense, thanks!

plugin/fzf.vim
```diff
@@ -680,7 +680,9 @@ function! s:split(dict)
     endif
     return [ppos, { '&l:wfw': &l:wfw, '&l:wfh': &l:wfh }, is_popup]
```
junegunn, line 681: We can also do this, right?

```vim
return [ppos, is_popup ? {} : { '&l:wfw': &l:wfw, '&l:wfh': &l:wfh }, is_popup]
```

PR #2045, opened 2020-05-18
Add support for Cygwin

prekageo:
This pull request adds fzf support for Cygwin.

A helper Python script is used to control and read from the terminal.


prekageo:
Any interest in this pull request?

junegunn:
Not really. I don't want to make fzf binary rely on an external script even if it's only required on Cygwin environment. Also, I'm not a Cygwin user, so I don't think I can maintain the code with confidence.

prekageo:
I understand. I have forked fzf here: https://github.com/prekageo/fzf-cygwin for anyone that might want to use it on Cygwin. I am closing this pull request.

PR #2054, opened 2020-05-22
Fix rendering in vim is broken with zoomed tmux.

seiyeah78:
### Description
When using fzf with tmux popup window options, rendering in vim is broken.
The cause seems to be in this [approach](https://github.com/junegunn/fzf/pull/433). 

### Screen capture
Before fix
before_fix
After fix
after_fix


junegunn:
Merged, thanks!

seiyeah78:
😄

PR #2064, opened 2020-05-30
README: Use --line-range instead of head in bat example

xerus2000:
From the bat man page:
```
       -r, --line-range <N:M>...

              Only print the specified range of lines for each file. For example:

              --line-range 30:40
                     prints lines 30 to 40

              --line-range :40
                     prints lines 1 to 40
```
I think this should be more efficient than head? 
And either way, it is the proper way, better to make it easy to find because it took me some time to find it :sweat_smile: 

Also, why 500 lines? What insane terminal is that where you can see 500 lines vertically without scrolling? ^^

### DIFF
--- a/README.md
+++ b/README.md
@@ -553,7 +553,7 @@ syntax-highlights the content of a file.
 - Highlight: http://www.andre-simon.de/doku/highlight/en/highlight.php
 
 ```bash
-fzf --preview 'bat --style=numbers --color=always --line-range :99 {}'
+fzf --preview 'bat --style=numbers --color=always --line-range :500 {}'
 ```
 
 You can customize the size, position, and border of the preview window using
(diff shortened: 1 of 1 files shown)


junegunn:
500, which is an arbitrary number that is supposed to be larger than the height of a normal terminal, allows you to scroll down the preview window for some degree, while still limiting the number of lines fzf has to process. There is no particular reason that it should be 500, but 99 seems not enough, e.g. pivot screen.

xerus2000:
how can I scroll the preview?

junegunn:
Using mouse scroll or using custom `--bind` keys.

xerus2000:
oh wow, I didn't even notice that - maybe add that to the README ;)

xerus2000:
alright, I set it back to 500

xerus2000:
and added a hint for scrolling

junegunn:
Merged, thanks!

PR #2077, opened 2020-06-09
Use fzf-redraw-prompt in history widget

jcassee:
I am using Powerlevel10k. When I use `fzf-tmux` with the history widget, it corrupts my prompt. After selecting from the widget, the command is printed, but it seems the cursor position is wrong. After typing anything, the command is duplicated like this:
```
❯ echo fooecho foo
```

Using `fzf-redraw-prompt` for the history widget fixes the problem.

### DIFF
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -109,7 +109,7 @@ fzf-history-widget() {
       zle vi-fetch-history -n $num
     fi
   fi
-  zle reset-prompt
+  zle fzf-redraw-prompt
   return $ret
 }
 zle     -N   fzf-history-widget


junegunn:
While installing Powerlevel10k, I noticed there are a bunch of options that one has to choose, and in the end, I couldn't reproduce the problem you described with my set of options. Please provide a minimal configuration file that can reproduce the issue.

jcassee:
@romkatv With Powerlevel10k disabled, the behavior is normal again.

But! If I load the default config that also fixes the problem. I tried disabling all prompt elements, but that did not change anything. I will investigate further.

PR #2087, opened 2020-06-20
fzf#wrap: fix message about signature

przepompownia:


przepompownia:
The message about using boolean values is misleading. Both `v:true` and `v:false` are not allowed by the current implementation.

junegunn:
Thanks. We can support v:true/false as well as 1/0.

przepompownia:
Thanks for fix.

PR #2096, opened 2020-06-25
preserve current directory in case someone changes it

yanlinsun:
I use ```autocmd BufEnter * silent! lcd %:p:h``` as suggested in this document <https://vim.fandom.com/wiki/Set_working_directory_to_the_current_file>.

When the '--multi' flag is set and I try to open multiple files, the current directory is changed after the first file is opened, which results in the following files failed to open because of the filenames are relative paths to "current directory".

I think this solution covers both 'autochdir' and 'BufEnter lcd' scenarios.

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -236,8 +236,8 @@ function! s:common_sink(action, lines) abort
   endif
   try
     let empty = empty(s:fzf_expand('%')) && line('$') == 1 && empty(getline(1)) && !&modified
-    " Preserve current directory in case current working directory is changed
-    " during the execution
+    " Preserve the current working directory in case it's changed during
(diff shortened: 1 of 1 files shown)


yanlinsun:
The solution is not ideal. The '%:p:h' may be different from the starting dir of files command. I'll try to make another change.

yanlinsun:
I cover the scenarios that I can think of. Please have a review, thanks.

junegunn:
Thanks, I made a few changes. See the commit message.

junegunn:
Also realized that the original patch breaks how relative paths are handled.

- `~/foo/bar` -> `/pwd/~/foo/bar`

Pushed the fix to your branch. Please test it and let me know what you think.

PR #2100, opened 2020-06-29
Void Linux installation instructions

anntnzrb:
[XBPS](https://voidlinux.org/usage/xbps/) is the package manager of [Void Linux](https://voidlinux.org/)

the package can be found by searching it via the [Void Packages](https://voidlinux.org/packages/) website "fzf" (unable to provide direct URL)


junegunn:
Merged, thanks.

Issue #1960, opened 2020-04-07
exact match that is also case-sensitive


mckellyln:
- [x] I have read through the manual page (`man fzf`)
- [x] I have the latest version of fzf
- [x] I have searched through the existing issues

## Info

- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [x] bash
    - [x] zsh
    - [ ] fish
    
## Problem / Steps to reproduce

This is not a problem, but a question/enhancement request.

When using fzf (with fd) to get a listing of files and dirs I can use 'wild for an exact match, but this still seems to be smart-case.
I can add +i but it is nice to have smart-case most of the time and have a single cmd that works for all types of searches.
Does it make sense that 'wild means exact + case-sensitive match ?  If not then perhaps a new ` or some other way to make it a true exact + case-sensitive match ?

thx for awesome fzf.
-m

dagadbm:
yes pls. this is very helpful specially when browsing through code using fzf.vim !

mckellyln:
Closing issue, but pull request is still active.

PR #2110, opened 2020-07-08
Add " like ' to be exact + case-sensitive search

mckellyln:
@junegunn what do you think ?
I use smart case all the time, but sometimes within that I want an exact + case-sensitive search.
This is issue: https://github.com/junegunn/fzf/issues/1960


mckellyln:
Closing for now, may re-submit PR at a later time.
thx for awesome fzf,
-m

PR #2121, opened 2020-07-26
Support `**` completion for `fc`

kflu:
This adds support for quick selecting, editing, and re-running history
commands via `fc`:

    $ fc **<TAB>
    fc>
     1000/1000
      3645  git st
      3644  vim .
      3643  ls -l
      ...

Upon selecting, e.g., "3644 vim.", the command becomes:

    $ fc 3644

Pressing <ENTER> then allows user's $EDITOR being open to edit that
history command, and re-run the editted command.

### DIFF
--- a/shell/completion.zsh
+++ b/shell/completion.zsh
@@ -260,6 +260,14 @@ _fzf_complete_kill_post() {
   awk '{print $2}'
 }
 
+_fzf_complete_fc() {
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks for your interest in this project, but we decided not to add more custom completion functions and features to this repository in order to avoid extra maintenance burden. You might want to define them in your configuration file or create a separate third-party project for serving the functions.

PR #2140, opened 2020-08-15
[zsh] Fix regular expression in history search function

ynakao:
# Summary

Fix the regular expression to capture the command containing asterisk.

# Details

After [this commit](https://github.com/junegunn/fzf/commit/e6d33f77da8cc36786c814f499f6cb3405e8c4c1#diff-961174188263ff616f70beae85ee0763), CTRL-R command history search has failed to display some commands in zsh.

The new method to cut the history number from the each line can't capture the command containing an asterisk(*) mark followed by the number.

If you have command history like an example below, CTRL-R only displays `ls`, `vi`, `man`, `git` and `ssh`... .

`$ fc -rl 1` output example
```shell
123* ls
122* cd
121* date
120 vi
119 man
118* git
117* which
116* emacs
115 ssh
...
```

I have little knowledge about perl, but I guess when it processes the `123* ls` line, capturing failure makes `'undef'` hash key. So the next commands(`cd` and `date`) including asterisks aren't be displayed because perl recognize them as the same commands.

While I don't know why `git` is displayed, my assumption is `undef`s that are not produced in the continuous block are treated differently by perl internals.

### DIFF
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -100,7 +100,7 @@ bindkey '\ec' fzf-cd-widget
 fzf-history-widget() {
   local selected num
   setopt localoptions noglobsubst noposixbuiltins pipefail no_aliases 2> /dev/null
-  selected=( $(fc -rl 1 | perl -ne 'print if !$seen{(/^\s*[0-9]+\s+(.*)/, $1)}++' |
+  selected=( $(fc -rl 1 | perl -ne 'print if !$seen{(/^\s*[0-9]+\**\s+(.*)/, $1)}++' |
     FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} $FZF_DEFAULT_OPTS -n2..,.. --tiebreak=index --bind=ctrl-r:toggle-sort $FZF_CTRL_R_OPTS --query=${(qqq)LBUFFER} +m" $(__fzfcmd)) )
   local ret=$?
   if [ -n "$selected" ]; then


junegunn:
Thanks. Do you know when `*` appears? My `fc` output doesn't have any item with `*`.

ynakao:
If I understand correctly, when you execute some commands in other terminal sessions, asterisks are added to those commands in the original session.

For example,

1. Open two terminal sessions side by side
2. In the first terminal, execute `ls`, `cd`, `date` in order. At this point, `fc -rl 1` shows history without *s.

```
3 date
2 cd
1 ls
```

3. In the second terminal, execute those commands in the same order `ls`, `cd`, `date`.
4. Go back to the first terminal, and see the output of `fc -rl 1`. *s are added.

```
3* date
2* cd
1* ls
```

I've searched for the zsh documentation mentioning this behavior, but can't find it so far.

junegunn:
Thanks, I can reproduce the problem by following your instruction after adding `setopt share_history`.

junegunn (approved):

junegunn:
Merged, thanks!

hotpxl:
Ah, I bumped into the same issue, thanks for the fix here.

According to `builtin.c` in zsh,
```
fprintf(f, "%5s%c ", buf,
    ent->node.flags & HIST_FOREIGN ? '*' : ' ');
}
```
it seems indeed that `fc` adds an asterisk for commands that came from another shell.

nh2:
A couple years later, I found the match for `\**` very surprising, so I am PRing a comment for it in PR #3006.

nh2:
> According to `builtin.c` in zsh,

Stable code link: https://github.com/zsh-users/zsh/blob/b7f280ab597b07bbb22fb741589fe7754fc5cc0c/Src/builtin.c#L1822-L1828

PR #2148, opened 2020-08-20
Check and ignore duplicate items

slim-bean:
Hello! Thank you so much for creating such a wonderful tool!

Apologies as I believe this PR to be a bit controversial, I realize the issue of de-duplication has been well discussed

#600, #49, #88, #492

And in previous discussions the decision was this functionality is better handled outside of fzf.

I have a use case where I'm finding this isn't really true, mainly because tooling outside of fzf for deduplication is either clunky or won't work they way I want at all.

I am streaming data into fzf via stdin, my use case is retrieving command history from an external service in batches and then sending them to fzf via stdin (ctrl-r support from an external source)

The problem with all the existing solutions I have seen for de-duplication outside fzf require all the items to be loaded into the intermediate command before they will be de-duplicated and sent to fzf.

Unfortunately this makes the user experience a lot worse in my use case, as I need to wait for all the remote entries to be fetched before they can be displayed.

As the external system in my case is able to send the entries in the correct order, it's way cooler and faster to immediately display the first batch of results (most of the time you want commands from recent history), and then let older commands continue to load in the background.  fzf already presents a really slick loading indicator while those results stream in!

By de-duplicating within fzf I an stream these results in for a really slick user experience while still removing dupes.

This PR is how I accomplished this, I would love to see if you are interested in merging this.  If so, I'm more than happy to finish work on this (make it an actual command line flag, add tests, documentation etc)

Thanks!

### DIFF
--- a/src/chunklist.go
+++ b/src/chunklist.go
@@ -1,6 +1,9 @@
 package fzf
 
-import "sync"
+import (
+	"bytes"
+	"sync"
+)
 
 // Chunk is a list of Items whose size has the upper limit of chunkSize
 type Chunk struct {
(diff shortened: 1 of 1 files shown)


junegunn:
> The problem with all the existing solutions I have seen for de-duplication outside fzf require all the items to be loaded into the intermediate command before they will be de-duplicated and sent to fzf.

That is not true. Try this command:

```sh
(echo 1; sleep 1; echo 2; sleep 1; echo 1; sleep 1; echo 3) | stdbuf -o0 awk '!seen[$0]++' | fzf
```

slim-bean:
I can tell you it was easier for me to modify fzf than it is for me to understand what that command is doing and how it works.

That being said it does work!  Curious why this isn't in the command completion scripts instead of perl?

I don't see any reason to keep this open the command you provided suits my needs.

If you ever are interested in adding this into fzf let me know i'd be happy to finish this PR.

Thanks!

PR #2150, opened 2020-08-22
Add a missing help for no-multi option.

VasilyFomin:
It's included in the man page already.


junegunn:
Thanks for the pull request. Actually, every flag has its `no-` companion, it's a little convention in fzf. e.g. `--no-cycle`, `--no-color`, `--no-ansi`, etc. And we don't want to show every one of them in `--help` output to keep it short and concise.

PR #2157, opened 2020-08-29
Minor fix in readme

sayanarijit:


junegunn:
`kill` completion actually works with or without `**` prefix.

> Fuzzy completion for PIDs is provided for kill command. In this case,
> there is no trigger sequence, just press tab key after kill command.

I haven't decided if we should "un-document" the exception.

sayanarijit:
Oh right... Missed this.

PR #2206, opened 2020-10-09
Add upleft and downright option for preview window.

arsane:
These options let fzf decide where to put the preview window by
checking current terminal's width/height; it's helpful when we
move the window from one monitor to another.

### DIFF
--- a/src/options.go
+++ b/src/options.go
@@ -80,7 +80,7 @@ const usage = `usage: fzf [options]
   Preview
     --preview=COMMAND     Command to preview highlighted line ({})
     --preview-window=OPT  Preview window layout (default: right:50%)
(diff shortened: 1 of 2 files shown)


junegunn:
Thanks, but I don't think this logic should be built into fzf. It's much more flexible to choose the right option outside of fzf.

```sh
if (( $(tput cols) > 150 )) || (( $(tput cols) > 1.5 * $(tput lines) )); then
  export FZF_DEFAULT_OPTS='--preview-window right'
else
  export FZF_DEFAULT_OPTS='--preview-window up'
fi
```

arsane:
It works with external options until you move your terminal from one screen to another, or open another terminal in tiling mode.
In such scenario, it won't resize automatically.

PR #2208, opened 2020-10-12
[fish] Fix parser handling of option-like args

0az:
Fixes error when option-like args are parsed (e.g. "-1", followed by `c-t` to trigger fzf).


junegunn:
Merged, thanks.

PR #2214, opened 2020-10-16
[key-bindings] find hidden files and directories with CTRL-T and ALT-C

dashezup:
Currently the `find` command in `shell/key-bindings.{bash,zsh,fish}` does not list hidden files and directories, from the command it looks like it supposed to list hidden files and folders but it does not because of don't have enough escapes.
> `command find -L . -mindepth 1 \\( -path '*/\\.*' -o -fstype 'sysfs' -o -fstype 'devfs' -o -fstype 'devtmpfs' -o -fstype 'proc' \\) -prune -o -type f -print -o -type d -print -o -type l -print 2> /dev/null | cut -b3-`

I'm not very familiar with escaping, I think the command needs two escapes and we need to have three backslash here. first one for the dot, second one for find, third one for eval. 
> `-path '*/\\.*'`

But from actual test; the result is different
|shell|\ in command|\ in script|
|---|---|---|
|bash|2|4|
|zsh|2|3|
|fish|3|5|

Having more backslash could also work, if anyone can explain how many backslash should be added here maybe we can add the same amount of backslash to both bash, zsh and fish. The current PR includes the minimal backslashes to be able to list hidden files and folders.

btw I think it should be able to select hidden files and directories, because when we type `ls <tab>`, `ls .<tab>`, `cd <tab>` and `cd .<tab>` but currently this script can't select hidden files/directories.

### DIFF
--- a/shell/key-bindings.bash
+++ b/shell/key-bindings.bash
@@ -14,7 +14,7 @@
 # Key bindings
 # ------------
 __fzf_select__() {
-  local cmd="${FZF_CTRL_T_COMMAND:-"command find -L . -mindepth 1 \\( -path '*/\\.*' -o -fstype 'sysfs' -o -fstype 'devfs' -o -fstype 'devtmpfs' -o -fstype 'proc' \\) -prune \
+  local cmd="${FZF_CTRL_T_COMMAND:-"command find -L . -mindepth 1 \\( -path '*/\\\\.*' -o -fstype 'sysfs' -o -fstype 'devfs' -o -fstype 'devtmpfs' -o -fstype 'proc' \\) -prune \
     -o -type f -print \
     -o -type d -print \
     -o -type l -print 2> /dev/null | cut -b3-"}"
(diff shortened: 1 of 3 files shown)


rayiik:
why don't
 you just replace find with fd (fdfind on Debian) -uu (list hidden and git ignore files) or --hidden its faster and for the example case-use functions fine (I've replaced find in both my binding files with fd) should be the same across the board for zsh fish and bash. or even setting it as  a variable substitution so if you need to (for some reason) use find instead you can set the variable and expand into that situation in bah "${_findcmd:-'fd -uu -t file}"
unless i need one of finds special unique functions i always replace it with fd as its faster (a lot) and generally is more stable.

dashezup:
1. user need to install `fd` to use it, while `find` is installed on most of linux distro and macOS. the shell keybinding plugin should works out of box
2. fd does not support `-fstype`
3. `find` is faster than `fd`
I tried to run this commands under my `$HOME`

|command|result|user|system|cpu|total|
|---|---|---|---|---|---|
|`time find -L . -mindepth 1 \( -path '*/\\.*' -o -fstype 'sysfs' -o -fstype 'devfs' -o -fstype 'devtmpfs' -o -fstype 'proc' \) -prune -o -type f -print -o -type d -print -o -type l -print >/dev/null 2>&1 \| wc -l`|1222577|0.09s|0.15s|6%|3.697|
|`time find -L . -mindepth 1 -type f,d,l >/dev/null 2>&1 \| wc -l`|1222588|0.09s|0.14s|7%|3.097|
|`time fd -L . --min-depth 1 -t f -t d -t l -H \| wc -l`|626372|0.18s|0.56|19%|3.755|

rayiik:
not sure why but none of those for commands produced results for me from hoe dir (0) on the wc -l  but i hear what you're saying, however, its fast-movingi towards most systems having fd available and i guess becasue i have the cpu i dont really notice it but its a fair point.  was more suggesting for personal use than a patch fix i was just wondering if there was a reason you werent using it and thats a pretty valid reason :) (still learning here and trying to get prespective on peoples reasonings and conventions). so thanks for the analysis.

junegunn:
> it looks like it supposed to list hidden files and folders

No, it's not. The option is followed by `-prune`, which means we want to exclude the ones that match the condition.

PR #2215, opened 2020-10-16
Implement streaming preview window

### LINKED ISSUE #2212: Preview loading

## Info

- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [x] bash
    - [x] zsh
    - [x] fish
    
While a preview script/command is executing, preview from previous item is displayed. I'm assuming it's because fzf is waiting for script to execute and then echoes its output? Is it possible to show some "loading" indicator until preview script finishes (for example, quering database or caching remote data)?

junegunn:
Preliminary implementation of the streaming preview window. Please test and report any bugs. Fix #2212.

```sh
# Will start rendering after 200ms, update every 100ms
fzf --preview 'for i in $(seq 100); do echo $i; sleep 0.01; done'

# Should print "Loading preview" message after 500ms
fzf --preview 'sleep 1; for i in $(seq 100); do echo $i; sleep 0.01; done'

# The first line should appear after 200ms
fzf --preview 'date; sleep 2; date'

# Should not render before enough lines for the scroll offset are ready
rg --line-number --no-heading --color=always ^ |
  fzf --delimiter : --ansi --preview-window '+{2}-/2' \
      --preview 'sleep 1; bat --style=numbers --color=always --pager=never --highlight-line={2} {1}'
```

/ping @p-kolacz @edi9999

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 CHANGELOG
 =========
 
+0.24.0
+------
+- fzf can render preview window before the command completes
+  ```sh
+  fzf --preview 'sleep 1; for i in $(seq 100); do echo $i; sleep 0.01; done'
+  ```
+
 0.23.1
 ------
 - Added `--preview-window` options for disabling flags
--- a/README.md
+++ b/README.md
@@ -582,9 +582,9 @@ and fzf will warn you about it. To suppress the warning message, we added
 
 ### Preview window
 
-When the `--preview` option is set, fzf automatically starts an external process 
-with the current line as the argument and shows the result in the split window. 
-Your `$SHELL` is used to execute the command with `$SHELL -c COMMAND`. 
+When the `--preview` option is set, fzf automatically starts an external process
+with the current line as the argument and shows the result in the split window.
+Your `$SHELL` is used to execute the command with `$SHELL -c COMMAND`.
 The window can be scrolled using the mouse or custom key bindings.
 
 ```bash
@@ -592,16 +592,8 @@ The window can be scrolled using the mouse or custom key bindings.
 fzf --preview 'cat {}'
 ```
 
-Since the preview window is updated only after the process is complete, it's
-important that the command finishes quickly.
-
-```bash
-# Use head instead of cat so that the command doesn't take too long to finish
-fzf --preview 'head -100 {}'
-```
-
 Preview window supports ANSI colors, so you can use any program that
-syntax-highlights the content of a file, such as 
+syntax-highlights the content of a file, such as
 [Bat](https://github.com/sharkdp/bat) or
 [Highlight](http://www.andre-simon.de/doku/highlight/en/highlight.php):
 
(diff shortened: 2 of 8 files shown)


edi9999:
I see some issues on some of my custom scripts.

Only the first line is shown even after the preview has fully loaded.

When I start to scroll to the bottom, then I see the second line and the ones after that.

I'm trying to create a simple example.

edi9999:
Actually it can be reproduced simply with the following : 

FZF_DEFAULT_OPTS="" ./bin/fzf --preview 'echo "1"; sleep 1; echo 2' --preview-window 'nohidden'

edi9999:
See recording : 

record_2020-10-16_213457

edi9999:
You can also reproduce the same issue (probably), with the following : 

```
FZF_DEFAULT_OPTS="" ./bin/fzf --preview 'for i in $(seq 10); do echo $i; sleep 0.13; done' 
```

Here, on my computer, only the first 3 numbers are shown (numbers 1 to 3).

with : 

```
FZF_DEFAULT_OPTS="" ./bin/fzf --preview 'for i in $(seq 10); do echo $i; sleep 0.11; done'
```

it writes number from 1 to 8.

edi9999:
with `FZF_DEFAULT_OPTS="" ./bin/fzf --preview 'for i in $(seq 10); do echo $i; sleep 0.10; done'`,

everything works fine, so there is something special happening near 1 tenth of a second.

junegunn:
Nice find, 63444b4dee580b65a017eb36d67ada40901433bc should fix that.

junegunn:
I'm also seeing an issue where fzf not terminating the running command when the cursor is moved. Need further investigation.

edi9999:
There's one behavior I'm not sure is intended : 

When I run the following : 

```
fzf ./bin/fzf --preview ' yes | head -n 20; sleep 5; echo foo ;  ' --preview-window 'up:nohidden'
```

It shows the 1/20 while loading, even though the preview does not have a scroll.

After the sleep and the echo foo however, the 1/20 disappears.

I suggest to put only the loading screen in this case.

Maybe to avoid some "jittering", it would make more sense to put the loading on the right of the pagination, or maybe one line below that.

junegunn:
This is the intended behavior as of now. You can see there's a braille character indicating that the stream is open (i.e. `⠋ 1/20`). We don't know if the final number of lines will exceed the height of the window or not at this point.

```sh
# `⠋ 1/20` disappears when the stream is closed
fzf --preview ' yes | head -n 20; sleep 5; echo foo ;  ' --preview-window 'up:nohidden'

# `1/1020` remains (spinner on the left gone)
fzf --preview ' yes | head -n 20; sleep 5; seq 1000; sleep 1' --preview-window 'up:nohidden'
```

One idea is to only print `⠋` without numbers before the content overflows.

edi9999:
> One idea is to only print ⠋ without numbers before the content overflows.

Yes, exactly, that is what I thought would make more sense, but I also would understand that you'd prefer to keep current behavior.

edi9999:
I have just tested again and it seems to work well for me.

junegunn:
I'm going to merge this as it all seems to be working fine and test it over the next few days. Please let me know if you find any issues.

edi9999:
Yes, sure I will !

PR #2223, opened 2020-10-26
chore(vim): replace manual border with native's in popup

kevinhwang91:
If `g:fzf_layout.window.border` is empty, and `--border` or `--no-border`
is absent, draw border manually by (neo)vim.

If `g:fzf_layout.window.border` is empty, but `--border` or `--no-border`
is existed, respect native option (border or no border) instead of
always drawing border by (neo)vim.

If `g:fzf_layout.window.border` is existed, ignore native options and
use (neo)vim's layout to draw border.

For now, the default border in the popup is the same as fzf's native border, we should respect the fzf's border options.
If I'm wrong please correct me.

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -670,6 +670,10 @@ function! s:split(dict)
         if !s:popup_support()
           throw 'Nvim 0.4+ or Vim 8.2.191+ with popupwin feature is required for pop-up window'
         end
+        let opts = $FZF_DEFAULT_OPTS.' '.s:evaluate_opts(get(a:dict, 'options', ''))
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks for bringing this up. You made me realize that we had to implement border rendering in VimScript, because back then `--border` of fzf was only for "horizontal" borders. But now, we have more `--border` options (d9b12111912a43e62f61051d40897a378a020cf5) and I think it's about time to simplify the VimScript and use the native `--border` option instead.

`--border` needs a few more options, but it's trivial to add those to fzf. Let me take a stab at it.

kevinhwang91:
Nice, It seems that it's time to close this PR, I'm waiting for your simplified vimscript!

junegunn:
Sure. I'll add `'border': 'none'` option. e73383fbbba172c1375c4b8d09328f30516d5333

PR #2224, opened 2020-10-27
Fix: barbled multibyte text(exe. Japanese).

nekowasabi:
OS : Windows10
vim : gvim 8.2

This pull request is fixed candidate's text with Multibyte language(exe. Japanese).

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -50,9 +50,6 @@ if s:is_win
   " Use utf-8 for fzf.vim commands
   " Return array of shell commands for cmd.exe
   function! s:enc_to_cp(str)
-    if s:is_win && has('gui')
-      return iconv(a:str, 'utf-8', &encoding)
-    endif
     if !has('iconv')
       return a:str
     endif
@@ -407,7 +404,7 @@ try
       let prefix = '( '.source.' )|'
     elseif type == 3
       let temps.input = s:fzf_tempname()
-      call writefile(map(source, '<SID>enc_to_cp(v:val)'), temps.input)
+      call writefile(source, temps.input)
       let prefix = (s:is_win ? 'type ' : 'cat ').fzf#shellescape(temps.input).'|'
     else
       throw 'Invalid source type'
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -50,6 +50,9 @@ if s:is_win
   " Use utf-8 for fzf.vim commands
   " Return array of shell commands for cmd.exe
   function! s:enc_to_cp(str)
+    if s:is_win && has('gui')
+      return iconv(a:str, 'utf-8', &encoding)
+    endif
     if !has('iconv')
       return a:str
     endif


junegunn:
Thanks, can you explain what problem this will solve? Maybe a screenshot before & after the fix will help.

nekowasabi:
Thank you for reply.

This sample is execution by `:History:<CR> ` 

before
Before this fix, there was a problem with the Japanese language rendering (encoding) and the display was not correct.

after

After the fix, Japanese strings should now be displayed correctly.

plugin/fzf.vim
```diff
@@ -50,6 +50,9 @@ if s:is_win
   " Use utf-8 for fzf.vim commands
   " Return array of shell commands for cmd.exe
   function! s:enc_to_cp(str)
+    if s:is_win && has('gui')
+      return iconv(a:str, 'utf-8', &encoding)
+    endif
```
junegunn, line 55: These lines should appear after line 58 because `iconv` may not be available.

junegunn:
Thanks for the comment. Yes, I can easily reproduce the problem on Windows GVim.

```vim
:call fzf#run({'source': ['日本']})
```

However, I'm not familiar with these Windows stuff, and I could really use some advice from @janlazo.

Maybe we need to add a separate function only for `source` array? (`utf-8` -> `encoding`)

junegunn:
Or maybe we shouldn't have applied `enc_to_cp` to `source` array? I mean, this also fixes the problem:

```diff
diff --git a/plugin/fzf.vim b/plugin/fzf.vim
index 9990787..9f280c0 100644
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -437,7 +437,7 @@ try
       let prefix = '( '.source.' )|'
     elseif type == 3
       let temps.input = s:fzf_tempname()
-      call writefile(map(source, '<SID>enc_to_cp(v:val)'), temps.input)
+      call writefile(source, temps.input)
       let prefix = (s:is_win ? 'type ' : 'cat ').fzf#shellescape(temps.input).'|'
     else
       throw 'Invalid source type'

```

nekowasabi:
Thank you.

I checked.

`+ call writefile(source, temps.input)`
This fix solves the problem.
This fix is better.

Should I redo the pull request to reflect the fix?

junegunn:
> This fix is better.

Agreed.

> Should I redo the pull request to reflect the fix?

You mean closing this and opening a new pull request? No, you don't have to. You can just push more commits to your repository to update this one.

janlazo:
`s:enc_to_cp` is meant for the text in the batchfile. Text in another file can remain be in  any encoding as long as fzf can understand it from stdin. Pipes don't change encoding, even on powershell, so the fix to not modify the text for the input file is best.

`set encoding=utf-8` is required for multi-byte characters so I don't think that we need a separate encoding function for text in the input file. It's safe to ignore other `&encoding` values.

IIRC, I used `s:enc_to_cp` for the input file because cmd.exe commands do not suppot utf-8 without `chcp 65001`. I didn't test if `cat.exe` from msys2/mingw/cygwin/busybox is better for utf-8 than `type` command.

junegunn:
@janlazo Thanks for the comment.

@nekowasabi Let me finish this up.

junegunn:
Merged, thanks!

nekowasabi:
Thanks for the merge!

PR #2246, opened 2020-11-11
[bash-completion] Optimize startup + two minor fixes

liskin:
Commit d4ad4a25 slowed loading of completion.bash significantly (on my
laptop from 10 ms to 30 ms), then 54891d11 improved that (to 20 ms) but
it still stands out as the heavy part of my .bashrc. Furthermore, an
omission in d4ad4a25 caused meaningless "complete _a" and "complete _v"
completions to be defined.

This pull request fixes that mistake, refactors the code to prevent
making it again, and most importantly optimizes
__fzf_orig_completion_filter so that loading completion.bash is as fast
as it was before d4ad4a25, that is under 10 ms on my laptop.

(I tried not to use any features not present in bash 3.2, but I didn't
test it with bash 3.2. Can someone try this on MacOS please?)

### DIFF
--- a/shell/completion.bash
+++ b/shell/completion.bash
@@ -53,7 +53,7 @@ __fzf_orig_completion() {
       comp="${BASH_REMATCH[1]}"
       f="${BASH_REMATCH[2]}"
       cmd="${BASH_REMATCH[3]}"
-      export "_fzf_orig_completion_${cmd//[^A-Za-z0-9_]/_}=${comp} %s ${cmd} #${f}"
+      printf -v "_fzf_orig_completion_${cmd//[^A-Za-z0-9_]/_}" "%s" "${comp} %s ${cmd} #${f}"
       if [[ "$l" = *" -o nospace "* ]] && [[ ! "$__fzf_nospace_commands" = *" $cmd "* ]]; then
         __fzf_nospace_commands="$__fzf_nospace_commands $cmd "
       fi
--- a/shell/completion.bash
+++ b/shell/completion.bash
@@ -353,7 +353,7 @@ _fzf_setup_completion() {
     return 1
   fi
   shift
-  eval "$(complete -p "$@" 2> /dev/null | grep -v "$fn" | __fzf_orig_completion_filter)"
+  eval "$(complete -p "$@" 2> /dev/null | sed '/-F/!d; / _fzf/d' | __fzf_orig_completion_filter)"
   for cmd in "$@"; do
     case "$kind" in
       dir)   __fzf_defc "$cmd" "$fn" "-o nospace -o dirnames" ;;
--- a/shell/completion.bash
+++ b/shell/completion.bash
@@ -47,7 +47,7 @@ __fzf_comprun() {
 }
 
 __fzf_orig_completion_filter() {
-  sed 's/^\(.*-F\) *\([^ ]*\).* \([^ ]*\)$/export _fzf_orig_completion_\3="\1 %s \3 #\2"; [[ "\1" = *" -o nospace "* ]] \&\& [[ ! "$__fzf_nospace_commands" = *" \3 "* ]] \&\& __fzf_nospace_commands="$__fzf_nospace_commands \3 ";/' |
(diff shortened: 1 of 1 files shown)


shell/completion.bash
```diff
@@ -46,9 +46,20 @@ __fzf_comprun() {
...
+  while read -r l; do
+    if [[ "$l" =~ ^(.*\ -F)\ *([^ ]*).*\ ([^ ]*)$ ]]; then
+      comp="${BASH_REMATCH[1]}"
+      f="${BASH_REMATCH[2]}"
+      cmd="${BASH_REMATCH[3]}"
+      printf -v "_fzf_orig_completion_${cmd//[^A-Za-z0-9_]/_}" "%s" "${comp} %s ${cmd} #${f}"
+      if [[ "$l" = *" -o nospace "* ]] && [[ ! "$__fzf_nospace_commands" = *" $cmd "* ]]; then
+        __fzf_nospace_commands="$__fzf_nospace_commands $cmd "
+      fi
+      :
```
junegunn, line 60: Do we need this line?

liskin: absolutely not, that's a leftover, sorry :-)

liskin: fixed & forcepushed

junegunn:
Thanks Tomáš, I really like what you did here. Bravo.

liskin:
It's a pleasure. Thank _you_ for maintaining this excellent piece of software!

wingpin6:
Hi guys, thanks for your efforts on this program and improving it.

Unfortunately for me, a commit within this PR has caused my terminal to crash when trying to tab. I've identified commit ef2c29d5d4974cc773149d368488a6d8a0fcc1fc through git bisect as the offending commit.

For reference I am currently using:
GNU bash, version 5.0.17(1)-release (x86_64-pc-linux-gnu)
on Ubuntu 20.04
Terminal choice does not seem to matter; I crash on both Alacritty (my daily driver) and Gnome Terminal.

Where would I find logs on how to diagnose this issue?

liskin:
@wingpin6 Could you please try running a nested `bash` in your terminal and reproducing this in it? This way the terminal won't terminate when the bash terminates and we might see an error message if there is any.

liskin:
@wingpin6 I tried whether I can reproduce it using ubuntu:20.04 in docker and I can't, so it seems this is in some way related to your environment or other bits of your bashrc. :-/

wingpin6:
@liskin Thank you for the tip; after doing nested bash --norc I found that
- sourcing only .bashrc did not crash
- sourcing only .fzf.bash did not crash
- sourcing .fzf.bash and then .bashrc did not crash
- However, sourcing .bashrc and then .fzf.bash caused it to crash

The error it outputted was Segmentation fault.

I looked deeper into this issue, and was able to fix it if I removed
`[ -f ~/.fzf.bash ] && source ~/.fzf.bash`
from my .bashrc

It seems something in that commit mentioned earlier may have sourced my .bashrc, which sources .fzf.bash and then causes an infinite loop?

I was able to fix my issue by not using .bashrc as my default bash config file, and then sourcing .bashrc and .fzf.bash within that new config file separately. 

Let me know if you need more info or if you can't replicate that.

liskin:
@wingpin6 Segmentation fault seems to be a bug in bash, provided it comes from bash and not from something else.

Could you get a stack trace from the core dump? I'm not sure if Ubuntu has systemd-coredump, but if yes, then this should be as easy as running `coredumpctl info` and sending the output. If you could also `apt install bash-dbgsym` (https://wiki.ubuntu.com/Debug%20Symbol%20Packages) and then do `coredumpctl debug` and then type `bt`into gdb, that would be wonderful.

If there's no systemd-coredump in Ubuntu, instead of running bash directly, run `gdb -ex run --args bash` and reproduce the crash and then when it crashes, get stack trace by typing `bt`.

I also looked at recent changes in bash in Ubuntu and it seems that this may possibly help: https://github.com/liskin/fzf/commit/6b4ab0a2af42c2103cb4b7b96f4fba076b4cd002, although that is really a wild guess, nothing else.

liskin:
Oh, and it seems I can replicate that now. The key is sourcing completion.bash twice!
No need to debug it then, I'll take care of it!

liskin:
@wingpin6 Fix in https://github.com/junegunn/fzf/pull/2250. Thanks for reporting this and sorry for breaking your setup.

(You might, however, consider sourcing completion.bash just once in the future, as that will be faster. :-))

wingpin6:
Thank you for the quick fix!

I agree this case is most likely uncommon. I have a personal config file that I use on all my machines and a work config file for my work machine. I source both of them with work config overriding anything I defined in my personal config, with some overlaps.

I happened to source .fzf.bash in both of them; not sure how that got into my work config file. But great to find out that I can remove that line now!

PR #2250, opened 2020-11-13
[bash-completion] Fix endless loop when completion.bash sourced twice

liskin:
I forgot to add the "not _fzf" check into __fzf_orig_completion, so
invoking it twice would rewrite the _fzf_orig_completion_xxx variables
and then cause an endless loop when completion is requested.

Fixes: https://github.com/junegunn/fzf/pull/2246
Reported-by: @wingpin6


junegunn:
Thanks for the quick fix!

PR #2251, opened 2020-11-14
[zsh] Reload shared history before searching

mpolden:
When history sharing across shells is enabled (`setopt SHARE_HISTORY`), history written by shell A won't be available in shell B until re-rendering the prompt in B (e.g. by pressing Enter at the prompt).

With this change, C-r will always reload the history file if history sharing is enabled, making the latest history searchable immediately.

### DIFF
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -106,6 +106,9 @@ bindkey '\ec' fzf-cd-widget
 fzf-history-widget() {
   local selected num
   setopt localoptions noglobsubst noposixbuiltins pipefail no_aliases 2> /dev/null
+  if [[ -o sharehistory && -z "$BUFFER" ]]; then
+      fc -R "$HISTFILE"
+  fi
   selected=( $(fc -rl 1 | perl -ne 'print if !$seen{(/^\s*[0-9]+\**\s+(.*)/, $1)}++' |
     FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} $FZF_DEFAULT_OPTS -n2..,.. --tiebreak=index --bind=ctrl-r:toggle-sort $FZF_CTRL_R_OPTS --query=${(qqq)LBUFFER} +m" $(__fzfcmd)) )
   local ret=$?
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -107,7 +107,7 @@ fzf-history-widget() {
   local selected num
   setopt localoptions noglobsubst noposixbuiltins pipefail no_aliases 2> /dev/null
   if [[ -o sharehistory && -z "$BUFFER" ]]; then
-      fc -R "$HISTFILE"
+    fc -R "$HISTFILE"
   fi
   selected=( $(fc -rl 1 | perl -ne 'print if !$seen{(/^\s*[0-9]+\**\s+(.*)/, $1)}++' |
     FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} $FZF_DEFAULT_OPTS -n2..,.. --tiebreak=index --bind=ctrl-r:toggle-sort $FZF_CTRL_R_OPTS --query=${(qqq)LBUFFER} +m" $(__fzfcmd)) )
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -106,9 +106,7 @@ bindkey '\ec' fzf-cd-widget
 fzf-history-widget() {
   local selected num
   setopt localoptions noglobsubst noposixbuiltins pipefail no_aliases 2> /dev/null
-  if [[ -o sharehistory && -z "$BUFFER" ]]; then
-    fc -RI "$HISTFILE"
-  fi
+  [[ -o sharehistory ]] && fc -RI
   selected=( $(fc -rl 1 | perl -ne 'print if !$seen{(/^\s*[0-9]+\**\s+(.*)/, $1)}++' |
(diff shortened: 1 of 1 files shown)


shell/key-bindings.zsh
```diff
@@ -106,6 +106,9 @@ bindkey '\ec' fzf-cd-widget
 fzf-history-widget() {
   local selected num
   setopt localoptions noglobsubst noposixbuiltins pipefail no_aliases 2> /dev/null
+  if [[ -o sharehistory && -z "$BUFFER" ]]; then
```
junegunn, line 109: Can you explain why we need `-z "$BUFFER"` condition?

shell/key-bindings.zsh
```diff
@@ -106,6 +106,9 @@ bindkey '\ec' fzf-cd-widget
...
+    fc -R "$HISTFILE"
```
junegunn, line 110: Isn't `fc -R` sufficient?

Excerpt from http://zsh.sourceforge.net/Doc/Release/Shell-Builtin-Commands.html

> ‘fc -R’ reads the history from the given file, ‘fc -W’ writes the history out to the given file, and ‘fc -A’ appends the history out to the given file. If no filename is specified, the $HISTFILE is assumed.

junegunn: Looks like `[[ -o sharehistory ]] && fc -R` is all we need, but please correct me if I'm missing something.

mpolden: It prevents reloading history if the edit buffer is non-empty. E.g you typed something before pressing C-r, it's not strictly necessary though.

mpolden:
Last commit only reloads new history events, this prevents the event number of existing history entries from changing (which happens when you reload the entire history).

shell/key-bindings.zsh
mpolden: And yes, looks like the "$HISTFILE" parameter can be omitted.

shell/key-bindings.zsh
mpolden: It prevents reloading history if the edit buffer is non-empty. E.g you typed something before pressing C-r, it's not strictly necessary though.

junegunn: Why do you want to prevent reloading in that case?

mpolden: I honestly don't remember, I've removed it. 🙂

junegunn (approved):

junegunn:
Merged, thanks!

KimNorgaard:
This change has introduced a noticeable latency when using CTRL-R. On my system (arch/5.9.12 on lenovo t470s) this amounts to around 0.6 seconds:
 ```
time (fc -RI)
( fc -RI; )  0.59s user 0.00s system 99% cpu 0.595 total
```
I would suggest making this behavior optional since having to wait half a second for each CTRL-R is kind of draining :)

**Comment by junegunn** [comment]
@KimNorgaard Thanks for letting me know. I'm going to have to revert this for now since 0.59s is unacceptable.

**Comment by junegunn** [comment]
And it looks like we don't need an option for this as one can a function that calls `fzf-history-widget` after `fc -RI` and re-bind it to `^R`.

**Comment by mpolden** [comment]
Reverting is fine, it's 0.071s for me though. @KimNorgaard Just curious, how big is your history? (`fc -l 0 | wc -l`)

**Comment by KimNorgaard** [comment]
It's 10000 lines which is my HISTSIZE.

**Comment by mpolden** [comment]
Is this a spinning disk by any chance? Strange that it would be so slow. On my 3 year old MBP:
```
~$ time (fc -RI)
( fc -RI; )  0.04s user 0.03s system 96% cpu 0.068 total
~$ time (fc -l 0 | wc -l)
   23659
( fc -l 0 | wc -l; )  0.08s user 0.01s system 105% cpu 0.086 total
~$ zsh --version
zsh 5.7.1 (x86_64-apple-darwin19.0)
```

**Comment by KimNorgaard** [comment]
Nah it's an NVME. `.zsh_history` is 580K in size. I just ran `strace --absolute-timestamps=precision:us -p <pid_of_zsh>` and although this is obviously much slower due to overhead from running strace, it clearly shows that for me it's the `seek` syscalls on `.zsh_history` that takes up all the time. I have no obvious explanation but 'll investigate further.

**Comment by KimNorgaard** [comment]
Right so I found out if I experiment with HISTSIZE things move a lot faster. I have both HISTSIZE and SAVEHIST set to 10000 but I'm not sure yet why that would impact searching history. Anyway if I set HISTSIZE to a value a bit higher than the actual number of saved commands, e.g. 10010 then things run smoothly.

**Comment by mpolden** [comment]
I'm guessing `fc -RI` has to shift your history when appending new events, because it ends up exceeding `$HISTSIZE`. Mine is 100 000, with only 23 000 entries so no shifting is required.

**Comment by KimNorgaard** [comment]
Yeah, I think you are right, however it's pretty normal to set your HISTSIZE and SAVEHIST to the same value in which case this is bound to happen sooner or later, right? As long as no shifting is going on everything will be fast, though.

**Comment by tvrg** [comment]
I noticed the same delay as @KimNorgaard and found this commit and the revert on master.
Not sure if it helps, but I noticed that the index number increases by 10000 (`$HISTSIZE`) on every invocation. It's immediately fixed when I run `unsetopt sharehistory`.

Peek 2020-12-17 10-28

**Comment by mpolden** [comment]
Index remains unchanged for me. What other options do you have (`setopt`)?

**Comment by tvrg** [comment]
Sorry for the long delay. Here are my setopts:
```
alwaystoend
autocd
autopushd
completeinword
extendedhistory
noflowcontrol
histexpiredupsfirst
histignoredups
histignorespace
histverify
incappendhistory
interactive
interactivecomments
login
longlistjobs
nonomatch
promptsubst
pushdignoredups
pushdminus
sharehistory
shinstdin
```

It's probably the default ohmyzsh configuration. IIRC the problem only happened with multiple tmux panes. But even then I couldn't reproduce it reliably.

PR #2252, opened 2020-11-14
[vim] Clean up temp files on interrupt

dmnks:
The clean-up is done in s:collect(), so let's make sure it's run before
we may terminate due to CTRL-C or ESC in exit_handler().

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -645,7 +645,8 @@ function! s:execute(dict, command, use_height, temps) abort
   endif
   let exit_status = v:shell_error
   redraw!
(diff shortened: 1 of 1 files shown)


dmnks:
Reproducer:

1. Start vim
2. Activate the search (e.g. :FZF)
3. Observe the `/tmp/v*` directory (there should be at least one file)
4. Press ESC or CTRL-C to close the window
5. Files in `/tmp/v*` are still there

After the patch, no files are present in the `/tmp/v*` directory.

junegunn:
Merged, thanks!

PR #2266, opened 2020-11-25
zsh: declare variable as local before assignment

ratijas:
Zsh with option `WARN_CREATE_GLOBAL` helps find bugs like this. Basically, builtin `read` assigns to whatever variable it is given, and by default when variable does not exists — it will be created in global scope, but in my case (with option turned on) it produces a warning:

```
__fsel:6: scalar parameter item created globally in function __fsel
```

### DIFF
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -45,6 +45,7 @@ __fsel() {
     -o -type d -print \
     -o -type l -print 2> /dev/null | cut -b3-"}"
   setopt localoptions pipefail no_aliases 2> /dev/null
+  local item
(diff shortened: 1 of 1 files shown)


junegunn:
Interesting.

So this is one of the cases where zsh behaves differently from bash.

```sh
# bash
$ : | foo=bar; echo foo=$foo
foo=
```

```sh
# zsh
$ : | foo=bar; echo foo=$foo
foo=bar
```

Thanks for the patch.

ratijas:
I'd guess, zsh executes as much as possible in the shell process itself. I've even done a research once, only to find out that zsh indeed does something similar to tail call elimination when spawning a pipleline. It is _very heavy_ on optimizations.

On the other hand, bash seems pretty stupid. It only does what was asked for.

PR #2269, opened 2020-11-28
Shellcheck fzf-tmux

VeryTastyTomato:
Use shellcheck on the `fzf-tmux` bash script.

I left one [warning _(Prefer mapfile or read -a to split command output (or quote to avoid splitting).)_](https://github.com/koalaman/shellcheck/wiki/SC2207) because I don’t know what would be the correct fix.
https://github.com/junegunn/fzf/blob/298342677107f19f2ef7035afcf5302fdf26aa51/bin/fzf-tmux#L157

### DIFF
--- a/bin/fzf-tmux
+++ b/bin/fzf-tmux
@@ -143,7 +143,7 @@ if [[ ! "$opt" =~ "-K -E" ]] && tmux list-panes -F '#F' | grep -q Z; then
   zoomed_without_popup=1
   original_window=$(tmux display-message -p "#{window_id}")
   tmp_window=$(tmux new-window -d -P -F "#{window_id}" "bash -c 'while :; do for c in \\| / - '\\;' do sleep 0.2; printf \"\\r\$c fzf-tmux is running\\r\"; done; done'")
-  tmux swap-pane -t $tmp_window \; select-window -t $tmp_window
+  tmux swap-pane -t "$tmp_window" \; select-window -t "$tmp_window"
 fi
 
 set -e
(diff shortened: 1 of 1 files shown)


VeryTastyTomato:
I just realized there are other bash files that could be checked against shellcheck. If that kind of changes is useful for you, I may open another PR to do them, or I can do them in this one and change the title.

junegunn:
Thanks for the request. I once tried to satisfy shellcheck myself, but I decided not to because

1. I'm pretty confident that the temporary file paths won't include white-spaces so quoting is unnecessary.

2. It makes the code less readable, for instance, see the code below. We have to sprinkle quotes all over the place to satisfy the checker and to be consistent.

```diff
diff --git a/bin/fzf-tmux b/bin/fzf-tmux
index 080f936..d957b47 100755
--- a/bin/fzf-tmux
+++ b/bin/fzf-tmux
@@ -213,7 +213,7 @@ if [[ -n "$term" ]] || [[ -t 0 ]]; then
   cat <<< "\"$fzf\" $opts > $fifo2; echo \$? > $fifo3 $close" >> $argsf
 else
   mkfifo $fifo1
-  cat <<< "\"$fzf\" $opts < $fifo1 > $fifo2; echo \$? > $fifo3 $close" >> $argsf
+  cat <<< "\"$fzf\" $opts < \"$fifo1\" > \"$fifo2\"; echo \$? > \"$fifo3\" \"$close\"" >> \"$argsf\"
   cat <&0 > $fifo1 &
 fi
 tmux set-window-option synchronize-panes off \;\
```

3. We can't satisfy the checker 100%. For example, we shouldn't quote `$opt` in line 208, as doing so will break `fzf-tmux -p`, but the checker doesn't know about it and complains.

So all in all, I don't think it's really worth it.

Issue #2101, opened 2020-06-30
Control + R history commands does not scroll with the mouse as all other fzf things do including Control + T


abhijitpjoshi:
- [ ] I have read through the manual page (`man fzf`)
- [X] I have the latest version of fzf
- [X] I have searched through the existing issues

## Info

- OS
    - [ ] Linux
    - [X] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [ ] bash
    - [X] zsh
    - [ ] fish
    
## Problem / Steps to reproduce

While I do Control + R, I get a history search widget as expected. but the history commands does not scroll with the mouse as all other fzf things do including Control + T. Am I missing anything?

below does not scroll through with the mouse. I have to use up or down keys to scroll through

![image](https://user-images.githubusercontent.com/41492347/86071957-846a5d00-ba4e-11ea-864e-f8161dc49654.png)

below control + T scrolls through with the mouse wheel

![image](https://user-images.githubusercontent.com/41492347/86072037-af54b100-ba4e-11ea-9e80-2e1add7d7b79.png)

junegunn:
Works for me. Can you reproduce the problem with a minimal config?

```sh
mkdir -p /tmp/fzf-test
cp ~/.fzf.zsh /tmp/fzf-test/.zshrc
ZDOTDIR=/tmp/fzf-test zsh
```

abhijitpjoshi:
It's still an issue for me. I don't remember if something has changed with my iterm2. but then I wonder how does it work correctly with Control + T

With minimal configuration you provided, it's still the same issue for me.
![image](https://user-images.githubusercontent.com/41492347/86388238-6aa16380-bc62-11ea-864b-43c1f59a3c82.png)

Here is a working snippet with Control + T with minimal configuration.

![image](https://user-images.githubusercontent.com/41492347/86388331-a0dee300-bc62-11ea-8c9f-e7c25ee1bb9d.png)

junegunn:
You seem to have configured the bindings with environment variables. Please post the output of

```sh
env | grep FZF
```

abhijitpjoshi:
Please see below -

![image](https://user-images.githubusercontent.com/41492347/86474071-1359cc80-bd10-11ea-9a01-50d110f290f4.png)

junegunn:
Are you using a non-official implementation of the key bindings? Because the screenshots show that you have non-default options enabled such as `--inline-info` and `--prompt`.

[...]

PR #2281, opened 2020-12-11
#2101 keep mouse mode in ctrl-r, alt-c in zsh

lincheney:
Fix for #2101 
zsh sends SIGCONT when running fzf in a pipe in certain cases, causing mouse mode to become disabled when using the ctrl-r and alt-c keybindings in zsh.
Work around this problem by using redirection instead.

The zsh completion scripts are already doing this (e.g. https://github.com/junegunn/fzf/blob/master/shell/completion.zsh#L219).


junegunn:
Merged, thanks!

PR #2287, opened 2020-12-15
[vim] Allow closing fzf window without confirmation

mezzarobba:
Just a minor improvement to the vim plugin. Thank you very much for this great tool!

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -838,6 +838,7 @@ function! s:execute_term(dict, command, temps) abort
         let term_opts.curwin = 1
       endif
       let fzf.buf = term_start([&shell, &shellcmdflag, command], term_opts)
+      call term_setkill(fzf.buf, 'term')
       if exists('&termwinkey')
         call setbufvar(fzf.buf, '&termwinkey', '<c-z>')
       endif


junegunn:
Hi, what problem does this solve? In my case, fzf doesn't ask for confirmation. Which OS are you running? Do you have a screenshot?

mezzarobba:
Hi, thank you for your reply!

I am running gvim 8.2.1913 under Debian. Here is a screenshot of the situation my patch aims to avoid. To reproduce:

1. launch ``gvim`` at the root of the fzf source tree,
2. run the ``:Files`` command from fzf-vim,
3. attempt to close the gvim window using the window manager.

fzf

junegunn:
I see, why don't we simply add it to `term_opts` i?

```diff
diff --git a/plugin/fzf.vim b/plugin/fzf.vim
index 4705115..9ce9ab4 100644
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -831,7 +831,7 @@ function! s:execute_term(dict, command, temps) abort
     if has('nvim')
       call termopen(command, fzf)
     else
-      let term_opts = {'exit_cb': function(fzf.on_exit)}
+      let term_opts = {'exit_cb': function(fzf.on_exit), 'term_kill': 'term'}
       if is_popup
         let term_opts.hidden = 1
       else
```

mezzarobba:
That's even better indeed, I had not noticed that it was possible.

junegunn:
Patch pushed to master. Thanks!

PR #1710, opened 2019-10-11
Support building on machines with `uname -m` == "aarch64"

stiletto:
I was attempting to build fzf inside termux on an android phone and it failed because the phone reports its architecture as "aarch64", not as "armv8l":

```
$ uname -a
Linux localhost 4.4.189-perf+ #1 SMP PREEMPT Wed Oct 2 11:49:20 UTC 2019 aarch64 Android
$ go env
GO111MODULE=""
GOARCH="arm64"
GOBIN=""
GOCACHE="/data/data/com.termux/files/home/.cache/go-build"
GOENV="/data/data/com.termux/files/home/.config/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="arm64"
GOHOSTOS="android"
GONOPROXY=""
GONOSUMDB=""
GOOS="android"
GOPATH="/data/data/com.termux/files/home/go"
GOPRIVATE=""
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/data/data/com.termux/files/usr/lib/go"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/data/data/com.termux/files/usr/lib/go/pkg/tool/android_arm64"
GCCGO="gccgo"
AR="ar"
CC="aarch64-linux-android-clang"
CXX="aarch64-linux-android-clang++"
CGO_ENABLED="1"
GOMOD="/data/data/com.termux/files/home/.fzf/go.mod"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/data/data/com.termux/files/usr/tmp/go-build401736825=/tmp/go-build -gno-record-gcc-switches"
$ make
Makefile:48: *** "Build on aarch64 is not supported, yet.".  Stop.
```

This pull request adds "aarch64" as an alias for "armv8l" so that attempting to build on an aarch64 machine will build for GOARCH=arm64.

PR #2291, opened 2020-12-21
Support building on machines with `uname -m` == "arm64"

lloeki:
Similarly to https://github.com/junegunn/fzf/pull/1710 except Darwin's `uname -m` on Apple Silicon M1 reports `arm64`, which is what Go picked up as well for their release:

```
$ uname -m
arm64
$ env PATH="/opt/go/bin:$PATH" go version
go version go1.16beta1 darwin/arm64
```

Once that hurdle is passed, `make` built a working `target/fzf-darwin_arm8` without issues.


lloeki:
One more note, the install script might need updating though:

```bash
archi=$(uname -sm)
binary_available=1
binary_error=""
case "$archi" in
  Darwin\ *64)     download fzf-$version-darwin_amd64.tar.gz  ;;
```

junegunn:
Merged, thanks!

> One more note, the install script might need updating though:

Yeah, thanks for the heads up. I'm currently building fzf binaries with Go 1.14 because of a performance regression (https://github.com/golang/go/issues/40727), but it looks like it doesn't support building darwin-arm64 binaries. I'll see what I can do.

lloeki:
Yep, building darwin-arm64 is a new Go 1.16 feature, for which the beta has just been released.

PR #2295, opened 2020-12-23
Add alt->, alt-< bindings and "bottom" action.

7ocb:
I'd like to have emacs-like "go last" and "go to first", which is bound on "alt->" and "alt-<", but as handling for these keys and for "bottom" action, there is PR with it.

### DIFF
--- a/src/options.go
+++ b/src/options.go
@@ -497,6 +497,10 @@ func parseKeyChords(str string, message string) map[int]string {
 			chord = tui.AltSpace
 		case "alt-/":
 			chord = tui.AltSlash
+		case "alt->":
+			chord = tui.AltMore
+		case "alt-<":
+			chord = tui.AltLess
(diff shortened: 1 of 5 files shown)


junegunn:
Thanks. Actually, I'm working on a more general solution that allows any alt key chord (`alt-ANY_SINGLE_CHAR`).

I have mixed feelings for `bottom` action you proposed as I rarely want to select the item with the lowest score. I'll think about it.

junegunn:
Implemented on master. Please test it and let me know if you run into any problems.

* New action `last`
* `top` is renamed to `first`
* `alt-ANY_CHAR` supported

```sh
fzf --bind alt-,:first,alt-.:last
fzf --bind 'alt-<:first,alt->:last'
# ...
```

7ocb:
@junegunn works for me, thanks.

PR #2306, opened 2021-01-03
Fix selection changed on terminal resize

7ocb:
Before:

If cursor placed at the end of list (i.e. maximum offset):

https://user-images.githubusercontent.com/8470556/103484891-3f5f2300-4e03-11eb-9f54-3a5eb5e7a0e5.mp4

After:

https://user-images.githubusercontent.com/8470556/103484901-57cf3d80-4e03-11eb-813b-af8272ad44ee.mp4

### DIFF
--- a/src/terminal.go
+++ b/src/terminal.go
@@ -2636,11 +2636,11 @@ func (t *Terminal) constrain() {
 	count := t.merger.Length()
 	// count of lines can be displayed
 	height := t.maxItems()
-	
-	t.cy = util.Constrain(t.cy, 0, count - 1)
+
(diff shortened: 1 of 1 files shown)


junegunn:
I've tested it in a dedicated terminal window and in a tmux pane. But I can't reproduce the problem. The cursor doesn't drift away from the initial selection.

junegunn:
The code was written 5 years ago, so it took me a while to recall why I wrote it that way. Looks like my priority was to keep the distance from the prompt to the cursor unchanged. But yeah, it's better not to change the selection. Thanks.

Issue #2256, opened 2020-11-18
Failed to run "fzf --version": ['<be><f8><b7><c3><ce>?<a3>^M']


jiangjianshan:
Hello, 

I'm using fzf which version is 0.24.3 on Vim 8.2.2000 on Windows 10. Recently I have got the following error when using 'noremap <silent> <leader>f :Rg <C-R><C-W><CR>' settings on _vimrc:

![image](https://user-images.githubusercontent.com/20179047/99508327-6c748500-29bf-11eb-89d3-dfed26c53a18.png)

To run 'fzf --version' on Command Prompt can got the correct version result, so I think fzf.exe hasn't been broken.
![image](https://user-images.githubusercontent.com/20179047/99509420-b1e58200-29c0-11eb-8b60-a470efc23403.png)

The settings for fzf.vim on _vimrc are here:
" ------------------------------------ start -------------------------
let g:fzf_tags_command = 'ctags -R'
let g:fzf_preview_window = 'right:60%'

" This is the default extra key bindings
let g:fzf_action = {
  \ 'ctrl-t': 'tab split',
  \ 'ctrl-x': 'split',
  \ 'ctrl-v': 'vsplit'
\}

let g:fzf_history_dir = '$HOME/.local/share/fzf-history'

let $FZF_DEFAULT_COMMAND = 'rg --files --ignore-case --hidden'
let $FZF_DEFAULT_OPTS = '--bind ?:toggle-preview --bind ctrl-j:down --bind ctrl-k:up --bind ctrl-d:half-page-down --bind ctrl-u:half-page-up'

let g:fzf_layout = { 'window': { 'width': 0.9, 'height': 0.8 } }

" Customize fzf colors to match your color scheme
let g:fzf_colors = {
  \ 'fg':      ['fg', 'Normal'],
  \ 'bg':      ['bg', 'Normal'],
  \ 'hl':      ['fg', 'Comment'],
  \ 'fg+':     ['fg', 'CursorLine', 'CursorColumn', 'Normal'],
  \ 'bg+':     ['bg', 'CursorLine', 'CursorColumn'],
  \ 'hl+':     ['fg', 'Statement'],
  \ 'info':    ['fg', 'PreProc'],
  \ 'border':  ['fg', 'Ignore'],
  \ 'prompt':  ['fg', 'Conditional'],
  \ 'pointer': ['fg', 'Exception'],
  \ 'marker':  ['fg', 'Keyword'],
[...]

PR #2308, opened 2021-01-04
Fixed the issue of "Failed to run fzf --version: [<be><f8><b7><c3><ce>?<a3>^M]"

jiangjianshan:
The first version of install.ps1 was created by me on last pull request on last year. But the original implement of it was missing the feature to set the access right of fzf.exe. It may cause the following issue:
1. Failed to run "fzf --version": ['<be><f8><b7><c3><ce>?<a3>^M']",  Please see issue #2256.
2. Failed to download fzf.exe because it will delete fzf.exe if the return result of "fzf --version" is wrong, see point 1.
Also,  the download progress hasn't been showed at previous implmentation. Ths pull request also fixed this.

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -43,7 +43,11 @@ function download {
   $url="https://github.com/junegunn/fzf/releases/download/$version/$file"
   $temp=$env:TMP + "\fzf.zip"
   [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
-  (New-Object Net.WebClient).DownloadFile($url, $ExecutionContext.SessionState.Path.GetUnresolvedProviderPathFromPSPath("$temp"))
+  if ($PSVersionTable.PSVersion.Major -ge 3) {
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks, I'll test the script when I get a chance.

install.ps1
```diff
@@ -53,7 +57,7 @@ function download {
     $binary_error="Failed to download $file"
     return
   }
-  check_binary >$null
+  echo y | cacls $fzf_base\bin\fzf.exe /p administrator:f ; check_binary >$null
```
junegunn, line 60: * I'm seeing `NOTE: Cacls is now deprecated, please use Icacls.` message when I run cacls. Should we use icacls instead?
* Does this work even when the user account is not "administrator"?

jiangjianshan:
Sorry for late reply. I have just came back from a long business trip. The excutable program cacls is worked for my laptop. I have just change to use icacls as your suggestion and test it on both admistrator and user account.

junegunn:
Merged, thanks!

PR #2309, opened 2021-01-04
Update README to upgrade using brew upgrade

nicolasbarra:
Unless there is some specific reason that FZF needs to be reinstalled, running brew update and then brew upgrade will upgrade all formula, including FZF. Reinstalling does not seem necessary for this use case and makes it seem as if users need to go to the additional step of reinstalling FZF when the usual brew upgrade and then brew install workflow will work correctly to upgrade FZF. However, if I have missed some specific reason why FZF needs to be reinstalled, I'll close this PR accordingly. Open to comments!

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,7 +171,7 @@ while. Please follow the instruction below depending on the installation
 method used.
 
 - git: `cd ~/.fzf && git pull && ./install`
-- brew: `brew update; brew upgrade`
+- brew: `brew update; brew upgrade fzf`
 - macports: `sudo port upgrade fzf`
 - chocolatey: `choco upgrade fzf`
 - vim-plug: `:PlugUpdate fzf`
--- a/README.md
+++ b/README.md
@@ -171,7 +171,7 @@ while. Please follow the instruction below depending on the installation
 method used.
 
 - git: `cd ~/.fzf && git pull && ./install`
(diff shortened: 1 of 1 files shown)


junegunn:
I think we should change it to `brew update; brew upgrade fzf`. No need to tell the users to upgrade all formulas.

> Unless there is some specific reason that FZF needs to be reinstalled

fzf can be installed with `--HEAD` flag, and in that case, `brew upgrade fzf` didn't work as expected. So we have instructed the users to "reinstall" it, to remove the old HEAD version and install the latest stable version. However, it doesn't seem to work that way with the latest Homebrew.

nicolasbarra:
If that's the case, perhaps we could say something like "run 'brew reinstall fzf' if this is the first time you are upgrading. After that, you may upgrade as usual with 'brew upgrade fzf'". Should I update the PR to reflect this?

junegunn:
With the current Homebrew, `brew reinstall fzf` will still install with `--HEAD`. You need to remove it first if you want to switch to the stable release. So the point of mentioning `reinstall` has become moot. `brew upgrade fzf` is good enough.

nicolasbarra:
Got it! Just updated the PR to say brew upgrade fzf. Let me know if there is anything else I should do!

junegunn:
Merged, thanks!

PR #2340, opened 2021-02-02
Don't run precmd hooks in cd widget

marlonrichert:
`precmd` hooks expect the Zsh Line Editor to not be active.
Running these when the ZLE is active can lead to unpredictable results.
See https://github.com/marlonrichert/zsh-autocomplete/issues/180

### DIFF
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -68,16 +68,6 @@ fzf-file-widget() {
 zle     -N   fzf-file-widget
 bindkey '^T' fzf-file-widget
 
-# Ensure precmds are run after cd
-fzf-redraw-prompt() {
-  local precmd
-  for precmd in $precmd_functions; do
-    $precmd
-  done
-  zle reset-prompt
-}
-zle -N fzf-redraw-prompt
-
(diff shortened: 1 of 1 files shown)


junegunn:
This reverts #1136 which fixed #915.

@aaronjensen Can you review this pull request?

marlonrichert:
@junegunn While perhaps it looks like I'm reverting #1136, this is not entirely true, because this PR does not regress #915. Rather, it changes the code to do things properly. With this PR, `precmd` hooks will still fire when using <kbd>Alt</kbd><kbd>c</kbd>, but now they will fire at the right time, just before the ZLE gets activated, rather than when it's already active. 

You can test this in Zsh as follows:
```
$ precmd_test() { print fired }
$ autoload -Uz add-zsh-hook 
$ add-zsh-hook precmd precmd_test
```
Then press <kbd>Alt</kbd><kbd>c</kbd> to change dirs. You will see that `fired` will get printed.

aaronjensen:
AFAICT it still works for my use case. 👍

junegunn:
Thanks for the clarification!

junegunn:
Merged, thanks!

PR #2348, opened 2021-02-11
[completion] Default behaviour to use fd if present else use find.

chhajedji:
Since using `fd` along with `fzf` is very common choice,  checking in `completion.{ba, z}sh` if `fd` is present. If yes then use it by default or else use the default `find` command.

### DIFF
--- a/README.md
+++ b/README.md
@@ -413,19 +413,6 @@ export FZF_COMPLETION_TRIGGER='~~'
 # Options to fzf command
 export FZF_COMPLETION_OPTS='+c -x'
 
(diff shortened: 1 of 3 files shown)


chhajedji:
Hi @junegunn,

What do you think about this?

junegunn:
Thanks for the patch. But I wouldn't choose a non-standard tool as the default. `fd` is a relatively new tool with a terse, non-descriptive name so we can't be sure if the "fd" program on the user's system is the "fd" we're talking about. We already show how to configure the behavior on the README page and I think it's enough.

PR #2349, opened 2021-02-11
Vim plugin fix for Cygwin 3.1.7 and above

### LINKED ISSUE #2153: Vim plugin doesn't work at all on Cygwin; removing special-casing in fzf.vim may fix it

## Info

- OS
    - [ ] Linux
    - [ ] Mac OS X
    - [X] Windows
    - [ ] Etc.
- Shell
    - [ ] bash
    - [ ] zsh
    - [ ] fish
    
I am running Vim with the fzf and fzf.vim plugins under Cygwin. Whenever I try to use one of the fzf.vim commands, such as `:Files` or `:Rg`, the following happens:

1. The terminal goes blank
2. A separate console window opens and closes again immediately. (No output seems to appear in it.)
3. The terminal un-blanks again, and I'm back where I started (e.g. editing a file).

All of this happens in about a second.

I've found that by commenting out the following `elseif` block [at line 590 of `plugin/fzf.vim`](https://github.com/junegunn/fzf/blob/c0a83b27eb0e2a9a1f631d04830a7dc6d762a951/plugin/fzf.vim#L590) in its entirety, I can restore what seems to be the correct behaviour, namely that the fzf file selection pane is shown:
```vim
  elseif has('win32unix') && $TERM !=# 'cygwin'
    let shellscript = s:fzf_tempname()
    call writefile([command], shellscript)
    let command = 'cmd.exe /C '.fzf#shellescape('set "TERM=" & start /WAIT sh -c '.shellscript)
    let a:temps.shellscript = shellscript
```
It seems to be treating Cygwin differently from other Windows-UNIX environments, and I'm wondering whether that's still the right thing to do. Cygwin now uses mintty as its default terminal, and that has `TERM=xterm` by default. In fact, `TERM=cygwin` doesn't even seem to be an option in mintty.

mburz:
The short story: this special handling doesn't work with any recent version of Cygwin. Without this special treatment, it just works.
Forcing the users to set `$TERM` to `"cygwin"` just to get fzf running under Cygwin, as suggested in one comment is not cool, and breaks on its own many other things.

Since there is no comment on why this special treatment was applied, it's hard to figure out what the initial intention was, and whether this was required in order to make things run at all, or to improve upon something lacking.

This PR fixes #2153

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -632,10 +632,21 @@ function! s:execute(dict, command, use_height, temps) abort
       return []
     endif
   elseif has('win32unix') && $TERM !=# 'cygwin'
-    let shellscript = s:fzf_tempname()
-    call writefile([command], shellscript)
-    let command = 'cmd.exe /C '.fzf#shellescape('set "TERM=" & start /WAIT sh -c '.shellscript)
-    let a:temps.shellscript = shellscript
+    if !exists('s:cygwin_317_or_above')
+      let uname_os = system('uname -o')
+      if v:shell_error || uname_os !~? 'Cygwin'
+        let s:cygwin_317_or_above = 0
+      else
+        let cygwin_version = substitute(system("uname -r |grep --only-matching '^[0-9.]\\+' |tr -d '.'"), '\_s*$', '', '')
+        let s:cygwin_317_or_above = !(v:shell_error || cygwin_version !~# '^\d\+$' || cygwin_version < 317)
+      endif
+    endif
+    if !s:cygwin_317_or_above
+      let shellscript = s:fzf_tempname()
+      call writefile([command], shellscript)
+      let command = 'cmd.exe /C '.fzf#shellescape('set "TERM=" & start /WAIT sh -c '.shellscript)
+      let a:temps.shellscript = shellscript
+    endif
   endif
   if a:use_height
     let stdin = has_key(a:dict, 'source') ? '' : '< /dev/tty'


junegunn:
Have you reviewed the original pull request? https://github.com/junegunn/fzf/pull/933
I believe @janlazo can help you here.

mburz:
Thanks @junegunn , will go through the whole conversation there and test things out. Would help, if I could find out what breaks, when this `elseif` branch gets tossed.
I tested already using multiple terminals, including via ssh and everything works well without the `elseif`.
Nothing works with the `elseif`. I can't even execute the temporarily created `shellscript` file because it is lacking the execute flag. I then introduced a `chmod` after `writefile`, but things don't work out, although the script gets executed.

@janlazo I more than welcome any suggestions, thank you!

mburz:
Some more information on what works and what doesn't with my patch.
I only use `mintty` when on the Windows maching, I'm normally in through `ssh` (using the `sshd` bundled with Cygwin).
Tried out 2 versions of `ConEmu`, one old 180626 and the newest 210206, just to figure out if my patch works.

Short story, without my patch `:FZF` doesn't work with any of them as expected.

Used cygwin version: `uname -a`
CYGWIN_NT-10.0 host123 3.1.7(0.340/5/3) 2020-08-22 17:48 x86_64 Cygwin

To make `shellscript` executable I added  with `call system('chmod +x ' . shellscript)` after the `call writefile` line.

Results:
Terminal | `shellscript` with `+x` | My patch | Result
--- | --- | --- | ---
`urxvt`, `ssh` into `Cygwin`, vim runs inside `tmux` | Yes | No | Doesn't work (the script is executed, the screen stays black, requires `C-L` once back to vim
`urxvt`, `ssh` into `Cygwin` | Yes | No | Doesn't work (the script is executed, the screen stays black, requires `C-L` once back to vim
`ConEmu 180626` | Yes | No | Works, but fzf is opened up in a new detached terminal window
`ConEmu 210206` | Yes | No | Works, but fzf is opened up in a new detached terminal window
`mintty` | Yes | No | Works, but`fzf is opened up in a new terminal window
`urxvt`, `ssh` into `Cygwin`, vim runs inside `tmux` | No | No | Doesn't work, the script fails to execute
`urxvt`, `ssh` into `Cygwin` | No | No | Doesn't work, the script fails to execute
`ConEmu 180626` | No | No | Doesn't work, the script fails to execute
`ConEmu 210206` | No | No | Doesn't work, the script fails to execute
`mintty` | No | No | Doesn't work, the script fails to execute
`urxvt`, `ssh` into `Cygwin`, vim runs inside `tmux` | No | Yes | Works as expected, fzf is opened up "inside" vim
`urxvt`, `ssh` into `Cygwin` | No | Yes | Works as expected, fzf is opened up "inside" vim
`ConEmu 180626` | No | Yes | Painfully slow, but works as expected, fzf is opened up "inside" vim
`ConEmu 210206` | No | Yes | Works as expected, fzf is opened up "inside" vim
`mintty` | No | Yes | Works as expected, fzf is opened up "inside" vim

So I'm yet to find a case where my patch doesn't work.

janlazo:
`Works, but`fzf is opened up in a new terminal window` is the intended behavior of the special case because fzf didn't work if `TERM` is set and not equal to `cygwin`, default `TERM` for git bash without spawning mintty.

If the OS/MSYS2/Cygwin/Mingw environment changed so that fzf works without issues regardless of `TERM` value, then the special case should account for that.

ssh or any form of remote access weren't taken into account for the special case.

plugin/fzf.vim
```diff
@@ -631,11 +631,6 @@ function! s:execute(dict, command, use_height, temps) abort
       call jobstart(cmd, fzf)
       return []
     endif
-  elseif has('win32unix') && $TERM !=# 'cygwin'
-    let shellscript = s:fzf_tempname()
-    call writefile([command], shellscript)
-    let command = 'cmd.exe /C '.fzf#shellescape('set "TERM=" & start /WAIT sh -c '.shellscript)
```
janlazo, line 637: `sh filename` should work here so `-c` is likely causing issues if `sh` cares about mingw/cygwin/msys2 file permissions. Read persmission is sufficient on Windows, unlike Unix.

mburz:
Was now able to reproduce the problem that requires that special handling using an older version of cygwin (3.0.7).
Updated my PR to only disable that handling starting the most recent Cygwin version (3.1.7), since that's the one I tested mostly against.

plugin/fzf.vim
```diff
@@ -632,10 +632,13 @@ function! s:execute(dict, command, use_height, temps) abort
       return []
     endif
   elseif has('win32unix') && $TERM !=# 'cygwin'
-    let shellscript = s:fzf_tempname()
-    call writefile([command], shellscript)
-    let command = 'cmd.exe /C '.fzf#shellescape('set "TERM=" & start /WAIT sh -c '.shellscript)
-    let a:temps.shellscript = shellscript
+    let cygwin_version = substitute(system("uname -r |grep --only-matching '^[0-9.]\\+' |tr -d '.'"), '\_s*$', '', '')
```
janlazo, line 635: Is this safe to cache?

mburz: Introduced local on-demand caching.
Good point, I assumed `win32unix` means it's Cygwin, need to look into this.

mburz: Added an explicit check for Cygwin

plugin/fzf.vim
```diff
@@ -632,10 +632,21 @@ function! s:execute(dict, command, use_height, temps) abort
       return []
     endif
   elseif has('win32unix') && $TERM !=# 'cygwin'
-    let shellscript = s:fzf_tempname()
-    call writefile([command], shellscript)
-    let command = 'cmd.exe /C '.fzf#shellescape('set "TERM=" & start /WAIT sh -c '.shellscript)
-    let a:temps.shellscript = shellscript
+    if !exists('s:cygwin_317_or_above')
+      let uname_os = system('uname -o')
```
junegunn, line 636: I would use `uname -s` instead since `-o` is not always available (see https://stackoverflow.com/a/27776822/755334)

plugin/fzf.vim
```diff
@@ -632,10 +632,21 @@ function! s:execute(dict, command, use_height, temps) abort
...
+      if v:shell_error || uname_os !~? 'Cygwin'
+        let s:cygwin_317_or_above = 0
+      else
+        let cygwin_version = substitute(system("uname -r |grep --only-matching '^[0-9.]\\+' |tr -d '.'"), '\_s*$', '', '')
```
junegunn, line 640: We can run a single command and parse the result in VimScript.

```vim
let uname = system('uname -sr')
let s:cygwin_317_or_above = uname =~? 'Cygwin' && s:version_requirement(matchstr(uname, '[0-9.]\+'), '3.1.7')
```

junegunn (changes requested):

mburz:
Will close this PR now. My fixes might help in some very particular cases, but overall the Cygwin support is just not good enough to pursue it.

ElArtista:
Noo I just wanted this to happen...

szevzol:
So I have to use cygwin as my TERM.... I'm sad.

Issue #2352, opened 2021-02-14
[regression] sink function calling fzf#run() puts vim in normal mode instead of terminal


antoinemadec:
- [x] I have read through the manual page (`man fzf`)
- [x] I have the latest version of fzf
- [x] I have searched through the existing issues

## Info

- OS
    - [x] Linux

- Shell
    - [x] bash
 
## Env
```sh
NVIM v0.5.0-dev+1085-g52397aaa0
> nvim --version
NVIM v0.5.0-dev+1085-g52397aaa0
> vim --version
VIM - Vi IMproved 8.2 (2019 Dec 12, compiled Jan 20 2021 22:47:34)
Included patches: 1-2380
```

## Problem / Steps to reproduce
When the sink function calls ```fzf#run()```, **vim and neovim** switch to normal mode instead of terminal mode.

This regression comes from this commit: https://github.com/junegunn/fzf/commit/1a76bdf8914195ab26ebac7e49a3ed8752dd69bd

minimal_vimrc:
```vim
call plug#begin('~/.vim/plugins')
Plug 'junegunn/fzf', {'dir': '~/.fzf','do': './install --all'}
call plug#end()
```

testcase.vim:
```vim
function! s:list_handler(...) abort
  call s:run_testcase()
endfunction

function! s:run_testcase() abort
  let opts = {'sink*': function('s:list_handler')}
  call fzf#run(fzf#wrap(opts))
endfunction

call s:run_testcase()
```

Steps to reproduce the issue:

1. nvim -u minimal_vimrc
2. source testcase.vim
3. press Enter

antoinemadec:
@ojroques what do you  think about this ?
Should your commit be kept, adapted or reverted ?

junegunn:
Can you test if this patch fixes the problem?

```diff
diff --git a/plugin/fzf.vim b/plugin/fzf.vim
index 763747c..1069af8 100644
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -764,6 +764,10 @@ function! s:split(dict)
   endtry
 endfunction
 
+noremap  <silent> <Plug>(fzf-normal) <Nop>
+noremap! <silent> <Plug>(fzf-normal) <Nop>
+tnoremap <silent> <expr> <Plug>(fzf-normal) &filetype == 'fzf' ? "\<C-L>" : "\<C-\>\<C-n>"
+
 function! s:execute_term(dict, command, temps) abort
   let winrest = winrestcmd()
[...]

PR #2364, opened 2021-02-20
Fix feedkeys() side effects

antoinemadec:
See #2352
See https://github.com/junegunn/fzf.vim/issues/1216

Fixes side effects cleanely.

All of those work after this fix:

- ```:Buffers```
- ```fzf#run()``` calling ```fzf#run()```
- RPC calls in the sink function

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -803,6 +803,8 @@ function! s:execute_term(dict, command, temps) abort
         " We use close instead of bd! since Vim does not close the split when
         " there's no other listed buffer (nvim +'set nobuflisted')
         close
+        " avoids side effects caused by previous feedkeys()
+        call feedkeys('', 'x')
       endif
       silent! execute 'tabnext' self.ppos.tab
       silent! execute self.ppos.win.'wincmd w'


junegunn:
Thanks, but I've noticed this makes Neovim unresponsive in a certain scenario.

```vim
call plug#begin('~/.vim/plugged')
Plug 'junegunn/fzf'
Plug 'junegunn/fzf.vim'
call plug#end()

function! s:list_handler(...) abort
  call s:run_testcase()
endfunction

function! s:run_testcase() abort
  let opts = {'sink*': function('s:list_handler')}
  call fzf#run(fzf#wrap(opts))
endfunction

command! Run call s:run_testcase()
```

- `:Run`
- Select any entry
- Second fzf opens; cursor is not visible
- Press CTRL-C and Neovim becomes unresponsive.

The problem is not reproducible on regular Vim.

junegunn:
Okay, I've just updated Neovim, and I'm seeing a different problem.

image

I have to press `i` to use the second fzf.

antoinemadec:
Ok, i will try it again on the minimal vimrc.
Eventually, if we don't find any clean way to fix those feddkeys side effects, i think we should remove the feedkeys all together.
The bugs it introduces outweigh what it is fixing in the first place.

antoinemadec:
@junegunn here is what I observe:

- neovim 0.4.4 (latest release) : bad behavior with and without my changes
- neovim 0.5.0 (up-to-date build) : behaves properly with my changes

```sh
> nvim --version
NVIM v0.5.0-dev+1102-g595f6e4d6
```

The testcase passed without any issues.

However, the ```:Buffers``` behavior when jumping in a terminal is inconsistent, it sometimes jump in terminal mode, sometimes in normal mode.

**I think we should remove all the feedkeys() altogether.**
What do you think?

junegunn:
Agreed. I'm going to remove feedkeys and reopen https://github.com/junegunn/fzf.vim/issues/1216.

PR #2369, opened 2021-02-24
[vim] Make `winblend` for popup window configurable

relastle:
## Make Neovim floating window's transparency configurable

```vim
let g:fzf_layout = {
  \ 'window': {
    \ 'width': 0.9,
    \ 'height': 0.95,
    \ 'border': 'rounded',
    \ 'winblend': 30,
  \ },
\ }
```

image

I personally think `winblend` is one of factors that make Neovim UI components rich,
so this PR just makes that property configurable via `g:fzf_layout`.

## Other minor fix

I noticed that some `border` related sources codes have been left in Neovim's `create_popup` function.
I think that from [this commit](https://github.com/junegunn/fzf/commit/2e8e63fb0b3b62e89472eebe9e86578598e1a541), `border` had been fully controlled by `fzf` binary itself, and script-local `s:create_popup` function is never been called against `opts` with `border` property.

### DIFF
--- a/README-VIM.md
+++ b/README-VIM.md
@@ -412,6 +412,7 @@ The latest versions of Vim and Neovim include builtin terminal emulator
 " - yoffset [float default 0.5 range [0 ~ 1]]
 " - border [string default 'rounded']: Border style
 "   - 'rounded' / 'sharp' / 'horizontal' / 'vertical' / 'top' / 'bottom' / 'left' / 'right'
+" - winblend [int default 0 range[0 ~ 100]]: Transparency of floating window  (Neovim only)
 let g:fzf_layout = { 'window': { 'width': 0.9, 'height': 0.6 } }
 ```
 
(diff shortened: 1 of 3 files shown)


junegunn:
Thanks for the suggestion. But I would suggest using `FileType fzf` autocmd to customize the settings of an fzf window.

```vim
if has('nvim')
  autocmd! FileType fzf setlocal winblend=30
endif
```

This approach is more flexible as you can set any other options as well, even the ones that are not yet implemented on Vim or Neovim. Also, you can do some extra stuff.

```vim
tnoremap <Plug>(nohl) <c-\><c-n>:nohl<cr>i
autocmd! FileType fzf set winblend=30 | call feedkeys("\<Plug>(nohl)")
```

relastle:
@junegunn 

Thank you for response, and your advice is highly reasonable (I was not able to come up with the idea.. 😢 )
I close this PR.

junegunn:
Thanks, we should still remove the `border` code you pointed out.

relastle:
https://github.com/junegunn/fzf/pull/2370

I opened another PR 🙏

PR #2370, opened 2021-02-25
[vim] Remove unnecessary `border` management in nvim floating window

relastle:
I noticed that some `border` related sources codes have been left in Neovim's `create_popup` function.
I think that from [this commit](https://github.com/junegunn/fzf/commit/2e8e63fb0b3b62e89472eebe9e86578598e1a541), `border` had been fully controlled by `fzf` binary itself, and script-local `s:create_popup` function is never been called against `opts` with `border` property.


junegunn (approved):

junegunn:
Merged, thanks!

PR #2379, opened 2021-03-04
[fzf-tmux] Adapt to tmux latest changes

solarizedalias:
[In a latest change](https://github.com/tmux/tmux/commit/c44750792a9683c5cd6f9df5a69e7417b88772d2), tmux deleted `-K`, `-R` flags to `display-popup`, `popup` command.
This PR updates `fzf-tmux` script to add support for new version .

### DIFF
--- a/bin/fzf-tmux
+++ b/bin/fzf-tmux
@@ -206,7 +206,7 @@ if [[ "$opt" =~ "-K -E" ]]; then
   fi
 
   # tmux dropped the support for `-K`, `-R` to popup command
-  if [[ ! "$(strings "$(which tmux)" | grep -A2 'display-popup')" =~ 'CEKc:d:h:R:t:w:x:y:' ]]; then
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks for the pull request.

---

`which tmux` doesn't work when `tmux` is an alias. Changing it to `type -P tmux` fixes the problem.

```sh
$ which tmux
$ type tmux
tmux is aliased to `tmux -2'
$ type -P tmux
/usr/local/bin/tmux
```

---

How about if we see the output of `--help`? It is slightly faster than using `strings`.

```sh
$ time strings $(type -P tmux) > /dev/null

real    0m0.013s
user    0m0.006s
sys     0m0.006s

$ time tmux display-popup --help 2> /dev/null

real    0m0.008s
user    0m0.003s
sys     0m0.004s
```

solarizedalias:
Thank you for the suggestions!
Indeed, it seems much better to check the usage text.
Technically, there's no `--help` but that wouldn't much likely to be relevant here.

junegunn:
Merged, thanks!

PR #2383, opened 2021-03-07
fish key bindings: make fzf-file-widget work with -option= prefix

exploide:
Currently, when having a commandline like `grep --exclude-from=~/Desktop`  using the `fzf-file-widget` does not work since it takes `--exclude-from=` as part of the path.

This PR makes the widget under fish shell work even when the path is prefixed with an option.


junegunn:
Branch rebased to the current master.

I'll review the code later in the day. Thanks!

junegunn:
Merged, thanks!

PR #2395, opened 2021-03-18
replace golang.org/x/crypto/ssh/terminal with golang.org/x/term

johejo:
See https://github.com/golang/go/issues/31044


junegunn:
Merged, thanks!

Issue #2328, opened 2021-01-19
Will future releases support 32bit again?


alechstong:
Or am I stuck at 0.22.0?

minimum requirement of current fzf.vim is 0.23.0, and I cannot find earlier releases. I'm using a patched fzf.vim for the moment.

I hope there would be a future 32bit version, or fzf.vim set the minimum requirement of fzf at 0.22.0.

junegunn:
Go 1.15 dropped support for building 32-bit binaries, so, unfortunately, no.

Have you tried building fzf from source? You can just `git clone` this repository and `make install`, if you have Go installed on your system.

See https://github.com/junegunn/fzf/blob/master/BUILD.md

alechstong:
Yeah, I should have tried this first. After adding a i686 target to Makefile, It works with go1.14.13.

Thank you for helping.

junegunn:
Oh, right. I forgot that I removed 32-bit targets in https://github.com/junegunn/fzf/commit/dba14d26303297322dc4426d38ebfbf1d76a6011. I should probably restore them.

junegunn:
Updated Makefile so you can build binaries without manually editing the file.

PR #2396, opened 2021-03-18
update golang to 1.16

johejo:


.github/workflows/macos.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     runs-on: macos-latest
     strategy:
       matrix:
-        go: [1.14, 1.15]
+        go: [1.15, 1.16]
```
junegunn, line 15: I'd like to keep 1.14 here because

> As announced in the Go 1.14 release notes, Go 1.15 drops support for 32-bit binaries on macOS, iOS, iPadOS, watchOS, and tvOS (the darwin/386 and darwin/arm ports).

And there are still users who actually need 32-bit binaries on macOS (#2328)

So please change this to `go: [1.14, 1.16]`

junegunn (changes requested):

junegunn:
Thanks. The test failure is not related to the change you made. I'll see why it keeps failing.

PR #2404, opened 2021-03-25
Update README (--phony -> --disabled)

pschmitt:
Since --phony was renamed to --disabled in 
https://github.com/junegunn/fzf/commit/d779ff7e6dcf068fbcf743bed45127c3b857ec92 
it makes sense to update the README accordingly ;)


junegunn:
Merged, thanks!

PR #2405, opened 2021-03-25
Use FZF_DEFAULT_COMMAND in more places

peterfpeterson:
The CTRL-T and ALT-C bindings were ignoring the FZF_DEFAULT_COMMAND when it was set. This left users wondering why "ignored" files/directories were still showing up in searches with these bindings.

### DIFF
--- a/shell/key-bindings.bash
+++ b/shell/key-bindings.bash
@@ -14,10 +14,15 @@
 # Key bindings
 # ------------
 __fzf_select__() {
-  local cmd="${FZF_CTRL_T_COMMAND:-"command find -L . -mindepth 1 \\( -path '*/\\.*' -o -fstype 'sysfs' -o -fstype 'devfs' -o -fstype 'devtmpfs' -o -fstype 'proc' \\) -prune \
-    -o -type f -print \
-    -o -type d -print \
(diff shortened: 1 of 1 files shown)


peterfpeterson:
@Keating950 is there something I need to do to get this either merged or rejected?

Keating950:
> @Keating950 is there something I need to do to get this either merged or rejected?

Don't know; I'm not a maintainer, it just happens that my PR is the current head commit. I'll tag @junegunn.

junegunn:
I don't think we should do this.

* Unfortunately, there is no guarantee that a user configures `FZF_DEFAULT_COMMAND` to list file paths under the current directory. It can be anything.
    * `export FZF_DEFAULT_COMMAND='cat ~/docs/items.txt'`
* Even if it is configured to produce a list of file paths, it will not be compatible with ALT-C binding which expects a list of directories excluding files.
* The options are clearly documented on the README page
    * https://github.com/junegunn/fzf#environment-variables
        * > ⚠️ This variable is not used by shell extensions due to the slight difference in requirements.
    * https://github.com/junegunn/fzf#key-bindings-for-command-line

peterfpeterson:
Thanks for pointing out where it was hiding in the documentation

PR #2410, opened 2021-03-27
Use the first FZF found in the path

afnanenayet:
I think it makes more sense to respect the user's path rather than look at a
hardcoded path first. This causes issues, for example, on ubuntu if you have
fzf installed with `apt` and a newer version of fzf installed in your path (I
used the install script provided by fzf).

This changes the order so the first fzf found in the path is used for the
command.

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -169,10 +169,10 @@ endfunction
 let s:checked = {}
 function! fzf#exec(...)
   if !exists('s:exec')
-    if executable(s:fzf_go)
-      let s:exec = s:fzf_go
-    elseif executable('fzf')
+    if executable('fzf')
       let s:exec = 'fzf'
+    elseif executable(s:fzf_go)
+      let s:exec = s:fzf_go
     elseif input('fzf executable not found. Download binary? (y/n) ') =~? '^y'
       redraw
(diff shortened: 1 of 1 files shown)


junegunn:
I was thinking of the opposite case; where the official Ubuntu package provides an outdated version of fzf but you locally install the latest version under `bin` directory.

The installation instruction for Vim currently [suggests calling `fzf#install()` function](https://github.com/junegunn/fzf#as-vim-plugin) which will download the latest binary in `bin` directory. So I thought this case would be more common. But yeah, there's no guarantee that is the case. Hmm.

afnanenayet:
I think it's more conventional to take whatever is in the path first, since people do all sorts of different things with their setups. For example, I install lots of binaries locally to `~/.local/bin` and I would prefer to use those over other system-provided binaries because they're newer.

My suggestion would be to make it the user's responsibility to add `~/.fzf/bin` to their path, or maybe even install the local binary to a more conventional directory like `~/.local/bin`.

junegunn:
A more solid solution would be to compare the `--version` of the binaries and choose the newer one. I'll see what I can do.

afnanenayet:
Yeah, I'd agree that makes the most sense

PR #2426, opened 2021-04-03
Add deselect-last option

MrAwesome:
I found myself really wanting a way to deselect the last-selected selection even if the query has been cleared/changed, and couldn't for the life of me come up with any way to hack it together given the commands in place. 

`deselect-last` simply deselects the last-selected selection, and repeated uses will pop back through the selections until there are none left, at which point it's a no-op.

Added the command, updated the docs, and added a test. Please just let me know if there's anything else I need to change here, or if anything doesn't look idiomatic (new to Go).

Local test run:
```[Sat 21/04/03 21:39 CST][pts/11]
<glenn@ronnie:~/MrAwesome/fzf>
$ make
GOARCH=amd64 go build -a -ldflags "-s -w -X main.version=0.26.0 -X main.revision=3318bae" -tags "" -o target/fzf-linux_amd64
[Sat 21/04/03 21:39 CST][pts/11]
<glenn@ronnie:~/MrAwesome/fzf>
$ cp target/fzf-linux_amd64 bin/fzf
[Sat 21/04/03 21:39 CST][pts/11]
<glenn@ronnie:~/MrAwesome/fzf>
$ ruby test/test_go.rb --name test_deselect_last
Run options: --name test_deselect_last --seed 31866
<snip>
Finished in 0.172537s, 5.7959 runs/s, 63.7544 assertions/s.

1 runs, 11 assertions, 0 failures, 0 errors, 0 skips
```

### DIFF
--- a/src/terminal.go
+++ b/src/terminal.go
@@ -282,6 +282,7 @@ const (
 	actEnableSearch
 	actSelect
 	actDeselect
+	actDeselectLast
 )
 
 type placeholderFlags struct {
@@ -2393,6 +2394,14 @@ func (t *Terminal) Loop() {
 			case actDeselect:
 				current := t.currentItem()
 				if t.multi > 0 && current != nil && t.deselectItemChanged(current) {
+					req(reqList, reqInfo)
+				}
+			case actDeselectLast:
+				if t.multi > 0 {
+					len_selected := int32(len(t.selected))
+					if len_selected > 0 {
+						t.deselectItem(t.sortSelected()[len_selected-1].item)
+					}
 					req(reqList, reqInfo)
 				}
 			case actToggle:
--- a/man/man1/fzf.1
+++ b/man/man1/fzf.1
@@ -800,6 +800,7 @@ A key or an event can be bound to one or more of the following actions.
     \fBdelete-char/eof\fR           \fIctrl-d\fR (same as \fBdelete-char\fR except aborts fzf if query is empty)
     \fBdeselect\fR
     \fBdeselect-all\fR              (deselect all matches)
+    \fBdeselect-last\fR             (deselect the last match)
     \fBdisable-search\fR            (disable search functionality)
     \fBdown\fR                      \fIctrl-j  ctrl-n  down\fR
     \fBenable-search\fR             (enable search functionality)
--- a/src/options.go
+++ b/src/options.go
@@ -842,6 +842,8 @@ func parseKeymap(keymap map[tui.Event][]action, str string) {
 				appendAction(actDeleteCharEOF)
 			case "deselect":
 				appendAction(actDeselect)
+			case "deselect-last":
+				appendAction(actDeselectLast)
 			case "end-of-line":
 				appendAction(actEndOfLine)
 			case "cancel":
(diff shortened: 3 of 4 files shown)


junegunn:
Thanks. Please rename the variable to satisfy golint.

> src/terminal.go:2401:6: don't use underscores in Go names; var len_selected should be lenSelected

By the way, can you give an example scenario where this would be really useful? I would personally use `deselect-all` if I want to undo the previous selections after changing the query, instead of "pop"ing them out one-by-one, especially because I probably wouldn't remember the selection order.

MrAwesome:
I'm working with this idea of using fzf as a sort of keyboard for Taiwanese Hokkien. Just started on it today so it's a bit rough, but here's the basic idea:



Pipe TW-US dictionary formatted a particular way into fzf, tab to select a chosen word and have it appear in the preview window as part of a sentence (without the definition), and have a live preview of the about-to-be-selected word at the end of the sentence you're constructing. Once you're happy with the sentence, mash enter and the indices of those chosen words are printed out for my original program to nom up and find the corresponding entries in the dict to get whatever information about those words I might need, parse/present them the way I want, and then send them off to X11 or what-have-you. 

Just started on it today, but I'm really finding it to be a gigantic improvement over "go to online dictionary site, plug in one word at a time, wait for it to load, curse, repeat". fzf handles the latin letter normalization delightfully well, and searching for words is pretty easy. Got some work to do to figure out how to prioritize exact matches / shorter words, but beyond that it's working about as well as I could ever ask for.

MrAwesome:
Oh, got carried away and forgot to say the use case. I want a keyboard shortcut, plus `backward-eof`, to delete the last word in the sentence - otherwise you have to search a sometimes pretty precise phrase to find the last word, tab, then clear again before going on to the next word.

MrAwesome:
> Oh, got carried away and forgot to say the use case. I want a keyboard shortcut, plus `backward-eof`, to delete the last word in the sentence - otherwise you have to search a sometimes pretty precise phrase to find the last word, tab, then clear again before going on to the next word.

Thinking about this some more, I am maybe trying to overload fzf's selection buffer by having it hold the state of my program. As clever as this feels, it does also feel a bit fragile. So it would maybe be best for me to just use `execute-silent` to send off selections elsewhere as they're made, and have `preview` just ask that file/process for the state of the built buffer.

I think it's best not to tie my program's functionality too deeply into a data structure I don't control that is intended for other uses. With that said, I'm happy to leave this up and push it through if anyone has a need/desire for this particular functionality. It did feel surprisingly intuitive to undo the last selections from anywhere (especially with `backward-eof`) instead of going back to each one in turn.

src/terminal.go
```diff
@@ -2395,6 +2396,14 @@ func (t *Terminal) Loop() {
 				if t.multi > 0 && current != nil && t.deselectItemChanged(current) {
 					req(reqList, reqInfo)
 				}
+			case actDeselectLast:
+				if t.multi > 0 {
+					lenSelected := int32(len(t.selected))
+					if lenSelected > 0 {
+						t.deselectItem(t.sortSelected()[lenSelected-1].item)
+					}
+					req(reqList, reqInfo)
```
junegunn, line 2405: `req` can be inside `if lenSelected > 0` block.

junegunn:
So are you still interested in pushing this to fzf? Or should we leave this open and see if anyone has similar requirements?
I'm not particularly against the idea, I just don't see myself using it in my workflow.

Issue #395, opened 2015-10-23
Feature: Set selected entries


ggVGc:
It would be useful to be able to pass a list of strings that should be selected when launching fzf.
And, related, to pass a string on which the cursor should start. I am currently building https://github.com/ggVGc/fzf_browser and with these options the UI could behave much nicer.

junegunn:
I also thought about those stuff, but I felt that maybe I was trying to squeeze too much out of fzf, beyond what it's supposed to do, unnecessarily making it more complex.

> pass a list of strings that should be selected

It's doable. But I thought it can be very confusing if one of the "pre-selected" items is not listed on the initial page.

> to pass a string on which the cursor should start

It's not as trivial as it seems since fzf works asynchronously. And I suppose you'll end up even wanting to preserve the scrolling offset of the window as well.

I'm not going to rush into a decision, rather I'll leave this issue open and see how the idea develops.

ggVGc:
Yeah, I understand there are some implementation design concerns with both of these, but I'm not sure it would be making "fzf unnecessarily too complex". I think it's more an ideological concern than a technical one(which is fine, fzf is your creation and should follow your ideas).
It's a great tool by itself, but even with current configuration options, also shines as a UI for other more complex applications. Writing `fzf_browser` around fzf was great, but as I mentioned would be even better with these two additions.
But maybe it's me thinking about fzf in a different light than you. In my opinion "bloat" is when you add things to your software that could as well have been implemented outside of it, or in an alternative way. When it comes to these two options, I don't see a way to work around it really.
So, if fzf wants to be a powerful building block within larger applications, I do think things like these options have a place.
Regarding implementation concerns I think it's not a big issue if pre-selected entries are off screen, since this can already happen during normal usage

And regarding the cursor position, I think a simple enough and intuitive implementation is:
* While loading,  if the passed string(upon which the cursor should be) is not in the list, then just stay at top. When it shows up, move the cursor to it.
* If the user has moved the cursor at all, or entered any text, cancel(i.e don't put the cursor on the selected string, even if it shows up) 

All these two things add is the ability to restore state upon re-launching fzf, not really introducing any new behaviour as far as I see. The main use case in my opinion is that the user has already built the state previously, so it shouldn't be surprising.

junegunn:
Well, I said "more complex" not "too complex", so there's difference in nuance :)

> So, if fzf wants to be a powerful building block within larger applications

[...]

PR #2440, opened 2021-04-13
preselector option, issue #395

Naheel-Azawy:
As [discussed](https://github.com/junegunn/fzf/issues/395#issuecomment-816934553) in issue #395, here's my quick implementation.

### DIFF
--- a/src/options.go
+++ b/src/options.go
@@ -92,6 +92,7 @@ const usage = `usage: fzf [options]
     -1, --select-1        Automatically select the only match
     -0, --exit-0          Exit immediately when there's no match
     -f, --filter=STR      Filter mode. Do not start interactive finder.
+    --preselector=FILE    File including the pointed and selected items
     --print-query         Print query as the first line
     --expect=KEYS         Comma-separated list of keys to complete fzf
     --read0               Read input delimited by ASCII NUL characters
@@ -207,6 +208,7 @@ type Options struct {
 	Pointer     string
 	Marker      string
 	Query       string
+	Preselector *Preselector
 	Select1     bool
 	Exit0       bool
 	Filter      *string
@@ -268,6 +270,7 @@ func defaultOptions() *Options {
 		Pointer:     ">",
 		Marker:      ">",
 		Query:       "",
+		Preselector: nil,
 		Select1:     false,
 		Exit0:       false,
 		Filter:      nil,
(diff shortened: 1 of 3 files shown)


Naheel-Azawy:
I don't have access to a macOS machine. Anyone have any clue why the macOS tests failed?

junegunn:
Thanks for the suggestion, but I'm not really into the idea.

It looks to me that the implementation is mostly for niche use cases such as yours, and I don't see much point in building a file manager with fzf. Because,
1. fzf was designed to be an interactive grep rather than a framework for building TUI applications so I think the requirement here is beyond the scope of the project and adding this would lead to feature bloat.
2. We already have nice terminal file managers such as [ranger](https://github.com/ranger/ranger), [nnn](https://github.com/jarun/nnn), and [xplr](https://github.com/sayanarijit/xplr) (which even states that it was inspired by fzf). The one built using fzf with duct tapes won't be able to compete with them in terms of features and functionality.

Naheel-Azawy:
I thought a lot before making this pull request. I was worried in particular about the feature bloat. But I've been using my file manager for more than a week and sometimes it gets so frustrating when it jumps back to the beginning. I thought about it and said to myself maybe adding one little option isn't much of a bloat. This might be just my personal opinion though.

I use fzf as a dmenu alternative too. This preselector could also help jump to different items in the menu depending on the context. And who knows, perhaps someone else find it useful somehow. Not sure if it would be considered as bloat or not in this case; I guess not.

As of why did I make yet another file manager, I have briefly discussed it in my [readme](https://github.com/Naheel-Azawy/fmz/blob/master/README.org#why). I tried many other file managers, except xplr, this is the first time I see it. They all are pretty cool. Every one of them has its advantages and disadvantages. My personal favorite used to be lf. But regardless of the file manager I use, I end up using fzf inside of it more than using the file manager itself. I'm not planning to compete with any of them. I'm just trying to build something simple that do not exceed a single posix shell script. Other file managers are, in my opinion, bigger and more complex that they should.

I believe that file managers can be simpler than what exist out there and I guess I could proved it with fmz, at least to myself.

I guess I talked a lot, just expressing my thoughts. Maybe I'm wrong and maybe a get rid of my fmz someday. But as of now, I'm pretty happy with it and I do find myself a bit more productive using it. It's fine if you still find it unnecessary. After all it's just a personal opinion. Let me know what you think.

Naheel-Azawy:
Hi @junegunn, it has been a while. Any thoughts on this?

PR #2443, opened 2021-04-15
Support relatively positioned windows in Neovim

tpict:
Neovim supports a `relative` option in `nvim_open_win()` that allows you to position windows relative to the current window or the cursor (as opposed to the entirety of the editor).

This PR:
- passes all options from `s:popup` to `s:create_popup` so that it can receive `relative` (plus other options like `anchor`)
- uses window width/height instead of `&columns`/`&lines` if the `relative` option is present

A configuration of `let g:fzf_layout = { 'window': { 'relative': 'win', 'width': 1.0, 'height': 0.4 }}]` looks like this:

image

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -972,21 +972,24 @@ else
 endif
 
 function! s:popup(opts) abort
+  let columns = has_key(a:opts, 'relative') ? winwidth(0) : &columns
+  let lines = has_key(a:opts, 'relative') ? winheight(0) : &lines
+
   " Size and position
-  let width = min([max([8, a:opts.width > 1 ? a:opts.width : float2nr(&columns * a:opts.width)]), &columns])
-  let height = min([max([4, a:opts.height > 1 ? a:opts.height : float2nr(&lines * a:opts.height)]), &lines - has('nvim')])
-  let row = float2nr(get(a:opts, 'yoffset', 0.5) * (&lines - height))
-  let col = float2nr(get(a:opts, 'xoffset', 0.5) * (&columns - width))
+  let width = min([max([8, a:opts.width > 1 ? a:opts.width : float2nr(&columns * a:opts.width)]), columns])
+  let height = min([max([4, a:opts.height > 1 ? a:opts.height : float2nr(&lines * a:opts.height)]), lines - has('nvim')])
+  let row = float2nr(get(a:opts, 'yoffset', 0.5) * (lines - height))
+  let col = float2nr(get(a:opts, 'xoffset', 0.5) * (columns - width))
 
   " Managing the differences
-  let row = min([max([0, row]), &lines - has('nvim') - height])
-  let col = min([max([0, col]), &columns - width])
+  let row = min([max([0, row]), lines - has('nvim') - height])
+  let col = min([max([0, col]), columns - width])
   let row += !has('nvim')
   let col += !has('nvim')
 
-  call s:create_popup('Normal', {
+  call s:create_popup('Normal', extend(a:opts, {
     \ 'row': row, 'col': col, 'width': width, 'height': height
-  \ })
+  \ }))
 endfunction
 
 let s:default_action = {
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -987,7 +987,7 @@ function! s:popup(opts) abort
   let row += !has('nvim')
   let col += !has('nvim')
 
-  call s:create_popup('Normal', extend(a:opts, {
+  call s:create_popup('Normal', extend(copy(a:opts), {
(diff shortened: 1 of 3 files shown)


plugin/fzf.vim
```diff
@@ -972,21 +972,24 @@ else
...
   " Managing the differences
-  let row = min([max([0, row]), &lines - has('nvim') - height])
-  let col = min([max([0, col]), &columns - width])
+  let row = min([max([0, row]), lines - has('nvim') - height])
+  let col = min([max([0, col]), columns - width])
   let row += !has('nvim')
   let col += !has('nvim')
 
-  call s:create_popup('Normal', {
+  call s:create_popup('Normal', extend(a:opts, {
```
junegunn, line 990: We should `extend(copy(a:opts), ...` to avoid mutating the user's `g:fzf_layout`.

junegunn (changes requested):

junegunn (changes requested):
This doesn't work when `g:fzf_layout` contains extra attributes such as `xoffset`, `yoffset`,  and `border`.

```
E5555: API call: Invalid key 'yoffset'
```

tpict:
I've opted to remove the three documented FZF-specific options from the `opts` dict rather than constructing the dictionary from scratch–I figure this will require less future maintenance if the Vim/Neovim APIs are extended.

plugin/fzf.vim
```diff
@@ -972,21 +972,47 @@ else
...
+  let columns = has_key(a:opts, 'relative') ? winwidth(0) : &columns
+  let lines = has_key(a:opts, 'relative') ? winheight(0) : &lines
+
   " Size and position
-  let width = min([max([8, a:opts.width > 1 ? a:opts.width : float2nr(&columns * a:opts.width)]), &columns])
-  let height = min([max([4, a:opts.height > 1 ? a:opts.height : float2nr(&lines * a:opts.height)]), &lines - has('nvim')])
-  let row = float2nr(get(a:opts, 'yoffset', 0.5) * (&lines - height))
-  let col = float2nr(get(a:opts, 'xoffset', 0.5) * (&columns - width))
+  let width = min([max([8, a:opts.width > 1 ? a:opts.width : float2nr(&columns * a:opts.width)]), columns])
+  let height = min([max([4, a:opts.height > 1 ? a:opts.height : float2nr(&lines * a:opts.height)]), lines - has('nvim')])
```
junegunn, line 1001: We still have `&columns` and `&lines` here.

junegunn (changes requested):

plugin/fzf.vim
```diff
@@ -972,21 +972,47 @@ else
...
+
+  if has_key(a:opts, 'yoffset')
+    let yoffset = a:opts.yoffset
+    call remove(create_opts, 'yoffset')
+  endif
+
+  if has_key(a:opts, 'border')
+    call remove(create_opts, 'border')
+  endif
+
```
junegunn, line 994: Can we also remove `highlight` property? It's deprecated but some users may still have it.

https://github.com/junegunn/fzf/commit/6e93eefc82d8fb749e97d094e6ad6fbbe09d289b

junegunn:
```vim
  " Remove known FZF options
  let xoffset = get(a:opts, 'xoffset', 0.5)
  let yoffset = get(a:opts, 'yoffset', 0.5)

  for key in ['xoffset', 'yoffset', 'border', 'highlight']
    if has_key(a:opts, key)
      call remove(create_opts, key)
    endif
  endfor
```

What do you think?

tpict: Looks good! I'll add a commit for it

junegunn:
I'm seeing a few issues with this implementation, and here are my thoughts.

* Position is relative to the current window even when `relative` is set to `editor` or `cursor` because it only looks for the `relative` key.
    * We can safely fix this and ignore `editor`, but what should happen on `cursor`? Should we allow it, or should we raise an error?
    * We are leaking the abstraction layer by passing arbitrary arguments to the underlying implementation, but we can't guarantee that those arguments would work as expected and conceptually compatible with our other parameters such as `xoffset` and `yoffset`.
* The current implementation only works in Neovim, but we can easily implement the same functionality on regular Vim using `win_screenpos(0)`
    * Neovim also provides the function, so we can have the unified implementation.

If we can provide the same functionality on both products, our API probably should not be determined by the API of one specific product (`nvim_open_win`). So I'm suggesting that we only allow boolean-valued `relative` property in `g:fzf_layout.window` and implement the relative positioning `win_screenpos(0)`.

Let me update the branch.

junegunn:
Merged, thanks!

tpict:
Thanks for wrapping this up!

PR #2447, opened 2021-04-17
Add argument specifying root directory for built-in file search

stinos:
I often (like, multiple times a day) find myself in directory A but wanting to select files with fzf in directory B. In those cases as far as I'm aware one either has to go to that directory or else construct a command similar to the default one and feed that into fzf. Moreover the default implementation for Windows with saracen's walker and it is way faster than anything readily available. Meaning that a custom command to list files outside of the current directory is either relatively slow, or requires a specialized tool or custom code. All this while the default way fzf uses to list files works fine in most cases (anecdotal only of course) so just adding the ability to select the root directory seems worth it.

### DIFF
--- a/man/man1/fzf.1
+++ b/man/man1/fzf.1
@@ -541,6 +541,9 @@ Exit immediately when there's no match
 Filter mode. Do not start interactive finder. When used with \fB--no-sort\fR,
 fzf becomes a fuzzy-version of grep.
 .TP
+.BI "-r, --root=" "STR"
+Root directory for listing files. Defaults to current directory.
+.TP
 .B "--print-query"
 Print query as the first line
 .TP
--- a/src/constants.go
+++ b/src/constants.go
@@ -57,10 +57,11 @@ const (
 var defaultCommand string
 
 func init() {
+	// The %s and %d get replaced with root directory and cut position
 	if !util.IsWindows() {
-		defaultCommand = `set -o pipefail; command find -L . -mindepth 1 \( -path '*/\.*' -o -fstype 'sysfs' -o -fstype 'devfs' -o -fstype 'devtmpfs' -o -fstype 'proc' \) -prune -o -type f -print -o -type l -print 2> /dev/null | cut -b3-`
+		defaultCommand = `set -o pipefail; command find -L %s -mindepth 1 \( -path '*/\.*' -o -fstype 'sysfs' -o -fstype 'devfs' -o -fstype 'devtmpfs' -o -fstype 'proc' \) -prune -o -type f -print -o -type l -print 2> /dev/null | cut -b%d-`
 	} else if os.Getenv("TERM") == "cygwin" {
-		defaultCommand = `sh -c "command find -L . -mindepth 1 -path '*/\.*' -prune -o -type f -print -o -type l -print 2> /dev/null | cut -b3-"`
+		defaultCommand = `sh -c "command find -L %s -mindepth 1 -path '*/\.*' -prune -o -type f -print -o -type l -print 2> /dev/null | cut -b%d-"`
 	}
 }
 
--- a/src/core.go
+++ b/src/core.go
@@ -141,7 +141,7 @@ func Run(opts *Options, version string, revision string) {
 		reader = NewReader(func(data []byte) bool {
 			return chunkList.Push(data)
 		}, eventBox, opts.ReadZero, opts.Filter == nil)
-		go reader.ReadSource()
+		go reader.ReadSource(opts.RootDir)
 	}
 
 	// Matcher
@@ -185,7 +185,7 @@ func Run(opts *Options, version string, revision string) {
 					}
 					return false
 				}, eventBox, opts.ReadZero, false)
-			reader.ReadSource()
+			reader.ReadSource(opts.RootDir)
 		} else {
 			eventBox.Unwatch(EvtReadNew)
 			eventBox.WaitFor(EvtReadFin)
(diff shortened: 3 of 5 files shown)


rayiik:
just wondering is there a reason why you dont set this the the FZF_DEFAULT_COMMAND? im genuinely curious i took a look at the changes but it seems unnecessary when this works, (unless im missing something and please do tell me as im not very fluent in go.) 
```bash
f1 () {
local FZF_DEFAULT_COMMAND
export FZF_DEFAULT_COMMAND="command ${1:-'.'} /"
fzf
}
f2 () {
local FZF_DEFAULT_COMMAND
FZF_DEFAULT_COMMAND="command ${1:-'.'} $HOME"
fzf
}
```
this way you can take it further and set path the $2 then you could control the variable and the path (any path) with very little work f1 query path.

stinos:
@rayiik I'm not really sure what you mean; how are these f1/f2 functions supposed to be used? I tried a bunch of things but they just start fzf in the current directory (tested with bash)?

Anyway I guess my main reasoning is: fzf has a built-in file search, so it seems convenient to be able to use that everywhere instead of only for the current working directory.

rayiik:
hmm that's weird (not the f1 f2 keys but f 1 typed out)  but f1 "search term" should search for files from root directory even if you're on a different branch and filter based on word hope this helps as ive tested on Kali wsl and arch. but basically it takes the first positional parameter if provided or it searches everything. hope this helps i figured seeing it would be easier :)
[asciicast](https://asciinema.org/a/413720)

rayiik:
*edited the original wasn't supposed to have the export in there, * this next part is mostly for others who may stumble across this but full breakdown FZF_DEFAULT_COMMAND is called  when no input is given to fzf (either by piping or redirection) when defined locally this temporarily overrides what ever else is set to the default command and the variable doesnt pass the function boundaries () {  .....  }  while defining the default command we use parameter substitution  ${1:-'.'} what this means is we use the first space separated (excluding quotes) word after the function f1 'word'  as the input for the variable (or in this case search term) we then define a default action if no $1 is given :-'.' by defaulting to . we're telling our search term to match any single character. after defining we call fzf without input to trigger the default command. 
```bash
#!/bin/bash
f1 () {
local FZF_DEFAULT_COMMAND
FZF_DEFAULT_COMMAND="fdfind -uu -t f -d 4 ${1:-'.'} /"
fzf
}
f2 () {
local FZF_DEFAULT_COMMAND
FZF_DEFAULT_COMMAND="fdfind -uu -t f -d 4 ${1:-'.'} $HOME"
fzf
}
```

stinos:
> but f1 "search term" should search for files from root directory

yeah I figured that. But that's not what I'm after: there's hardly any difference between entering the search term on the commandline or in fzf itself, i.e. there's no gain in speed or usability. What I want is to specify the directory where fzf starts its search. So basically like your _f1 function. Difference being that this PR allows `fzf -r <dir>` and that's it, no need for custom functions etc.

rayiik:
Yea but thats my point instead of writting 4 lines of code to produce the same result (arguably with less typing) is it really more efficent to write/ add 28 lines of code instead? Just seems like an excessive alternative when theres already a simple solution if im missing something please let me know as im still learning most of this and i only have minor go experience

stinos:
> Yea but thats my point instead of writting 4 lines of code to produce the same result (arguably with less typing) is it really more efficent to write/ add 28 lines of code instead

That's a bit of an unfair comparison since it doesn't count all lines of code needed for `fdfind` which needs to be installed separately :) Also it's not just about efficiency, but about convenience by leveraging an existing part of the application to do what it's built for instead of having to resolve to custom code to do the exact same thing.

rayiik:
Maybe it isnt a fair comparison, but consider this, I used fdfind because its fast lets say we dont set the default command fzf still calls an external element to process and populate the list as fzf itself just filters (as far as i understand). Which i believe defaults to find on linux and  /dir or get-childitem on windows. this can still be achieved in a similar manner as a make file with
```bash
Origdir="$(pwd)"
Searchdir="${1:-"$Origdir"}"
If [[ "$Searchdir" -ne "$Origdir" ]];
         then
       cd "$Searchdir" || exit 2
fzf
cd “$Origdir“ || exit 3
fi
'''
I guess my biggest concern is the kinds effects this might have on my powershell scripts vs the functionality it provides :) but i do understand what you mean and im just play a bit of devils advocate to get a better understanding of my ideals and other trains of though when it comes to programming so thanks for discussing :)

junegunn:
Thanks for the suggestion, but I don't want to add this option to fzf.

> requires a specialized tool or custom code

I can understand your hesitation to install another program, but fzf was **designed to be used with other programs** from the very start, following the Unix philosophy. Also, having this option may give the wrong impression to the audience that fzf is a "find" alternative, a very primitive one with lack of crucial features for a "find"-like program, when it's actually more like "grep". Maybe I should have called it "fuzzy filter" rather than "fuzzy finder" in the first place to be clear (So then it's fz..f? 😄 )

As mentioned above, [fdfind (or fd)](https://github.com/sharkdp/fd) is a very nice "find" alternative I've always recommended to the users. I believe we should guide the users to install such specialized programs if they don't like the basic file system traversal fzf provides.

stinos:
> Also, having this option may give the wrong impression to the audience that fzf is a "find" alternative, a very primitive one lacking crucial features for a "find"-like program, when it's actually more like "grep"

Fair enough, but why does fzf have any finding functionality at all then? Just to 'get started'?

junegunn:
> Just to 'get started'?

Yes, but if I were to start writing fzf today, I probably won't include it.

Issue #2465, opened 2021-04-24
[FEATURE] dynamic `replace-query`


githorse:
- [x] I have read through the manual page (`man fzf`)
- [x] I have the latest version of fzf
- [x] I have searched through the existing issues

## Info

- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [x] bash
    - [x] zsh
    - [ ] fish
    
## Idea

I'd like to create a "search engine" widget similar to Alfred/Albert or the Firefox/Chrome omnibar. To make this work, I want the user to be able to type in (customizable) triggers to perform different searches/actions:

```
ps:<space>    # trigger process search mode
files:<space> # trigger file search mode
dirs:<space>  # trigger directory search mode
etc.
```
I can achieve the *functionality* of this already by using `change` and `reload` to switch the logic based on the trigger sequence:

```
fzf --bind 'change:reload(custom-search {q})`
```

But I can't quite achieve the *look* that I want, because the trigger sequence of course is still displayed in the query.

![image](https://user-images.githubusercontent.com/8271372/115969459-06aa9d00-a50b-11eb-9b53-d0d66d5168cb.png)

Whereas what I really want is to either 

1) delete the trigger from the query string and absorb it into a [dynamic prompt](https://github.com/junegunn/fzf/issues/2299), or
2) rewrite the query string itself, with ANSI colors.

Either of those options would allow me to mimic the omnibar search engine functionality to do something like this:

![image](https://user-images.githubusercontent.com/8271372/115968616-c1846c00-a506-11eb-9250-b6aa13cf639c.png)

![image](https://user-images.githubusercontent.com/8271372/115968628-c9441080-a506-11eb-8779-f62974661fbf.png)

I'm guessing this would look something like:

```
fzf --bind 'change:replace-query(format_query {q})'
```

Note that we can get pretty close here with the `--header-lines` option. But presumably being able to rewrite the query at will would open up other nifty features (syntax highlighting?). There are some complications, like what to do with the cursor if you're not at the end of the line (maybe pass the cursor position as a second argument to `format_query`?)

See also [my feature request for multi-character trigger sequences](TBD).

jbriales:
I second this. I was myself looking for a similar feature which would let me work around a nr of limitations I'm finding when using `reload`.
[...]

PR #2559, opened 2021-07-16
Add change-query execution action

SpicyLemon:
Add a `change-query(...)` action. This action will execute the given command and replace the current query with the result.

closes: #2465


src/terminal.go
```diff
@@ -1758,6 +1759,46 @@ func (t *Terminal) executeCommand(template string, forcePlus bool, background bo
...
+func (t *Terminal) executeChangeQuery(template string) {
+	valid, list := t.buildPlusList(template, false)
+	if !valid {
+		return
+	}
+	command := t.replacePlaceholder(template, false, string(t.input), list)
+	cmd := util.ExecCommand(command, false)
+	newQuery := []rune{}
+	t.executing.Set(true)
+	cmd.Stdin = tui.TtyIn()
```
SpicyLemon, line 1771: I put this line in because it's in the `executeCommand` func above. I'm not exactly sure what it does but am guessing that it provides user input to interactive commands being executed. I haven't yet tried out an interactive command in the new `change-query(...)` action, though.

src/terminal.go
```diff
@@ -1758,6 +1759,46 @@ func (t *Terminal) executeCommand(template string, forcePlus bool, background bo
...
+	out, _ := cmd.StdoutPipe()
+	t.tui.Pause(false)
```
SpicyLemon, line 1773: I'm not entirely sure what `t.tui.Pause` and `t.tui.Resume` do, so I included them based on what `executeCommand` does. But they might not be actually needed.

src/terminal.go
```diff
@@ -1758,6 +1759,46 @@ func (t *Terminal) executeCommand(template string, forcePlus bool, background bo
...
+			newQuery = append(newQuery, []rune(line)...)
+			if err != nil {
+				break
+			}
+		}
+	}
+	err = cmd.Wait()
+	if err != nil {
+		newQuery = append(newQuery, []rune(err.Error())...)
+	}
```
SpicyLemon, line 1791: I based this off of a combination of what's done for the preview pane and `executeCommand`. I'm not too familiar with the behavior of either the `exec` or `bufio` libraries, though, so I'm not sure if there's a better way.

I initially tried `cmd.Run()` but I must have been missing something because I couldn't figure out how to get any output from it.

I didn't use `cmd. CombinedOutput()` because it returns a byte array, and I didn't want to have to try to correctly convert that to a rune array (needed for `t.input`). But converting strings to rune arrays seems straight-forward.

src/terminal.go
```diff
@@ -1758,6 +1759,46 @@ func (t *Terminal) executeCommand(template string, forcePlus bool, background bo
...
+	if newQuery[len(newQuery)-1] == rune('\n') {
+		newQuery = newQuery[:len(newQuery)-1]
+	}
```
SpicyLemon, line 1794: I was torn on adding this, but decided to put it in.

While testing using `sed 's/^/_/' <<< {q}` as the command, I found that the query was getting a space at the end. It was the newline being added by `sed` being converted somewhere after being set in here.

A lot of commands do some weird stuff to figure out whether or not to include a final newline, and trying to fix it in the command is often clunky at best. So I added the lines in here to do it. I figure a trailing newline is almost always meaningless as far as a query goes. And since it was being turned into a space anyway, it's almost always incorrect to include. Usually, I'd stay away from automatic manipulation like this, but in this case, I think it's warranted. It's much more likely that a user would be confused without it than with it.

SpicyLemon (commented):
Some notes on uncertainties encountered and decisions made while making this change.

src/terminal.go
```diff
@@ -1758,6 +1759,46 @@ func (t *Terminal) executeCommand(template string, forcePlus bool, background bo
...
+	out, _ := cmd.StdoutPipe()
+	t.tui.Pause(false)
+	cmd.Stderr = cmd.Stdout
+	reader := bufio.NewReader(out)
+	err := cmd.Start()
+	if err != nil {
+		newQuery = append(newQuery, []rune(err.Error())...)
+	} else {
+		for {
+			line, err := reader.ReadString('\n')
```
vovcacik, line 1781: This works windows and *nix line endings. But then you may have `\n` or `\r\n` at the end of the string.

src/terminal.go
vovcacik: This also needs to trim `\r\n`.

Use case:
Press ctrl+x to update query with the file extension of current item:
- when I run `fzf.exe --delimiter "\." --bind "ctrl-x:change-query(echo {2})"` and execute it on "test.txt" item in fzf, the query is changed to `"txt"\r`. The `\r` is zero-width, but it does break the filtering. (The double quotes break the filtering as well, but these are straightforward to deal with)

PR #2564, opened 2021-07-20
add bind action

christian-schulze:
This is a proof of concept `bind` action, which as an example, can be used to rebind the `change` event.

Building on your third advanced [ripgrep](https://github.com/christian-schulze/fzf/blob/master/ADVANCED.md#switching-to-fzf-only-search-mode) example, this allows switching back and forth between `ripgrep` and `fzf` search modes:
```bash
#!/usr/bin/env bash

# Two-phase filtering with Ripgrep and fzf
#
# 1. Search for text in files using Ripgrep
# 2. Interactively restart Ripgrep with reload action
#    * Press ctrl-f to switch to fzf-only filtering
#    * Press ctrl-r to switch back to ripgrep
# 3. Open the file in Vim
RG_PREFIX="rg --column --line-number --no-heading --color=always --smart-case "
INITIAL_QUERY="${*:-}"
IFS=: read -ra selected < <(
  FZF_DEFAULT_COMMAND="$RG_PREFIX $(printf %q "$INITIAL_QUERY")" \
  fzf --ansi \
      --color "hl:-1:underline,hl+:-1:underline:reverse" \
      --disabled \
      --query "$INITIAL_QUERY" \
      --bind "change:reload:sleep 0.1; $RG_PREFIX {q} || true" \
      --bind "ctrl-f:unbind(change)+change-prompt(2. fzf> )+enable-search+clear-query" \
      --bind "ctrl-r:change-prompt(1. ripgrep> )+disable-search+clear-query+bind(change:reload:sleep 0.1; $RG_PREFIX {q} || true)" \
      --prompt "1. ripgrep> " \
      --delimiter : \
      --preview "bat --color=always {1} --highlight-line {2}" \
      --preview-window "up,60%,border-bottom,+{2}+3/3,~3"
)
[ -n "${selected[0]}" ] && vim "${selected[0]}" "+${selected[1]}"
```

What do you think, is this something you would consider adding?

### DIFF
--- a/src/options.go
+++ b/src/options.go
@@ -780,7 +780,7 @@ func init() {
 	// Backreferences are not supported.
 	// "~!@#$%^&*;/|".each_char.map { |c| Regexp.escape(c) }.map { |c| "#{c}[^#{c}]*#{c}" }.join('|')
 	executeRegexp = regexp.MustCompile(
-		`(?si)[:+](execute(?:-multi|-silent)?|reload|preview|change-prompt|unbind):.+|[:+](execute(?:-multi|-silent)?|reload|preview|change-prompt|unbind)(\([^)]*\)|\[[^\]]*\]|~[^~]*~|![^!]*!|@[^@]*@|\#[^\#]*\#|\$[^\$]*\$|%[^%]*%|\^[^\^]*\^|&[^&]*&|\*[^\*]*\*|;[^;]*;|/[^/]*/|\|[^\|]*\|)`)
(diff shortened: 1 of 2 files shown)


junegunn:
I would avoid doing this. We can't properly cope with nested bind actions.

Instead, we can consider adding `rebind(...)` so that you can restore a binding that was previously disabled via `unbind`.

christian-schulze:
ok cool, i will look into how this will work 👍

PR #2567, opened 2021-07-25
Allow Fzf to be embedded into any go project

LivinParadoX:
Hello,

I've been using fzf for a while now, both from the cli and by embedding it into programs the recommended way (https://junegunn.kr/2016/02/using-fzf-in-your-program).
However, now I'm learning Go and I wanted to use fzf in a program without needing to rely on it being on the target system.
I first tried to do that client-side by using Fzf's public API, but a lot of necessary parts are private (e.g. defaultOptions).
Failing that, I forked your repo and created two new functions:
 - Fzf(in <-chan string, out chan<- Out, options string) {} is largely based on Run from core.go, except it replaces os.StdIn with a string channel, and os.StdOut and os.StdErr by an output channel (containing both strings and errors).
 - FzfFilter(choices []string, options string) (chosen []string, err error) {} is a helper function doing what I'll mostly need, and it does not require any private stuff.

Since I did not want to refactor any of your code (I don't want to introduce any bug), everything is contained in a new file.
However, since Run is a large function, there are a lot of redundancy, I think you should consider splitting it.

In order to break out of the started goroutines (such as matcher.Loop()) without needing to refactor important parts, I used an ugly defer, panic and recover trick. I believe it works as intended without leaking any ressource.

There are a lot of calls to os.Exit in options.go and elsewhere, the functions will crash the whole program if those are triggered.

There may be other unwanted side effects I am not aware of, and they are probably the reason why you did not want to do this in the first place...

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -14,6 +14,25 @@ builds:
       - amd64
     ldflags:
       - "-s -w -X main.version={{ .Version }} -X main.revision={{ .ShortCommit }}"
+    hooks:
+      post: |
+        sh -c '
+        cat > /tmp/fzf-gon-amd64.hcl << EOF
+        source = ["./dist/fzf-macos_darwin_amd64/fzf"]
+        bundle_id = "kr.junegunn.fzf"
+        apple_id {
+          username = "junegunn.c@gmail.com"
+          password = "@env:AC_PASSWORD"
+        }
+        sign {
+          application_identity = "Developer ID Application: Junegunn Choi (Y254DRW44Z)"
+        }
+        zip {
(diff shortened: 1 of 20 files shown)


LivinParadoX:
Actually, please disregard this PR.
I will:
 - Fork the devel branch instead of master
 - Split the Run function into smaller parts
 - Write Fzf and FzfFilter using only those parts that are needed without redundancy, directly into core.go
 - Write the small needed parts in reader.go
 - Refactor the os.Exit I can find into panics, and catch them in the highest-level functions (Run and Fzf).

Hopefully this will achieve the following benefits:
 - Less code redundancy
 - Not breaking anything
 - Introduce as few changes as possible

PR #2577, opened 2021-08-05
Document using which to set rtp for vim

hfaulds:
This might not be worth documenting but I ended up doing this so I could share my `.vimrc` between linux and mac whilst using package managers that install fzf in different locations and I thought it might be useful to other's.


hfaulds:
Well this is slightly embarrassing, apparently I only tested this on Linux and it doesn't resolve the homebrew symlinks on macos so doesn't actually work

PR #2578, opened 2021-08-08
Minor readme change

leontepe:
`CTRL-K` moves up and `CTRL-J` moves down, not the other way around (same for `CTRL-P` and `CTRL-N`)


junegunn:
Thanks!

PR #2598, opened 2021-08-24
Minor cleanup of fzf-tmux

a1346054:


bin/fzf-tmux
```diff
@@ -7,8 +7,8 @@ fail() {
   exit 2
 }
 
-fzf="$(command -v fzf 2> /dev/null)" || fzf="$(dirname "$0")/fzf"
-[[ -x "$fzf" ]] || fail 'fzf executable not found'
+fzf=$(command -v fzf 2>/dev/null) || fzf="$(dirname "$0")/fzf"
```
junegunn, line 10: Why do we remove the space before `/dev/null`?

junegunn: Related? https://github.com/koalaman/shellcheck/issues/1344#issuecomment-443542106

a1346054: Was just a stylistic choice, I removed it now.

a1346054:
I rebased the pull request on latest master to fix merge conflicts, and reworked the commits to make more minimal changes, avoiding stylistic fixes.

It makes the scripts more robust, and tiny bit faster because bash doesn't have to start up the globbing logic when variables are properly quoted.

a1346054:
Rebased to fix merge conflicts.

By the way, many of the underquoting issues could be fixed instead by using `[[ ... ]]` instead of `[ ... ]` but I chose to not do that change here yet.

Issue #2603, opened 2021-08-29
"export:310: not valid in this context: --bind" error when 


sdondley:
- [x] I have read through the manual page (`man fzf`)
- [x] I have the latest version of fzf
- [x] I have searched through the existing issues

## Info

- OS
    - [ ] Linux
    - [x ] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [ ] bash
    - [x] zsh
    - [ ] fish

## Problem / Steps to reproduce

DISCLAIMER, I searched the manual but did not read the whole thing. Search on the web turned up nothing with my exact error.

Notes: rg and fzf have are installed and updated via brew.

Following the example in the README, copy and paste the following for integration with rg:

```
INITIAL_QUERY=""
RG_PREFIX="rg --column --line-number --no-heading --color=always --smart-case "
FZF_DEFAULT_COMMAND="$RG_PREFIX '$INITIAL_QUERY'" \
  fzf --bind "change:reload:$RG_PREFIX {q} || true" \
      --ansi --disabled --query "$INITIAL_QUERY" \
      --height=50% --layout=reverse
```

When the shell is reloaded, I get the following error:

`.zshrc:export:310: not valid in this context: --bind`

sdondley:
I forgot to mention that i'm using ohymyz.sh with the fzf plugin. And per the instructions in the plugin README file, I have placed the keyword `export` in front of `FZF_DEFUALT_COMMAND`. I tried removing `export` but then fzf with rg doesn't work at all.

sdondley:
OK, I don't think this is a problem with fzf but with my zsh config. I'm going to close.

sdondley:
Reopening because there is a similar issue in the queue at #2481. I updated the patched files (shell/completion.zsh and shell/key-bindings.sh) and ran the install.sh script and reloaded zsh but I still get the same error:

`.zshrc:export:310: not valid in this context: --bind`

Not sure what else to do at this point.

sdondley:
OK, the command were not designed to go into the shell's config file. Putting them in a separate script solved the problem.

PR #2605, opened 2021-08-30
provide guidance on implementing the shell snippet

sdondley:
This is to help newer shell users avoid confusion. See #2603

### DIFF
--- a/README.md
+++ b/README.md
@@ -578,6 +578,9 @@ expression `{q}`. Also, note that we used `--disabled` option so that fzf
 doesn't perform any secondary filtering.
 
 ```sh
+# Add this snippet into a separate shell script or wrap it in a function.
+# Do not place it directly into a shell configuration file.
+
 INITIAL_QUERY=""
 RG_PREFIX="rg --column --line-number --no-heading --color=always --smart-case "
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks for the suggestion, but I think it's stating the obvious. If we are to do this, we'll have to repeat the comment in every example in the page.

oryband:
> Thanks for the suggestion, but I think it's stating the obvious. If we are to do this, we'll have to repeat the comment in every example in the page.

Hi. I am very confused about the error, and don't understand what is obvious. How should I fix this problem? I tried adding this to my `.zshrc` and still getting the same error:
```sh
fzf_default_command() {
INITIAL_QUERY=""
RG_PREFIX="rg --column --line-number --no-heading --color=always --smart-case "
export FZF_DEFAULT_COMMAND="$RG_PREFIX '$INITIAL_QUERY'" \
	fzf --bind "change:reload:$RG_PREFIX {q} || true" \
	--ansi --disabled --query "$INITIAL_QUERY" \
	--height=50% --layout=reverse
}
fzf_default_command
```

oryband:
@sdondley :arrow_up_small:

PR #2614, opened 2021-09-21
Minor changes

vovcacik:
This is few minor things I came across lately.

- fmt.Println in tests doesn't honor verbosity flags like t.Log does
- the keyboard event works just fine for space character


junegunn:
Thanks!

PR #2615, opened 2021-09-21
Change the tests to run on Windows

vovcacik:
I needed to write some tests, but the existing tests were failing on Windows. In this pull request you can find changes that make the tests run and pass on both platforms.

The first commit prepares the ground and make some strings in the tests parametric. I'm using the text/templates package from Go. The second commit then contains all changes related to Windows platform.


src/reader_test.go
```diff
@@ -23,7 +23,7 @@ func TestReadFromCommand(t *testing.T) {
 
 	// Normal command
 	reader.fin(reader.readFromCommand(nil, `echo abc && echo def`))
-	if len(strs) != 2 || strs[0] != "abc" || strs[1] != "def" {
+	if len(strs) != 2 || strs[0] != util.OS.Sieve("abc", "abc ") || strs[1] != "def" {
```
junegunn, line 26: I would simply change the command so that it yields the same result on both systems.

i.e. `echo abc&&echo def`

src/util/templates.go
```diff
@@ -0,0 +1,18 @@
+package util
+
+import (
+	"bytes"
+	"text/template"
+)
+
+// Helper function to parse, execute and convert "text/template" to string. Panics on error.
+func TemplateToString(format string, data interface{}) string {
```
junegunn, line 9: I would remove this file and put this function in terminal_test.go at the moment.

src/util/os.go
```diff
@@ -0,0 +1,18 @@
...
+import (
+	"runtime"
+)
+
+type osName string
+
+const OS = osName(runtime.GOOS)
+
+// Returns one of the arguments, based on operating system
+func (os osName) Sieve(onUnix interface{}, onWindows interface{}) interface{} {
```
junegunn, line 12: This file can be removed too. Once we modify the command in reader_test.go, this function is only needed in terminal_test.go. No need to extend the public API.

junegunn (changes requested):

src/util/os.go
vovcacik: I have addressed the issues in the commit below. But I removed the Sieve method, it felt wrong to keep it with only one occurrence.

junegunn (approved):

junegunn:
Merged, thanks!

PR #2616, opened 2021-09-21
Align keyboard and mouse support in FullscreenRenderer with LightRenderer

vovcacik:
This pull request is focused on fixing some issues and adding new features in the FullscreenRenderer. Most of the branch is about keyboard events, but the last commit also implements mouse events.

Keyboard
-----------
- I tried to bring the keyboard events in FullscreenRenderer as close as possible to the LightRenderer
- additions:
  - extensive testing of the tcell events that fzf uses as source for keyboard events
  - ability to input AltGr characters on foreign keyboards
- fixes:
  - CtrlSpace event didn't work
  - CtrlCaret event wasn't implement
  - there was big discrepancy in how backspace key is handled between FullscreenRenderer and LightRenderer. The CtrlH event was emitted even in the cases where BSpace or AltBS was appropriate.

Mouse
-------
The mouse events in LightRenderer are consumed by the shell in Windows, so I didn't have reference point. Based on existing code it should be close to what user would expect, but I noticed something that looked like Shift+Click support in LightRenderer, so this implementation may still lag behind the unix one.

### DIFF
--- a/src/tui/tui.go
+++ b/src/tui/tui.go
@@ -5,8 +5,6 @@ import (
 	"os"
 	"strconv"
 	"time"
-
-	"github.com/gdamore/tcell"
 )
 
 // Types of user action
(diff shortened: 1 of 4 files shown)


junegunn:
Thanks. I'll review the patch when I get some time.

/cc @kelleyma49

src/tui/tui.go
```diff
@@ -5,6 +5,8 @@ import (
 	"os"
 	"strconv"
 	"time"
+
+	"github.com/gdamore/tcell"
```
junegunn, line 9: This makes non-Windows binaries, which don't need tcell, unnecessarily larger.

```
-rwxr-xr-x  1 jg  staff  2785696 Sep 29 17:29 fzf-darwin_amd64
-rwxr-xr-x  1 jg  staff  3198592 Sep 29 17:28 fzf-darwin_amd64-tcell-linked
```

An easy, albeit non-elegant, solution would be to define it as a non-member variable in `tcell.go` (similarly to `_screen` https://github.com/junegunn/fzf/blob/f4f47f5fe32db500765b47d937f5eb8fa27be130/src/tui/tcell.go#L123-L125),  because FullscreenRenderer is effectively a singleton.

vovcacik: Thanks for pointing it out, I didn't realize it.

vovcacik:
I am bit puzzled by the failing integration check. I've rerun the workflow on the fork with same result though.

junegunn:
Merged, thanks for the contribution!

Issue #2609, opened 2021-09-07
[Issue] Expanding {n} automatically adds quotes and character escapes


pidgeon777:
- [x] I have read through the manual page (`man fzf`)
- [x] I have the latest version of fzf
- [x] I have searched through the existing issues

## Info

- OS
    - [ ] Linux
    - [ ] Mac OS X
    - [x] Windows
    - [ ] Etc.
- Shell
    - [ ] bash
    - [ ] zsh
    - [ ] fish
    - [x] PowerShell

## Problem / Steps to reproduce

I'm using this command:

```powershell
fzf --preview="bat --style=numbers --color=always {2}" --bind "ctrl-b:execute(notepad++ -n{1} {2})"
```

This is executed when applying `Ctrl+B` on a result:

```powershell
notepad++ -n"12" "C:\\Work\\Test Folder\\File.txt"
```

I see that `{2}` is automatically expanded with the properly escaped path.

But `{1}` is causing issues. I would need to expand it without adding any quote, also without making use of any further tool such as **sed**/**awk**.

How could this be achieved in **FZF**?

junegunn:
Is this a problem specific to Windows environment? Because `-n"12"` isn't a problem on *nix shells.

i.e. There's no difference between `ls -al | sort -nk"5"` and `ls -al | sort -nk5`

pidgeon777:
The fact is that I wish to open the **FZF** selected file with **Notepad++**. It accepts these arguments:

```powershell
notepad++ -n<LINE_NUM> <FILE_PATH>
```

`<LINE_NUM>` must not include quotes (`"`), otherwise, it won't work. Thus, I need to output `{n}` without quotes.

pidgeon777:
Obviously, on Windows, this is not an issue strictly related to **Notepad++**, but potentially to every command-line application accepting arguments.

Thus, this issue should be kept into great account, in my opinion.

pidgeon777:
@junegunn unfortunately on Windows this "issue" makes it impossible to execute any command for example through:

```powershell
type "C:\Scripts\FZF\Main.bat" | fzf --delimiter="\t" --bind "enter:execute({2})"
````

`Main.bat` content:

````bat
p    ping test.it
e    notepad++ "C:\Work\Test.txt"
````

When trying for example to run the `e` command, this will be wrongly returned by the `{2}` operator instead:

````bat
"notepad++ \"C:\\Work\\Test.txt\""
````

[...]

PR #2624, opened 2021-10-05
Add more tests of placeholder flags and simplify its logic

vovcacik:
Hi, this is some work I have done while playing with #2609. No changes though, just tests, refactoring and bit of docs.

Notable:
- Most of the value is in the tests that improve the coverage of placeholder flags.
- But I have also refactored `replacePlaceholder` function, that seemed rather difficult to understand/edit at first. I have got tests before I touched anything; all the logic should be equivalent.


vovcacik:
(the failing test should be just a hiccup, I've run the workflows successfully before PR)

PR #2651, opened 2021-10-30
Fix README word order

preslavrachev:


junegunn:
Thanks, but this was a not typo. Unix filter is a common term for describing a certain class of programs. (e.g. https://www.ibm.com/docs/en/zos/2.3.0?topic=program-writing-unix-filter)

PR #2674, opened 2021-11-30
Enable alt+c bash binding with version > 4 syntax.

lucasteligioridis:
# What?

This patch introduces the same backwards compatible logic as the other
bash bindings, I believe the logic for alt+c was perhaps just missed in
a previous patch to use the bash > 4 syntax, which is cleaner and
actually resolves an issue with the terminal bash prompt overlapping in
a strange way / line wrapping. (with vi-mode enabled on bash)

This is mostly a cosmetic fix but also fixes another issue with the fzf
command exiting cleanly when aborted.

See some screenshots showing the issue I was having, too me way too
long to figure out that this was the issue. Full disclosure I'm not sure if
this will cause issues on bash versions < 4 or whether or not there was
something else in my configuration that was conflicting. Ultimately, this
resolved the issue you can see was occurring in the screenshots.

## Broken / malformed prompt

Screen Shot 2021-11-30 at 10 38 29 pm
Screen Shot 2021-11-30 at 10 38 44 pm

## After patch

Screen Shot 2021-11-30 at 10 43 37 pm

### DIFF
--- a/shell/key-bindings.bash
+++ b/shell/key-bindings.bash
@@ -76,6 +76,11 @@ if (( BASH_VERSINFO[0] < 4 )); then
   bind -m emacs-standard '"\C-r": "\C-e \C-u\C-y\ey\C-u"$(__fzf_history__)"\e\C-e\er"'
   bind -m vi-command '"\C-r": "\C-z\C-r\C-z"'
   bind -m vi-insert '"\C-r": "\C-z\C-r\C-z"'
+
+   # ALT-C - cd into the selected directory
+   bind -m emacs-standard '"\ec": " \C-b\C-k \C-u`__fzf_cd__`\e\C-e\er\C-m\C-y\C-h\e \C-y\ey\C-x\C-x\C-d"'
+   bind -m vi-command '"\ec": "\C-z\ec\C-z"'
+   bind -m vi-insert '"\ec": "\C-z\ec\C-z"'
 else
   # CTRL-T - Paste the selected file path into the command line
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks, I'm seeing this issue where my prompt (`$PS1`) is not updated after changing directory. Any idea?

See https://asciinema.org/a/XMIFveAAEmIDlV7LLhyLlk23Y

lucasteligioridis:
Hmmm something i missed perhaps.
Let me take a look again and I'll push up a fix. Leave it with me for the moment.

Thanks for testing @junegunn, thought I checked this fine yesterday and saw that exact same issue. I'll reset my PS1 to something more basic to remove any of my own configuration out of the equation.

lucasteligioridis:
Ah yes I see the same issue, this is what I saw yesterday with my actual `$PS1` not updating. 👍🏼

lucasteligioridis:
On a fair bit of research and testing, I don't think this is actually possible using the `bind -x` parameter, it's really the core of how READLINE works. There are a bunch of "hacky" ways to get this "almost" working, but it's probably not worth even worrying about to be honest and would add a bunch of extra code for not a huge amount of value. Was really just trying to cleanup the way the bash prompt handled the binding like the other commands, I'm super disappointed this can't be easily done in bash.

Ultimately there is no way for the prompt to re-evaluate correctly that I can see as it expects a user input / return carriage. As I see it that isn't possible with `bind -x`, you'd have to fall back onto macros to achieve this. 

Appreciate checking this out and testing anyway ❤️

PR #2678, opened 2021-12-03
Fix fish update fish_user_paths failure

wrrn:
Updating the `fish_user_paths` always failed because there was an errant backslash in the command. This would cause the fish parser to fail.

### DIFF
--- a/install
+++ b/install
@@ -274,7 +274,7 @@ done
 if [[ "$shells" =~ fish ]]; then
   echo -n "Update fish_user_paths ... "
   fish << EOF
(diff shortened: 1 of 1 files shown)


junegunn:
The code works for me.

```
Update fish_user_paths ... OK
```

`\` prefix is not an error. See https://stackoverflow.com/questions/15691977/why-start-a-shell-command-with-a-backslash

wrrn:
After reading that stackoverflow link above it looks like it should be a noop in non-interactive shells. I'm not sure why mine was failing then. Thanks for taking the time respond.

Issue #1440, opened 2018-11-28
--with-nth implies it wont effect searching but it does


i-am-eva:
- Category
    - [x] fzf binary
    - [ ] fzf-tmux script
    - [ ] Key bindings
    - [ ] Completion
    - [ ] Vim
    - [ ] Neovim
    - [ ] Etc.
- OS
    - [ ] Linux
    - [x] Mac OS X
    - [ ] Windows
    - [ ] Windows Subsystem for Linux
    - [ ] Etc.
- Shell
    - [ ] bash
    - [x] zsh
    - [ ] fish

**From the manual**

       -n, --nth=N[,..]
              Comma-separated  list  of  field  index expressions for limiting
              search scope.  See FIELD INDEX EXPRESSION for the details.

       --with-nth=N[,..]
              Transform the  presentation  of  each  line  using  field  index
              expressions

To me, this reads that "`--with-nth` will alter what you *see* in the list, while `--nth` will alter what you *search*. This behaviour doesn't seem to occur though, so either (to my brain) the docs need a bit of a tune up or there is a bug.

**Here is a demo**

```
printf "all_files\nfile_a_only_1\nfile_a_only_2\n" > file-a
printf "all_files\nfile_b_only_1\nfile_b_only_2\nfile_b_and_file_c\n" > file-b
printf "all_files\nfile_c_only_1\nfile_c_only_2\nfile_b_and_file_c\nfile_c_only_3\n" > file-c

ls | xargs -I {} sh -c 'printf "{}:"; cat {} | tr "\n" " "; printf "\n"' | fzf -e --delimiter ":" --nth 1,2 --with-nth 1 --preview="cat {1}" --preview-window="up:40"
```

**What I want to happen**

I want to see a list of filename (`file-a`, `file-b`, `file-c`) and be able to filter them by *content*. I.e: type `file_b_only` and see the list view highlight `file-b` and see `file-b` in the preview window. 

**What does happen**

As soon as I get to `file_`, all results disappear because none of the filenames match, even though the content *is* passed to `fzf`.

If you run the same command but with `--with-nth 1,2`, 

```
ls | xargs -I {} sh -c 'printf "{}:"; cat {} | tr "\n" " "; printf "\n"' |  fzf -e --delimiter ":" --nth 1,2 --with-nth 1,2 --preview="cat {1}" --preview-window="up:40"
```

[...]

Issue #102, opened 2014-10-24
Feature Request: Option to hide some fields


jdevera:
## The request

I would like to somehow be able to dissociate what is shown to what is returned. For example, the output of `git --no-pager log --format=oneline --abbrev-commit -n 5` in this repository

```
6fd6fff [vim] Ignore 'dir' option if empty
d1387bf Use IO.console when possible
4c923a2 [uninstall] Remove both patterns of source command (#97)
4ee85f1 [install] Join line numbers when multiple matches found
829c7f9 Merge branch 'mjwhitta-master'
```

I'd like to be able to show only the commit messages, without the sha, but at the same time, when I choose one, I want the sha to be in the output. So fzf would show me this:

```
[vim] Ignore 'dir' option if empty
Use IO.console when possible
[uninstall] Remove both patterns of source command (#97)
[install] Join line numbers when multiple matches found
Merge branch 'mjwhitta-master'
```

And when I choose the last one, I get this, the full line:

```
829c7f9 Merge branch 'mjwhitta-master'
```
## Possible interface

The interface could, for example, be similar to that for limiting the search scope. Only in this case we either choose what to hide, or what to show.

tagwint:
There's always someone who likes the opposite :)
I mean - to show N fields but only pass one (some) of them (specified by -nth ) to the output.
A kind of -xnth, which would not only specify  the fields to make selection on, but also forced to only take those fields from selection. 
Generalizing the idea, inside an item line i'd like to distinguish fields i can make filtering on from the  fields that are accepted e.g.

John  23 single baseball
Bob 29 married guitar
Suzi 22 single cooking
Mary 30 single baseball
Dan 44 single guitar

I would like select all singles who like baseball but only printout their name and age.
Perhaps there's a way to achieve that w/o implementing a new -xntht option? 
I could not manage that using --with-nth and -nth - they seem to contradict for use case like that

junegunn:
@tagwint 

> Perhaps there's a way to achieve that w/o implementing a new -xntht option? 

You can post-process the output. e.g. awk, sed.

tagwint:
Sure I can, and I also can chain fzf calls. There's nearly always more than one way of doing things :)

junegunn:
@tagwint Yes, and so there's no point of adding the option. It's not like `--nth` or `--with-nth` which cannot be done with pre or post processors.

Bugswriter:
I have this file: 
http://0x0.st/iCwK.pls
[...]

PR #2682, opened 2021-12-08
Reapply PR 1447 to fix with-nth

TamaMcGlinn:
I reapplied [PR1447](https://github.com/junegunn/fzf/pull/1447) (as a single commit) without changing anything. This is just to make it easier for others to try this feature out, since the original author removed his fork and hence the PR's underlying commits.

The context and discussion is [here](https://github.com/junegunn/fzf/issues/102). See also #1440.

### DIFF
--- a/src/core.go
+++ b/src/core.go
@@ -96,6 +96,15 @@ func Run(opts *Options, version string, revision string) {
 			}
 			item.text, item.colors = ansiProcessor(data)
 			item.text.Index = itemIndex
+			// Pattern.go basicMatch and extendedMatch will tokenize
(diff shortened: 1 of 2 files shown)


junegunn:
https://github.com/junegunn/fzf/issues/102#issuecomment-988449994

Tests are failing because of the broken backward compatibility. Not only that, the patch doesn't properly address the 
required changes in highlighting of matched substrings and horizontal scrolling according to it. Feel free to use your fork if you're okay with those deficiencies, but I can't merge this because we don't break backward compatibility.

PR #2691, opened 2021-12-17
fzf-tmux: Executes fzf in env-based shell

shivanshs9:
This allows executing fzf in a user-defined shell, rather than just using the bash. I noticed that both fzf and fzf-tmux has different behavior due to different shell environment (fzf runs in the commandline shell, but fzf-tmux always executes fzf in bash).

This behavior got prominent when using [fzf.fish plugin](https://github.com/PatrickF1/fzf.fish) that had a fish autoloaded function to preview file contents. In fzf it ran succesfully, but in fzf-tmux it threw an error due to missing function in bash.


junegunn:
What is the output of `tmux show -g default-shell`?

Issue #2718, opened 2022-01-29
Why "directory completion (`cd **`)" and "cd key binding (`ALT-C`)" has different finding strategy?


kevin61416:
- [x] I have read through the manual page (`man fzf`)
- [x] I have the latest version of fzf
- [x] I have searched through the existing issues

## Info

- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [ ] bash
    - [ ] zsh
    - [x] fish

## Problem / Steps to reproduce

In **completion.xsh**, the default finding command is:
```
command find -L "$1" \
    -name .git -prune -o -name .hg -prune -o -name .svn -prune -o -type d \
    -a -not -path "$1" -print 2> /dev/null | sed 's@^\./@@'
```
In my opinion, Its purpose is to skip common **version control** directories.

---

On the other hand, the default finding command in **key-bindings.xsh** is:
```
command find -L . -mindepth 1 \\( -path '*/\\.*' -o -fstype 'sysfs' -o -fstype 'devfs' -o -fstype 'devtmpfs' -o -fstype 'proc' \\) -prune \
    -o -type d -print 2> /dev/null | cut -b3-
```
In my opinion, Its purpose is to skip hidden directories and common **virtual file systems** directories.

---
I thought they are quite similar features, but their finding strategy is very different.
What is the reason for this different behavior? Thanks!

junegunn:
Hmm, I don't have a good answer. It doesn't look like there's an inherent reason they should differ. They seemed to have gotten out of sync at some point. Probably an oversight.

This is an old PR that changed *key-bindings*, but not *completion*.

* https://github.com/junegunn/fzf/commit/53d5d9d1624d23175eca4c1e1bd93bbbb7b41b0c

kevin61416:
Thanks for the quick reply and PR information.  😆
Do you think it is worth to sync these two commands? Maybe a new env variable `FZF_FIND_DIR_COMMAND` which has a placeholder `$1` to accept a base search path (default value is '.') 
For compatibility, it can use some expansion syntax like this: `${FZF_FIND_DIR_COMMAND:-${FZF_ALT_C_COMMAND:-"command find -L . ......"}}`.
In this way, it can maintain consistence between these commands and provide users more flexibility to change default command in `_fzf_compgen_dir()`

junegunn:
[...]

PR #2721, opened 2022-02-01
feat: add support for `_fzf_find_dir` function to zsh shell

JS-Zheng:
By introducing a new `_fzf_find_dir` function, fzf can sync the finding strategy between `_fzf_compgen_dir` and `fzf-cd-widget` without  breaking compatibility (#2718).

### DIFF
--- a/shell/completion.zsh
+++ b/shell/completion.zsh
@@ -88,9 +88,13 @@ fi
 
 if ! declare -f _fzf_compgen_dir > /dev/null; then
   _fzf_compgen_dir() {
-    command find -L "$1" \
-      -name .git -prune -o -name .hg -prune -o -name .svn -prune -o -type d \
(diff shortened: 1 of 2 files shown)


junegunn:
Thanks for the suggestion, but I don't think we need another way to configure the behavior. We can simply call _fzf_compgen_dir in FZF_ALT_C_COMMAND.

```sh
_fzf_compgen_dir() {
  echo "$1"
  echo foo
  echo bar
}

export FZF_ALT_C_COMMAND='_fzf_compgen_dir .'
```

JS-Zheng:
Yep, it's a concise way. 👍 
By the way, the reason why I introduce this little redundancy is just for compatibility and semantic.

junegunn:
I see. The problem with having another configuration method is that it can confuse users that have both `$FZF_ALT_C_COMMAND` and `_fzf_find_dir` defined.

PR #2723, opened 2022-02-03
[vim] Let users decide to use --height instead of terminal buffer

liskin:
Commit 85ae745910 forced the use of terminal buffers whenever they're
available to improve consistency and avoid confusion. This, however, has
drawbacks:

* Using a terminal buffer is slightly slower (not noticeable to most
  people but visible to me).

* It breaks certain features of fzf.vim, requiring workarounds like
  https://github.com/junegunn/fzf.vim/pull/1363. And, unfortunately, it
  also breaks https://github.com/junegunn/fzf.vim/pull/1196 (a fix for
  :Snippets in insert-mode which relies on `<Cmd>` and the cursor
  staying behind the end-of-line) which I have no workaround for yet.

This commit adds `g:fzf_prefer_height` to optionally restore the old
behaviour.

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -487,6 +487,7 @@ try
   endif
 
   let prefer_tmux = get(g:, 'fzf_prefer_tmux', 0) || has_key(dict, 'tmux')
+  let prefer_height = get(g:, 'fzf_prefer_height', 0)
   let use_height = has_key(dict, 'down') && !has('gui_running') &&
         \ !(has('nvim') || s:is_win || has('win32unix') || s:present(dict, 'up', 'left', 'right', 'window')) &&
         \ executable('tput') && filereadable('/dev/tty')
@@ -497,6 +498,8 @@ try
   let use_tmux = (has_key(dict, 'tmux') || (!use_height && !use_term || prefer_tmux) && !has('win32unix') && s:splittable(dict)) && s:tmux_enabled()
   if prefer_tmux && use_tmux
     let use_height = 0
+    let use_term = 0
+  elseif prefer_height && use_height
     let use_term = 0
   endif
   if use_term


junegunn:
I'm okay with having this option if we keep it undocumented as we may decide to remove it in the future.

> Using a terminal buffer is slightly slower

Indeed. It is especially so on the popup window. I'm currently using regular Vim, but IIRC, the rendering of the terminal buffer on Neovim was slightly better.

By the way, if you use tmux, I recommend that you set up tmux layout as it's more responsive and flexible. I like tmux popup as it can display across multiple tmux split panes.

```vim
if exists('$TMUX')
  " Or use a pop-up window
  let g:fzf_layout = { 'tmux': '-p90%,60%' }

else
  let g:fzf_layout = { 'window': { 'width': 0.9, 'height': 0.6 } }
endif
```

junegunn:
I'm okay with having this option if we keep it undocumented as we may decide to remove it in the future.

> Using a terminal buffer is slightly slower

Indeed. It is especially so on the popup window. I'm currently using regular Vim, but IIRC, the rendering of the terminal buffer on Neovim was slightly better.

By the way, if you use tmux, I recommend that you set up tmux layout as it's more responsive and flexible. I like tmux popup as it can display across multiple tmux split panes.

```vim
if exists('$TMUX')
  " Using tmux pop-up window
  let g:fzf_layout = { 'tmux': '-p90%,60%' }

  " Or tmux split pane below the current pane
  " let g:fzf_layout = { 'tmux': '-d40%' }
else
  let g:fzf_layout = { 'window': { 'width': 0.9, 'height': 0.6 } }
endif
```

> a fix for :Snippets in insert-mode

I don't use UltiSnips anymore so I'm not super motivated in fixing the command, but I'm well aware it has some issues. I'll see if I can improve it when I get some time.

liskin:
> Indeed. It is especially so on the popup window. I'm currently using regular Vim, but IIRC, the rendering of the terminal buffer on Neovim was slightly better.

Yeah, I'm on regular vim too still.

> By the way, if you use tmux, I recommend that you set up tmux layout as it's more responsive and flexible. I like tmux popup as it can display across multiple tmux split panes.

I use it less often than many tmux users do (with a tiling WM I often don't need it), but I'll try your config snippet as it appears to be fast and less visually disruptive than `--height`. I still need stuff to work outside of tmux, though. :-/

> I don't use UltiSnips anymore so I'm not super motivated in fixing the command, but I'm well aware it has some issues. I'll see if I can improve it when I get some time.

I'm okay with running my own branch but if you ever get to it, I'll be happy to test whatever you come up with or chat/brainstorm or whatever. :-)

---

Thanks!

PR #2727, opened 2022-02-06
[zsh]: bring keybind functionalities on par with Fish

berbiche:
Prior to this commit, the zsh keybindings did not support a prefix search and did not support using an initial directory for commands.

For instance, the following operation was impossible prior to this commit:
Pressing `Ctrl-T` while the cursor was after a directory would not start searching from the given directory, i.e. `> ls src/_` (where underscore denotes the current cursor) would always search for files in '.'.

With this commit, zsh's behavior is now inline with Fish's behavior and will correctly start searching in the existing directory.

The `M-c` (alt-c) behavior was also fixed to be compatible with Fish's behavior.

---

The features implemented in this commit were only tested on `zsh 5.8`.
I do not have enough knowledge about ZSH to assert that this PR will not introduce regressions.

### DIFF
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -38,20 +38,72 @@ fi
 
 [[ -o interactive ]] || return 0
 
-# CTRL-T - Paste the selected file path(s) into the command line
-__fsel() {
-  local cmd="${FZF_CTRL_T_COMMAND:-"command find -L . -mindepth 1 \\( -path '*/\\.*' -o -fstype 'sysfs' -o -fstype 'devfs' -o -fstype 'devtmpfs' -o -fstype 'proc' \\) -prune \
-    -o -type f -print \
-    -o -type d -print \
-    -o -type l -print 2> /dev/null | cut -b3-"}"
-  setopt localoptions pipefail no_aliases 2> /dev/null
-  local item
-  eval "$cmd" | FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} --reverse --bind=ctrl-z:ignore $FZF_DEFAULT_OPTS $FZF_CTRL_T_OPTS" $(__fzfcmd) -m "$@" | while read item; do
-    echo -n "${(q)item} "
-  done
-  local ret=$?
-  echo
-  return $ret
+# Find the longest existing filepath from input string
+# Adapted from the Fish function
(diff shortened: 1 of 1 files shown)


shell/key-bindings.zsh
```diff
@@ -38,29 +38,116 @@ fi
...
+  if [[ "$dir" != "." ]]; then
+    while [[ ! -d "$dir" ]]; do
+      dir=${dir:h}
+    done
+  fi
+  echo "$dir"
+}
+
+__fzf_parse_commandline() {
+  autoload -Uz split-shell-arguments
```
berbiche, line 55: Should this be moved outside of the function, after the `interactive` check?

berbiche:
Found the issue thanks to the tests.
1. The builtin function (`modify-current-argument` used to modify the argument where the user has their cursor on does not work if there is only whitespace.
2. And if there is enough arguments, but the cursor is on whitespace, the function modifies the previous argument instead.

Given the license of ZSH is "MIT-like" I could copy the source of the function and fix the warts.
I'm not sure how to retain the copyright or give proper credit to the authors if I proceed with copying the source.

junegunn:
Have you considered using the fuzzy completion instead (i.e. `ls src/**<tab>` or `cd src/**<tab>`)? CTRL-T was initially designed to be a simple, context-free key binding that should work identically on all three shells, bash, zsh, and fish.

The fish version of it, however, was extended afterward to look at the prefix, because unfortunately, there is no fuzzy completion for fish. 

I'd like to keep these key bindings as simple as possible so that the users can read the source code and easily understand how they are implemented.

RayZ0rr:
I agree with @junegunn . I use ctrl-t in bash and zsh to find files irrespective already present commands in the commandline.

berbiche:
> Have you considered using the fuzzy completion instead (i.e. `ls src/**<tab>` or `cd src/**<tab>`)? CTRL-T was initially designed to be a simple, context-free key binding that should work identically on all three shells, bash, zsh, and fish.

Indeed, I forgot about this feature.

> I'd like to keep these key bindings as simple as possible so that the users can read the source code and easily understand how they are implemented.

Understandable, under these circumstances I will close this PR since the new features introduced diverge from the intended behavior.

Thanks again for the review and the comments! 😄

p.s. for on-lookers, this PR's functionality is incomplete and broken.

PR #2749, opened 2022-03-02
next-selected

mtempling:
The reason for this PR is that we've started to use fzf as a viewer for code traces that are produced using DMCE (https://github.com/PatrikAAberg/dmce). Having the code trace in fzf one would like to select multiple lines (--multi) and have the ability to jump between the selections. Note that we are using the fzf preview window to display the matching source code.

Btw, thx for fzf! It's such a great piece of software.

### DIFF
--- a/man/man1/fzf.1
+++ b/man/man1/fzf.1
@@ -850,10 +850,12 @@ A key or an event can be bound to one or more of the following actions.
     \fBkill-word\fR                  \fIalt-d\fR
     \fBlast\fR                       (move to the last match)
     \fBnext-history\fR               (\fIctrl-n\fR on \fB--history\fR)
+    \fBnext-selected\fR
     \fBpage-down\fR                  \fIpgdn\fR
     \fBpage-up\fR                    \fIpgup\fR
     \fBhalf-page-down\fR
     \fBhalf-page-up\fR
+    \fBprev-selected\fR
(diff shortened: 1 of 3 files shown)


junegunn:
Thanks for the suggestion. While I agree that the suggested feature has valid use cases, as you might have realized, the current data structure is not ideal for implementing it. We might run into performance issues if the number of matches becomes very large. I'll see what we can do when I get some time.

AXGKl:
Looking at the code I think a `previous-selected` would be only a few lines more and it would be awesome to hop back and forth between selections.

AXGKl:
That was fast, many thanks. For me personally this is a milestone for log analysis with fzf. Match error string => select all + clear-query => dig context around occurrences. Lets hope the master accepts :-)

junegunn:
I added `next-selected` and `prev-selected` actions. The implementation is a bit different than yours that `next` always goes to the next selected item below the current line, and `prev` goes to the selected item above the current line, regardless of `--layout`. I believe it will make more sense to most users. Also, it doesn't go through the whole list but breaks when a selected item is found, so it is less likely to make fzf sluggish.

There is a potential for users, I think, to find the "next" and "prev" actions confusing, as they may assume these actions will respect the order in which items were selected. To avoid this confusion, we can consider renaming the actions to `up-selected` and `down-selected` (though I don't like the names).

PR #2777, opened 2022-04-01
[zsh] Replace perl with awk

jirutka:
Unlike awk, which is even defined in POSIX, perl is not pre-installed on all *nix systems. This awk command is functionally equivalent to the original perl command.

~~If extended_history is enabled, the history widget will display:~~ (EDIT: rejected)
~~12  2022-04-02   echo "older command"~~
~~123  today'16:30  echo "today's command"~~

### DIFF
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -97,7 +97,7 @@ bindkey -M viins '\ec' fzf-cd-widget
 fzf-history-widget() {
   local selected num
   setopt localoptions noglobsubst noposixbuiltins pipefail no_aliases 2> /dev/null
-  selected=( $(fc -rl 1 | perl -ne 'print if !$seen{(/^\s*[0-9]+\**\s+(.*)/, $1)}++' |
+  selected=( $(fc -rl 1 | awk '{ cmd=$0; sub(/^\s*[0-9]+\**\s+/, "", cmd); if (!seen[cmd]++) print $0 }' |
     FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} $FZF_DEFAULT_OPTS -n2..,.. --tiebreak=index --bind=ctrl-r:toggle-sort,ctrl-z:ignore $FZF_CTRL_R_OPTS --query=${(qqq)LBUFFER} +m" $(__fzfcmd)) )
   local ret=$?
   if [ -n "$selected" ]; then


ryboe:
This is great! `perl` isn't present on Alpine Linux distributions out of the box. You have to add a [38 MB package](https://pkgs.alpinelinux.org/package/edge/main/x86_64/perl) and a bunch of subpackages. It would be great if I didn't have to install that.

ryboe:
@jirutka That's incredible! That solves my exact problem! Thank you so much for making that happen 🙇

junegunn:
FWIW, awk version and perl version are not identical.

```sh
echo $'3 echo -e "foo\tbar"\n2 echo -e "foo bar"\n1 echo -e "foo   bar"' | awk '{ line=$0; $1=""; if (!seen[$0]++) print line }'

echo $'3 echo -e "foo\tbar"\n2 echo -e "foo bar"\n1 echo -e "foo   bar"' | perl -ne 'print if !$seen{(/^\s*[0-9]+\**\s+(.*)/, $1)}++'
```

jirutka:
> FWIW, awk version and perl version are not identical.

Fixed.

junegunn:
> is it really a bug?

Yes. Printing `foo\tbar` is definitely not the same as printing `foo bar`.

I'm okay with simply replacing perl with awk, but I don't want anything beyond that because my goal here is keeping the three implementations (for bash, zsh, and fish) as similar and simple as possible.

jirutka:
Hm, okay, I removed the second commit.

For the record, the commit was:

```diff
From: Jakub Jirutka <jakub@jirutka.cz>
Date: Sat, 2 Apr 2022 02:18:55 +0200
Subject: [PATCH] [zsh] include date or time in Zsh history if available

If extended_history is enabled, the history widget will display:

     12  2022-04-02   echo "older command"
    123  today'16:30  echo "today's command"
---
 shell/key-bindings.zsh | 19 +++++++++++++++++--
 1 file changed, 17 insertions(+), 2 deletions(-)

diff --git a/shell/key-bindings.zsh b/shell/key-bindings.zsh
index 88a9a9e..d883eeb 100644
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -97,8 +97,23 @@ bindkey -M viins '\ec' fzf-cd-widget
 fzf-history-widget() {
   local selected num
   setopt localoptions noglobsubst noposixbuiltins pipefail no_aliases 2> /dev/null
-  selected=( $(fc -rl 1 | awk '{ cmd=$0; sub(/^\s*[0-9]+\**\s+/, "", cmd); if (!seen[cmd]++) print $0 }' |
-    FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} $FZF_DEFAULT_OPTS -n2..,.. --tiebreak=index --bind=ctrl-r:toggle-sort,ctrl-z:ignore $FZF_CTRL_R_OPTS --query=${(qqq)LBUFFER} +m" $(__fzfcmd)) )
+  local awk_filter='{ cmd=$0; sub(/^\s*[0-9]+\**\s+/, "", cmd); if (!seen[cmd]++) print $0 }'  # filter out duplicates
+  local n=2 fc_opts=''
+  if [[ -o extended_history ]]; then
+    local today=$(date +%Y-%m-%d)
+    # For today's commands, replace date ($2) with "today", otherwise remove time ($3).
+    # And filter out duplicates.
+    awk_filter='{
+      if ($2 == "'$today'") sub($2 " ", "today'\''")
+      else sub($3, "")
+      line=$0; $1=""; $2=""; $3=""
+      if (!seen[$0]++) print line
+    }'
+    fc_opts='-i'
+    n=3
+  fi
+  selected=( $(fc -rl $fc_opts 1 | awk "$awk_filter" |
+    FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} $FZF_DEFAULT_OPTS -n$n..,.. --tiebreak=index --bind=ctrl-r:toggle-sort,ctrl-z:ignore $FZF_CTRL_R_OPTS --query=${(qqq)LBUFFER} +m" $(__fzfcmd)) )
   local ret=$?
   if [ -n "$selected" ]; then
     num=$selected[1]
```

junegunn:
Merged, thanks!

(We're still using perl in the bash script.)

ryboe:
Great work! Thanks very much to both of you!

PR #2784, opened 2022-04-06
Make complex commands slightly more friendly to work with

jan-warchol:
- extract logical parts to separate variables (e.g. $opts)
- put options in $opts in similar order
- move +/-m into $opts (at the end, so they won't be overridden)
- split pipelines into multiple lines
- remove "echo" that seems to be redundant

All this should help with readability and also result in cleaner diffs
when changes are made.

What do you think? Is it helpful?


junegunn:
The patch look good, thanks!

PR #2796, opened 2022-04-18
make usage(--help) neater.

Pandademic:
also:
* clean up atoi error msg.

### DIFF
--- a/src/options.go
+++ b/src/options.go
@@ -16,97 +16,97 @@ import (
 )
 
 const usage = `usage: fzf [options]
+  	OPTIONS:
+	  Search:
+	    -x, --extended        Extended-search mode
+				  (enabled by default; +x or --no-extended to disable)
+	    -e, --exact           Enable Exact-match
+	    --algo=TYPE           Fuzzy matching algorithm: [v1|v2] (default: v2)
+	    -i                    Case-insensitive match (default: smart-case match)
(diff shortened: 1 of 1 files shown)


src/options.go
```diff
@@ -16,97 +16,97 @@ import (
...
+	  Preview
+	    --preview=COMMAND     Command to preview highlighted line ({})
+	    --preview-window=OPT  Preview window layout (default: right:50%)
+				  [up|down|left|right][,SIZE[%]]
+				  [,[no]wrap][,[no]cycle][,[no]follow][,[no]hidden]
+				  [,border-BORDER_OPT]
+				  [,+SCROLL[OFFSETS][/DENOM]][,~HEADER_LINES]
+				  [,default]
+
+	  Scripting:
```
Jiehong, line 94: you're inconsistent in your use of `:` for each section

electron271 (approved):
Looks good, but not sure how much having any instance of `not` be uppercase will help, seems a bit more aggressive.

Pandademic:
@electron271 , I feel it makes it more easy to see the "not" right away , as it stands out more.

Thanks for your review!

Pandademic:
@electron271 , sorry for the late reply.Can you suggest any such options?

electron271:
I dunno, haven't used fzf in a while

PR #2799, opened 2022-04-22
[shell] use builtin cd to avoid conflicts

ajeetdsouza:
Ever since https://github.com/junegunn/fzf/pull/2659 was merged, the ALT-C command completes to `cd -- /selected/path`.  The `--` was added to prevent directories starting with `-` from being treated as options.

However, this also creates problems for people using tools like [zoxide](https://github.com/ajeetdsouza/zoxide) to replace their `cd` command (https://github.com/ajeetdsouza/zoxide/issues/332). A more correct approach would be to use `builtin cd`.


junegunn:
Thanks!

PR #2368, opened 2021-02-23
ansi: speed up parsing by roughly 7.5x

charlievieth:
This commit speeds up the parsing/processing of ANSI escape codes by roughly 7.5x. The speedup is mostly accomplished by replacing the regex with dedicated parsing logic ([`nextAnsiEscapeSequence()`](https://github.com/charlievieth/fzf/blob/3d4a123fa2c7f4daf9dc822fbc3e52214172d106/src/ansi.go#L122-L184)) and reducing the number of allocations in [`extractColor()`](https://github.com/charlievieth/fzf/blob/3d4a123fa2c7f4daf9dc822fbc3e52214172d106/src/ansi.go#L186-L276).

#### Benchmarks
```
name             old time/op    new time/op     delta
ExtractColor-16    4.89µs ± 5%     0.64µs ± 2%   -86.87%  (p=0.000 n=9+9)

name             old speed      new speed       delta
ExtractColor-16  25.6MB/s ± 5%  194.6MB/s ± 2%  +661.43%  (p=0.000 n=9+9)

name             old alloc/op   new alloc/op    delta
ExtractColor-16    1.37kB ± 0%     0.31kB ± 0%   -77.31%  (p=0.000 n=10+10)

name             old allocs/op  new allocs/op   delta
ExtractColor-16      48.0 ± 0%        4.0 ± 0%   -91.67%  (p=0.000 n=10+10)
```

#### TODO
* ANSI performance can be further improved by reducing conversions between `string` and `[]byte`, for example most callers of `extractColor` convert the string response to `[]byte`:
   * [`core.go:72:74`](https://github.com/junegunn/fzf/blob/29851c18aa47ad4b4cbf92358cfbe11935f0c033/src/core.go#L72-L74)
   * [`core.go:79:81`](https://github.com/junegunn/fzf/blob/29851c18aa47ad4b4cbf92358cfbe11935f0c033/src/core.go#L79-L81)
   * [`terminal.go:539:540`](https://github.com/junegunn/fzf/blob/29851c18aa47ad4b4cbf92358cfbe11935f0c033/src/terminal.go#L539-L540)
   * [`terminal.go:1021:1024`](https://github.com/junegunn/fzf/blob/29851c18aa47ad4b4cbf92358cfbe11935f0c033/src/terminal.go#L1021-L1024)

PR #2813, opened 2022-05-07
fzf: enable cpu, mem, block, and mutex profiling

charlievieth:
This commit enables cpu, mem, block, and mutex profling of the FZF
executable. To support flushing the profiles at program exit it adds
util.AtExit to register "at exit" functions and mandates that util.Exit
is used instead of os.Exit to stop the program.


src/util/atexit.go
```diff
@@ -0,0 +1,45 @@
...
+	if fn == nil {
+		panic("AtExit called with nil func")
+	}
+	atExitFuncs.mu.Lock()
+	defer atExitFuncs.mu.Unlock()
+	once := &sync.Once{}
+	atExitFuncs.fns = append(atExitFuncs.fns, func() {
+		once.Do(fn)
+	})
+}
```
junegunn, line 25: Please correct me if I'm wrong, but I don't think we need a mutex here, unless we're exporting the code for other programs, which fzf isn't doing.

charlievieth: Good catch, I removed the mutexes.

src/options.go
```diff
@@ -110,6 +113,11 @@ const usage = `usage: fzf [options]
     FZF_DEFAULT_OPTS      Default options
                           (e.g. '--layout=reverse --inline-info')
 
+  Development
+    --cpuprofile=FILE      Write a CPU profile to the specified file.
+    --memprofile=FILE      Write an allocation profile to the file before exiting.
+    --blockprofile=FILE    Write a goroutine blocking profile to the specified file.
+    --mutexprofile=FILE    Write a mutex contention profile to the specified file.
```
junegunn, line 120: I'd like to not show these options in `--help` option. Because ordinary users don't ever need to know them. Can you add these to the man page instead?

charlievieth: Done

charlievieth:
@junegunn Thank you again for pulling in my fastwalk lib and for helping debug it. 

I saw that you were doing some perf work recently (https://github.com/junegunn/fzf/commit/c30e486b648c1540a660802a5015aeca208274fa, https://github.com/junegunn/fzf/commit/5234c3759adfb68ec7ec5232347ae0d9b4e91243, https://github.com/junegunn/fzf/commit/41b3511ad9a885504a41eb4292fbceeba86e56f0) and was wondering if you'd find this PR useful. No need to pull it in if you don't want to, but I've used it in the past to help look for FZF perf issues (https://github.com/junegunn/fzf/pull/2368).

junegunn:
Thanks for doing this. I would definitely find these options useful. But I'm not sure yet if we should make them available to end users, so I want to keep them undocumented for now.

src/options.go
```diff
@@ -2299,6 +2372,12 @@ func ParseOptions() *Options {
 	errorContext = ""
 	parseOptions(opts, os.Args[1:])
 
+	if err := opts.initProfiling(); err != nil {
+		fmt.Fprintln(os.Stderr, "Error: failed to start pprof profiles:", err)
+		util.Exit(1)
```
junegunn, line 2377: According to the spec, this should be 2.

https://github.com/junegunn/fzf/blob/62963dcefd6d74510ae2692cd0955177bae57543/man/man1/fzf.1#L928-L935

```suggestion
		util.Exit(2)
```

src/options.go
```diff
@@ -140,7 +142,6 @@ const usage = `usage: fzf [options]
     FZF_DEFAULT_OPTS       Default options (e.g. '--layout=reverse --info=inline')
     FZF_DEFAULT_OPTS_FILE  Location of the file to read default options from
     FZF_API_KEY            X-API-Key header for HTTP server (--listen)
```
junegunn, line 144: The extra line at the end was intentional.

```suggestion
    FZF_API_KEY            X-API-Key header for HTTP server (--listen)

```

src/options.go
junegunn: Oh, I think we can instead use `errorExit` function.

src/options.go
```diff
@@ -1974,6 +1979,14 @@ func parseOptions(opts *Options, allArgs []string) {
...
 		case "--version":
 			opts.Version = true
+		case "--cpuprofile":
+			opts.CPUProfile = nextString(allArgs, &i, "file path required: cpu")
+		case "--memprofile":
+			opts.MEMProfile = nextString(allArgs, &i, "file path required: mem")
+		case "--blockprofile":
+			opts.BlockProfile = nextString(allArgs, &i, "file path required: block")
+		case "--mutexprofile":
+			opts.MutexProfile = nextString(allArgs, &i, "file path required: mutex")
```
junegunn, line 1989: How about naming these with the same prefix? i.e. `--profile-cpu`, `--profile-mem`, etc.

charlievieth: Done.

src/options.go
```diff
@@ -2299,6 +2373,12 @@ func ParseOptions() *Options {
 	errorContext = ""
 	parseOptions(opts, os.Args[1:])
 
+	if err := opts.initProfiling(); err != nil {
+		fmt.Fprintln(os.Stderr, "Error: failed to start pprof profiles:", err)
+		util.Exit(exitError)
```
junegunn, line 2378:
```suggestion
		errorExit("failed to start pprof profiles: " + err.Error())
```

junegunn:
Hmm, it seems that the inclusion of pprof package slightly increases the binary size. 3861794 -> 4115682 on my machine. It's not a big difference, but it's an undocumented feature only the developers will use, so...

Would it be possible to enable this only when a certain tag is given? Like we include tcell library only when tcell tag is specifed.

https://github.com/junegunn/fzf/blob/62963dcefd6d74510ae2692cd0955177bae57543/src/tui/tcell.go#L1

```sh
# Profiling options will be no-op without dev tag
make install

# Profiling feature enabled in this binary
TAGS=dev make install
```

charlievieth:
> Would it be possible to enable this only when a certain tag is given? Like we include tcell library only when tcell tag is specifed.

Just pushed https://github.com/junegunn/fzf/commit/1e48c1ec11688df6f57e223df0846e209b892bea which makes profiling a no-op unless the "pprof" build tag and warns the user otherwise.

junegunn:
Thanks, have a nice weekend!

junegunn (approved):
Looks good to me, thanks!

Please let me know if there's anything else you want to address. I'll merge this after your confirmation.

charlievieth:
@junegunn apologies for the delay in getting back to you and thank you for merging this!

PR #2829, opened 2022-05-23
Add zsh completions for fzf flags

jamespeapen:
The fzf executable itself did not have any completions so I've added these based on the man page. Some descriptions have been abbreviated to maintain short lines. 

The colorscheme flags needs further work to get proper completion of all its options working. Currently it only works to complete the base scheme. I'm still learning zsh completion so I wasn't able to figure it out yet.

This script could be installed into the `$fpath` on install or manually moved to a completion directory by the user. I can try an auto install script if this looks good or add the installation steps into the README.

### DIFF
--- a/completions/_fzf
+++ b/completions/_fzf
@@ -1,0 +1,133 @@
+#compdef fzf
+
+_fzf () {
+    _arguments -s -S \
+        --version"[Show version of fzf]" \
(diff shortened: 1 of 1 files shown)


EnginnerDog:
Even my grandma's knitting is more complicated than this PR.

tjex:
@EnginnerDog is that for or against the validity of the PR?

PR #2844, opened 2022-06-09
Hide previewer when screen width < 40

### LINKED ISSUE #2804: [FR] Hide preview window if there is less horizontal space space

## Info

- OS
    - [x] Linux
    - [x] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [ ] bash
    - [x] zsh
    - [ ] fish

fzf currently squeezes the preview window into the frame even if the terminal has been resized to a very small width. It would be great if there was an option to hide the preview window automatically when there is insufficient space.

fivemoreminix:
This is a one-liner solution, that does not take into account the calculated margins, only the true screenWidth.

This is because it would require a refactor of the resizeWindows() function to prevent a cyclic depencency, wherein the previewer's enabled is determined by margins, and the margins size is determined by whether the previewer is enabled.

This hardly affects anything when the padding and margins combined are less than 3 or 4 cells in width.

Fixes #2804

### DIFF
--- a/src/terminal.go
+++ b/src/terminal.go
@@ -801,6 +801,7 @@ func (t *Terminal) resizeWindows() {
 		}
 	}
 
+	t.previewer.enabled = screenWidth > 39 // Hide previewer when window is too narrow
 	previewVisible := t.isPreviewEnabled() && t.previewOpts.size.size > 0
 	minAreaWidth := minWidth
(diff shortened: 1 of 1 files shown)


fivemoreminix:
Just to clarify, what is "hardly affected" is the width of the preview when it is shown.

PR #2852, opened 2022-06-17
Don't export PATH in ~/.fzf.{bash,zsh}

tanishyadav:
There is no use exporting PATH when it is already exported. Moreover, it
causes things like `typeset -U path` in zsh to break if done before
sourcing "~/.fzf.zsh".


junegunn:
Thanks!

PR #2871, opened 2022-07-11
fix `fzf-tmux` to have `fzf`'s completion

jzacsh:
this isn't the _full_ fix but I left an inline comment in case this ever
bugs a future user and they're wondering where the last few flags'
completions are (answer: they're not added yet, as they're just
diminishing return for _this_ particular quick itch-scratching).

### DIFF
--- a/shell/completion.bash
+++ b/shell/completion.bash
@@ -304,6 +304,13 @@ _fzf_alias_completion() {
 
 # fzf options
 complete -o default -F _fzf_opts_completion fzf
+if command -v fzf-tmux >/dev/null; then
+  complete -o default -F _fzf_opts_completion fzf-tmux
+  # fzf-tmux is a thin fzf wrapper that has only a few more options than fzf
+  # istself. As a quick improvement we'll take fzf's completion and adding the
+  # few extra fzf-tmux specific options (like `-w WIDTH`) leave as a future
+  # patch.
+fi
 
 d_cmds="${FZF_COMPLETION_DIR_COMMANDS:-cd pushd rmdir}"
 a_cmds="
--- a/shell/completion.bash
+++ b/shell/completion.bash
@@ -304,13 +304,10 @@ _fzf_alias_completion() {
 
 # fzf options
 complete -o default -F _fzf_opts_completion fzf
-if command -v fzf-tmux >/dev/null; then
-  complete -o default -F _fzf_opts_completion fzf-tmux
-  # fzf-tmux is a thin fzf wrapper that has only a few more options than fzf
-  # istself. As a quick improvement we'll take fzf's completion and adding the
-  # few extra fzf-tmux specific options (like `-w WIDTH`) leave as a future
-  # patch.
-fi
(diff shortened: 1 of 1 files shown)


shell/completion.bash
```diff
@@ -304,6 +304,13 @@ _fzf_alias_completion() {
 
 # fzf options
 complete -o default -F _fzf_opts_completion fzf
+if command -v fzf-tmux >/dev/null; then
+  complete -o default -F _fzf_opts_completion fzf-tmux
+  # fzf-tmux is a thin fzf wrapper that has only a few more options than fzf
+  # istself. As a quick improvement we'll take fzf's completion and adding the
+  # few extra fzf-tmux specific options (like `-w WIDTH`) leave as a future
+  # patch.
+fi
```
junegunn, line 313: I think it's okay to remove the if condtion.

```suggestion
# fzf-tmux is a thin fzf wrapper that has only a few more options than fzf
# itself. As a quick improvement we'll take fzf's completion.
# TODO: Add fzf-tmux specific options (e.g. `-w WIDTH`)
complete -o default -F _fzf_opts_completion fzf-tmux
```

Anyway, please note that the fzf completion hasn't been updated for quite a while, so it's missing a lot of new options.

jzacsh: > I think it's okay to remove the if condition.

Done.

> Anyway, please note that the fzf completion hasn't been updated for quite a while, so it's missing a lot of new options.

Good to know. I'll leave that to some separate, future PR. Perhaps someone will have the energy to automate generation of the options (say for [lists like this](https://github.com/junegunn/fzf/blob/51fdaad002a5ad827bd5ebfac43386592005d02c/shell/completion.bash#L71-111) to not be hardcoded).

jzacsh:
Huh I [see this assertion](https://github.com/junegunn/fzf/blob/d56f605b6338842b415ac1ff578f4316455815fe/test/test_go.rb#L2336) has [failed](https://github.com/junegunn/fzf/runs/7295268082?check_suite_focus=true)

Here are the lines: https://github.com/junegunn/fzf/blob/d56f605b6338842b415ac1ff578f4316455815fe/test/test_go.rb#L2333-L2336

If I'm understanding correctly this test is showing that, within tmux, `ctrl+r` isn't triggering the fzf UI. Not sure how this could've been broken by may change (particularly because I set a completion for `fzf-tmux` on my own machine and that doesn't in any way impact my `ctrl+r` behavior). I'll wait and see what the latest commit's run shows (maybe it was just a flaky test run the first time).

junegunn:
There are some flaky test cases; they usually pass on retry.

jzacsh:
yup! seems to be passing now: https://github.com/junegunn/fzf/runs/7312921952?check_suite_focus=true

jzacsh:
oh, whoops - I mistakenly thought a review was required - sorry about that. Feel free to mark resolved (doesn't look like I can un-request a review) and I'll go ahead and merge.

junegunn (approved):
No problem. Thanks.

PR #2885, opened 2022-07-21
fix: defaultCommand retrieves dotfiles in cwd; document defaultCommand in readme and man

alex-courtis:
The default find command was not printing dotfiles in the current directory.

Added the default to readme and man pages: it was very difficult to fix via `$FZF_DEFAULT_COMMAND` as I did not know what the default was without digging into source.

### DIFF
--- a/src/constants.go
+++ b/src/constants.go
@@ -58,9 +58,9 @@ var defaultCommand string
 
 func init() {
 	if !util.IsWindows() {
-		defaultCommand = `set -o pipefail; command find -L . -mindepth 1 \( -path '*/\.*' -o -fstype 'sysfs' -o -fstype 'devfs' -o -fstype 'devtmpfs' -o -fstype 'proc' \) -prune -o -type f -print -o -type l -print 2> /dev/null | cut -b3-`
+		defaultCommand = `set -o pipefail; command find -L . -mindepth 1 \( -path '*/\.*' -o -fstype 'sysfs' -o -fstype 'devfs' -o -fstype 'devtmpfs' -o -fstype 'proc' \) -prune -print -o -type f -print -o -type l -print 2> /dev/null | cut -b3-`
 	} else if os.Getenv("TERM") == "cygwin" {
-		defaultCommand = `sh -c "command find -L . -mindepth 1 -path '*/\.*' -prune -o -type f -print -o -type l -print 2> /dev/null | cut -b3-"`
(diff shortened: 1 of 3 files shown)


alex-courtis:
Tested on linux zsh but not windows.

src/constants.go
```diff
@@ -57,10 +57,11 @@ const (
 var defaultCommand string
 
 func init() {
+	// please update README.md and man/man1/fzf.1 when changig defaultCommand
 	if !util.IsWindows() {
-		defaultCommand = `set -o pipefail; command find -L . -mindepth 1 \( -path '*/\.*' -o -fstype 'sysfs' -o -fstype 'devfs' -o -fstype 'devtmpfs' -o -fstype 'proc' \) -prune -o -type f -print -o -type l -print 2> /dev/null | cut -b3-`
+		defaultCommand = `set -o pipefail; command find -L . -mindepth 1 \( -path '*/\.*' -o -fstype 'sysfs' -o -fstype 'devfs' -o -fstype 'devtmpfs' -o -fstype 'proc' \) -prune -print -o -type f -print -o -type l -print 2> /dev/null | cut -b3-`
```
nkouevda, line 62: this indeed prints dotfiles, but 1) it also prints dotdirs, and 2) it still prunes dotdirs, so files in them are _not_ printed

e.g. in a git repo, this will print `.git` (which shouldn't happen) and it won't print `.git/HEAD` (which should be printed)

instead of adding `-print` after `-prune`, i believe you want to remove this part: `-path '*/\.*' -o` (or for the windows case, the whole `-path '*/\.*' -prune -o`)

then dotfiles will be printed like all other files (via `-type f -print`), and dotdirs will be neither printed nor pruned

alex-courtis: Thanks for looking. Printing dotdirs but not their contents is indeed incorrect/nonsense behaviour.

This works as described:
```
set -o pipefail; command find -L . -mindepth 1 \( -fstype 'sysfs' -o -fstype 'devfs' -o -fstype 'devtmpfs' -o -fstype 'proc' \) -prune -o -type f -print -o -type l -print 2> /dev/null | cut -b3-"
```

The only issue is that it looks inside dotdirs. That may be undesirable from a performance/correctness point of view.

e.g. in my $HOME with the default command, it finds 1.3 million or so files, very quickly. With dotdir searching, it finds 57 million after many minutes.

`.cache` and `.local` are the problematic ones. In my case, steam lives under `.local`

Perhaps we might present the dotdirs version in the man page as an alternative. @junegunn what are your thoughts?

junegunn:
See #3649

alex-courtis:
> See #3649

Many thanks!

PR #2889, opened 2022-07-23
Add option to force 24 bit colors

Kamholtz:
Add an option `g:fzf_force_24_bit_colors` which would allow windows vim users to force the use of 24 bit colours. Currently there is a condition in `s:get_colors` which prevents this from happening,  and the issue here https://github.com/junegunn/fzf.vim/issues/1152 on fzf.vim seems to indicate a change has not been made because it breaks compatibility with `cmd.exe`. This would allow users to just force 24 bit colours to get around the poor contrast issues that occur when fzf falls back onto 256 ANSI colors.

### DIFF
--- a/doc/fzf.txt
+++ b/doc/fzf.txt
@@ -130,6 +130,8 @@ Note that the environment variables `FZF_DEFAULT_COMMAND` and
    - Customizes fzf colors to match the current color scheme
  - `g:fzf_history_dir`
    - Enables history feature
+ - `g:fzf_force_24_bit_colors`
+   - Force use of 24 bit colors
 
 
 Examples~
--- a/doc/fzf.txt
+++ b/doc/fzf.txt
@@ -130,8 +130,6 @@ Note that the environment variables `FZF_DEFAULT_COMMAND` and
    - Customizes fzf colors to match the current color scheme
  - `g:fzf_history_dir`
    - Enables history feature
- - `g:fzf_force_24_bit_colors`
-   - Force use of 24 bit colors
 
 
 Examples~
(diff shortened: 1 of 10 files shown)


doc/fzf.txt
```diff
@@ -130,6 +130,8 @@ Note that the environment variables `FZF_DEFAULT_COMMAND` and
    - Customizes fzf colors to match the current color scheme
  - `g:fzf_history_dir`
    - Enables history feature
+ - `g:fzf_force_24_bit_colors`
+   - Force use of 24 bit colors
```
junegunn, line 134: So if this is a temporary workaround that we're eventually going to remove, I'll leave this option undocumented.

```suggestion
```

plugin/fzf.vim
```diff
@@ -342,7 +342,8 @@ function! s:common_sink(action, lines) abort
 endfunction
 
 function! s:get_color(attr, ...)
-  let gui = !s:is_win && !has('win32unix') && has('termguicolors') && &termguicolors
+  " Force 24 bit colors: g:fzf_force_24_bit_colors
+  let gui = (exists('g:fzf_force_24_bit_colors') && g:fzf_force_24_bit_colors) || (!s:is_win && !has('win32unix') && has('termguicolors') && &termguicolors)
```
junegunn, line 346: I would reuse the name of an existing Vim option, rather than coin a new name.

```suggestion
  " Force 24 bit colors: g:fzf_force_termguicolors (temporary workaround for https://github.com/junegunn/fzf.vim/issues/1152)
  let gui = get(g:, 'fzf_force_termguicolors', 0) || (!s:is_win && !has('win32unix') && has('termguicolors') && &termguicolors)
```

junegunn (changes requested):

Kamholtz:
Those changes have been added and I have confirmed that the branch still works as expected.

junegunn:
Merged, thanks!

Issue #2583, opened 2021-08-13
Conflict with bash-completion: issue with alias


f3fora:
- [x] I have read through the manual page (`man fzf`)
- [x] I have the latest version of fzf
- [x] I have searched through the existing issues

## Info

- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [x] bash
    - [ ] zsh
    - [ ] fish
    
## Problem / Steps to reproduce

I  use also `bash-completion`.

Minimal `.bashrc`
```
alias rm='echo "r u sure?"'  # or another random alias

source /usr/share/fzf/key-bindings.bash
source /usr/share/fzf/completion.bash
```

This works:
```
rm <TAB>
```

This doesn't:
```
\rm <TAB>
```
and raises `bash: warning: programmable_completion: \rm: possible retry loop`

Maybe related to #1227

AKSoo:
Duplicate of #2474. Not sure of the owner's opinion, but it seems to be an issue with bash-completion.

PR #2891, opened 2022-07-24
fixes #2474 bash completion: possible retry loop

### LINKED ISSUE #2474: bash programmable_completion: possible retry loop

## Info

- OS
    - [ ] Linux
    - [x] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [x] bash
    - [ ] zsh
    - [ ] fish
    
I have the following alias in my .bash_profile because I like the verbose output _most_ of the time: `alias rm='rm -v'`. However, sometimes I want to skip all that output - deleting a `node_modules` directory, for example. So, I'll start my command with a backslash to skip the alias, like: `\rm -drf node_modules`.

I recently upgraded to fzf 0.27 (I'm not sure what version I was on previously - 0.24 maybe?) and started getting `-bash: warning: programmable_completion: \rm: possible retry loop` if I try to tab-complete something like `\rm -drf nod<tab>`. If I disable fzf's bash completion, I don't get this error. I think I tracked it down to [this line](https://github.com/junegunn/fzf/blob/0.27.0/shell/completion.bash#L164): if I change it to the following, it works, but I'm not sure if that's the best solution:

```bash
  cmd="${COMP_WORDS[0]}"
  if [[ $cmd == \\* ]]; then
    cmd="${cmd:1}"
  fi
  cmd="${cmd//[^A-Za-z0-9_=]/_}"
```

Happy to open a PR if it looks good, though!

In addition to fzf's completion, I also have https://github.com/scop/bash-completion installed. If I `set -x` before attempting the completion, I can see that it seems to get stuck in a loop calling into fzf's completion, then scop's, then fzf's, etc, so, maybe it's some interaction between the two? Scop's completions don't have anything defined explicitly for `rm` as far as I can tell.

bmatcuk:
I opened #2474 more than a year ago and came to the conclusion that it was more [bash-completion's](https://github.com/scop/bash-completion) fault than fzf's. However, it does not appear it's an easy fix in bash-completion, and it's a fairly easy fix in fzf. It also appears other people are running into it (ex: #2583) - I, myself, just ran into it again on a new machine - so maybe we should just fix it?

Open to other suggestions. This PR is my proposal from the ticket long ago =)

### DIFF
--- a/shell/completion.bash
+++ b/shell/completion.bash
@@ -161,7 +161,11 @@ _fzf_handle_dynamic_completion() {
 
 __fzf_generic_path_completion() {
   local cur base dir leftover matches trigger cmd
-  cmd="${COMP_WORDS[0]//[^A-Za-z0-9_=]/_}"
+  cmd="${COMP_WORDS[0]}"
+  if [[ $cmd == \\* ]]; then
+    cmd="${cmd:1}"
+  fi
+  cmd="${cmd//[^A-Za-z0-9_=]/_}"
   COMPREPLY=()
   trigger=${FZF_COMPLETION_TRIGGER-'**'}
   cur="${COMP_WORDS[COMP_CWORD]}"


junegunn:
Merged, thanks!

PR #2906, opened 2022-08-02
[zsh] Fix the problem command history deduplication not working on macOS

kissge:
Currently `fzf-history-widget` simply shows duplicated candidates, because unlike GNU awk `\s` is not supported by Apple's awk.
This PR fixes this issue by modifying regex so that it works with both GNU and macOS awk.

```sh-session
$ echo '123  hello' | awk '{ sub(/^\s*[0-9]+\s+/, "", $0); print $0 }'
123  hello

$ echo '123  hello' | awk '{ sub(/^[ \t]*[0-9]+[ \t]+/, "", $0); print $0 }'
hello

$ awk --version
awk version 20200816
```


junegunn:
Merged, thanks! This was a regression caused by #2777.

PR #2908, opened 2022-08-06
fzf-tmux: disable tmux popup window border if fzf has already enabled its border

ytian81:
Before this PR, inside tmux, `fzf-tmux -p` will show double border if fzf has already enabled its border

image

After this PR, we will check if we've enabled `--border` in `FZF_DEFAULT_OPTS`. If so, we can turn off tmux popup window border 

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@ CHANGELOG
 0.32.1
 ------
 - Fixed incorrect ordering of `--tiebreak=chunk`
+- fzf-tmux will show fzf border instead of tmux popup border (requires tmux 3.3)
+  ```sh
+  fzf-tmux -p70%
+  fzf-tmux -p70% --color=border:bright-red
+  fzf-tmux -p100%,60% --color=border:bright-yellow --border=horizontal
+  fzf-tmux -p70%,100% --color=border:bright-green --border=vertical
+  ```
 
 0.32.0
 ------
--- a/bin/fzf-tmux
+++ b/bin/fzf-tmux
@@ -213,6 +213,11 @@ if [[ "$opt" =~ "-K -E" ]]; then
     opt="${opt} -R"
   fi
 
+  # disable tmux pop-up window border if fzf has already enabled border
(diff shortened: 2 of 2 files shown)


junegunn:
Thanks, I wasn't aware of `-B` option as I was using tmux 3.2.

I think we should use this option by default, regardless of `$FZF_DEFAULT_OPTS` or the arguments given to `fzf-tmux`. This is what I'm suggesting. What do you think?

```diff
diff --git a/bin/fzf-tmux b/bin/fzf-tmux
index 6a18cf8..fc864c2 100755
--- a/bin/fzf-tmux
+++ b/bin/fzf-tmux
@@ -136,7 +136,7 @@ if [[ -z "$TMUX" ]]; then
 fi
 
 # --height option is not allowed. CTRL-Z is also disabled.
-args=("${args[@]}" "--no-height" "--bind=ctrl-z:ignore")
+args=("--border" "${args[@]}" "--no-height" "--bind=ctrl-z:ignore")
 
 # Handle zoomed tmux pane without popup options by moving it to a temp window
 if [[ ! "$opt" =~ "-K -E" ]] && tmux list-panes -F '#F' | grep -q Z; then
@@ -213,7 +213,7 @@ if [[ "$opt" =~ "-K -E" ]]; then
     opt="${opt} -R"
   fi
 
-  tmux popup -d "$PWD" "${tmux_args[@]}" $opt "bash $argsf" > /dev/null 2>&1
+  tmux popup -d "$PWD" "${tmux_args[@]}" -B $opt "bash $argsf" > /dev/null 2>&1
   exit $?
 fi
 
```

ytian81:
That's what I have locally! I indeed prefer using fzf border instead of tmux popup window border!

junegunn:
One concern is that tmux 3.3 is relatively new at this point and many users might be still using 3.2 which doesn't support `-B`. We could check the version by running `tmux -V`, but it will make fzf-tmux slightly slower to start.

junegunn:
Merged with the version check, thanks!

PR #2925, opened 2022-08-13
silence awk: cmd. line:1: warning

colicarr:
When using Includes with ssh/config
esp:
---config--- 
Include config.d/*
Include config.d/**/*
---
ssh host<tab>
causes a warning and obfuscates the completions.
awk: cmd. line:1: warning: command line argument `/Users/yourHome/.ssh/config.d/SetOfHosts' is a directory: skipped

AWK has a long history of having strong feelings against directories, going back to the POSIX version, and would terminate.
This is an order of magnitude along that of clowns' dislike of mimes but with age awk's strong feelings toward directories "mellowed" into warnings (gawk).

Sometimes we just need bash completions without the editorial opinions from awk and this appears to be the best move to silence a AWK.


junegunn:
Do you still have the issue with the latest version of the code?

PR #2952, opened 2022-09-01
Fix bash completion var or aliase containing newlines

knutze:
This PR fixes the following minor problem.

In bash completion, if an environment variable contains a newline, the completion list contains the value (not the name) of the variable. The same problem exists with aliases.

Ex)

```bash
$ export AAA="
aaa_line1
aaa_line2
"

$ unset **<TAB>
> < 91/91 (0)
> AAA
  aaa_line1
  aaa_line2
  "
  ...
```


junegunn (approved):

junegunn:
Merged, thanks!

PR #2972, opened 2022-09-19
FZF command run fails with "cmd.exe: Bad address" on Cygwin with TERM=xterm-256color

sankl:
If I run `:FZF` in vim under Cygwin in mintty with TERM=xterm-256color, the separate launched window quickly closes. Debugging shows that launched process terminates due to sh.exe reporting `cmd.exe: Bad address`. 
This pull request fixes the problem by using full path to cmd.exe.


sankl:
It seems to be working fine now, closing. Thanks!

PR #2991, opened 2022-10-03
doc(man): add a hint about which UI element is the finder info

Delapouite:
Hi

While reading the description of the `--info` flag, it's not immediately obvious that the "finder info" is in fact the UI element representing the match counters.

Thanks


junegunn:
Thanks!

Issue #2992, opened 2022-10-07
0.34.0 breaks for Vim on Windows


AceofSpades5757:
- [ ] I have read through the manual page (`man fzf`)
- [x] I have the latest version of fzf
- [x] I have searched through the existing issues

## Info

- OS
    - [ ] Linux
    - [ ] Mac OS X
    - [x] Windows
    - [ ] Etc.
- Shell
    - [ ] bash
    - [ ] zsh
    - [ ] fish
    - [x] cmd.exe

## Problem / Steps to reproduce

fzf 0.34.0 doesn't work on Vim 8.2 or 9.

1. Install Vim 8.2 or 9
2. Add `fzf.vim` and `fzf`. You can leave the latest version for the moment.
3. Run `fzf/install/install.ps1`
    * Using vim-plug: `powershell.exe -ExecutionPolicy Bypass -file "$Env:USERPROFILE\vimfiles\plugged\fzf\install.ps1"` 
    * I get errors using `fzf#install()`. Probably due to execution policy issues.
5. Try to use any of the Vim commands associated with `fzf.vim`. e.g. `:Commands`

A new buffer briefly pops up, but no errors pop up. All related Vim command quietly fail.

Following the same steps, but using version 0.33.0 works as expected.

AceofSpades5757:
Yes, `0.34.0 (04d0b02)` located at `$HOME\.fzf\bin\fzf.exe` on path.

It looks like Vim commands will only mostly work when `0.34.0` is installed on the system, but the `fzf` plugin is set to tag `0.33.0`.

junegunn:
Can you check the output of the following commands in Vim?

* `:echo fzf#exec()`
* `:echo system(fzf#exec() .. ' --version')`

> 2\. Add `fzf.vim` and `fzf`.

So you're using vim-plug, what are the exact `Plug` statements you use for them?

AceofSpades5757:
`:echo fzf#exec()`

* `C:\Users\Kyle L. Davis\.vim\plugged\fzf/bin/fzf`
* If I remove that bin directory, it'll use the `fzf` on `PATH` and simply out `fzf`.

`:echo system(fzf#exec() .. ' --version')`

* `'C:\Users\Kyle' is not recognized as an internal or external command, operable program or batch file.`

`echo system(shellescape(fzf#exec()) .. ' --version')` (to fix the space issue)

* `0.34.0 (04d0b02)`

**Plug Statements**

```vimL
[...]

PR #2993, opened 2022-10-07
Update fzf#install to handle spaces on Windows

AceofSpades5757:
Didn't see any issues addressing this problem, or any other pull requests. Should fix problems for Windows users that have a space in the plugin path.

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -143,7 +143,7 @@ function! fzf#install()
     if !filereadable(script)
       throw script.' not found'
     endif
-    let script = 'powershell -ExecutionPolicy Bypass -file ' . script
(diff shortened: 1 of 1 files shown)


junegunn:
So this will fix #2992, right?

AceofSpades5757:
No, I'm still not sure what's causing the issue with #2992. For the moment, I'm avoiding using `fzf#install()` since using a manual install (`~\.fzf\bin\fzf.exe`) works just fine. This was a separate issue I noticed for a while but didn't have time to investigate. I have a space in my home directory path, so this fixes that for myself and anyone else with a similar issue.

junegunn:
I see, thanks for the explanation. The patch is merged, thanks!

PR #2999, opened 2022-10-12
Fix unbound variables in shell completion

cevhyruz:
when `set -o nounset` is enabled
completion complains about these variables being unset:

`orig_var`
`__fzf_no_space_commands`

### DIFF
--- a/shell/completion.bash
+++ b/shell/completion.bash
@@ -37,7 +37,7 @@ bind '"\e[0n": redraw-current-line' 2> /dev/null
 __fzf_comprun() {
   if [[ "$(type -t _fzf_comprun 2>&1)" = function ]]; then
     _fzf_comprun "$@"
-  elif [[ -n "$TMUX_PANE" ]] && { [[ "${FZF_TMUX:-0}" != 0 ]] || [[ -n "${FZF_TMUX_OPTS-}" ]]; }; then
+  elif [[ -n "${TMUX_PANE-}" ]] && { [[ "${FZF_TMUX:-0}" != 0 ]] || [[ -n "${FZF_TMUX_OPTS-}" ]]; }; then
     shift
     fzf-tmux ${FZF_TMUX_OPTS:--d${FZF_TMUX_HEIGHT:-40%}} -- "$@"
   else
@@ -180,7 +180,7 @@ __fzf_generic_path_completion() {
         leftover=${leftover/#\/}
         [[ -z "${dir-}" ]] && dir='.'
         [[ "${dir-}" != "/" ]] && dir="${dir/%\//}"
-        matches=$(eval "$1 $(printf %q "${dir-}")" | FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} --reverse --bind=ctrl-z:ignore $FZF_DEFAULT_OPTS ${FZF_COMPLETION_OPTS-} $2" __fzf_comprun "$4" -q "$leftover" | while read -r item; do
+        matches=$(eval "$1 $(printf %q "${dir-}")" | FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} --reverse --bind=ctrl-z:ignore ${FZF_DEFAULT_OPTS-} ${FZF_COMPLETION_OPTS-} $2" __fzf_comprun "$4" -q "$leftover" | while read -r item; do
           printf "%q " "${item%$3}$3"
         done)
         matches=${matches% }
@@ -236,7 +236,7 @@ _fzf_complete() {
   if [[ "$cur" == *"$trigger" ]]; then
     cur=${cur:0:${#cur}-${#trigger}}
 
-    selected=$(FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} --reverse --bind=ctrl-z:ignore $FZF_DEFAULT_OPTS ${FZF_COMPLETION_OPTS-} $str_arg" __fzf_comprun "${rest[0]}" "${args[@]}" -q "$cur" | $post | tr '\n' ' ')
+    selected=$(FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} --reverse --bind=ctrl-z:ignore ${FZF_DEFAULT_OPTS-} ${FZF_COMPLETION_OPTS-} $str_arg" __fzf_comprun "${rest[0]}" "${args[@]}" -q "$cur" | $post | tr '\n' ' ')
     selected=${selected% } # Strip trailing space not to repeat "-o nospace"
     if [[ -n "$selected" ]]; then
       COMPREPLY=("$selected")
--- a/shell/completion.bash
+++ b/shell/completion.bash
@@ -139,17 +139,17 @@ _fzf_handle_dynamic_completion() {
   shift
   orig_cmd="$1"
   orig_var="_fzf_orig_completion_$cmd"
-  orig="${!orig_var##*#}"
+  orig="${!orig_var-##*#}"
   if [[ -n "$orig" ]] && type "$orig" > /dev/null 2>&1; then
     $orig "$@"
-  elif [[ -n "$_fzf_completion_loader" ]]; then
+  elif [[ -n "${_fzf_completion_loader-}" ]]; then
     orig_complete=$(complete -p "$orig_cmd" 2> /dev/null)
     _completion_loader "$@"
     ret=$?
     # _completion_loader may not have updated completion for the command
     if [[ "$(complete -p "$orig_cmd" 2> /dev/null)" != "$orig_complete" ]]; then
       __fzf_orig_completion < <(complete -p "$orig_cmd" 2> /dev/null)
-      if [[ "$__fzf_nospace_commands" = *" $orig_cmd "* ]]; then
+      if [[ "${__fzf_nospace_commands-}" = *" $orig_cmd "* ]]; then
         eval "${orig_complete/ -F / -o nospace -F }"
       else
         eval "$orig_complete"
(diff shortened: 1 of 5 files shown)


junegunn:
Thanks, but are you setting `set -o nounset` in an interactive session? Wouldn't it cause a lot of problems with existing scripts? Also, if we apply this patch, bash will successfully load `completion.bash` file, but the completion still won't properly.

```
$ set -o nounset
$ . shell/completion.bash
$ ssh **-bash: FZF_DEFAULT_OPTS: unbound variable
$ ls **-bash: dir: unbound variable
```

cevhyruz:
> are you setting set -o nounset in an interactive session?
Yeah, I like it when I'm able to see if a variable is not created yet, rather than printing nothing when echoing.

> Wouldn't it cause a lot of problems with existing scripts?
what do you mean by existing scripts?

junegunn:
> what do you mean by existing scripts?

Many scripts in the wild are written with the assumption that `set -u` is not set in an interactive session, so setting it is going to break them. You're lucky if you haven't run into such a script.

> This feature is rather controversial, and has many extremely vocal advocates and opponents
> (https://mywiki.wooledge.org/BashFAQ/112)

* https://github.com/gentoo/gentoo-functions/pull/1
* https://github.com/bmc/lib-sh/issues/2
* https://github.com/rupa/z/pull/299
* https://github.com/magicmonty/bash-git-prompt/issues/389
* https://github.com/akesterson/cmdarg/pull/24
* https://github.com/rvm/rvm/issues/4342

Making sure that a legacy script works with `set -u` can be a lot of work and it seems that the maintainers of the scripts are not so sure if it's worth it given that most users historically don't have `set -u` on interactive sessions.

But anyway, as I said above, the patch alone won't make fuzzy completion work (i.e. `ls **<tab>`)? Have you tested that?

cevhyruz:
> Making sure that a legacy script works with set -u can be a lot of work and it seems that the maintainers of the  scripts are not so sure if it's worth it given that most users historically don't have set -u on interactive sessions

I see the point. Well, It wont hurt to support both cases is it? 

The only thing to lookout out for sure is maintaining the way it is which can be solve by adding a little test script.

> But anyway, as I said above, the patch alone won't make fuzzy completion work (i.e. ls **<tab>)? Have you tested that?

Got it working on my end now there seems a couple of little refactoring needed. Will push later after testing. 

Just wanna know what your side in this?

junegunn:
> Just wanna know what are your thoughts about this?

Not just the fuzzy completion, none of the key bindings (CTRL-T, ALT-C, and CTRL-R) works with `set -u` and you're the first one to report this compatibility issue says a lot about common users' configuration. Also, zsh implementation has the same problem, but no complaints so far.

> Well, It wont hurt to support both cases is it?

The code will have things like `${XXX-}` all over the place, they'll likely make the code less readable, and harder to maintain as we always need to put extra care whenever we change the code.

But after all, if we can make things (key bindings and fuzzy completion of bash and zsh) work with `set -u` with not too many intrusive changes, I'm okay with the direction.

shell/completion.bash
```diff
@@ -175,16 +175,16 @@ __fzf_generic_path_completion() {
...
     while true; do
-      if [[ -z "$dir" ]] || [[ -d "$dir" ]]; then
-        leftover=${base/#"$dir"}
+      if [[ -z "${dir-}" ]] || [[ -d "${dir-}" ]]; then
+        leftover=${base/#"${dir-}"}
         leftover=${leftover/#\/}
-        [[ -z "$dir" ]] && dir='.'
-        [[ "$dir" != "/" ]] && dir="${dir/%\//}"
-        matches=$(eval "$1 $(printf %q "$dir")" | FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} --reverse --bind=ctrl-z:ignore $FZF_DEFAULT_OPTS $FZF_COMPLETION_OPTS $2" __fzf_comprun "$4" -q "$leftover" | while read -r item; do
+        [[ -z "${dir-}" ]] && dir='.'
```
junegunn, line 181: Do we need to write `${dir-}` after this line? `$dir` is guaranteed to be defined.

shell/completion.bash
```diff
@@ -175,16 +175,16 @@ __fzf_generic_path_completion() {
...
+        [[ "${dir-}" != "/" ]] && dir="${dir/%\//}"
+        matches=$(eval "$1 $(printf %q "${dir-}")" | FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} --reverse --bind=ctrl-z:ignore $FZF_DEFAULT_OPTS ${FZF_COMPLETION_OPTS-} $2" __fzf_comprun "$4" -q "$leftover" | while read -r item; do
```
junegunn, line 183: `$FZF_DEFAULT_OPTS` is not handled.

shell/key-bindings.zsh
```diff
@@ -55,7 +55,7 @@ __fsel() {
 }
 
 __fzfcmd() {
-  [ -n "$TMUX_PANE" ] && { [ "${FZF_TMUX:-0}" != 0 ] || [ -n "$FZF_TMUX_OPTS" ]; } &&
+  [ -n "$TMUX_PANE" ] && { [ "${FZF_TMUX:-0}" != 0 ] || [ -n "${FZF_TMUX_OPTS-}" ]; } &&
```
junegunn, line 58: `$TMUX_PANE`

shell/key-bindings.bash
```diff
@@ -19,7 +19,7 @@ __fzf_select__() {
     -o -type f -print \
     -o -type d -print \
     -o -type l -print 2> /dev/null | cut -b3-"}"
-  opts="--height ${FZF_TMUX_HEIGHT:-40%} --bind=ctrl-z:ignore --reverse $FZF_DEFAULT_OPTS $FZF_CTRL_T_OPTS -m"
+  opts="--height ${FZF_TMUX_HEIGHT:-40%} --bind=ctrl-z:ignore --reverse $FZF_DEFAULT_OPTS ${FZF_CTRL_T_OPTS-} -m"
```
junegunn, line 22:
```suggestion
  opts="--height ${FZF_TMUX_HEIGHT:-40%} --bind=ctrl-z:ignore --reverse ${FZF_DEFAULT_OPTS-} ${FZF_CTRL_T_OPTS-} -m"
```

shell/key-bindings.bash
```diff
@@ -30,7 +30,7 @@ __fzf_select__() {
 if [[ $- =~ i ]]; then
 
 __fzfcmd() {
-  [[ -n "$TMUX_PANE" ]] && { [[ "${FZF_TMUX:-0}" != 0 ]] || [[ -n "$FZF_TMUX_OPTS" ]]; } &&
+  [[ -n "$TMUX_PANE" ]] && { [[ "${FZF_TMUX:-0}" != 0 ]] || [[ -n "${FZF_TMUX_OPTS-}" ]]; } &&
```
junegunn, line 33:
```suggestion
  [[ -n "${TMUX_PANE-}" ]] && { [[ "${FZF_TMUX:-0}" != 0 ]] || [[ -n "${FZF_TMUX_OPTS-}" ]]; } &&
```

shell/completion.bash
cevhyruz: bash complains about `dir` being unbound when using completion like `ls **`

junegunn: I mean **after this line**. I've tested and bash doesn't complain.

However, it's much simpler to just add `dir=` line before line 176 and just referencing it as `$dir` as before.

junegunn:
```diff
diff --git a/shell/completion.bash b/shell/completion.bash
index ece458e..364c3b4 100644
--- a/shell/completion.bash
+++ b/shell/completion.bash
@@ -173,13 +173,14 @@ __fzf_generic_path_completion() {
     base=${cur:0:${#cur}-${#trigger}}
     eval "base=$base"
 
+    dir=
     [[ $base = *"/"* ]] && dir="$base"
     while true; do
-      if [[ -z "${dir-}" ]] || [[ -d "${dir-}" ]]; then
-        leftover=${base/#"${dir-}"}
+      if [[ -z "$dir" ]] || [[ -d "$dir" ]]; then
+        leftover=${base/#"$dir"}
         leftover=${leftover/#\/}
-        [[ -z "${dir-}" ]] && dir='.'
-        [[ "${dir-}" != "/" ]] && dir="${dir/%\//}"
+        [[ -z "$dir" ]] && dir='.'
+        [[ "$dir" != "/" ]] && dir="${dir/%\//}"
         matches=$(eval "$1 $(printf %q "${dir-}")" | FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} --reverse --bind=ctrl-z:ignore ${FZF_DEFAULT_OPTS-} ${FZF_COMPLETION_OPTS-} $2" __fzf_comprun "$4" -q "$leftover" | while read -r item; do
           printf "%q " "${item%$3}$3"
         done)
@@ -193,8 +194,8 @@ __fzf_generic_path_completion() {
         printf '\e[5n'
         return 0
       fi
-      dir=$(dirname "${dir-}")
-      [[ "${dir-}" =~ /$ ]] || dir="${dir-}"/
+      dir=$(dirname "$dir")
+      [[ "$dir" =~ /$ ]] || dir="$dir"/
     done
   else
     shift
```

cevhyruz: ah! yes we this wont be needed if we instantiate an empty var.

shell/completion.bash
```diff
@@ -139,17 +139,17 @@ _fzf_handle_dynamic_completion() {
   shift
   orig_cmd="$1"
   orig_var="_fzf_orig_completion_$cmd"
-  orig="${!orig_var##*#}"
+  orig="${!orig_var-##*#}"
```
junegunn, line 142: Turns out this has a bug and I'm getting `bash: warning: programmable_completion: ls: possible retry loop`.

```sh
foo='complete -F %s ls #_longopt'
orig_var=foo

echo "${!orig_var##*#}"
  # GOOD: _longopt

echo "${!orig_var-##*#}"
  # BAD: complete -F %s ls #_longopt

# Workaround: split the expression into two
orig="${!orig_var-}"
orig="${orig##*#}"
echo "$orig"
  # GOOD: _longopt
```

shell/completion.bash
```diff
@@ -173,18 +173,19 @@ __fzf_generic_path_completion() {
     base=${cur:0:${#cur}-${#trigger}}
     eval "base=$base"
 
+    dir=
     [[ $base = *"/"* ]] && dir="$base"
     while true; do
       if [[ -z "$dir" ]] || [[ -d "$dir" ]]; then
-        leftover=${base/#"$dir"}
+        leftover=${base/#"${dir-}"}
```
junegunn, line 180:
```suggestion
        leftover=${base/#"${dir}"}
```

shell/completion.bash
```diff
@@ -173,18 +173,19 @@ __fzf_generic_path_completion() {
...
         leftover=${leftover/#\/}
         [[ -z "$dir" ]] && dir='.'
         [[ "$dir" != "/" ]] && dir="${dir/%\//}"
-        matches=$(eval "$1 $(printf %q "$dir")" | FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} --reverse --bind=ctrl-z:ignore $FZF_DEFAULT_OPTS $FZF_COMPLETION_OPTS $2" __fzf_comprun "$4" -q "$leftover" | while read -r item; do
+        matches=$(eval "$1 $(printf %q "${dir-}")" | FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} --reverse --bind=ctrl-z:ignore ${FZF_DEFAULT_OPTS-} ${FZF_COMPLETION_OPTS-} $2" __fzf_comprun "$4" -q "$leftover" | while read -r item; do
```
junegunn, line 184:
```suggestion
        matches=$(eval "$1 $(printf %q "$dir")" | FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} --reverse --bind=ctrl-z:ignore ${FZF_DEFAULT_OPTS-} ${FZF_COMPLETION_OPTS-} $2" __fzf_comprun "$4" -q "$leftover" | while read -r item; do
```

shell/completion.bash
cevhyruz: I'm still getting `bash: warning: programmable_completion: ls: possible retry loop` even with the workaround. I'm only getting it though when  [scop/bash-completion](https://github.com/scop/bash-completion) is sourced.

cevhyruz: I'm still getting `bash: warning: programmable_completion: ls: possible retry loop` even with the workaround. I'm only getting it though when  [scop/bash-completion](https://github.com/scop/bash-completion) is sourced.

cevhyruz: This may be the one causing it https://github.com/scop/bash-completion/issues/522

cevhyruz: nevermind. It is working now with the workaround.

junegunn:
Please allow me to push more commits to your branch.

(https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork)

cevhyruz:
looks good to me, what do you think?

junegunn:
Looks good to me too. Thanks for the work.

PR #3002, opened 2022-10-12
Issues 1432

Adnn:
This is the other part of the issue 1432 opened in the fzf.vim repository.

It fixes the feature of `:Buffers` in Neovim running on Windows' git-bash, while making sure not to break when running in Windows' cmd.

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -484,7 +484,8 @@ try
     elseif type == 3
       let temps.input = s:fzf_tempname()
       call s:writefile(source, temps.input)
-      let source_command = (s:is_win ? 'type ' : 'cat ').fzf#shellescape(temps.input)
+      "let source_command = (s:is_win ? 'type ' : 'cat ').fzf#shellescape(temps.input)
+      let source_command = 'cat '.substitute(temps.input, '\', '/', 'g')
     else
       throw 'Invalid source type'
     endif
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -37,6 +37,9 @@ endif
 if s:is_win
   let s:term_marker = '&::FZF'
 
+  " This is not bulletproof, but cmd.exe does not have SHELL env var.
+  let s:is_gitbash = has_key(environ(), 'SHELL')
+
   function! s:fzf_call(fn, ...)
     let shellslash = &shellslash
     try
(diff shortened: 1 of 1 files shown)


plugin/fzf.vim
```diff
@@ -484,7 +487,8 @@ try
     elseif type == 3
       let temps.input = s:fzf_tempname()
       call s:writefile(source, temps.input)
-      let source_command = (s:is_win ? 'type ' : 'cat ').fzf#shellescape(temps.input)
+      let source_command = (s:is_win && !s:is_gitbash ? 'type ' : 'cat ')
```
junegunn, line 490: If I'm not mistaken, `environ()` is a recent addition to Vim (after Vim 8), so it's probably safer to use `exists('$VARNAME')` not to break this on older Vim versions.

```suggestion
      let source_command = (s:is_win && !exists('$SHELL') ? 'type ' : 'cat ')
```

plugin/fzf.vim
```diff
@@ -37,6 +37,9 @@ endif
 if s:is_win
   let s:term_marker = '&::FZF'
 
+  " This is not bulletproof, but cmd.exe does not have SHELL env var.
+  let s:is_gitbash = has_key(environ(), 'SHELL')
+
```
junegunn, line 42:
```suggestion
```

junegunn:
I wonder if it's possible to treat git-bash environment like a Unix shell environment. Does this work? Just a wild guess.

```diff
diff --git a/plugin/fzf.vim b/plugin/fzf.vim
index 40f01a0..2190209 100644
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -26,7 +26,7 @@ if exists('g:loaded_fzf')
 endif
 let g:loaded_fzf = 1
 
-let s:is_win = has('win32') || has('win64')
+let s:is_win = (has('win32') || has('win64')) && !exists('$SHELL')
 if s:is_win && &shellslash
   set noshellslash
   let s:base_dir = expand('<sfile>:h:h')
```

Adnn:
I reverted while using `exists('$SHELL')`, so all commands work on my test environment.

plugin/fzf.vim
```diff
@@ -484,7 +486,9 @@ try
     elseif type == 3
       let temps.input = s:fzf_tempname()
       call s:writefile(source, temps.input)
-      let source_command = (s:is_win ? 'type ' : 'cat ').fzf#shellescape(temps.input)
+      " Disable shell escape for git bash, as it breaks the command here
+      let source_command = (s:is_win_cmd ? 'type ' : 'cat ')
+        \.(!s:is_win || !exists('$SHELL') ? fzf#shellescape(temps.input) : substitute(temps.input, '\', '/', 'g'))
```
junegunn, line 491: Shouldn't we still `shellescape` (not `fzf#shellescape`) `temps.input` for safety?

junegunn: The expression is getting a bit too complicated. How about splitting it into if-else statements?
```sh

if !s:is_win
  let source_command = ...
elseif exists('$SHELL')
  " Git bash
  let source_command = ...
else
  " cmd.exe
  let source_command = ...
end
```

PR #3004, opened 2022-10-13
Don't expand user defined functions in shell scripts

ad-chaos:
Useful when, for example, doing
```zsh
fzf() {
    file_dir=$(command fzf "$@")
    if [[ $? -eq 0 ]]; then
        nvim $file_dir
    fi
}
```
or any other convenient wrappers around fzf, but don't want them to be
part of the `CTRL+T`, `CTRL+r` or `ALT+c` invocations of fzf.

### DIFF
--- a/shell/completion.zsh
+++ b/shell/completion.zsh
@@ -108,7 +108,7 @@ __fzf_comprun() {
     fi
   else
     shift
-    fzf "$@"
+    command fzf "$@"
   fi
 }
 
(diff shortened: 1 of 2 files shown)


junegunn:
I think you should just rename your function. There can be users who rely on the current behavior for valid reasons, like dynamically applying default options.

```sh
fzf() {
  if night; then
    /usr/local/bin/fzf --color dark "$@"
  else
    /usr/local/bin/fzf --color light "$@"
  fi
}
```

ad-chaos:
fair enough 👍

PR #3006, opened 2022-10-14
key-bindings.zsh: Document regex handling undocumented zsh behaviour

nh2:
See https://github.com/junegunn/fzf/pull/2140


junegunn:
Thanks, but I don't think this is necessary. One can `git blame` the line and get to the commit without too much difficulty.

e.g. If you use [gv.vim](https://github.com/junegunn/gv.vim), you select the line and `:GV` will give you this.

image

PR #3062, opened 2022-11-26
More util test

zeek0x:


junegunn:
Thanks, we will eventually remove 16 and 32 variants using Go generics in the future, but it wouldn't hurt to have more tests at this moment.

src/util/util_test.go
```diff
@@ -64,3 +175,12 @@ func TestTruncate(t *testing.T) {
 		t.Errorf("Expected: 6, actual: %d", width)
 	}
 }
+
+func TestRepeatToFill(t *testing.T) {
+	if RepeatToFill("abcde", 10, 50) != strings.Repeat("abcde", 5) {
+		t.Error("Expected:", strings.Repeat("abcde", 5))
+	}
+	if RepeatToFill("abcde", 10, 42) != strings.Repeat("abcde", 4) + "abcde"[:2] {
+		t.Error("Expected:", strings.Repeat("abcde", 4) + "abcde"[:2])
```
junegunn, line 184:
```suggestion
	if RepeatToFill("abcde", 10, 42) != strings.Repeat("abcde", 4)+"abcde"[:2] {
		t.Error("Expected:", strings.Repeat("abcde", 4)+"abcde"[:2])
```

zeek0x:
@junegunn I'll send PR use generics if possible, but may I do at this stage ?

junegunn:
Thanks but I don't see any compelling reason to switch to generics now. Those functions are trivial and they haven't been updated over the years without any problem. Using generics will only make fzf harder to build on systems with old versions of Go.

PR #3064, opened 2022-11-29
README.md: clarify what FZF_*_OPTS is used for

bjornfor:
At first I thought they were appended to FZF_*_COMMAND. Let's make it
clear that these are passed to `fzf` itself.


junegunn:
Thanks.

PR #3065, opened 2022-11-29
Refactor util.go

zeek0x:
- add Min16, Constrain16, AsUint32
- unify Min*, Max* inequalities  to `>` or `<` excluding `>=` or `<=`
- omit types of function arguments

### DIFF
--- a/src/util/util.go
+++ b/src/util/util.go
@@ -12,7 +12,7 @@ import (
 )
 
 // RunesWidth returns runes width
-func RunesWidth(runes []rune, prefixWidth int, tabstop int, limit int) (int, int) {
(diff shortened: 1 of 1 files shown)


junegunn:
> * sort functions

I'd rather not do this, it makes it hard to see the diff or track the changes (git blame) in a line range.

You see it's hard to see if the moved functions are the same as before or not: https://github.com/junegunn/fzf/pull/3065/files

junegunn:
> omit types of function arguments

I'm just being curious. Is this the recommended style these days? Are you following a specific Go style guide? The code we're discussing here dates back to early 2015. I don't recall following any style guides back then. I wonder if the community standard has evolved over time.

PR #3076, opened 2022-12-05
Add --complete-1 option

4z3:
This adds the `--complete-1` options that will cause the interactive finder to exit as soon as a single match is found.


avidseeker:
I think this featured is pushed to upstream now: https://github.com/junegunn/fzf/commit/8ec917b1c36ed1cf6ace64419307aa85fda4bee3

Issue #2019, opened 2020-05-06
Same concept as --bind but for UNIX signals


edi9999:
- [x] I have read through the manual page (`man fzf`)
- [x] I have the latest version of fzf
- [x] I have searched through the existing issues

## Info

- OS
    - [x] Linux
    - [] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [x] bash
    - [ ] zsh
    - [ ] fish
    
## Problem / Steps to reproduce

This is a bit similar as what is described in https://github.com/junegunn/fzf/issues/1789, so maybe it still does not make sense to handle signals as events.

I use fzf for multiple cases, and in one particular usecase, I am using fzf on some json data that comes from a given file, which contains an array of items.

Since this data takes some time to refresh from its "production" source, I actually use a cached version of the data for the fzf script which is present on my local machine.

However, in parallel, when the script runs, I also do a refresh of the data so that the data is uptodate when I run the script the next time. 

Refreshing the data and then running fzf would add some seconds of delay (approximately 5 probably), and since I use that command quite often, I think it makes sense to do both steps in parallel.

Now the problem I have is that quite often, I'm interested in the updated data, and since it doesn't show the uptodate version at first, it means that I usually start the script once and then again because my cached version is too much outdated.

I would find it cool to have the possibility to tell 

# My current script : 

```
get_data() {
  jq '.[] | .name + " " + .version' <~/tmp/data.json
}
update_data() {
   # some refreshing command with curl that updates ~/tmp/data.json
}
update_data &
get_data | fzf
```

# What I would like to have : 

```
get_data() {
  jq '.[] | .name + " " + .version' <~/tmp/data.json
}
update_data() {
   # some refreshing command with curl
   kill -SIGUSR1 "$fzf_pid"
}
update_data &
fzf < <(get_data) --bind 'sigusr1:reload'  &
fzf_pid="$!"
wait "$fzf_pid"
```

doronbehar:
I'd also be interested in such a feature. I'd like to make an already running fzf session reload data when an external program does something - I'd like the external program to tell fzf to reload the list. (I came to search for such an issue after trying to solve [this](https://gitlab.com/doronbehar/taskwarrior-fzf/-/issues/10) and [this](https://github.com/junegunn/fzf.vim/pull/1044#issuecomment-650979014)).

edi9999:
Hello @junegunn ,

I have a usecase where I would like to reopen this issue.
[...]

PR #3094, opened 2022-12-17
Add --listen=HTTP_PORT option to receive actions

junegunn:
```sh
# Start HTTP server on port 6266
fzf --listen 6266

# Send actions to the server
curl -XPOST localhost:6266 -d 'reload(seq 100)+change-prompt(hundred> )'
```
Feedbacks are welcome. Do you like it? Are you going to use it? What are you planning to do with this?

### Drawbacks

This will significantly increase the size of the binary (2.8MB to 5.7MB for darwin arm64). [UPX](https://upx.github.io/) might help with the issue. (However, https://github.com/upx/upx/issues/601)

### Alternatives

* Using signals as suggested in #2019 
     * Limited to only two bindings, SIGUSR1 and SIGUSR2
     * No SIGUSR1 or SIGUSR2 on Windows
* Non-HTTP socket server or named pipes
     * Less expressive and less extensible. We can extend the HTTP server to support other HTTP verbs in the future.
     * Likely less binary size? Haven't confirmed.
     * Portability issue

### DIFF
--- a/src/server.go
+++ b/src/server.go
@@ -97,7 +97,7 @@ func handleHttpRequest(conn net.Conn, channel chan []*action) string {
 	}
 
 	errorMessage := ""
-	actions := parseSingleActionList(strings.TrimSpace(string(body)), func(message string) {
+	actions := parseSingleActionList(strings.Trim(string(body), "\r\n"), func(message string) {
 		errorMessage = message
 	})
 	if len(errorMessage) > 0 {
--- a/src/server.go
+++ b/src/server.go
@@ -1,0 +1,112 @@
+package fzf
+
+import (
+	"bufio"
+	"bytes"
+	"errors"
+	"fmt"
+	"net"
+	"strconv"
+	"strings"
+)
+
+const (
+	crlf           = "\r\n"
+	httpOk         = "HTTP/1.1 200 OK" + crlf
+	httpBadRequest = "HTTP/1.1 400 Bad Request" + crlf
+)
+
+func startHttpServer(port int, channel chan []*action) {
+	if port == 0 {
+		return
+	}
+
+	listener, err := net.Listen("tcp", fmt.Sprintf(":%d", port))
+	if err != nil {
+		return
+	}
+
+	go func() {
+		for {
+			conn, err := listener.Accept()
+			if err != nil {
+				if errors.Is(err, net.ErrClosed) {
+					break
+				} else {
+					continue
+				}
+			}
+			conn.Write([]byte(handleHttpRequest(conn, channel)))
+			conn.Close()
+		}
+		listener.Close()
+	}()
+}
+
+// Here we are writing a simplistic HTTP server without using net/http
+// package to reduce the size of the binary.
+//
+// * No --listen:            2.8MB
+// * --listen with net/http: 5.7MB
+// * --listen w/o net/http:  3.3MB
+func handleHttpRequest(conn net.Conn, channel chan []*action) string {
+	line := 0
+	headerRead := false
+	contentLength := 0
+	body := ""
+	bad := func(message string) string {
+		message += "\n"
+		return httpBadRequest + fmt.Sprintf("Content-Length: %d%s", len(message), crlf+crlf+message)
+	}
+	scanner := bufio.NewScanner(conn)
+	scanner.Split(func(data []byte, atEOF bool) (int, []byte, error) {
+		found := bytes.Index(data, []byte(crlf))
+		if found >= 0 {
+			token := data[:found+len(crlf)]
+			return len(token), token, nil
+		}
+		if atEOF || len(body)+len(data) >= contentLength {
+			return 0, data, bufio.ErrFinalToken
+		}
+		return 0, nil, nil
+	})
+
+	for scanner.Scan() {
+		text := scanner.Text()
+		if line == 0 && !strings.HasPrefix(text, "POST / HTTP") {
+			return bad("invalid request method")
+		}
+		if text == crlf {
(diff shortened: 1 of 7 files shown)


edi9999:
Hi, interesting !

Yes, I think it is cool.
I will use it mainly for one purpose : to reload data from a source that updates itself in a background.
I'm not sure how the cursor position will adapt if the number of lines increases/decreases. It seems a lot like the React loop keys problem : https://reactjs.org/docs/lists-and-keys.html and I'm pretty sure it would be way to complex to handle keys like React does.

It is totally true that SIGNALS are too platform specific, Windows doesn't have them.

Specifically to me, the increase of binary size is not an issue, but it is true that just for one single feature, doubling the binary size seems like a lot.

Maybe it would be possible to simply use tcp (probably not since it doesn't give you enough garantee) ?

Another idea would be to have a websocket server instead of an HTTP server if we predict that there will be several messages sent in the future.

doronbehar:
> Feedbacks are welcome. Do you like it? Are you going to use it? What are you planning to do with this?

I like this feature a lot! I think the usage I was originally intending for this feature was deleting Vim buffers from `fzf.vim`'s `:Buffers`. This feature is needed for this feature because naively using `--bind=X:execute('nvr ...")+reload(...)` did not work for me from some reason. The alternative I was thinking about was mapping <kbd>X</kbd> from within Vim to delete the buffer (and then only tell `fzf` to reload the list via the server / UNIX signal), and not via `fzf`'s `--bind`.

I'm not sure it will be easy or even possible to implement this feature even with it merged.. I also hadn't had any more usages in mind.

The general usage @edi9999 mentioned:

> I will use it mainly for one purpose : to reload data from a source that updates itself in a background.

Is basically a generalization of my usage.

junegunn:
I rewrote the server without using `net/http` package. The implementation is simplistic and probably brittle, but the size of the binary has reduced to 3.3MB, which I think is acceptable.

@edi9999 

> I'm not sure how the cursor position will adapt

The cursor stays at the same vertical position after `reload` unless you update the query while reloading is in progress. It's an issue unrelated to this addition of `--listen`.

```sh
seq 100 | fzf --query 5 --sync --bind 'start:up+up+up+reload:sleep 1; seq 1000'
```

edi9999:
Looks great to me !
Awesome that you could use directly tcp and minimize the binary size.

src/server.go
```diff
@@ -0,0 +1,112 @@
...
+				length, err := strconv.Atoi(strings.TrimSpace(pair[1]))
+				if err != nil {
+					return bad("invalid content length")
+				}
+				contentLength = length
+			}
+		} else if contentLength <= 0 {
+			break
+		} else {
+			body += text
```
balki, line 94: There is no length/limit checking. If a malicious client keeps on sending garbage text, it gets added to `body`, till the process runs out of memory. I don't see a use-case where one would use this feature where the client cannot be trusted. However if it is used, then this could cause a crash. I think `net/http` does some limit checking.

Also if not fully validating content-length, why bother parsing? Some clients may not send it.

junegunn: > There is no length/limit checking

It is done here:

https://github.com/junegunn/fzf/blob/fd1f7665a77f3bd062586f0f3b54d4ff9863dd4b/src/server.go#L70-L71

I have tested this with `curl`, and if we don't stop after getting more than the specified content-length, the loop endlessly hangs trying to read more from the client, and curl never gets the response.

> I don't see a use-case where one would use this feature where the client cannot be trusted.

That's the assumption. `fzf < /dev/random` will easly make fzf consume an infinite amount of memory.

Anyway, thanks for chiming in. Is there anything you propose to better the implementation?

junegunn: https://github.com/junegunn/fzf/commit/1a9761736ed2190aa7536d4b6754f617f46cfd9c is an attempt to make the code slightly more robust.

edi9999:
Oops, just one thing has changed : 

Now I got some message like this : 

```
unknown action: execute(tmux split-window 'og_tmp_dir=/tmp/og_tmp_gIZ99X og | jq "del(.. | .loadedContent?, .fileName?,.functionName?,.lineNumber?,.columnNumber?)" --color-output | less --RAW-CONTROL-CHARS')
```

In previous versions it worked well. I don't know exactly how it worked previously.

junegunn:
That's because of https://github.com/junegunn/fzf/commit/def011c0291741b0ab3566a01d939374712719c8.

It was an attempt to reject incorrect inputs like `execute(echo foo)bar`, but it ended up breaking things that used to work. I'll see what I can do.

Having said that, see `ACTION ARGUMENT` section of the man page, you are better off writing `execute:...` if you're not chaining more actions (i.e. `execute(...)+foo+bar`).

junegunn:
But technically, fzf is incapable of processing `)` characters in `execute(...)` form. It used to work because of luck.

junegunn:
@edi9999 Please test if 73162a4bc3403e3532afa8a95008a3b4a00d554a helps in that case.

src/server.go
balki:
```suggestion
                        if len(body) > contentLength {
                            return bad("too much content")
                        }
			body += text
```
Improvements looks good but above check should resolve the infinite memory case fully.

junegunn: I don't think we have any memory problem now. Can you write an example of a malicious TCP request that can cause a memory problem?

edi9999:
I saw that you then also commited 8e283f512acb74faef6f1760927e1ef4da1b5263

I tested with this version and my binding is indeed parsed correctly again.

PR #3112, opened 2023-01-04
Refactor ansi.go

unrealjo:
Simplify matchControlSequence()  :
-  Convert switch to if statement
-  Avoid re-declaring c in each iteration

### DIFF
--- a/src/ansi.go
+++ b/src/ansi.go
@@ -110,12 +110,11 @@ func matchControlSequence(s string) int {
 	//                     ^ match starting here
 	//
 	i := 2 // prefix matched in nextAnsiEscapeSequence()
+    var c byte
 	for ; i < len(s); i++ {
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks, but we intentionally chose to use switch statement due to performance considerations. See https://github.com/junegunn/fzf/commit/209d5e8e9089d33694e40faa81107e07ff4f171f. The minor performance gain observed might get nullified as the Go compiler advances, but I don't see a good reason to revert the change at this time.

As for `c` variable, I don't see any problem with the current style which is used throughout the code base. I actually prefer the current code which is one-line shorter.

unrealjo:
My goal is to make the code cleaner and simpler, However, I appreciate your point, I will try my best to make important changes from now on.

PR #3128, opened 2023-01-17
Expose custom options for fzf#run

artemvmin:
With the exception of buffer open actions and preview toggling, I haven't found a way to declare action bindings (i.e. movement, selection toggling, etc.) from my vimrc. All resources I've been able to find indicate that this should be done through zsh, which is heavy-handed.

This is a simple mechanism to pass additional options for the vim plugin, without modifying the shell environment. I use this to pass bindings and other fzf options that I want applied unilaterally, such as `--layout reverse`.

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -463,6 +463,9 @@ try
   let dict   = exists('a:1') ? copy(a:1) : {}
   let temps  = { 'result': s:fzf_tempname() }
   let optstr = s:evaluate_opts(get(dict, 'options', ''))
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks, but you can just put this in your vim configuration file if you don't want to define it in your shell configuration file for some reason.

```vim
let $FZF_DEFAULT_OPTS = '--layout reverse'
```

artemvmin:
That's exactly what I was looking for. Thank you.

PR #3160, opened 2023-02-09
Update install script with misc fixes

ChaosCypher:
Misc quote fixes and additional safety checks

### DIFF
--- a/install
+++ b/install
@@ -7,14 +7,13 @@ auto_completion=
 key_bindings=
 update_config=2
 shells="bash zsh fish"
-prefix='~/.fzf'
+prefix="$HOME/.fzf"
 prefix_expand=~/.fzf
 fish_dir=${XDG_CONFIG_HOME:-$HOME/.config}/fish
 
 help() {
   cat << EOF
 usage: $0 [OPTIONS]
-
     --help               Show this message
     --bin                Download fzf binary only; Do not generate ~/.fzf.{bash,zsh}
     --all                Download fzf binary and update configuration files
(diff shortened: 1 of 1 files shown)


install
```diff
@@ -7,14 +7,13 @@ auto_completion=
 key_bindings=
 update_config=2
 shells="bash zsh fish"
-prefix='~/.fzf'
+prefix="$HOME/.fzf"
```
junegunn, line 10: Any reason you changed this? This was intentional. I wanted the line to be added to look like `[ -f ~/.fzf.bash ] && source ~/.fzf.bash` so you can bring your configuration around different computers with different user names and it still works.

ChaosCypher: Hello @junegunn, I changed it because [this](https://github.com/junegunn/fzf/blob/4982111f6eafe8c9221c7fce3df2cf5c9e7ff7b9/install#L1) sets the script to use bash which has a built-in for $HOME. This would still work if you change computers and usernames.

junegunn: That's not my point. The actual username may differ on different machines. Have you actually compared the resulting file before and after the patch?

ChaosCypher: Hi @junegunn perhaps we can remove this change from the PR, and discuss this in a Discussion/Issue thread separately if that works? Assuming this is the only change that's raising a concern.

ChaosCypher: Hey @junegunn I just wanted to followup to see if there were any other concerns with this PR besides the one you mentioned? I've removed that change to invite a separate discussion behind the reasoning.

junegunn: It's not clear which problem this pull request is trying to solve. Can you give me an example? I can't think of an example where the omitted quotes can be a problem. Also, this pull request introduces another behavioral change in addition to the one I mentioned above, which makes me wonder if you have properly understood the intention of each code line. We can't just blindly apply the suggestions from the linters like shellcheck.

PR #3180, opened 2023-02-20
Add support for IBM s390x architecture

julianruess:


junegunn:
Interesting. We can also build and release s390x binaries if that matters to you. What do you think?

For that, we need to update two files:

* https://github.com/junegunn/fzf/blob/b1459c79cfc004dd7944f4db3ffb5a4ae20569ca/.goreleaser.yml#L71-L76
* https://github.com/junegunn/fzf/blob/b1459c79cfc004dd7944f4db3ffb5a4ae20569ca/install#L173-L180

What does `uname -sm` print on the machine?

julianruess:
Hi, yes, it would be really nice to also build and release the binaries!
uname -sm prints: "Linux s390x"
Thank you very much!

junegunn:
Merged, thanks. s390x binaries will be provided from the next version.

PR #3204, opened 2023-03-09
Adding `$FWDPROXY_CURL` environment variable for `curl`

skanti:
Adding `$FWDPROXY_CURL` environment variable for `curl` allows download and installation of `fzf` dependencies through company internet proxies.

### DIFF
--- a/install
+++ b/install
@@ -115,10 +115,10 @@ link_fzf_in_path() {
 try_curl() {
   command -v curl > /dev/null &&
   if [[ $1 =~ tar.gz$ ]]; then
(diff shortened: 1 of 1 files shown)


junegunn:
Why not use `$http[s]_proxy`?

skanti:
`FWDPROXY_CURL` would include e.g. `-p -x --resolve --proxy-cert --proxy-key` proxy parameters for curl. Setting `https_proxy` does not work if you need `--proxy-cert` or `--proxy-key`

Issue #2714, opened 2022-01-13
Bash completion for kill not working on BSD


lcheylus:
- [X] I have read through the manual page (`man fzf`)
- [X] I have the latest version of fzf
- [X] I have searched through the existing issues

## Info

- OS
    - [X] BSD (FreeBSD, OpenBSD)
- Shell
    - [X] bash
    - [ ] zsh
    - [ ] fish

## Problem / Steps to reproduce

When using FZF completion with bash (via `shell/completion.bash`), completion does not work for `kill` command (`kill **`).

In `shell/completion.bash`, `ps -ef` is used in `_fzf_proc_completion` function but this syntax is not compliant on BSD OS (see FreeBSD manpage for `ps` https://www.freebsd.org/cgi/man.cgi?query=ps). I propose to replace `ps -ef` command by `ps aux`, compatible with BSD and Linux OS.

lcheylus:
Issue solved with https://github.com/junegunn/fzf/pull/3219 : `kill **` command works on OpenBSD (and also FreeBSD) with version 0.39.0.

PR #3219, opened 2023-03-23
Support kill completion for busybox

ChingChang9:
The GNU `ps -ef` lists the processes with fields in the order of `user pid time args`, but the busybox `ps` lists them in the order of `pid user time args`. This makes the tab completion for kill expand to the user instead of the pid.

Specifying the order explicitly allows us to support busybox without changing the behaviour for GNU.

### DIFF
--- a/shell/completion.bash
+++ b/shell/completion.bash
@@ -270,8 +270,13 @@ _fzf_complete_kill() {
 }
 
 _fzf_proc_completion() {
+  if test $(readlink /bin/ps); then
+    local ps_args='-o user,pid,time,args'
+  else
+    local ps_args='-ef'
+  fi
   _fzf_complete -m --preview 'echo {}' --preview-window down:3:wrap --min-height 15 -- "$@" < <(
-    command ps -o user,pid,time,args | sed 1d
+    command ps $ps_args | sed 1d
   )
 }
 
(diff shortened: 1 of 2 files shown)


junegunn:
The resulting list is not identical.

```sh
$ command ps -o user,pid,time,args | wc -l
5
$ command ps -ef | wc -l
181
```

ChingChang9:
Oh oops `-ef` and `-o` give the same processes on busybox but apparently not on GNU. I've changed it again so now it passes different arguments based on whether the user is running busybox or GNU coreutils. It may be a workaround but I think it works.

junegunn:
Also related: #2714

Should we try to find a command that works (almost) the same on all (or most) platforms before introducing code branches?

* `ps -eo user,pid,ppid,start,time,args` seems like a good start
    * Huh, `start` is not supported on BusyBox; then we have `ps -eo user,pid,ppid,time,args` which is unfortunate.
    * But even so, does it work on FreeBSD? 
* `ps aux` should work on all platforms, but it gives a different order of columns on BusyBox 😕

junegunn:
Hey, please check out 302e21fd5813a3c4c7f6014367988829fc92770a; it will make kill completion work as expected on BusyBox.

ChingChang9:
Thank you!

PR #3223, opened 2023-03-26
Add bat to bash autocomplete commands

StenAL:
Bat is a common alternative to `cat`, it's even referenced multiple times in fzf docs. This makes `bat **` work by default.


junegunn:
I'm a bit hesitant to grow the list because it will increase the start-up time of the shell, and the recommended way to add support for a specific command is to use `_fzf_setup_completion` function as documented on https://github.com/junegunn/fzf#supported-commands.

Having said that, I currently don't have a good criterion for determining which ones should be on the list. But as you said, `bat` is mentioned on our documentation, so I'm going to merge this. Thanks.

PR #3257, opened 2023-04-20
Add check for powershell 7 users

PavelSarlov:
If you are using Powershell 7 under Windows (meaning you have `shell` set to `pwsh`), FZF will fail to run with the following error:
image

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -164,7 +164,7 @@ function s:get_version(bin)
   if has_key(s:versions, a:bin)
     return s:versions[a:bin]
   end
-  let command = (&shell =~ 'powershell' || &shell =~ 'pwsh' ? '&' : '') . s:fzf_call('shellescape', a:bin) . ' --version --no-height'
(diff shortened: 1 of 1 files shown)


plugin/fzf.vim
```diff
@@ -164,7 +164,7 @@ function s:get_version(bin)
   if has_key(s:versions, a:bin)
     return s:versions[a:bin]
   end
-  let command = (&shell =~ 'powershell' ? '&' : '') . s:fzf_call('shellescape', a:bin) . ' --version --no-height'
+  let command = (&shell =~ 'powershell' || &shell =~ 'pwsh' ? '&' : '') . s:fzf_call('shellescape', a:bin) . ' --version --no-height'
```
junegunn, line 167: Thanks. I'd suggest using regex like so:

```suggestion
  let command = (&shell =~ 'powershell\|pwsh' ? '&' : '') . s:fzf_call('shellescape', a:bin) . ' --version --no-height'
```

junegunn:
Merged, thanks.

PR #3275, opened 2023-05-02
README: Add marcosnils/bin tool as installation method.

pataquets:
Just started using to manage all my standalone binaries and works beautifully.
I've discovered `bin` on [sysz](https://github.com/joehillen/sysz)'s [readme](https://github.com/joehillen/sysz#using-bin). Convenient, easy, nifty.

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,8 +155,14 @@ page][windows-wiki].
 
 [windows-wiki]: https://github.com/junegunn/fzf/wiki/Windows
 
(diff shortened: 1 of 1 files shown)


junegunn:
Thank you, but I'd like to avoid suggesting the method as it doesn't install anything other than the binary when fzf ships with additional files (fzf-tmux, fzf.vim, and man pages).

PR #3278, opened 2023-05-03
Several functions have been added. Modify Function `__fzf_cd__()` and…

A4-Tacks:
Fix: When the function `__fzf_cd_()` or function `__fzf_select__()` in `key-bindings.bash` has too many searches at once, `find` is confirmed by `fzf` downstream of the pipeline and the pipeline is closed from downstream, which in most cases will block the operation.
Now after the `fzf` search is completed, the pipeline will be killed upstream so that it will not be blocked.
At the same time, the completion function brought by the file `completion.bash` also has the same problem, but I cannot modify it because it is a bit complex.

### DIFF
--- a/shell/key-bindings.bash
+++ b/shell/key-bindings.bash
@@ -13,18 +13,39 @@
 
 # Key bindings
 # ------------
+__fzf_new_tmp_file() {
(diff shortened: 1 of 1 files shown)


A4-Tacks:
I conducted another test afterwards and I was unable to solve the problem

Issue #3240, opened 2023-04-06
Feature Request: SHIFT_DELETE to Delete The Current Line


WIZARDELF:
It is handy to use FZF to delete stale history items during walking through FZF hits with a shortcut, e.g. SHIFT_DELETE.

I personally use `history` to do such maintenance, but not convenient.

Issue #1933, opened 2020-03-21
[Feature] Add shift-pgup and shift-pgdn keybinds


anasbouzid:
- [x] I have read through the manual page (`man fzf`)
- [x] I have the latest version of fzf
- [x] I have searched through the existing issues

## Info

- OS
    - [x] Linux
    - [x] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [x] bash
    - [x] zsh
    - [x] fish

I'd like to use `shift-pgup (shift-page-up)` and `shift-pgdn (shift-page-down)` key bindings to have preview scrolling by page. `fzf` already supports `shift-down` and `shift-up`, but scrolling by page would be a useful feature.

0ion9:
`--bind`.

junegunn:
My terminal doesn't recognize shift-pgdn or shift-pgup. Maybe they are not part of the standard? Run this Ruby script and see what escape sequences those key chords generate.

```ruby
#!/usr/bin/env ruby

require 'io/console'

loop do
  chars = $stdin.getch + ($stdin.read_nonblock(1024) rescue '')
  break if [3, 4, 26].map(&:chr).include?(chars)

  p chars
end
```

0ion9:
Oh, that's a fair point, perhaps I misunderstood OP.

In xterm, which doesn't consume shift+pg[up][down], `fish_key_reader` ( a `fish` builtin tool designed to do approximately the same as your snippet, junegunn) does not register shift+pgup /shift+pgdn

In sakura, these keys are just consumed by the terminal (to scroll the terminal up/down).
Sakura is libvterm based;I also tested another libvterm-based terminal, `pangoterm`. That has the same behaviour as sakura, so possibly libvterm-based terminals bind these key combos by default. Lilyterm (using the legacy version, libvte) also captured these keys.

I took the time to survey a few more:

* lxterminal (libvterm based)  : Captures these keys
* evilvte (libvte legacy): Captures these keys
* gnome-terminal (libvte3): Captures these keys
* konsole (kpty based): Captures these keys

So I suspect these cannot normally be read, although possibly emulators with a rebinding option might change that.

junegunn:
@0ion9 Thanks for the investigation! FYI, I have tested these on iTerm2 and Terminal.app on macOS. iTerm2 doesn't recognize them at all, and Terminal.app reads them as plain pgup and pgdn.

[...]

PR #3284, opened 2023-05-06
Add key combinations for ctrl-delete and shift-delete

Syphdias:
Currently there is not option to bind ctrl-delete and shift-delete. As suggested by issue #3240, shift-delete could be used to bind "delete entry from history" as it is a common way to do so in other applications, e.g. browsers.

This, however, does only implement to use the key combination itself and does not assign a default action to any of them. This does enable to call one's all predefined actions. With the exec action this can expanded like the issue #3240 suggested.
If desirable, the key combinations could later get a default behavior.

I mainly wanted to use shift-delete like suggested in #3240. My approach would be to use the exec() action to build this functionality myself. I would not be comfortable deleting from the uses history due to expansion/escape issue and deleting too much or the wrong thing (for me I would risk it).

Two issues I would like to point to:
1. I am not sure which tests need to be expanded on and how. I could use some feedback on this, if the current tests, are not sufficient.
2. I modified the switch case for `Del`. In my testing, pressing only `Del` yielded an `r.buffer` with a length of 4. When combined with <kbd>Ctrl</kbd> or <kbd>Shift</kbd> the lenght went up to 6. Index 3 was always `;` though. So I skipped it for now. Should I check for for this explicitly and if so. Should I start a new switch to keep it analogous to other code and easier expansion, if there is ever another key combination that needs it?

### DIFF
--- a/man/man1/fzf.1
+++ b/man/man1/fzf.1
@@ -870,6 +870,8 @@ e.g.
 .br
 \fIctrl-space\fR
 .br
+\fIctrl-delete\fR
+.br
 \fIctrl-\\\fR
 .br
 \fIctrl-]\fR
@@ -937,6 +939,8 @@ e.g.
 \fIshift-left\fR
 .br
 \fIshift-right\fR
+.br
+\fIshift-delete\fR
 .br
 \fIalt-shift-up\fR
 .br
--- a/src/options.go
+++ b/src/options.go
@@ -612,6 +612,8 @@ func parseKeyChordsImpl(str string, message string, exit func(string)) map[tui.E
 			add(tui.BSpace)
 		case "ctrl-space":
 			add(tui.CtrlSpace)
+		case "ctrl-delete":
+			add(tui.CtrlDelete)
 		case "ctrl-^", "ctrl-6":
 			add(tui.CtrlCaret)
 		case "ctrl-/", "ctrl-_":
@@ -682,6 +684,8 @@ func parseKeyChordsImpl(str string, message string, exit func(string)) map[tui.E
 			add(tui.SLeft)
 		case "shift-right":
 			add(tui.SRight)
+		case "shift-delete":
+			add(tui.SDelete)
 		case "left-click":
 			add(tui.LeftClick)
 		case "right-click":
--- a/src/tui/light.go
+++ b/src/tui/light.go
@@ -430,7 +430,19 @@ func (r *LightRenderer) escSequence(sz *int) Event {
 				}
 				return Event{Invalid, 0, nil} // INS
 			case '3':
-				return Event{Del, 0, nil}
+				if r.buffer[3] == '~' {
+					return Event{Del, 0, nil}
+				}
+				if len(r.buffer) == 6 && r.buffer[5] == '~' {
+					*sz = 6
+					switch r.buffer[4] {
+					case '5':
+						return Event{CtrlDelete, 0, nil}
+					case '2':
+						return Event{SDelete, 0, nil}
+					}
+				}
+				return Event{Invalid, 0, nil}
 			case '4':
 				return Event{End, 0, nil}
 			case '5':
--- a/src/tui/tcell.go
+++ b/src/tui/tcell.go
@@ -346,8 +346,6 @@ func (r *FullscreenRenderer) GetChar() Event {
 		// section 2: Ctrl+[ \]_]
 		case tcell.KeyCtrlSpace:
 			return Event{CtrlSpace, 0, nil}
-		case tcell.KeyCtrlDelete:
-			return Event{CtrlDelete, 0, nil}
 		case tcell.KeyCtrlBackslash:
 			return Event{CtrlBackSlash, 0, nil}
 		case tcell.KeyCtrlRightSq:
@@ -408,8 +406,6 @@ func (r *FullscreenRenderer) GetChar() Event {
 				return Event{AltRight, 0, nil}
 			}
 			return Event{Right, 0, nil}
-		case tcell.KeySDelete:
-			return Event{SDelete, 0, nil}
 
 		// section 5: (Insert|Home|Delete|End|PgUp|PgDn|BackTab|F1-F12)
 		case tcell.KeyInsert:
@@ -417,6 +413,12 @@ func (r *FullscreenRenderer) GetChar() Event {
 		case tcell.KeyHome:
 			return Event{Home, 0, nil}
 		case tcell.KeyDelete:
+			if ctrl {
+				return Event{CtrlDelete, 0, nil}
+			}
+			if shift {
+				return Event{SDelete, 0, nil}
+			}
 			return Event{Del, 0, nil}
 		case tcell.KeyEnd:
 			return Event{End, 0, nil}
--- a/src/tui/tcell.go
+++ b/src/tui/tcell.go
@@ -346,6 +346,8 @@ func (r *FullscreenRenderer) GetChar() Event {
 		// section 2: Ctrl+[ \]_]
 		case tcell.KeyCtrlSpace:
 			return Event{CtrlSpace, 0, nil}
+		case tcell.KeyCtrlDelete:
+			return Event{CtrlDelete, 0, nil}
 		case tcell.KeyCtrlBackslash:
 			return Event{CtrlBackSlash, 0, nil}
 		case tcell.KeyCtrlRightSq:
@@ -406,6 +408,8 @@ func (r *FullscreenRenderer) GetChar() Event {
 				return Event{AltRight, 0, nil}
 			}
 			return Event{Right, 0, nil}
+		case tcell.KeySDelete:
+			return Event{SDelete, 0, nil}
 
 		// section 5: (Insert|Home|Delete|End|PgUp|PgDn|BackTab|F1-F12)
 		case tcell.KeyInsert:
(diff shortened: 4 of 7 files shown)


Syphdias:
Hey @junegunn, I was wondering if you find this PR useful.

Since https://github.com/junegunn/fzf/issues/1933 I assume the reason these key combinations are not already a thing, might be an issue with some terminals. Unfortunately, I cannot test if these keybinds work for iTerm or Terminal.app. 

Also, is there an option to use a "raw" mode for arbitrary keys? What I mean by that is to not be bound by the predefined keys manual definition, but use the escape sequence directly, e.g. `^[[3~` for <kbd>Delete</kbd>. I don't think this is possible right now. Issues I foresee would be that is is quite advanced and might lead to bug reports, reporting that some terminals behave different. This would need to be clarified in the man page.

TL;DR: Is this PR ([ctrl-delete and shift-delete](https://github.com/junegunn/fzf/pull/3284#top)) interesting to you? Would you be interested in accepting a PR that implements binding arbitrary keys?

junegunn:
> Is this PR ([ctrl-delete and shift-delete](https://github.com/junegunn/fzf/pull/3284#top)) interesting to you?

I mainly use iTerm2 and it doesn't appear to recognize those keys, so I'm not personally going to use it. But I have confirmed that it works on Terminal.app, so I'm going to merge this.

junegunn:
> Would you be interested in accepting a PR that implements binding arbitrary keys?

I honestly think it's a bit too much for this program.

Syphdias:
> > Would you be interested in accepting a PR that implements binding arbitrary keys?
> 
> I honestly think it's a bit too much for this program.

Okay thanks, I'd rather know before I spend the time :)


> I mainly use iTerm2 and it doesn't appear to recognize those keys, so I'm not personally going to use it. But I have confirmed that it works on Terminal.app, so I'm going to merge this.

Should I add an disclaimer that some terminals might not recognise these key combinations or is this a general concern with all of the keys?

junegunn:
Need to address these errors in tcell build (tcell is only used for Windows). Let me see...

```
jg@:~/github/fzf (add-ctrl-delete-and-shift-delete)> go run -tags tcell main.go
# github.com/junegunn/fzf/src/tui
src/tui/tcell.go:349:14: undefined: tcell.KeyCtrlDelete
src/tui/tcell.go:411:14: undefined: tcell.KeySDelete
```

junegunn:
> Should I add an disclaimer that some terminals might not recognise these key combinations or is this a general concern with all of the keys?

The latter. Let's just see if anyone complains.

junegunn:
Merged, thanks. Let me know if you have see issues.

eggbean:
Is this supposed to work with the `ctrl-r` history? I would like to delete lines with typos, for example, but `shift-delete` doesn't do anything (v0.41.1). Do I have to the binding somehow? Thanks.

eggbean:
Thanks, but I cannot get this to do what I thought it would.

    export FZF_CTRL_R_OPTS=" \
      --bind 'shift-delete:execute(echo {} >/tmp/test)'"

This writes the entry to `/tmp/test` but it still remains in the history list. I thought that the entry was supposed to be removed and I don't understand the need for the `/tmp/test` file.

Syphdias:
This is not the solution to the problem of removing the entry from the history. And I don't have the solution for it.
I only provided a starting-off point, so you could tackle the problem yourself.

You need to find something to replace `echo {} >/tmp/test` with the right command/script, to do what you want.

eggbean:
I see. Well I tried this hoping that it'd work (after backing up my history file), but it doesn't. I would prefer to avoid having to use a separate script.

    export FZF_CTRL_R_OPTS=" \
      --bind 'shift-delete:execute(LC_ALL=C sed -i "/{}/d" $HISTFILE)'"

ref: https://stackoverflow.com/a/13661794/140872

eggbean:
It's possible that the command number changes before the command is executed, so I tried this:

    export FZF_CTRL_R_OPTS=" \
      --bind 'shift-delete:execute(str="$(echo {} | cut -d " " -f2-)" && LC_ALL=C sed -i "/$str/d" $HISTFILE)'"

Hopefully someone can come up with some better ideas.

Syphdias:
If it was an easy problem, I would have included it in my PR :wink:.

What I can tell you is, that you have better chances with `'shift-delete:execute(LC_ALL=C sed -i /{}/d $HISTFILE)'` (no double quotes). But this is not perfect. It will fail for commands with slashes, quotes, variables (e.g. $1), and multi-line commands, which is not acceptable for me.

eggbean:
I've made some progress (zsh):

```
export FZF_CTRL_R_OPTS=" \
  --bind 'shift-delete:execute(echo {} >/tmp/test)'"

% cat fzf-history-delete.sh
#!/bin/bash -x
str="$(echo "$*" | awk '{$1=""}1' | colrm 1 1 | sed -e 's/[]\/$*.^[]/\\&/g')"
LC_ALL=C sed -i "/$str$/d" "$HISTFILE"

% cp $HISTFILE{,.bak}

* Use `Ctrl-R` , `shift-delete`

% cat /tmp/test
 1774  fc -p $HISTFILE

% HISTFILE=$HISTFILE ./fzf-history-delete.sh $(cat /tmp/test)
++ echo '1774 fc -p $HISTFILE'
++ colrm 1 1
++ awk '{$1=""}1'
++ sed -e 's/[]\/$*.^[]/\\&/g'
+ str='fc -p \$HISTFILE'
+ LC_ALL=C
+ sed -i '/fc -p \$HISTFILE$/d' /home/jason/.cache/zsh/zsh_history

% fc -R $HISTFILE
```

This seems to work with everything I've tried with it, in zsh. Doesn't seem to work in bash.

```
export FZF_CTRL_R_OPTS=" \
  --bind 'shift-delete:execute(HISTFILE=$HISTFILE fzf-history-delete.sh {})'"
```

This doesn't seem to work.

I tried making it script-less, but this nuked the history file.

```
export FZF_CTRL_R_OPTS=" \                                                                                   
  --bind 'shift-delete:execute(HISTFILE=$HISTFILE str="$(echo {} | awk '{$1=""}1' | colrm 1 1 | sed -e 's/[]\/$*.^[]/\\&/g')" ;\
    LC_ALL=C sed -i "/$str$/d" "$HISTFILE")'"
```

I'll add a test to avoid empty `$str` and simplify the `echo|awk|colrm` with `shift`.

eggbean:
I've made some progress (zsh):

```
export FZF_CTRL_R_OPTS=" \
  --bind 'shift-delete:execute(echo {} >/tmp/test)'"
```

```
% cat fzf-history-delete.sh
#!/bin/bash -x
str="$(echo "$*" | awk '{$1=""}1' | colrm 1 1 | sed -e 's/[]\/$*.^[]/\\&/g')"
LC_ALL=C sed -i "/$str$/d" "$HISTFILE"

% cp $HISTFILE{,.bak}

* Use `Ctrl-R` , `shift-delete`

% cat /tmp/test
 1774  fc -R $HISTFILE

% HISTFILE=$HISTFILE ./fzf-history-delete.sh $(cat /tmp/test)
++ echo '1774 fc -R $HISTFILE'
++ colrm 1 1
++ awk '{$1=""}1'
++ sed -e 's/[]\/$*.^[]/\\&/g'
+ str='fc -R \$HISTFILE'
+ LC_ALL=C
+ sed -i '/fc -p \$HISTFILE$/d' /home/jason/.cache/zsh/zsh_history

% fc -R $HISTFILE   # Reload history
```

This seems to work with everything I've tried with it, in zsh. Doesn't seem to work in bash.

```
export FZF_CTRL_R_OPTS=" \
  --bind 'shift-delete:execute(HISTFILE=$HISTFILE fzf-history-delete.sh {})'"
```

This doesn't seem to work.

I tried making it script-less, but this nuked the history file.

```
export FZF_CTRL_R_OPTS=" \
  --bind 'shift-delete:execute(str="$(echo {} | awk '{$1=""}1' | colrm 1 1 | sed -e 's/[]\/$*.^[]/\\&/g')" ; \
    LC_ALL=C sed -i "/$str$/d" "$HISTFILE"; \
    fc -R "$HISTFILE")'"
```

I'll add a test to avoid empty `$str` and simplify the `echo|awk|colrm` with `shift`.

eggbean:
I've made some progress (zsh):

```
export FZF_CTRL_R_OPTS=" \
  --bind 'shift-delete:execute(echo {} >/tmp/test)'"
```

```
% cat fzf-history-delete.sh
#!/bin/bash -x
str="$(echo "$*" | awk '{$1=""}1' | colrm 1 1 | sed -e 's/[]\/$*.^[]/\\&/g')"
LC_ALL=C sed -i "/$str$/d" "$HISTFILE"

% cp $HISTFILE{,.bak}

* Use `Ctrl-R` , `shift-delete`

% cat /tmp/test
 1774  fc -R $HISTFILE

% HISTFILE=$HISTFILE ./fzf-history-delete.sh $(cat /tmp/test)
++ echo '1774 fc -R $HISTFILE'
++ colrm 1 1
++ awk '{$1=""}1'
++ sed -e 's/[]\/$*.^[]/\\&/g'
+ str='fc -R \$HISTFILE'
+ LC_ALL=C
+ sed -i '/fc -p \$HISTFILE$/d' /home/jason/.cache/zsh/zsh_history

% fc -R $HISTFILE   # Reload history
```

This seems to work with everything I've tried with it, in zsh. Doesn't seem to work in bash.

```
export FZF_CTRL_R_OPTS=" \
  --bind 'shift-delete:execute(HISTFILE=$HISTFILE fzf-history-delete.sh {})'"
```

This doesn't seem to work.

I tried making it script-less, but this nuked the history file.

```
export FZF_CTRL_R_OPTS=" \
  --bind 'shift-delete:execute(str="$(echo {} | awk '{$1=""}1' | colrm 1 1 | sed -e 's/[]\/$*.^[]/\\&/g')" ; \
    LC_ALL=C sed -i "/$str$/d" "$HISTFILE"; \
    fc -R "$HISTFILE")'"
```

I'll should test to avoid empty `$str`, or whatever is going on and simplify the `echo|awk|colrm` with `shift`.

Issue #3077, opened 2022-12-07
Feature request: drop dependency for Perl in `key-bindings.bash`


jlebon:
- [x] I have read through the manual page (`man fzf`)
- [x] I have the latest version of fzf
- [x] I have searched through the existing issues

## Info

- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [x] bash
    - [ ] zsh
    - [ ] fish

## Problem / Steps to reproduce

`__fzf_history__()` in `key-bindings.bash` requires Perl. In more minimal environments, Perl may not be installed. It would be nice to be able to drop that dependency. Looking at [the exact code there](https://github.com/junegunn/fzf/blob/6b207bbf2bd916d06586cdbbf980ae507c01f2b2/shell/key-bindings.bash#L54), it seems like it's just doing some frequency counting of the bash command history. Could we have that be built into fzf perhaps and gated behind a flag?

glensc:
this doesn't seem like big deal to rewrite in awk or so

jlebon:
Looks like the latest effort is stuck on performance issues: https://github.com/junegunn/fzf/pull/3313#issuecomment-1573049093. I added a suggestion there.

step-:
Author of #3395 (closed), #3310 (bash) and #3313 (awk) here. I read your suggestion in #3310. You're suggesting to implement an ad-hoc filter that would do what the perl filter (or my awk filter) does now. Am I right?

jlebon:
> Author of #3395 (closed), #3310 (bash) and #3313 (awk) here. I read your suggestion in #3310. You're suggesting to implement an ad-hoc filter that would do what the perl filter (or my awk filter) does now. Am I right?

Right, I'm suggesting implementing the filter in golang behind some hidden switch. But I'm not a maintainer :) I would try to get @junegunn's opinion on it before doing this.

step-:
can be closed due to #3313 merge

step-:
TLDR; <https://github.com/junegunn/fzf/pull/3313#issuecomment-1573049093>

glensc:
@kaddkaka broken about locale, add `LC_ALL=C` before perl command to mute them. probably worth sending PR to this repo too.

kaddkaka:
> @kaddkaka broken about locale, add `LC_ALL=C` before perl command to mute them. probably worth sending PR to this repo too.

[...]

Issue #2260, opened 2020-11-20
CTRL-R show nothing before typing a command on a new bash


FabriceSh44:
- [x] I have read through the manual page (`man fzf`)
- [x] I have the latest version of fzf
- [x] I have searched through the existing issues

## Info

- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [x] bash
    - [ ] zsh
    - [ ] fish
    
## Problem / Steps to reproduce

1. run "bash"
2. press CTRL-R => preview shows 0/0 +S
3. CTRL-C
4. run "echo test"
5. press CTRL-R => preview show 1000 elements

If I run a new bash and "history" I see 1000 elements. It seems that I need to run at least one command to have something in CTRL+R.
Doesn't have this issue with[ key-bindings.bash version:9c293bb82b6be333c857d288d10aab730220a0e7](https://github.com/junegunn/fzf/blob/9c293bb82b6be333c857d288d10aab730220a0e7/shell/key-bindings.bash)

junegunn:
We use `fc -lnr -2147483648` to generate the list. Check if you see an empty list when you run the command.

FabriceSh44:
`fc -lnr -2147483648 `  shows the list alright

junegunn:
You can try debugging the problem by starting bash with `-x`.

FabriceSh44:
when pressing CTRL R, I get :
```
++ __fzf_history__                                             
++ local output                                                
+++ builtin fc -lnr -2147483648
++++ __fzfcmd                                                  
++++ HISTTIMEFORMAT=                                           
++++ builtin history 1                                         
++++ '[' -n %29 ']'                                            
++++ '[' 0 '!=' 0 ']'                                          
++++ '[' -n '' ']'                                             
++++ echo fzf                                                  
+++ FZF_DEFAULT_OPTS='--height 40% --height 50% --multi --preview '\''(bat --style=numbers --color=always {} || cat {}) 2> /dev/null | head -300'\'' --preview-window='\''right:hidden:wrap'\'' --bind='\''f2:toggle-preview,ctrl-d:preview-page-down,ctrl-u:p
review-page-up'\'' -n2..,.. --tiebreak=index --bind=ctrl-r:toggle-sort  +m --read0'                                            
+++ fzf --query ''                                             
+++ last_hist=' 1000  fc -lnr -2147483648'
[...]

PR #3295, opened 2023-05-14
replace perl with awk in the bash history widget

step-:
While awk is POSIX, perl isn't pre-installed on all *nix flavors. This commit replaces perl with awk thus it eliminates the perl dependency.

This commit passed the current test suite with zero errors :
* `make error` => all test sections 'ok'
* `make docker-test` => Finished in 50.929904s, 4.1822 runs/s, 35.3427 assertions/s. 213 runs, 1800 assertions, 0 failures, 0 errors, 0 skips

See also:  #2777 #1934 # #3077 #2260

Test suite was executed with ruby 3.2.2, minitest 5.18.0, and tmux 3.3a. The current [Dockerfile](https://github.com/junegunn/fzf/blob/94999101e358385f3ca67a6ec9512f549196b802/Dockerfile) runs with older ruby and minitest versions.

Note: unlike the original perl script, which gets `$HISTCOUNT` from the environment, this awk script reads the initial history count from line 1 of stdin. Then follow the history list, and finally a `\t`. The reason for `\t` is to fix a small, pre-existing issue I noticed while developing this patch (see the next comment for additional information).

### DIFF
--- a/shell/key-bindings.bash
+++ b/shell/key-bindings.bash
@@ -51,10 +51,14 @@ __fzf_cd__() {
 __fzf_history__() {
   local output opts script
   opts="--height ${FZF_TMUX_HEIGHT:-40%} --bind=ctrl-z:ignore ${FZF_DEFAULT_OPTS-} -n2..,.. --scheme=history --bind=ctrl-r:toggle-sort ${FZF_CTRL_R_OPTS-} +m --read0"
-  script='BEGIN { getc; $/ = "\n\t"; $HISTCOUNT = $ENV{last_hist} + 1 } s/^[ *]//; print $HISTCOUNT - $. . "\t$_" if !$seen{$_}++'
+  script='BEGIN { RS = "\n\t"; getline; histcount = $1; getline } !seen[$0]++ { sub(/^[ *]/, ""); printf "%d\t%s\0", histcount - nr++, $0 }'
   output=$(
-    builtin fc -lnr -2147483648 |
-      last_hist=$(HISTTIMEFORMAT='' builtin history 1) perl -n -l0 -e "$script" |
+    {
+      HISTTIMEFORMAT=$'\n\t' builtin history 1 # set histcount & normalize input for RS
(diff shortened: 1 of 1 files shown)


step-:
**side note about a side issue**

The commit message mentions a small issue that I uncovered while developing this PR.  I believe the issue is unrelated to this patch.

It's a small issue, mainly cosmetic: when the first element of the history list is displayed in fzf's search window, the element ends with a dandling LF character.  To reproduce this issue I removed `~/.bash_history` in a terminal, then started bash in another terminal, ran a command, like `ls`, finally pressed C-R to start the fzf search history widget.

This commit implements a work-arond to make the dandling LF disappers from the search history window. A `\t` is appended to awk's input stream. The awk script interprets the dangling `\n` (LF) followed by `\t` as a record separator (`RS`). The end result is an empty line, which disappears in the search history window.

junegunn:
Thanks for the patch, but it doesn't work with the awk from macOS. To make it work, I had to use GNU awk (gawk) which must be separately installed.

image

step-:
I can take a look at macOS compatibility, if you're interested in this patch. Are you?
Nothing stands out to me as being gnu-ish in the awk script... it's very basic awk, tested on busybox awk, gawk --posix and mawk. But again, I can take a look at macOS. What's the version string of the failing macOS awk?

step-:
Is the failed test output by docker-test? I ran docker-test on linux only. Do I need to run docker-test also on macOS for a complete test?

junegunn:
> if you're still interested in this patch. Are you?

I don't personally need it, but if it works and it doesn't do any harm, then why not?

> What's the version string of the failing macOS awk?

```
$ awk --version
awk version 20200816
```

> Is the failed test output that you included by docker-test?

No, I just manually tested it on my macbook. Everything is shown as a single item.

image

step-:
I tried, but failed, to make this commit work with the default awk version for macOS.  Thinking out of the box, I decided to try a bash-only solution instead, which is now PR #3310.  This PR can be closed.  It works on Linux, but fails on macOS unless GNU awk (or mawk / busybox awk) is installed.

PR #3299, opened 2023-05-17
Fix bug in ripgrep-fzf example

girishji:
In the example 'Switching between Ripgrep mode and fzf mode', when file is selected for editing using Vim and when fzf was invoked inside Vim (through fzf-vim plugin's :Rg command) the preview window would contain ripgrep artifacts from remapped command from FZF_DEFAULT_COMMAND_SAVED variable.
Solution is to simply restore the original value of this variable before invoking Vim.

M  ADVANCED.md


junegunn:
Good point. Let me see if there's a simpler way to address the issue.

Issue #3309, opened 2023-05-25
[macOS] missing history item with bash 3.2


step-:
- [x] I have read through the manual page (`man fzf`)
- [x] I have the latest version of fzf
- [x] I have searched through the existing issues

## Info

- OS
    - [ ] Linux
    - [x] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [x] bash
    - [ ] zsh
    - [ ] fish

## Problem / Steps to reproduce

I reproduced this issue on macOS Catalina with bash 3.2, and Linux with bash 3.2. It matters mostly for macOS, in which 3.2 is the default bash version.

On macOS (Linux) the history applet doesn't work correctly with bash 3.2. There are two issues:
* the widget does not list the most-recently typed command;
* the numeric index shown by the widget is off by one.

To reproduce, first add some history and make sure to take the key binding definition from the [source code](https://github.com/junegunn/fzf/blob/d2bde205f0f760f46e87c913450f28bb395b6e11/shell/key-bindings.bash#L82) for bash 3.

```
bash-3.2$ : 1
bash-3.2$ bind -m emacs-standard '"\C-r": "\C-e \C-u\C-y\ey\C-u"$(__fzf_history__)"\e\C-e\er"'
bash-3.2$ <c-r>
...
27	: 1
<esc>
```
You can see that the `bind` command wasn't shown although it should have.
Now run the history widget directly instead of via the `<c-r>` key.

```
bash-3.2$ __fzf_history__
bash-3.2$ bind -m emacs-standard '"\C-r": "\C-e \C-u\C-y\ey\C-u"$(__fzf_history__)"\e\C-e\er"'
...
26	: 1
27	bind -m emacs-standard '"\C-r": "\C-e \C-u\C-y\ey\C-u"$(__fzf_history__)"\e\C-e\er"'
<esc>
```
This time `bind` was there but the widget showed different line numbers!

So the history index shown by `<c-r>` ends up being off by one, which can mislead a user recalling commands with the `!<index>` builtin.
```
bash-3.2$ : 2
bash-3.2$ : 3
bash-3.2$ : 4
bash-3.2$ <c-r>
26	bind -m emacs-standard '"\C-r": "\C-e \C-u\C-y\ey\C-u"$(__fzf_history__)"\e\C-e\er"'
27	__fzf_history__
28	: 2
29	: 3
<esc>
bash-3.2$ !29
: 4
```

[...]

PR #3310, opened 2023-05-25
replace perl with bash in the bash history widget

step-:
While awk is POSIX, perl isn't pre-installed on all *nix flavors. This commit replaces perl with a bash script to eliminate dependency on perl.

This commit passed the current test suite with zero errors :
* `make error` => all test sections 'ok'
* `make docker-test` => Finished in 52.150847s, 4.1227 runs/s, 36.0684 assertions/s. 215 runs, 1881 assertions, 0 failures, 0 errors, 0 skips

See also:  #3295 (same idea replacing awk for perl).

I manually tested this commit on macOS Catalina with bash 3.2, and Linux with bash 3.2 and 4.4. While testing on Linux I reported issue #3309.

Note: This commit needs bash support for associative arrays, available since bash 4, to filter out duplicate history items. On macOS, where 3.2 is the default bash version, fzf will not de-duplicate items unless a newer bash version is installed.

### DIFF
--- a/shell/key-bindings.bash
+++ b/shell/key-bindings.bash
@@ -49,12 +49,50 @@ __fzf_cd__() {
 }
 
 __fzf_history__() {
-  local output opts script
+  local output opts
   opts="--height ${FZF_TMUX_HEIGHT:-40%} --bind=ctrl-z:ignore ${FZF_DEFAULT_OPTS-} -n2..,.. --scheme=history --bind=ctrl-r:toggle-sort ${FZF_CTRL_R_OPTS-} +m --read0"
-  script='BEGIN { getc; $/ = "\n\t"; $HISTCOUNT = $ENV{last_hist} + 1 } s/^[ *]//; print $HISTCOUNT - $. . "\t$_" if !$seen{$_}++'
   output=$(
+    # sequence of $'\t '<multi-line-entry>$'\n'
     builtin fc -lnr -2147483648 |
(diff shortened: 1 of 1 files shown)


step-:
The integration test action is still failing on the same ruby test. I noticed another difference in the test setup: the action runs ruby tests with `LC_ALL=C`, unlike the DockerFile.

junegunn:
Thanks for working on this. But I don't think this is going to make it. I liked your previous awk version because the required code change was minimal, but that is clearly not true here; I don't want to maintain these extra code lines just for a fairly small number of users.

The real deal breaker is the performance. I have 170K+ entries in my command history, and with the original Perl version, the processing is instantaneous, but with this version, it takes longer than a minute to fully process the list, and that is not acceptable.

step-:
> Thanks for working on this. But I don't think this is going to make it.
> The real deal breaker is the performance.

I would do the same in your place.

> I liked your previous awk version because the required code change was minimal.

I liked it too better than this version but the deal breaker with macOS awk is that I know no way to ascertain the last character read in streaming mode. If I had that in my pocket the awk script could work in macOS too with minimal changes.  I will keep thinking of a way to do it with awk, which is a favorite tool of mine.

> I have 170K+ entries in my command history.

Wow. Good to know to raise to the challenge!

step-:
@glensc, TLDR; <https://github.com/junegunn/fzf/pull/3313#issuecomment-1731031927>

PR #3327, opened 2023-06-08
Add new border style: 'thinblock'

mikesmithgh:
I saw that you recently added the block border and would like to introduce a thinblock border as well. It is using the symbols from https://www.unicode.org/charts/PDF/U1FB00.pdf Symbols for Legacy Computing, ONE EIGHTH BLOCK elements.

I searched for all places with `BorderBlock` and added `BorderThinBlock`. One thing to note about the symbols is that they a font specific, and may not exist or may not align perfectly depending on the font and terminal emulator. In the case of Kitty terminal, box drawings are handled by Kitty rather than the font so it appears correctly regardless of the font.

**MonaLisa font missing the corner symbols in Alacritty:**
Alacritty MonaLisa
**Iosevka font in Alacritty**
Alacritty Iosevka
**Iosevka font in Wezterm**
Wezterm Iosevka
**Iosevka font in iTerm2**
iTerm2 Iosevka
**MonaLisa font in Kitty**
Kitty MonaLisa

MonaLisa Font: https://www.monolisa.dev/
Iosevka Font: https://github.com/ryanoasis/nerd-fonts/releases/download/v3.0.2/Iosevka.zip

Please let me know what you think. Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,7 +29,7 @@ CHANGELOG
 ------
 - Added `zero` event that is triggered when there's no match
   ```sh
- https://en.wikipedia.org/wiki/Symbols_for_Legacy_Computing#:~:text=Symbols%20for%20Legacy%20Computing%20is,and%20in%20Teletext%20broadcasting%20standards # Reload the candidate list when there's no match
+  # Reload the candidate list when there's no match
   echo $RANDOM | fzf --bind 'zero:reload(echo $RANDOM)+clear-query' --height 3
   ```
 - New actions
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,16 @@ CHANGELOG
 
 0.41.2
 ------
-- Added new border style `thinblock` which uses symbols for legacy computing [one eighth block elements](https://en.wikipedia.org/wiki/Symbols_for_Legacy_Computing)
+- Added new border style `thinblock` which uses symbols for legacy computing
+  [one eighth block elements](https://en.wikipedia.org/wiki/Symbols_for_Legacy_Computing)
+    - Similarly to `block`, this style is suitable when using a different
+      background color because the window is completely contained within the border.
+      ```sh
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -25,7 +29,7 @@ CHANGELOG
 ------
 - Added `zero` event that is triggered when there's no match
   ```sh
-  # Reload the candidate list when there's no match
+ https://en.wikipedia.org/wiki/Symbols_for_Legacy_Computing#:~:text=Symbols%20for%20Legacy%20Computing%20is,and%20in%20Teletext%20broadcasting%20standards # Reload the candidate list when there's no match
```
junegunn, line 32: An accidental edit?

junegunn:
Thanks, I like it. I would like to have a more terse name than `thinblock`, but I can't think of a better one.

Unfortunately, it's not properly rendered on iTerm2 with the default font I use, but I can confirm that it's properly supported in other emulators.

### On Hyper terminal

image

mikesmithgh:
> Thanks, I like it. I would like to have a more terse name than `thinblock`, but I can't think of a better one.
> 
> Unfortunately, it's not properly rendered on iTerm2 with the default font I use, but I can confirm that it's properly supported in other emulators.
> 
> ### On Hyper terminal
> image

Thanks. Yeah, I wasn't sold on `thinblock` either 😂. I think `one-eighth-block` would be the most accurate. But, I kind of like `narrow`. Here are a couple that came to the top of my head:

```
--border=one-eighth-block
--border=narrow-block
--border=narrow
--border=thinblock
--border=solid
--border=thin
--border=line
--border=outline
--border=edge
```

Another option is to allow a custom block and the user specifies the characters in a comma delimited list.

junegunn:
Let's just go with thinblock. I think we need to keep `block` in its name as we can see it as an alternative to the existing, more widely supported `block` and it shares the same properties of it (i.e. good for different background color)

junegunn:
Merged, thanks!

PR #3342, opened 2023-06-23
feat: optimisation of EqualMatch method

Sanix-Darker:
Testing some optimisations i wanted to check :

    - Combined whitespace stripping: The leading and trailing whitespaces are stripped in a single step, reducing the number of method calls and condition checks.

    - Reduced string conversions: Instead of converting the entire text and pattern to strings, only the necessary substrings are used. This avoids unnecessary conversions and improves performance.

    - Early return: The function returns early if the normalized characters or string comparison fails, eliminating the need for the match boolean variable and reducing unnecessary computations.

### DIFF
--- a/src/algo/algo.go
+++ b/src/algo/algo.go
@@ -902,14 +902,11 @@ func EqualMatch(caseSensitive bool, normalize bool, forward bool, text *util.Cha
 		return Result{-1, -1, 0}, nil
 	}
 
-	// Strip leading whitespaces
-	trimmedLen := 0
+	// Strip leading and trailing whitespaces
+	trimmedLen, trimmedEndLen := 0, 0
 	if !unicode.IsSpace(pattern[0]) {
 		trimmedLen = text.LeadingWhitespaces()
 	}
-
-	// Strip trailing whitespaces
-	trimmedEndLen := 0
 	if !unicode.IsSpace(pattern[lenPattern-1]) {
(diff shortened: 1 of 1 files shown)


Sanix-Darker:
Hello, I didn't create an issue for this PR, sorry for that, do you want me to provide more details with benchmarks ? @junegunn

junegunn:
> I didn't create an issue for this PR

No problem with that.

> more details with benchmarks

That would be nice, yes. But if the suggested change is succinct and it makes the code objectively better, then we can merge it even if it doesn't bring a significant performance boost.

---

Just out of curiosity, can you tell me why you started looking into improving EqualMatch? which is 1. rarely used (only triggered when you type in `^PAT$`), 2. too simple to have performance issues in real-world use cases, at least from my experience.

Sanix-Darker:
> just out of curiosity, can you tell me why you started looking into improving EqualMatch?

I am a GIGA big fan and a daily user of fzf, so I was wondering how it works, I started reading the code and I wanted to provide my optimization "point of view" on some methods for this module, but in multiple small and different PRs, that's why I started with the smallest and easiest one `EqualMatch` here!

>  2. too simple to have performance issues in real-world use cases, at least from my experience.
You're right for that, I will be glad to work on other "most important perfs issues" from your experience if you have some to provide me !

PR #3359, opened 2023-07-09
fix NONE color

blockchainian:


junegunn:
Thanks, but this is not compatible with the current model where you can specify multiple highlight groups.

```vim
let g:fzf_colors = {
\   'gutter':  ['bg', 'Conceal', 'Normal', 'Comment']
\}

highlight Normal ctermbg=233
highlight Conceal ctermbg=NONE
```

`['bg', 'Conceal', 'Normal', 'Comment']` translates to "Use the background color of the first highlight group that has a background color", and that is 233 of `Normal`. But with your patch, it will choose `-1` for `Conceal`, ignoring the rest of the list.

PR #3364, opened 2023-07-11
Check if SHELL supports FZF_DEFAULT_COMMAND

### LINKED ISSUE #3339: bash dependency?
Hi,

While installing fzf on OpenBSD, I noticed that bash was being pulled as a dependency. The mantainer of the port pointed me to [this line](https://github.com/junegunn/fzf/blob/4cd621e877cb3a8e44b12ba3a7ce58709862922f/src/reader.go#L101) where bash seems to be hardcoded. Now since fzf claims to have no dependencies in the README, I wonder if this is how it's supposed to be.

Most linux distros include bash as the default shell, so having it as a dependency doesn't really make a difference. However for OpenBSD for example that includes ksh as its default shell, having bash installed becomes useless (in my opinion) and redundant.

Would anyone please care to explain why bash is required? Are there any plans to have this dependency removed? Should the README maybe be updated?

Thank you.

thyssentishman:
When FZF_DEFAULT_COMMAND is unset, it defaults to a command that depends on 'set -o pipefail'. Since not every shell supports this option, bash was hardcoded as the default shell to be used for this specific case. With this commit, other shells that support this option are first compared to SHELL before depending/falling back to bash.

This commit closes #3339

### DIFF
--- a/src/constants.go
+++ b/src/constants.go
@@ -54,6 +54,7 @@ const (
 	defaultJumpLabels string = "asdfghjklqwertyuiopzxcvbnm1234567890ASDFGHJKLQWERTYUIOPZXCVBNM`~;:,<.>/?'\"!@#$%^&*()[{]}-_=+"
 )
 
+var supportedShells = []string {"bash", "zsh", "ksh", "ash", "hush", "yash"}
 var defaultCommand string
 
 func init() {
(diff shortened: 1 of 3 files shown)


thyssentishman:
@junegunn here is an attempt at implementing your suggestion in #3339. I'm new to this codebase and go in general, so apologies in advance for any possible mispractice.

junegunn:
Thanks, I have had an almost identical patch locally, so let me finish this up.

thyssentishman:
Awesome, thank you very much!

PR #3368, opened 2023-07-12
Feature: Feedback files

boazy:
## Summary

Provide a simple and secure method for child-processes started by fzf to send back actions that fzf should execute after they exit. This allows child processes, especially those launched with `execute` or `execute-silent`, to provide custom logic that controls the way fzf reacts.

## How does it work

This pull-requests changes FZF, so that on every execute-type action (such as `execute`, `execute-silent`,
`transform-*`, etc.), it generates a random temporary filename. The file itself is not created, but the executed process can create the file and fill in with actions (separated by either newlines or `+`s). After the executed command returns, fzf will parse the feedback file if present and enqueue the commands on `t.serverChan`, so they would be executed by fzf later.

### Existing Alternatives

The only existing alternative that I know that can be used to implement equivalent functionality is `fzf --listen`. This is a powerful features, but requires additional complexity and (more importantly) a compromise on security.
Listening on localhost without any sort of authentication exposes the user to several threats:

1. A malicious script on the internet can bypass CORS and execute arbitrary commands on a running fzf instance by employing a [DNS rebinding attack](https://en.wikipedia.org/wiki/DNS_rebinding). (This vulnerability can mitigated by limiting the host headers accepted by fzf)
2. Non-browser applications have no concept of CORS, but often allow arbitrary POST requests to arbitrary URLs which can be manipulated by remote code.
3. Any non-containerized process running on the same machine, can force the fzf process to execute arbitrary shell commands, even if the other process is running as a different, non-privileged, user. This creates an opportunity for privilege escalation.

### Security considerations

The implementation relies on the appending a random suffix to the generated feedback filename. This provides enough security due to the following reasons:

1. The random suffix is generated using 192 bits from cryptographically secure RNG (`crypto/rand`), which is secure enough against unlimited-ratio brute-forcing.
2. The feedback file is never created by fzf, so malicious processes cannot scan for the file.
3. Even if a malicious process is somehow fast enough to find the feedback file after it was created by the fzf child process, but before it is consumed by fzf, it cannot write to that that file unless it is given write permission by the child process which created it.
4. The feedback filename is exposed through an environment variable, so it is available only to fzf child processes.

### Navigate directory hierarchy

This example builds a very simple file browser that can navigate into directories (including '..') when selecting them with enter, while selecting a file just accepts it:

```bash
read -r -d '' RELOAD_OR_ACCEPT_CMD <<-'EOF'
	ITEM='{}'
	if [[ $ITEM == ".." ]]; then
		dirname "$(<$CWD_FILE)" > $CWD_FILE
		echo 'reload(echo .. ; ls -1p "$(<$CWD_FILE)")' > $FZF_FEEDBACK_FILE
	elif [[ $ITEM == */ ]]; then
		echo "$(<$CWD_FILE)/${ITEM%/}" > $CWD_FILE
		echo 'reload(echo .. ; ls -1p "$(<$CWD_FILE)")' > $FZF_FEEDBACK_FILE
	fi
	else
		echo accept > $FZF_FEEDBACK_FILE
	fi
EOF

(echo .. ; ls -1p) | \
	CWD_FILE=$(mktemp | tee >(pwd > $(cat))) \
	./fzf \
		--reverse \
		--bind="enter:execute-silent@$RELOAD_OR_ACCEPT_CMD@"
```

### Context menus with embedded fzf

The example below implements a very simple context menu that lets you either accept or copy the selected item.

```bash
fzf --bind 'tab:execute[echo "Accept\taccept\nCopy\texecute-silent(echo {} | pbcopy)+abort" | fzf -d"\t" --with-nth=1 | awk -F "\\t" "{ print \$2 }" > $FZF_FEEDBACK_FILE]'
```

This toy example betrays the main driver behind my pull-request. I would be happy if I could implement Alfred-like action menu inside FZF.

### DIFF
--- a/src/terminal.go
+++ b/src/terminal.go
@@ -2,6 +2,8 @@ package fzf
 
 import (
 	"bufio"
+	"crypto/rand"
+	"encoding/base32"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -9,6 +11,7 @@ import (
 	"os"
 	"os/exec"
 	"os/signal"
+	"path"
 	"regexp"
 	"sort"
 	"strconv"
@@ -250,6 +253,7 @@ type Terminal struct {
 	theme              *tui.ColorTheme
 	tui                tui.Renderer
 	executing          *util.AtomicBool
+	feedbackFilename   string
 }
 
 type selectedItem struct {
@@ -721,6 +725,9 @@ func (t *Terminal) environ() []string {
 	env := os.Environ()
 	if t.listenPort != nil {
 		env = append(env, fmt.Sprintf("FZF_PORT=%d", *t.listenPort))
+	}
+	if t.feedbackFilename != "" {
+		env = append(env, fmt.Sprintf("FZF_FEEDBACK_FILE=%s", t.feedbackFilename))
 	}
 	return env
 }
@@ -2377,6 +2384,7 @@ func (t *Terminal) executeCommand(template string, forcePlus bool, background bo
 	if !valid && !capture {
 		return line
 	}
+	t.setupFeedback()
 	command := t.replacePlaceholder(template, forcePlus, string(t.input), list)
 	cmd := util.ExecCommand(command, false)
 	cmd.Env = t.environ()
@@ -2408,6 +2416,7 @@ func (t *Terminal) executeCommand(template string, forcePlus bool, background bo
 		}
 	}
 	t.executing.Set(false)
+	t.consumeFeedback()
 	cleanTemporaryFiles()
 	return line
 }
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks for the suggestion, but I would take a different approach.

Setting up and cleaning up the file whenever running an external command feels a bit wasteful. For example, few would want to perform extra side-effects on `transform-prompt`.

I would instead add a separate action just for the purpose (not sure about the name, for some reason, `feedback` doesn't really click with me. `act(...)`, `perform(...)`, `execute-perform(...)`, `execute-effect(...)`, `transform(...)`, `transform-state(...)`, ... 🤔 ) where you don't have to write to a file, but you can just `echo` the list of actions to the standard output, and fzf will run them on completion.

```sh
fzf --bind 'tab:transform:echo "Accept\taccept\nCopy\tbecome:pbcopy <<< {}" | fzf -d"\t" --with-nth=1 | awk -F "\\t" "{ print \$2 }"'
```

---

And thanks for pointing out the security concerns for `--listen`. I personally wouldn't worry too much about it, but those are valid points. We can maybe consider adding `--safe-listen` which forbids process execution.

boazy:
Thank you for the feedback.

I wanted to add a new command at first, but I thought you might be against adding new commands to fzf. I agree that this feature has very little value for most execute-like commands, but I think I will still need to support at least a version for `execute`, and `execute-silent` (after all, I gave useful example for both), so that means adding two actions. Perhaps `act` and `act-silent` or `perform` and `perform-silent`?

Regarding the feedback file - capturing the commands to execute from stdout would work perfectly for the silent version, but not for the non-silent one. For instance, this would not work:

```
fzf --bind 'tab:transform:echo "Accept\taccept\nCopy\tbecome:pbcopy <<< {}" | fzf -d"\t" --with-nth=1 | awk -F "\\t" "{ print \$2 }"'
```

Since the second fzf launched inside the action, needs stdout to display itself. There needs to be some way to separate the command's normal output from the list of actions to execute. That can either be a named pipe, unnamed pipe (possibly passing a write fd to the child process?), a socket or a file. From all options a file seemed to be the cleanest and easiest to use option to me. Standard output is cleaner of course, but it just won't work on the non-silent case.

----

Regarding making listen safer, the best option in my opinion is to allow the user to set an optional API Key. If they can specify an API key using a command line argument or an environment variable and then fzf would require requests to specify the same key in the `X-API-Key` (or non-standared `Authorization`?) header, then there should be no way to exploit fzf without knowing that key. All you need to do to be safe is:

```shell
FZF_API_KEY=$(base64 < /dev/urandom) fzf --listen 
```

The advantage of an environment variable over a command line argument, is that it is naturally accessible by all child processes (just like `FZF_PORT`). Child processes should probably be ok to access fzf in almost every conceivable case.

junegunn:
To be honest, I'm not confident fzf really needs this feature because fzf is basically an interactive filter rather than an omnipotent framework for sophisticated TUI programs, while it's tempting to see it that way (sadly because of the feature bloat fzf has gone through, should we keep bloating it even more?). Maybe you're expecting too much out of it.

>  I would be happy if I could implement Alfred-like action menu inside FZF.

I'm not sure if you'll be happy with the result even if you manage to do it. You'll see a flickering of the screen when switching between instances of fzf. If you're on tmux, you can try using fzf-tmux like so

```sh
fzf --bind 'enter:execute-silent:seq 10 | fzf-tmux -p'
```

> ```shell
> FZF_API_KEY=$(base64 < /dev/urandom) fzf --listen
> ```

I like this approach better (making the existing feature safer) than adding a yet another similar mechanism.

boazy:
> To be honest, I'm not confident fzf really needs this feature because fzf is basically an interactive filter rather than an omnipotent framework for sophisticated TUI programs, while it's tempting to see it that way (sadly because of the feature bloat fzf has gone through, should we keep bloating it even more?). Maybe you're expecting too much out of it.

I'm not expecting fzf to be a sophisticated framework for TUI programs, of course. I can limit this PR for to add a single action that goes through stdout (similar to execute-silient).


> I'm not sure if you'll be happy with the result even if you manage to do it. You'll see a flickering of the screen when switching between instances of fzf. If you're on tmux, you can try using fzf-tmux like so.

I wasn't very happy with the result, but it worked. I had a few ideas to get around that, like reloading the context menu and then reloading the list again if necessary, but they have their own issues. I also thought about `tmux`, but I have been avoiding it so far, since I never managed to get used to it.

> I like this approach better (making the existing feature safer) than adding a yet another similar mechanism.
This is pretty simple to do. I will send a PR.

PR #3369, opened 2023-07-14
avoid history activation error

cherkavi:
avoid next error
```
unknown option: --scheme=history
```


junegunn:
You should upgrade your fzf installation.

PR #3370, opened 2023-07-14
Fix deprecations of ioutil

barthr:
This PR addresses the recent deprecations in Go 1.16. I replaced them to the suggested versions (some of these methods where already under the hood calling the correct methods).

### DIFF
--- a/src/history.go
+++ b/src/history.go
@@ -2,7 +2,6 @@ package fzf
 
 import (
 	"errors"
-	"io/ioutil"
 	"os"
 	"strings"
 )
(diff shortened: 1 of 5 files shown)


junegunn:
Thanks, I wasn't aware of the deprecation. Did you use some kind of a linter or did you manually look for `ioutil` references?

barthr:
I did look for manual ioutil references. However, the trigger came from a static analyzer (the one included in intelij, so not aware which command line tool provides this functionality). 

To add to that, I think using something like https://github.com/golangci/golangci-lint can be quite beneficial to the project, since it runs quite some linters. This can also be added as ci step to run on every PR. If this is something you are interest in, I can make a proof of concept PR.

PR #3374, opened 2023-07-17
Add API Keys for fzf --listen

boazy:
## Summary

Provide optional hardened security for the FZF server. If the `FZF_API_KEY` environment variable is set, FZF would refuse any request that does not contain the specified API key. The API key can be any string, and should be the specified using the `x-api-key` header.

## Background

`fzf --listen` currently accepts all requests, without any kind of authentication. This exposes the user to several threats:

1. A malicious script on the internet can bypass CORS and execute arbitrary commands on a running fzf instance by employing a [DNS rebinding attack](https://en.wikipedia.org/wiki/DNS_rebinding).
2. Non-browser applications have no concept of CORS, but often allow arbitrary POST requests to arbitrary URLs which can be manipulated by remote code.
3. Any non-containerized process running on the same machine, can force the fzf process to execute arbitrary shell commands, even if the other process is running as a different, non-privileged, user. This creates an opportunity for privilege escalation on multi-user machines.

See #3368 for more details.

## Alternative Approaches

1. We can protect against the DNS rebinding vulnerability by blocking any request that doesn't specify `localhost` in the `Host` header. But this approach would only protect fzf against DNS rebinding, and it could also interfering with legitimate use cases where you want to bind another domain to the loopback interface.

2. The API key could be made mandatory, and randomly auto-generated if not explicitly specified. This approach would break all current usage of the `--listen` flag.

3. The API key could be specified with a new command line argument (e.g. `--listen-key` or `--listen-api-key`). I believe using an environment variable is cleaner, since it naturally propagates to child processes.

4. A different header name could be used. There are no standard headers for API authorization, but `x-api-key` is the closest thing we have to a popular choice for API keys (i.e. single-credential, directly-supplied secret keys for calling an API).

5. We could use the standard `Authorization` header, but it offers no standard scheme for API keys. The `bearer` scheme is meant for tokens (especially OAuth 2 tokens), while the `Basic` scheme requires two values (username and password) which makes it more complicated to use.

### DIFF
--- a/src/server.go
+++ b/src/server.go
@@ -3,9 +3,11 @@ package fzf
 import (
 	"bufio"
 	"bytes"
+	"crypto/subtle"
 	"errors"
 	"fmt"
 	"net"
+	"os"
 	"strconv"
 	"strings"
 	"time"
@@ -15,9 +17,15 @@ const (
 	crlf             = "\r\n"
 	httpOk           = "HTTP/1.1 200 OK" + crlf
 	httpBadRequest   = "HTTP/1.1 400 Bad Request" + crlf
+	httpUnauthorized = "HTTP/1.1 401 Unauthorized" + crlf
 	httpReadTimeout  = 10 * time.Second
 	maxContentLength = 1024 * 1024
 )
+
+type httpServer struct {
+	apiKey  []byte
+	channel chan []*action
+}
 
 func startHttpServer(port int, channel chan []*action) (error, int) {
 	if port < 0 {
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks! We need to mention this variable on the man page. Also can you add a test case (in test_go.rb)? Let me know if you need any assistance.

junegunn:
Thanks, let me finish this up.

junegunn:
Merged, thanks!

boazy:
Thank you!

PR #3382, opened 2023-07-26
key-bindings.bash: disable pipefail in Alt+C

ardnew:
With `set -o pipefail` enabled globally, the Alt+C keybinding silently fails without changing directory or indicating an error.

This change temporarily disables the shell option locally within the scope of `__fzf_cd__()`, as expected by the existing pipelines, and then restores it to the user's original setting prior to invoking Alt+C.

Requires Bash 4.4 at least, which is more than safe to assume here :)

> Reference: https://unix.stackexchange.com/a/406332

------
Proof that this technique works can be demonstrated with:
```bash
check(){ echo $*: $(set -o|grep pipefail); }
f() { check at enter f; local -; set +o pipefail; check at exit f; }
check before call f
f
check after call f
```

### DIFF
--- a/shell/key-bindings.bash
+++ b/shell/key-bindings.bash
@@ -14,7 +14,8 @@
 # Key bindings
 # ------------
 __fzf_select__() {
-  local cmd opts
+  local cmd opts -
+  set +o pipefail # restored on return because we declared $- local above
   cmd="${FZF_CTRL_T_COMMAND:-"command find -L . -mindepth 1 \\( -path '*/\\.*' -o -fstype 'sysfs' -o -fstype 'devfs' -o -fstype 'devtmpfs' -o -fstype 'proc' \\) -prune \
     -o -type f -print \
     -o -type d -print \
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks.

> Alt+C keybinding silently fails without changing directory or indicating an error.

If I'm not mistaken, it doesn't work only when an entry is selected while the list is still being populated by the underlying command because `eval "$cmd"` is interrupted.

https://github.com/junegunn/fzf/blob/dd59b8c7b91082a4ae3a669979980abfaa3b470e/shell/key-bindings.bash#L48

It works well if `$cmd` is complete.

> Requires Bash 4.4 at least

Unfortunately, we can't enforce this because macOS ships with bash 3 due to its GPL license and we don't want to stop supporting the default bash version of macOS.

Having said that, an easy fix would be to make `$cmd` not fail like so:

```diff
diff --git a/shell/key-bindings.bash b/shell/key-bindings.bash
index d83f9d3..58bb1db 100644
--- a/shell/key-bindings.bash
+++ b/shell/key-bindings.bash
@@ -45,7 +45,7 @@ __fzf_cd__() {
   cmd="${FZF_ALT_C_COMMAND:-"command find -L . -mindepth 1 \\( -path '*/\\.*' -o -fstype 'sysfs' -o -fstype 'devfs' -o -fstype 'devtmpfs' -o -fstype 'proc' \\) -prune \
     -o -type d -print 2> /dev/null | cut -b3-"}"
   opts="--height ${FZF_TMUX_HEIGHT:-40%} --bind=ctrl-z:ignore --reverse ${FZF_DEFAULT_OPTS-} ${FZF_ALT_C_OPTS-} +m"
-  dir=$(eval "$cmd" | FZF_DEFAULT_OPTS="$opts" $(__fzfcmd)) && printf 'builtin cd -- %q' "$dir"
+  dir=$(eval "$cmd || :" | FZF_DEFAULT_OPTS="$opts" $(__fzfcmd)) && printf 'builtin cd -- %q' "$dir"
 }
 
 __fzf_history__() {
```

junegunn:
Hi, I've decided to disable `pipefail` only inside the command substitutions where it can be a problem. Thanks.

PR #3403, opened 2023-08-17
Change Winget Releaser job to `ubuntu-latest`

sitiom:
Winget Releaser now supports non-Windows runners, and `ubuntu-latest` is generally faster.


junegunn:
Thanks!

Issue #3416, opened 2023-08-27
Context aware completion for ssh command


rebrendov:
- [x] I have read through the manual page (`man fzf`)
- [x] I have the latest version of fzf
- [x] I have searched through the existing issues

## Info

- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [ ] bash
    - [x] zsh
    - [ ] fish

## Problem / Steps to reproduce
1) Extend completion for ssh command so it offers matches based on the flags provided. 
2) Change completion behavior when username is provided as mentioned in [this](https://github.com/junegunn/fzf/discussions/3277) discussion.

Here's some demonstration of the old vs new behavior
***Old***
[![asciicast](https://asciinema.org/a/MAPO0csRAQ2Oo8IEpFrqt4af4.svg)](https://asciinema.org/a/MAPO0csRAQ2Oo8IEpFrqt4af4)
***New***
[![asciicast](https://asciinema.org/a/fIOUO20N8yRjDZnwuDLrBgbsM.svg)](https://asciinema.org/a/fIOUO20N8yRjDZnwuDLrBgbsM)

I already wrote a zsh implementation for personal use. If you think this behavior worth adding to the main repo I'll add implementations for bash and fish and create a pull request. Otherwise feel free to close this issue (I've read your comments on other PRs and know that you aim to keep complexity as low as possible).

junegunn:
That looks great. Thanks for sharing.

> If you think this behavior worth adding to the main repo

It depends on the complexity of the code. Can you show me your code?

junegunn:
This one?

https://github.com/rebrendov/fzf/commit/e243f87ccf27685d1a754f92abc3bdfa5908ca67

The code looks not too complicated and quite manageable. I think we can give it a try.

Issue #2366, opened 2021-02-21
Customize ssh known_hosts source


HsuanTingLu:
- [x] I have read through the manual page (`man fzf`)
- [x] I have the latest version of fzf
- [x] I have searched through the existing issues

## Info

- OS
    - [ ] Linux
    - [x] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [ ] bash
    - [ ] zsh
    - [ ] fish
    
## Problem / Steps to reproduce

fzf designates the default `~/.ssh/known_hosts` file in [line 287](https://github.com/junegunn/fzf/blob/a367dfb22e63e41b52f68056a7b3325bab3ba180/shell/completion.bash#L287) in `completion.bash`
https://github.com/junegunn/fzf/blob/a367dfb22e63e41b52f68056a7b3325bab3ba180/shell/completion.bash#L284-L291

I'm wondering if we can add a config variable such as `FZF_KNOWN_HOSTS` to allow custom known_hosts files like `google_compute_known_hosts` or `school_lab_known_hosts` to be searched too

PR #3424, opened 2023-09-04
Context aware completion for ssh command

rebrendov:
Solves #3416. 

Changes for zsh were tested on ArchLinux machine and changes for bash were tested on ArchLinux and Ubuntu Server machines.

For bash had to copy-paste `_fzf_complete` call from `_fzf_host_completion` to `_fzf_ssh_completion` to make sure that `_fzf_ssh_completion_post` is called.


rebrendov:
Seems that new logic broke the tests. I'll try to fix it.

rebrendov:
Ok, made sure `make docker-test` is passing.

shell/completion.bash
```diff
@@ -237,8 +237,10 @@ _fzf_complete() {
   cur="${COMP_WORDS[COMP_CWORD]}"
   if [[ "$cur" == *"$trigger" ]]; then
     cur=${cur:0:${#cur}-${#trigger}}
+    [ -z "${prefix:-}" ] && prefix="${cur}"
+    prefix=${prefix%% }
 
-    selected=$(FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} --reverse --bind=ctrl-z:ignore ${FZF_DEFAULT_OPTS-} ${FZF_COMPLETION_OPTS-} $str_arg" __fzf_comprun "${rest[0]}" "${args[@]}" -q "$cur" | $post | tr '\n' ' ')
+    selected=$(FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} --reverse --bind=ctrl-z:ignore ${FZF_DEFAULT_OPTS-} ${FZF_COMPLETION_OPTS-} $str_arg" __fzf_comprun "${rest[0]}" "${args[@]}" -q "$prefix" | $post | tr '\n' ' ')
```
junegunn, line 243: Not sure I follow the logic here. Can you explain why you introduced `$prefix`?

junegunn: Bash version doesn't seem to work as advertised.

image

Zsh version does work, but I'm getting warning messages.

image

junegunn: Also, `ssh -i **<tab>` correctly brings up a list of files, but `ssh -i**<tab>` doesn't.

rebrendov: That's weird, your bash completion demonstrates prior behavior. Maybe old version of shell/completion.bash got sourced in your terminal after the new one somehow? Can you check, and if it's not the case provide more details about your bash setup? (list of options with `echo $-` would be great)
bash-test

As for the `ssh -i**<tab>` for me this looks like I'm trying to find completion for a flag, not for an argument, so I didn't really thought about handling this case. If you think this is a valid use-case I can add some logic to check, if the word being completed is a flag or not and then handle it the same way as `ssh -i **<tab>`.

I also was thinking about putting some tests for the ssh completion in the `test/test_go.rb` but for that I would need to add an option to specify additional custom path for `known_hosts`. What do you think?

By the way,

junegunn: I don't understand how it's working on your machine, because if I'm not mistaken, bash doesn't support non-greedy quantifiers which you used in line 305.

```sh
jg@:~/github/fzf (rebrendov/master)> echo $BASH_VERSION
5.1.16(1)-release
jg@:~/github/fzf (rebrendov/master)> echo $-
himBHs
jg@:~/github/fzf (rebrendov/master)> [[ foobar =~ fo+?b ]] && echo 1 || echo nope
nope
jg@:~/github/fzf (rebrendov/master)> [[ foobar =~ fo+b ]] && echo 1 || echo nope
1
```

rebrendov: I'm not sure why it worked on my systems, but from what I found in docs it seems like bash uses regex implementation from libc. So the reason of working on my end and not working on yours might be just a different versions of libc (for me it's `ldd (Ubuntu GLIBC 2.35-0ubuntu3.1) 2.35`, `ldd (GNU libc) 2.38` and `ldd (Ubuntu GLIBC 2.31-0ubuntu9.1) 2.31`).

To avoid inconsistencies I got rid of matching groups entirely left just `=~ '@'` check.

junegunn: > I also was thinking about putting some tests for the ssh completion in the test/test_go.rb but for that I would need to add an option to specify additional custom path for known_hosts for ssh_completion. That would solve #2366.

You know I'm hesitant to add options, and I need more careful consideration before making the decision. Having said that, I imagine you can still write a set of tests for the new behavior because most systems should have `127.0.0.1` in the list.

junegunn: > To avoid inconsistencies I got rid of matching groups entirely left just `=~ '@'` check.

Thanks, it's working as expected now. I'll review the code in more detail and see if there's any room for improvement.

junegunn:
I tried to simplify the code and make both implementations as similar as possible.

* Prepend the user part to the input so the post function is not required
* [bash] Remove the reliance on `$prefix` variable across the functions
* [bash] Fix the bug where the part after `@` is lost. e.g. `ssh jg@10**<tab>` 

Let me know what you think.

junegunn:
Merged with a test case. Thanks.

rebrendov:
Sorry, I forgot to send a comment. The simplified code looks great. Thank you for a useful tool!

PR #3434, opened 2023-09-10
Makefile: build 32-bit binary on armv8l

thesamesam:
armv8l is always 32-bit and should implement the armv7 ISA, so just use the same filename as for armv7.


junegunn:
Can you explain why you want to change it?

thesamesam:
Does the improved commit message help?

junegunn:
Absolutely, thanks.

FWIW, [the official binaries](https://github.com/junegunn/fzf/releases/tag/0.42.0) are built using [goreleaser](https://github.com/rebrendov/fzf/blob/master/.goreleaser.yml). It doesn't use the Makefile.

PR #3444, opened 2023-09-23
[shell] Don’t print (and strip) the `./`-prefix

calestyo:
Since `find` is always performed on only one starting point (namely `.`), its `-printf`-option together with the `%P`-directive can be used to print the pathname without the leading `./`-prefix in the first place.

This makes stripping them (via `cut` respectively `sed`) unnecessary and should give some performance improvement.

### DIFF
--- a/shell/key-bindings.bash
+++ b/shell/key-bindings.bash
@@ -16,9 +16,9 @@
 __fzf_select__() {
   local cmd opts
   cmd="${FZF_CTRL_T_COMMAND:-"command find -L . -mindepth 1 \\( -path '*/\\.*' -o -fstype 'sysfs' -o -fstype 'devfs' -o -fstype 'devtmpfs' -o -fstype 'proc' \\) -prune \
-    -o -type f -print \
-    -o -type d -print \
-    -o -type l -print 2> /dev/null | cut -b3-"}"
+    -o -type f -printf '%P\\n' \
+    -o -type d -printf '%P\\n' \
+    -o -type l -printf '%P\\n' 2> /dev/null"}"
   opts="--height ${FZF_TMUX_HEIGHT:-40%} --bind=ctrl-z:ignore --reverse --scheme=path ${FZF_DEFAULT_OPTS-} ${FZF_CTRL_T_OPTS-} -m"
   eval "$cmd" |
     FZF_DEFAULT_OPTS="$opts" $(__fzfcmd) "$@" |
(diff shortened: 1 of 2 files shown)


calestyo:
Beware that I have not checked this especially for `fish` and `zsh`. ;-)

btw: I just see that you also do the `sed 's@^\./@@'` in the completion scripts, just as you did in `key-bindings.fish` (where I dropped it now).
So I guess we should change these places to use `-printf` as well, especially the `sed` expression should likely be `'s@^\\./@@'` and not just `'s@^\./@@'` (though it makes no difference, except perhaps performance, as the first character would always be `.` anyway.

calestyo:
Not so sure about the completion scripts...:
https://github.com/junegunn/fzf/blob/00809909ae586a56dcf97060a361b2c95aa4aa22/shell/completion.bash#L16-L30

You only strip a leading `./`, but `find`’s starting point is `$1` in each case... and I presume this may be directories other than `.`, too?
If so, things should of course stay there (i.e. the completion scripts) as they are.

calestyo:
I've also reverted the change for `key-bindings.fish` in which `find` seems to have starting point directories other than `.`.<br>
Still, the `sed` pattern should probably be changed there to use `'s@^\\./@@'`, cause otherwise if `$dir` is a one letter dir, it would remove it, even if it's not `.` - guess that shouldn't happen, should it?

LangLangBart:
- will be problematic for macOS users, who use the a different version of `find` by default
  - according to the man page the `-printf` option is not supported
  - [macOS14.0.sdk find.1](https://github.com/itsnebulalol/sdks/blob/main/macOS14.0.sdk/usr/share/man/man1/find.1)

calestyo:
Uhm... well it's always problematic to support unusual platforms that are in many cases not even POSIX compatible.

Question is, how much should others "suffer" (okay, here it's not really suffering ^^) to suit a few platforms that user their own (and incompatible) implementations of widely used tools.

Also, looking at e.g. https://ss64.com/osx/printf.html (which I assume is `printf(1)`??), Apple doesn't even seem to support `%q` in `printf`... which is however used by `key-bindings.bash` ... so would that even work right now on Apple?

LangLangBart:
> Edit: Okay the last point is moot, as printf would be bash’s built-in printf.

```sh
man printf
# Some shells may provide a builtin printf command which is similar or identical to 
# this utility.  Consult the builtin(1) manual page.
```

- `zsh` (default shell on macOS) covers `%q` as well
  - https://zsh.sourceforge.io/Doc/Release/Shell-Builtin-Commands.html#index-printf 

---

> how much should others "suffer" (okay, here it's not really suffering ^^)
- it is wrong to inflict unnecessary suffering on macOS users
- Instead, one can assign the `FZF_…COMMAND` environment variables to another command, such as `fd` with the `--strip-cwd-prefix` option. This allows users seeking improved performance to find a workaround, and ensures that `fzf` remains functional for macOS users

junegunn:
Thanks but we don't accept changes that would degrade the user experience. For years, we have taken great care to support both macOS and Linux even if it means using less optimal and more verbose code.

(FWIW, I'm a macOS user and I have developed fzf mostly on it.)

calestyo:
@junegunn You've seen my comments above, that the `sed` pattern might still be wrong as it is now?

calestyo:
I mean this one:
https://github.com/junegunn/fzf/blob/76364ea767cca7ca8f6570a049fbb8d3fff751a9/shell/key-bindings.fish#L31

Depends on how `fish` does it's escaping, if that's like in POSIX, it should be fine, though.

PR #3447, opened 2023-09-23
[shell] don’t needlessly escape `.` in shell pattern

calestyo:
`find`’s `-path`-option is described to use shell patterns (i.e. POSIX’ pattern matching notation).

In that, `.` is not a special character, thus escaping it shouldn’t be necessary.

### DIFF
--- a/shell/key-bindings.bash
+++ b/shell/key-bindings.bash
@@ -15,7 +15,7 @@
 # ------------
 __fzf_select__() {
   local cmd opts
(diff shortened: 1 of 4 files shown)


calestyo:
Again, I've only checked this with `bash`, where AFAICS, it works the same without the (IMO) needless escaping of the `.`,… but please double check. :-)

junegunn:
I think you're right. I don't remember why I escaped it in the first place, it was a long time ago. Probably by a mistake. Thanks.

junegunn:
Merged, thanks!

Issue #580, opened 2016-05-28
C-r is slow with `set blink-matching-paren on` in .inputrc


ptzz:
- Category
    - [ ] fzf binary
    - [ ] fzf-tmux script
    - [x] Key bindings
    - [ ] Completion
    - [ ] Vim
    - [ ] Neovim
    - [ ] Etc.
- OS
    - [ ] Linux
    - [x] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [x] bash
    - [ ] zsh
    - [ ] fish

There is a delay when accessing history with `C-r`, if `set blink-matching-paren on` is in .inputrc.
The reason seems to be the parentheses in the bound command: `bind '"\C-r": " \C-e\C-u$(__fzf_history__)\e\C-e\e^\er"'`.

I have tried to resolve it using `bind -x` (to execute the command silently so there are no parentheses to match), but the errors I get then is beyond my readline and bash level.

junegunn:
Thanks for the report. Hmm, an easy fix would be to change the command substitutions to use backticks instead of `$(...)`. What do you think? Interestingly the change is not needed for `vi`-mode where `blink-matching-paren` doesn't seem to work.

```diff
diff --git a/shell/key-bindings.bash b/shell/key-bindings.bash
index 5ff20bb..41e707e 100644
--- a/shell/key-bindings.bash
+++ b/shell/key-bindings.bash
@@ -67,7 +67,7 @@ __fzf_use_tmux__() {
 [ $BASH_VERSINFO -gt 3 ] && __use_bind_x=1 || __use_bind_x=0
 __fzf_use_tmux__ && __use_tmux=1 || __use_tmux=0
 
-if [[ $'\n'$(set -o) != *$'\n'vi*on* ]]; then
+if [[ ! -o vi ]]; then
   # Required to refresh the prompt after fzf
   bind '"\er": redraw-current-line'
   bind '"\e^": history-expand-line'
@@ -82,10 +82,10 @@ if [[ $'\n'$(set -o) != *$'\n'vi*on* ]]; then
   fi
 
   # CTRL-R - Paste the selected command from history into the command line
-  bind '"\C-r": " \C-e\C-u$(__fzf_history__)\e\C-e\e^\er"'
+  bind '"\C-r": " \C-e\C-u`__fzf_history__`\e\C-e\e^\er"'
 
   # ALT-C - cd into the selected directory
[...]

PR #3448, opened 2023-09-23
Various bash improvements

calestyo:
Again... someone should thoroughly double check these, especially I have no `bash` versions from the primordial times of the universe around. ;-)

Speaking of which... does it really make sense to keep code for `bash` versions < 4? I mean bash 4 was something like early 2009.<br>
I'd just drop all code for those ancient versions, especially as all the use of `\C-b\C-k`, etc.  may anyway easily break, if a user has re-assigned them to something else.

### DIFF
--- a/shell/completion.bash
+++ b/shell/completion.bash
@@ -12,7 +12,7 @@
 if [[ $- =~ i ]]; then
 
 # To use custom commands instead of find, override _fzf_compgen_{path,dir}
-if ! declare -f _fzf_compgen_path > /dev/null; then
+if ! declare -F _fzf_compgen_path > /dev/null; then
   _fzf_compgen_path() {
     echo "$1"
     command find -L "$1" \
@@ -21,7 +21,7 @@ if ! declare -f _fzf_compgen_path > /dev/null; then
   }
 fi
 
-if ! declare -f _fzf_compgen_dir > /dev/null; then
+if ! declare -F _fzf_compgen_dir > /dev/null; then
   _fzf_compgen_dir() {
     command find -L "$1" \
       -name .git -prune -o -name .hg -prune -o -name .svn -prune -o -type d \


calestyo:
btw: There's also one further difference in the 3:
https://github.com/junegunn/fzf/blob/00809909ae586a56dcf97060a361b2c95aa4aa22/shell/key-bindings.bash#L111
and
https://github.com/junegunn/fzf/blob/00809909ae586a56dcf97060a361b2c95aa4aa22/shell/key-bindings.bash#L132
use:
```
" \C-b…`__fzf_select__`\e…"
```
while:
https://github.com/junegunn/fzf/blob/00809909ae586a56dcf97060a361b2c95aa4aa22/shell/key-bindings.bash#L116
uses
```
"\C-e …"$(__fzf_history__)"\e…"
```

So no extra leading ` ` and some extra `"` (apart from the different style of command substitution)… is that by intention?

calestyo:
@junegunn … any ideas on these? :-D

calestyo:
Added another minor improvement.

junegunn:
Using `$(...)` instead of backticks was a mistake. We intentionally use backticks to avoid delay from `blink-matching-paren`.

See https://github.com/junegunn/fzf/issues/580

Fixed it on master: https://github.com/junegunn/fzf/commit/a3ff49aaf18eceefe4bf8b1df5bf275b3d982493

junegunn:
> Speaking of which... does it really make sense to keep code for bash versions < 4?

Yes, because macOS ships with bash 3. https://apple.stackexchange.com/a/208405/125495

The code is short and it has been stable for quite a while. And considering that I have no plans to extend the feature set, I don't see much benefit in pruning a few lines of code and breaking backward compatibility for some, albeit small, number of users.

calestyo:
> Using $(...) instead of backticks was a mistake. We intentionally use backticks to avoid delay from blink-matching-paren.

Funny... did you ever find out, why they behave different there?

> Fixed it on master: a3ff49a

Thx.

> Yes, because macOS ships with bash 3. https://apple.stackexchange.com/a/208405/125495

Sigh ^^... but okay for me.


I've rebased my branch onto current master and dropped the now obsolete commit that touched command substitution.

Could you please review the remaining commits? :-)

shell/key-bindings.bash
```diff
@@ -126,11 +133,11 @@ else
   bind -m emacs-standard -x '"\C-r": __fzf_history__'
   bind -m vi-command -x '"\C-r": __fzf_history__'
   bind -m vi-insert -x '"\C-r": __fzf_history__'
-fi
 
-# ALT-C - cd into the selected directory
-bind -m emacs-standard '"\ec": " \C-b\C-k \C-u`__fzf_cd__`\e\C-e\er\C-m\C-y\C-h\e \C-y\ey\C-x\C-x\C-d"'
-bind -m vi-command '"\ec": "\C-z\ec\C-z"'
-bind -m vi-insert '"\ec": "\C-z\ec\C-z"'
```
junegunn, line 134: We decided to keep the pre-bash4 implementation for ALT-C because of a critical UX issue. See https://github.com/junegunn/fzf/pull/2674.

Also, your version doesn't work and the test cases are failing. Have you tested it?

calestyo: Oops... I haven't noticed that it doesn't change the prompt, and was since then working on other branches.

I think I may have a solution for the issue, namely doing something like:
```sh
__fzf_cd__() {
  …
  dir=$(set +o pipefail; eval "$cmd" | FZF_DEFAULT_OPTS="$opts" $(__fzfcmd)) && builtin cd "$dir"
  bind '"\e[0n": "\C-j"' 2> /dev/null
  printf '\e[5n'
}
```
of course that's just a sketch and would require to restore the original `bind`.

Do you think that could be a way to go, or are there any issues with it, that I don't see right now?

But speaking of which, AFAICS, all he current code that uses the `\e[0n`/`\e[5n` trick sets the bind only once. Wouldn't it be generally better to set it again every time? I mean in case there are other scripts loaded, which use it too?


As for the failing tests... could be that some of the tests were really failing because of that bug, but I've had generally failing/flaky tests, even on `master` (see https://github.com/junegunn/fzf/pull/3449#issuecomment-1734544997).

calestyo: With the above implementation idea, a directory change looks like this:
```
calestyo@host:~$ 
calestyo@host:~/dir$ 
```
That is, the line where `Alt-C` was used is empty and the one afterwards has magically a new CWD.

I'd prefer if it would print `cd dir` in that line, instead of being empty, so that one can follow the transcript more easily.

But so far I'm only able to get it printed before the prompt :-/

calestyo: The above has of course still the problem that `C-j` may be bound to something else :-(

calestyo: Ooops... I hadn't noticed that this didn’t change the CWD, and worked on other branches since then and not looked back.

As for the tests, I've seen failures and flaky tests even when working on `master`, see https://github.com/junegunn/fzf/pull/3449#issuecomment-1734544997.
Not sure what I'm doing wrong there.

Okay, using `bind -x` seems indeed not so easy. `redraw-current-line` only re-draws the line, but does not re-evaluate `PS1`, so while the CWD has in fact changed, the old one is still shown in the prompt.<br>
Anything that causes enter to be printed wouldn’t be much better than the current solution (and cause the current line to be executed.

While searching for solutions, I stumbled over `${PS1@P}` - not sure if that could somehow be used to re-draw the prompt manually.<br>
Some people suggested sending `SIGWINCH` to the shell, but that didn’t work for me.

https://superuser.com/a/1662149 worked for me, but it would also require us override more bindings, and needs more shell vars. Same with https://stackoverflow.com/a/40426702.

The current implementation still seems fragile to me (in case anyone has chanced the meaning of any of those keybindings, which I for example actually have for `C-h`).<br>
Maybe we could somehow save the current keybindings, remove them (to be sure defaults are used), and restore them afters wards?

calestyo: @junegunn, I've removed all commits except the minor fa6b43fa9d3d7435209c3b773945980d764e18c5.

As for switching `Alt-C` to use `bind -x`:<br>
I still think there is some value in doing this, namely to get it working for any systems where the user has reassigned the keyseqs you use in:
https://github.com/junegunn/fzf/blob/4fdc08295b90485b7ee30a35d77177699c070113/shell/key-bindings.bash#L132-L134
and also to avoid having to set up helper bindings as in:
https://github.com/junegunn/fzf/blob/4fdc08295b90485b7ee30a35d77177699c070113/shell/key-bindings.bash#L102-L107
which may overwrite the user’s own bindings.

I think I may have found a working solution, at least for bash versions that already have `bind -x`.<br>
It's not yet perfect and I'm not enough of a bash/readline expert to determine whether it really works properly and in all cases, so I wrote mails to [help-bash](https://lists.gnu.org/mailman/listinfo/help-bash), asking for their advice:
- https://lists.gnu.org/archive/html/help-bash/2023-10/msg00000.html
- https://lists.gnu.org/archive/html/help-bash/2023-10/msg00001.html
if you're curious, you can already look there, how I think we could perhaps do it.

At least this time I actually did some testing O:-) ... and it looked good, except for the open points mentioned in the mail to help-bash.

I think we should handle any possible improvements to `Alt-C` in a separate issue, so if you're fine with the remaining minor commit of this PR, please merge it.<br>
At least the tests seem to ran through this time.

junegunn (approved):

junegunn:
Merged, thanks.

> I think we should handle any possible improvements to Alt-C in a separate issue

Ideally yes, but we need to evaluate whether the added complexity is really worth it. Users can make all sorts of unconventional choices and our bindings are not immune to them. But I have maintained this code for many years, yet no one has ever complained that ALT-C does not work because of their creative bindings. I think that says something.

calestyo:
Oh and thanks for merging.

junegunn:
> Well, I for example, have already assigned `C-h` and it breaks... so: "complain".😜

Can you tell me what you are binding it to, and how it makes ALT-C behave?

Issue #2123, opened 2020-07-27
Expand SSH hostname to IP where necessary (telnet for example)


drws:
- [x] I have read through the manual page (`man fzf`)
- [x] I have the latest version of fzf
- [x] I have searched through the existing issues

## Info

- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [x] bash
    - [x] zsh
    - [x] fish
    
## Problem / Steps to reproduce

SSH hosts configuration file ~/.ssh/config is parsed when the fzf "host" completion is used. The configured names of hosts and their respective addresses (IP/domain) are parsed into two separate fuzzy completion hits. However, the configured names cannot be used with non-SSH commands as they do not use the SSH config file (telnet being an example) and so these fuzzy completion hits are not useful. It would be a practical feature if in these cases fzf expanded SSH names into addresses (after selecting the name from the list of hits).

Such a solution could be implemented with a new "host_sshexpand" completion mode, but if this is not desired, the existing "host" one could be refit for this purpose and provide the described expansion by default. And then the SSH-type commands (ssh, scp, dbclient, ...) could be treated as special cases where this additional expansion is not necessary.

antofthy:
What is the fzf "host" completion you refer to?

I worked out my own solution.  See https://github.com/junegunn/fzf/issues/2316

HsuanTingLu:
> It would be a practical feature if in these cases fzf expanded SSH names into addresses (after selecting the name from the list of hits).
>
> Such a solution could be implemented with a new "host_sshexpand" completion mode, but if this is not desired, the existing "host" one could be refit for this purpose and provide the described expansion by default.

You should customize telnet completion instead.
After all hosts in ssh configs aren't for stuff like telnet anyway, you should set these in places like `/etc/hosts`.

Repurposing a semi-public function hurts backward compatibility and people who use `_fzf_host_completion` probably won't be grateful about this surprise.

drws:
I use fzf also because there is no need to edit `/etc/hosts`, so this would be counterproductive. Primarily I suggested an additional function (`_fzf_hostssh_completion` for example) that would expand SSH hosts into IPs after calling `_fzf_host_completion` and receiving input. There is no need to break compatibility.

antofthy:
You may like to have a look at...
https://github.com/junegunn/fzf/issues/2316

drws:
[...]

PR #3454, opened 2023-09-26
Improve host completion

### LINKED ISSUE #3450: feature request: host completion improvements

## Info

- OS
    - [X] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [X] bash
    - [ ] zsh
    - [ ] fish

Hey.

I think the current host completion has a number of areas where it could be improved:
1. It's not (easily) configurable, e.g. it would be nice to have a user switch that disables v4 and/or v6 addresses, and perhaps options that do sorting
2. It lacks many names that are shown by bash-completion’s `_known_hosts_real()`, which is quite powerful and IIRC even checks `ssh_config` for other configured `GlobalKnownHostsFile` and `UserKnownHostsFile`. `fzf`’s current completion seems to ignore any host (which are more or less all of them) that I have in the `GlobalKnownHostsFile`. `_known_hosts_real()` also has many more features, I think it can optionally use Avahi and ruptime.

For (2), I think it would be best if `fzf` would simply use `_known_hosts_real()` - why re-inventing the wheel, if bash-completion guys already maintain that properly?

Based on the 0.42.0 `fzf`, I tried roughly the following:
```
_fzf_host_completion() {
  _fzf_complete +m -- "$@" < <(
        _known_hosts_real ''
        printf '%s\n' "${COMPREPLY[@]}" | sort -u -V
  ) 
} 
```
and at a first glance it seemed to work:
- The `sort-u` is necessary as `COMPREPLY` contained many duplicates (no idea why).
- `-V` is just for sorting ... ideally this would be configurable,... plus I'd like a sort mode that sort by domains[0].

Some open points:
- Is `_known_hosts_real()` available for sure? I mean bash-completion does load completions for commands dynamically (i.e. only when the command is actually completed), but I'd blindly guess that `_known_hosts_real()` is always available and we don't need to "import" it first.
- Do we have any side-effects? At least `COMPREPLY` is set. Haven't checked whether `_known_hosts_real()` uses other non-local variables. As long as the function is called within process substitution `<(…)` this shouldn't be a problem, but I've seen that in edfdcc8cee31e227a5f7eda9c1faa4893757645c you placed the printing code in a separate function, so if that would ever be called outside process substitution it might cause unwanted side effects.
- What about the config env vars recognised by `_known_hosts_real()`, like `COMP_KNOWN_HOSTS_WITH_AVAHI`? Should we just use them directly, or clean them and initialise them only by our own e.g. `FZF_KNOWN_HOSTS_WITH_AVAHI` counterparts (so that people could configure bash-completion and fzf differently)?

calestyo:
Hey @junegunn.

This would fix #3450 and incorporates a number of improvements and ideas:
- IMO, `__fzf_list_hosts()` should be considered to be a function for user customisation, for which we provide however some reasonable default.
  This includes, that it’s up to that function to sort the host list it produces (if desired so) and the remove duplicates, if any.
- To keep things simple for users, I moved out the `username@` prefix handling from it into `_fzf_complete_ssh()` in c64fe9d4ea89e80405837772e7d8b0c876efbeec.
- fbe4c5408a21c981f774feeaaac81af302c9b7b5, makes `completion.bash` to use bash-completion’s `_known_hosts_real()`, if available, and falls back to the previous code for that, if not.

Some open points:
- For `completion.bash`, shall we remove the old host-list code altogether and only use `_known_hosts_real()`? I guess probably not… we need the same code anyway for zsh, and in principle `fzf`’s completion stuff can be used without bash-completion (though I doubt many do).
- I'd like to provide an easy way for users to override the sorting of the lists generated by `__fzf_list_hosts()`, so that one could make it easily configurable, whether e.g. natural sort is used or not, or easily hook in something that does what I've described [here](https://github.com/junegunn/fzf/issues/3450#issuecomment-1732762255) or perhaps a sorting that places all IPv4/6 addresses in the end, etc. pp..
  What do you think about introducing e.g. `FZF_COMPLETION_HOSTNAME_SORT_CMD`, which I'd then default to `sort -u --version-sort`?
- Possible side effects of invoking `_known_hosts_real()`.
  I went through it’s code, and if I didn’t miss anything, the only side effect is that it sets `COMPREPLY`, which I therefore made `local`. But of course, any bash-completion version may change that.
  Another approach would be to invoke `_known_hosts_real` and the `printf` in a `( … )` subshell. Would you prefer that?

### DIFF
--- a/shell/completion.bash
+++ b/shell/completion.bash
@@ -412,10 +412,22 @@ _fzf_proc_completion_post() {
 # desired sorting and with any duplicates removed, to standard output.
 if ! declare -F __fzf_list_hosts > /dev/null; then
   __fzf_list_hosts() {
-    command cat <(command tail -n +1 ~/.ssh/config ~/.ssh/config.d/* /etc/ssh/ssh_config 2> /dev/null | command grep -i '^\s*host\(name\)\? ' | command awk '{for (i = 2; i <= NF; i++) print $1 " " $i}' | command grep -v '[*?%]') \
-      <(command grep -oE '^[[a-z0-9.,:-]+' ~/.ssh/known_hosts 2> /dev/null | command tr ',' '\n' | command tr -d '[' | command awk '{ print $1 " " $1 }') \
-      <(command grep -v '^\s*\(#\|$\)' /etc/hosts 2> /dev/null | command grep -Fv '0.0.0.0') |
-      command awk '{if (length($2) > 0) {print $2}}' | command sort -u
+    if declare -F _known_hosts_real > /dev/null; then
+      # if available, use bash-completions’s _known_hosts_real() for getting the list of hosts
+
+      # set the local attribute for any non-local variable that is set by _known_hosts_real()
+      local COMPREPLY=()
+
+      _known_hosts_real ''
+      printf '%s\n' "${COMPREPLY[@]}" | command sort -u --version-sort
+    else
+      # otherwise, get a list of hosts on our own
+
+      command cat <(command tail -n +1 ~/.ssh/config ~/.ssh/config.d/* /etc/ssh/ssh_config 2> /dev/null | command grep -i '^\s*host\(name\)\? ' | command awk '{for (i = 2; i <= NF; i++) print $1 " " $i}' | command grep -v '[*?%]') \
+        <(command grep -oE '^[[a-z0-9.,:-]+' ~/.ssh/known_hosts 2> /dev/null | command tr ',' '\n' | command tr -d '[' | command awk '{ print $1 " " $1 }') \
+        <(command grep -v '^\s*\(#\|$\)' /etc/hosts 2> /dev/null | command grep -Fv '0.0.0.0') |
+        command awk '{if (length($2) > 0) {print $2}}' | command sort -u
+    fi
   }
 fi
 
--- a/shell/completion.bash
+++ b/shell/completion.bash
@@ -409,8 +409,8 @@ _fzf_proc_completion_post() {
 
 __fzf_list_hosts() {
   command cat <(command tail -n +1 ~/.ssh/config ~/.ssh/config.d/* /etc/ssh/ssh_config 2> /dev/null | command grep -i '^\s*host\(name\)\? ' | command awk '{for (i = 2; i <= NF; i++) print $1 " " $i}' | command grep -v '[*?%]') \
-    <(command grep -oE '^[[a-z0-9.,:-]+' ~/.ssh/known_hosts | command tr ',' '\n' | command tr -d '[' | command awk '{ print $1 " " $1 }') \
-    <(command grep -v '^\s*\(#\|$\)' /etc/hosts | command grep -Fv '0.0.0.0') |
+    <(command grep -oE '^[[a-z0-9.,:-]+' ~/.ssh/known_hosts 2> /dev/null | command tr ',' '\n' | command tr -d '[' | command awk '{ print $1 " " $1 }') \
+    <(command grep -v '^\s*\(#\|$\)' /etc/hosts 2> /dev/null | command grep -Fv '0.0.0.0') |
     command awk -v "user=$1" '{if (length($2) > 0) {print user $2}}' | command sort -u
 }
 
--- a/shell/completion.bash
+++ b/shell/completion.bash
@@ -411,24 +411,26 @@ _fzf_proc_completion_post() {
 # The function is expected to print hostnames, one per line as well as in the
 # desired sorting and with any duplicates removed, to standard output.
 if ! declare -F __fzf_list_hosts > /dev/null; then
-  __fzf_list_hosts() {
-    if declare -F _known_hosts_real > /dev/null; then
-      # if available, use bash-completions’s _known_hosts_real() for getting the list of hosts
+  if declare -F _known_hosts_real > /dev/null; then
+    # if available, use bash-completions’s _known_hosts_real() for getting the list of hosts
 
+    __fzf_list_hosts() {
       # set the local attribute for any non-local variable that is set by _known_hosts_real()
       local COMPREPLY=()
 
       _known_hosts_real ''
       printf '%s\n' "${COMPREPLY[@]}" | command sort -u --version-sort
-    else
-      # otherwise, get a list of hosts on our own
+    }
+  else
+    # otherwise, get a list of hosts on our own
 
+    __fzf_list_hosts() {
       command cat <(command tail -n +1 ~/.ssh/config ~/.ssh/config.d/* /etc/ssh/ssh_config 2> /dev/null | command grep -i '^\s*host\(name\)\? ' | command awk '{for (i = 2; i <= NF; i++) print $1 " " $i}' | command grep -v '[*?%]') \
         <(command grep -oE '^[[a-z0-9.,:-]+' ~/.ssh/known_hosts 2> /dev/null | command tr ',' '\n' | command tr -d '[' | command awk '{ print $1 " " $1 }') \
         <(command grep -v '^\s*\(#\|$\)' /etc/hosts 2> /dev/null | command grep -Fv '0.0.0.0') |
         command awk '{if (length($2) > 0) {print $2}}' | command sort -u
-    fi
-  }
+    }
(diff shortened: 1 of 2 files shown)


calestyo:
If you'd agree to the idea with `FZF_COMPLETION_HOSTNAME_SORT_CMD` we could provide an example file or so, where some common sortings are given.

Perhaps one should also name it raterh `FZF_COMPLETION_HOSTNAME_POSTPROCESS_CMD`, which would reflect better, that such a command could also be used to e.g. filter IPv4 and/or 6 address literals.

calestyo:
Last force push was a minor update to one of the commits.
I now set `COMPREPLY=()` rather than `=''`.

I think with the way bash implements `local` this shouldn’t matter, as the variable is created completely new (and not initialised by an outer-scope one of the same name).

Still `COMPREPLY` is an array so it feels better to initialise it with `=()`.

calestyo:
> Perhaps one should also name it rather `FZF_COMPLETION_HOSTNAME_POSTPROCESS_CMD`, which would reflect better, that such a command could also be used to e.g. filter IPv4 and/or 6 address literals.

Maybe that could/should be further extended:<br>
I saw #2123, and if I understand the reporter’s problem correctly (see https://github.com/junegunn/fzf/issues/2123#issuecomment-1737992805), it could be solved by providing e.g.:
- `FZF_COMPLETION_HOSTNAME_POSTPROCESS_CMD`
- `FZF_COMPLETION_SSH_HOSTNAME_POSTPROCESS_CMD`
- and if needed more like `FZF_COMPLETION_<use-case>_HOSTNAME_POSTPROCESS_CMD`

I think it wouldn't be too hard to actually do that.

The default for any `FZF_COMPLETION_<use-case>_HOSTNAME_POSTPROCESS_CMD` could be to use the value of `FZF_COMPLETION_HOSTNAME_POSTPROCESS_CMD`.<br>
That means i a user just wants to some sorting (as proposed above), he'd only need t set the command for that in `FZF_COMPLETION_HOSTNAME_POSTPROCESS_CMD`, and all others would "inherit" it automatically.

But **if** a user wants to e.g. filter hosts based on what's being completed for, e.g. `ssh`, he could e.g. do:
```
FZF_COMPLETION_SSH_HOSTNAME_POSTPROCESS_CMD="grep -v '^[A-Z]' | someSorting"
```
or even:
```
FZF_COMPLETION_SSH_HOSTNAME_POSTPROCESS_CMD="grep -v '^[A-Z]' | eval \"\$FZF_COMPLETION_HOSTNAME_POSTPROCESS_CMD\""
```

calestyo:
Found a small bug in the version for ZSH, which doesn't have the same `declare -F` as bash.

Now the tests run through, though locally I still have flaky tests (namely `TestGoFZF#test_kill_default_command_on_abort`, `TestGoFZF#test_history` and `TestGoFZF#test_kill_reload_command_on_abort` fail every now and then).

calestyo:
btw: The idea of a convenience script that I had over in https://github.com/junegunn/fzf/issues/3459#issuecomment-1742149923 might also be of use here:

We could simply ship a copy (which we update from time 2 time) of bash-completions’ `_known_hosts_real()` (which, at least as of my Debian version of bash-completion would require tree further functions `_included_ssh_config_files`, `__expand_tilde_by_ref`, `__ltrim_colon_completions`).

Unfortunately, they don't seem to work out of the box for zsh.

calestyo:
Rebased that onto all the recent changes in `master`, and also added missing `command `s to (hopefully) all places of the commits.

calestyo:
@junegunn Any thoughts about this whole series?

junegunn:
I'm fine with allowing advanced users to override `__fzf_list_hosts`, but I feel strongly that anything beyond that is unnecessary.

`__fzf_list_hosts` and `_fzf_complete_ssh` are just 20 lines of code as of now. Advanced users can easily read the source code and define their own completion function with unlimited freedom. It's faster to just copy the code and change it as they like than to figure out what `FZF_SOMETHING_SOMETHING` is and how to use it and think inside that box.

junegunn:
Merged, thanks!

shell/completion.bash
```diff
@@ -407,15 +407,34 @@ _fzf_proc_completion_post() {
...
+if ! declare -F __fzf_list_hosts > /dev/null; then
+  if declare -F _known_hosts_real > /dev/null; then
+    # if available, use bash-completions’s _known_hosts_real() for getting the list of hosts
+
+    __fzf_list_hosts() {
+      # set the local attribute for any non-local variable that is set by _known_hosts_real()
+      local COMPREPLY=()
+
+      _known_hosts_real ''
+      printf '%s\n' "${COMPREPLY[@]}" | command sort -u --version-sort
```
junegunn, line 422: This is fine for now. But I'll have to reconsider this if they decide to change the way the function works in the future.

calestyo: I think they're consolidating their function naming right now, so me might need to adapt that soon to work with the new name and fall back to the old one... but that shouldn't be too hard.

Other than that... I'd rather not expect any deeper changes to what the function does, do you?

calestyo:
> I'm fine with allowing advanced users to override __fzf_list_hosts, but I feel strongly that anything beyond that is unnecessary.
> __fzf_list_hosts and _fzf_complete_ssh are just 20 lines of code as of now. Advanced users can easily read the source code and define their own completion function with unlimited freedom. It's faster to just copy the code and change it as they like than to figure out what FZF_SOMETHING_SOMETHING is and how to use it and think inside that box.

I spent quite some time thinking recently, how fzf's shell/completion integration could be made more generic & powerful and at the same time be simpler for "us" (I mean the fzf side) to maintain.<br>
Haven't fully sorted my thoughts yet but sooner or later I'll write it all up and post an RFC here.

shell/completion.bash
junegunn: > I think they're consolidating their function naming right now, so me might need to adapt that soon to work with the new name and fall back to the old one... but that shouldn't be too hard.

Oh, that's a deal breaker for me. I don't want to ship a product that might not work because of a sudden change in an implicit dependency.

junegunn: I'm not familiar with their code, and I just assumed  _known_hosts_real was one of their documented public API that is guaranteed to not break backward-compatible. If that's not the case, I'm not going to make fzf depend on it.

calestyo: Well, bash-completion is not the C standard or so... I think it's reasonable to expect that they have some level of stability in their ABI (since people use it), but nothing is ever set fully in stone… not even POSIX and the likes.

I've also only noticed it a few days ago (when looking into that `eval` issue we had), that they're renaming their functions... and didn't really think any further that this might also affect `_known_hosts_real()`, but it does … in https://github.com/scop/bash-completion/commit/d3119216c5de651e39e1026d0e81faf8949d46f6 they’ve renamed it to `_comp_compgen_known_hosts()`.

But actually I think this consolidation is quite reasonable for them to do… the many different names and prefixes they've started to use over time were much more messy and likely that people accidentally use such names for something else.<br>
I've already mention that earlier here, I think hat `fzf` might also benefit from such systematic consolidation (e.g. names like `FZF_CTRL_T_COMMAND` are pretty vague (an ideally the keybinding should be configurable and not fixed to `Ctrl-T`).<br>
It may even make sense for `fzf` to integrate in the naming schema used by bash-completion to some extent... e.g. if we'd always use `_comp_fzf_*` or so for functions.

Anyway I don't see the big deal breaker here?!

It's not that they'd do this every year or so... and that consolidation they do seems to rather make this stable/official now.<br>
APIs do change, or the code is either dead are - less likely - already perfect.

Especially, it's trivially easy to support both (see #3476).<br>
Also, there's already much more complex to read code, that we have now in order to support different versions. The `mawk`-version check or the code to support old bash versions.<br>
Even if we would start using more bash-completion functions (e.g. also for the `eval` thing), I'd guess their total number would be rather limited, so I think this is pretty manageable.

Similarly, even if I could persuade you to e.g. rename `FZF_CTRL_T_COMMAND` and the likes (which I guess might be a tough job to do ;-) ), it would be easy to fall back to the old names, and e.g. print deprecation warnings for a while.

Issue #3458, opened 2023-09-28
`command <someCommand>` missing in several places throughout the shell code


calestyo:
- [X] I have read through the manual page (`man fzf`)
- [X] I have the latest version of fzf
- [X] I have searched through the existing issues

## Info

- OS
    - [X] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [X] bash
    - [ ] zsh
    - [ ] fish

## Problem / Steps to reproduce

Hey.

AFAIU, you use `command <someCommand>` in the sourced shell files to prevent undesired alias subsitution, which would likely break everything - just as e.g. bash-completion does as well and which of course is a good thing.

Looking at the shell code (`completion.bash` and `key-bindings.bash` at least), there are however numerous places where this isn't done, e.g.:
https://github.com/junegunn/fzf/blob/a3ff49aaf18eceefe4bf8b1df5bf275b3d982493/shell/completion.bash#L17

https://github.com/junegunn/fzf/blob/a3ff49aaf18eceefe4bf8b1df5bf275b3d982493/shell/completion.bash#L51

https://github.com/junegunn/fzf/blob/a3ff49aaf18eceefe4bf8b1df5bf275b3d982493/shell/completion.bash#L175

https://github.com/junegunn/fzf/blob/a3ff49aaf18eceefe4bf8b1df5bf275b3d982493/shell/completion.bash#L284-L287

I could try to write a patch that adds `command`, but:
- only after my other PRs have been either merged or rejected (otherwise it gets to messy)
- we need to decide on what we protect via `command`

I think for cases like `awk` or `tr` it makes sense to use `command`. I also have seen cases, where people alias `echo`/`printf`. Not sure about things like `read`, `break` or `eval`. I'd tend to perhaps use `command` for `read` but not for any of the special built-ins.

Also, function names like in:
https://github.com/junegunn/fzf/blob/a3ff49aaf18eceefe4bf8b1df5bf275b3d982493/shell/completion.bash#L185
may be aliased, too.<br>
We could in principle avoid alias substitution in such cases by quoting the function name, not sure whether this should be done though.

Cheers,
Chris.

junegunn:
> * we need to decide on what we protect via `command`

That's the hard part. An easy solution is to put `command` only when someone actually complains and raises an issue. That's how we've been handling the issue so far.

> I also have seen cases, where people alias `echo`/`printf`.

Wow.

calestyo:
[...]

PR #3462, opened 2023-10-05
use `command` to “protect” further commands

calestyo:
Hey.

This would be for #3458.

- The current commit (hopefully) adds `command ` to all commands that are not built-ins or functions.<br>
I haven't really tested it yet ;-) ... and guess anyway that it will be rather quite some effort to test all cases where `command ` is now used.

- A future commit might use `builtin` to protect further things like `printf` or `echo`. But arguably, it's always a trade off how far we go here (code readability vs. actually adding some benefit).
  But I do think, that at least "regular" (non-built-in) commands should all be "protected".

- Also, I made no effort to do the same for fish or zsh. Since this adds no real new feature, I hope this is okay.

- But I did notice that all zsh files have:
   https://github.com/junegunn/fzf/blob/488a236b7a4d01e910e0ac80b4794a7a054fb3a8/shell/completion.zsh#L70
   https://github.com/junegunn/fzf/blob/488a236b7a4d01e910e0ac80b4794a7a054fb3a8/shell/key-bindings.zsh#L35
   Not really sure how that works, but shouldn't it anyway make the `command ` superfluous, at least in terms of aliases (probably not in terms of functions)?

### DIFF
--- a/shell/completion.bash
+++ b/shell/completion.bash
@@ -18,7 +18,7 @@ if ! declare -F _fzf_compgen_path > /dev/null; then
     echo "$1"
     command find -L "$1" \
       -name .git -prune -o -name .hg -prune -o -name .svn -prune -o \( -type d -o -type f -o -type l \) \
-      -a -not -path "$1" -print 2> /dev/null | sed 's@^\./@@'
+      -a -not -path "$1" -print 2> /dev/null | command sed 's@^\./@@'
   }
 fi
 
@@ -26,7 +26,7 @@ if ! declare -F _fzf_compgen_dir > /dev/null; then
   _fzf_compgen_dir() {
     command find -L "$1" \
       -name .git -prune -o -name .hg -prune -o -name .svn -prune -o -type d \
-      -a -not -path "$1" -print 2> /dev/null | sed 's@^\./@@'
+      -a -not -path "$1" -print 2> /dev/null | command sed 's@^\./@@'
   }
 fi
 
@@ -196,7 +196,7 @@ __fzf_generic_path_completion() {
         printf '\e[5n'
         return 0
       fi
-      dir=$(dirname "$dir")
+      dir=$(command dirname "$dir")
       [[ "$dir" =~ /$ ]] || dir="$dir"/
     done
   else
@@ -230,8 +230,8 @@ _fzf_complete() {
   fi
 
   local cur selected trigger cmd post
-  post="$(caller 0 | awk '{print $2}')_post"
-  type -t "$post" > /dev/null 2>&1 || post=cat
+  post="$(caller 0 | command awk '{print $2}')_post"
+  type -t "$post" > /dev/null 2>&1 || post='command cat'
 
   cmd="${COMP_WORDS[0]//[^A-Za-z0-9_=]/_}"
   trigger=${FZF_COMPLETION_TRIGGER-'**'}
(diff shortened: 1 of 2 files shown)


calestyo:
@junegunn … I did some (very very limited) testing now... and the test suite also runs through.

So once @step- has made his fix of the issue described at https://github.com/junegunn/fzf/pull/3313#issuecomment-1747934690, I'd rebase this branch and we could IMO merge it (after you double and triple checked it ;-) ).

shell/completion.bash
```diff
@@ -39,10 +39,10 @@ __fzf_comprun() {
     _fzf_comprun "$@"
   elif [[ -n "${TMUX_PANE-}" ]] && { [[ "${FZF_TMUX:-0}" != 0 ]] || [[ -n "${FZF_TMUX_OPTS-}" ]]; }; then
     shift
-    fzf-tmux ${FZF_TMUX_OPTS:--d${FZF_TMUX_HEIGHT:-40%}} -- "$@"
+    command fzf-tmux ${FZF_TMUX_OPTS:--d${FZF_TMUX_HEIGHT:-40%}} -- "$@"
   else
     shift
-    fzf "$@"
+    command fzf "$@"
```
junegunn, line 45: I would allow users to use their versions of fzf and fzf-tmux, whether it's an alias or a function. By using a function, you can dynamically apply different sets of options depending on some conditions, e.g. light theme during daytime, dark theme at night.

junegunn:
```suggestion
    fzf "$@"
```

shell/completion.bash
```diff
@@ -39,10 +39,10 @@ __fzf_comprun() {
     _fzf_comprun "$@"
   elif [[ -n "${TMUX_PANE-}" ]] && { [[ "${FZF_TMUX:-0}" != 0 ]] || [[ -n "${FZF_TMUX_OPTS-}" ]]; }; then
     shift
-    fzf-tmux ${FZF_TMUX_OPTS:--d${FZF_TMUX_HEIGHT:-40%}} -- "$@"
+    command fzf-tmux ${FZF_TMUX_OPTS:--d${FZF_TMUX_HEIGHT:-40%}} -- "$@"
```
junegunn, line 42:
```suggestion
    fzf-tmux ${FZF_TMUX_OPTS:--d${FZF_TMUX_HEIGHT:-40%}} -- "$@"
```

shell/completion.bash
calestyo: Hmm I was actually wondering about the same thing when I edited it, and had even planned to point that out and ask you for your opinion.

Thought now that you write it, I'm no longer sure whether it should really be using a potential alias/function.

1. In case if `fzf`/`fzf-tmux`, alias and function would most likely be used in the same way (if set by a user) - I mean to override the default behaviour. Whether via an alias or a function doesn't really matter, IMO.
2. Why would a user set an alias/function on either of those? I mean most likely to change some default behaviour, for which we in principle already have `FZF_DEFAULT_OPTS`. But that is used for both, `fzf`and `fzf-tmux`, ain’t it? So I could imagine a user might set up an alias/function to specifically target just one of these.
3. I'd also think, that users would rather not complete replace `fzf`/`fzf-tmux`, like with another program, or something that completely changes behaviour - I think it's more likely that they just set options to it.
4. Now, for the purpose of completion or key-binding functions... do we want to take such override effect?
  I mean what are people likely to set? Maybe things like `‐‐scheme=` or `--height` or so?
  But for those I'd rather say that we **don't** want them to take effect for the our completion or key-binding functions, do we?
  I rather think that **if** a user has such specific needs that wants him to override `fzf`/`fzf-tmux` in `__fzf_comprun()`, he should either override the function itself, or - if enough there was enough demand for that, we should add a specific env var that sets options for the case where this is used.

But of course, no choice is perfect. So up to you, what we do.

calestyo:
Rebased the commit on current master, though it still has the `command ` on `fzf`/`fzf-tmux` until we come to a decision above.

shell/completion.bash
calestyo: see above

shell/completion.bash
junegunn: As I said above, dynamic theming is a good example where users may feel $FZF_DEFAULT_OPTS isn't enough. After all, it doesn't matter what we think is a better way to do things. For a user who relies on an alias or a function, this is a breaking change. And we don't break user scripts unless there is a very compelling reason to do so.

calestyo: Rebased to current master, with the “protection” removed from `fzf`/`fzf-tmux`.

Please review & merge.

shell/key-bindings.bash
```diff
@@ -32,7 +32,7 @@ __fzf_select__() {
 
 __fzfcmd() {
   [[ -n "${TMUX_PANE-}" ]] && { [[ "${FZF_TMUX:-0}" != 0 ]] || [[ -n "${FZF_TMUX_OPTS-}" ]]; } &&
-    echo "fzf-tmux ${FZF_TMUX_OPTS:--d${FZF_TMUX_HEIGHT:-40%}} -- " || echo "fzf"
+    echo "command fzf-tmux ${FZF_TMUX_OPTS:--d${FZF_TMUX_HEIGHT:-40%}} -- " || echo "command fzf"
```
junegunn, line 35:
```suggestion
    echo "fzf-tmux ${FZF_TMUX_OPTS:--d${FZF_TMUX_HEIGHT:-40%}} -- " || echo "fzf"
```

junegunn:
Merged, thanks!

shell/key-bindings.bash
calestyo: sorry... had forgotten messed that up and forgotten those two.

PR #3463, opened 2023-10-06
[shell] improve mawk detection

step-:
* Use the all-compatible mawk `-W version` option. https://github.com/junegunn/fzf/pull/3313#issuecomment-1747934690.
* Do not remap the history key if no dependent commands is installed (perl, awk or mawk in this order).
* Run the command and not a function consistently with #3462.

The version check bash code relies on the following mawk source code, extracted from mawk 1.3.4 20230322.

```
version.c:
18-  #include "init.h"
19-  #include "patchlev.h"
20-
21:  #define	 VERSION_STRING	 \
22-    "mawk %d.%d%s %s\n\
23-  Copyright 2008-2022,2023, Thomas E. Dickey\n\
24-  Copyright 1991-1996,2014, Michael D. Brennan\n\n"
....
30-  void
31-  print_version(FILE *fp)
32-  {
33:      fprintf(fp, VERSION_STRING, PATCH_BASE, PATCH_LEVEL, PATCH_STRING, DATE_STRING);
34-      fflush(fp);
35-
36-  #define SHOW_RANDOM "random-funcs:"

patchlev.h:
13-  /*
14-   * $MawkId: patchlev.h,v 1.128 2023/03/23 00:23:57 tom Exp $
15-   */
16:  #define  PATCH_BASE	1
17-  #define  PATCH_LEVEL	3
18-  #define  PATCH_STRING	".4"
19-  #define  DATE_STRING    "20230322"
```

### DIFF
--- a/shell/key-bindings.bash
+++ b/shell/key-bindings.bash
@@ -48,7 +48,7 @@ __fzf_cd__() {
   dir=$(set +o pipefail; eval "$cmd" | FZF_DEFAULT_OPTS="$opts" $(__fzfcmd)) && printf 'builtin cd -- %q' "$dir"
 }
 
-if command -v perl > /dev/null; then
+if command -v perl >&- ; then
   __fzf_history__() {
     local output opts script
     opts="--height ${FZF_TMUX_HEIGHT:-40%} --bind=ctrl-z:ignore ${FZF_DEFAULT_OPTS-} -n2..,.. --scheme=history --bind=ctrl-r:toggle-sort ${FZF_CTRL_R_OPTS-} +m --read0"
@@ -56,7 +56,7 @@ if command -v perl > /dev/null; then
     output=$(
       set +o pipefail
       builtin fc -lnr -2147483648 |
-        last_hist=$(HISTTIMEFORMAT='' builtin history 1) perl -n -l0 -e "$script" |
+        last_hist=$(HISTTIMEFORMAT='' builtin history 1) command perl -n -l0 -e "$script" |
         FZF_DEFAULT_OPTS="$opts" $(__fzfcmd) --query "$READLINE_LINE"
     ) || return
     READLINE_LINE=${output#*$'\t'}
@@ -66,16 +66,14 @@ if command -v perl > /dev/null; then
       READLINE_POINT=0x7fffffff
     fi
   }
-else # awk
+elif command -v awk >&- || command -v mawk >&- ; then # awk - fallback for POSIX systems
   __fzf_history__() {
-    local output opts script
+    local output opts script n x y z d
     if [[ -z $__fzf_awk ]]; then
       __fzf_awk=awk
-      # if installed, mawk is faster
-      command -v mawk > /dev/null &&
-        mawk --version |          # at least 1.3.4
-          awk 'NR == 1 { split($2, a, "."); v=(a[1]*1000000+ a[2]*1000+ a[3]*1); exit !(v >= 1003004) }' &&
-          __fzf_awk=mawk
+      # choose the faster mawk if: it's installed && build date >= 20230322 && version >= 1.3.4
+      IFS=' .' read n x y z d <<< $(command mawk -W version 2>&-)
+      [[ $n == mawk ]] && (( d >= 20230302 && (x *1000 +y) *1000 +z >= 1003004 )) && __fzf_awk=mawk
     fi
     opts="--height ${FZF_TMUX_HEIGHT:-40%} --bind=ctrl-z:ignore ${FZF_DEFAULT_OPTS-} -n2..,.. --scheme=history --bind=ctrl-r:toggle-sort ${FZF_CTRL_R_OPTS-} +m --read0"
     [[ $(HISTTIMEFORMAT='' builtin history 1) =~ [[:digit:]]+ ]]    # how many history entries
@@ -86,8 +84,8 @@ else # awk
     END { if (NR) P(b) }'
     output=$(
       set +o pipefail
-      builtin fc -lnr -2147483648 2> /dev/null |   # ( $'\t '<lines>$'\n' )* ; <lines> ::= [^\n]* ( $'\n'<lines> )*
-        $__fzf_awk "$script"                          |   # ( <counter>$'\t'<lines>$'\000' )*
+      builtin fc -lnr -2147483648 2>&-   |   # ( $'\t '<lines>$'\n' )* ; <lines> ::= [^\n]* ( $'\n'<lines> )*
+        command $__fzf_awk "$script"     |   # ( <counter>$'\t'<lines>$'\000' )*
         FZF_DEFAULT_OPTS="$opts" $(__fzfcmd) --query "$READLINE_LINE"
     ) || return
     READLINE_LINE=${output#*$'\t'}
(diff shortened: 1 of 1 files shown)


calestyo:
Other than the comments made in the code above, I think it looks mostly good, I especially like that you saved at least one execution of `awk`, though:
- I'd split it up in more commits, you do now various mostly unrelated things in one commit, adding the `command`, fixing the version check, changing the output redirection (which I would however drop completely), make the code handle the situation when both `perl` and `[m]awk` aren't available.
  I think it's easier to understand and follow in smaller pieces
- Handling the case that `awk` isn't available is in principle fine by me, though at least POSIX wise, systems are expected to have `awk`. So assuming POSIX systems, handling this is unnecessary.
- btw: I think the `(x *1000 +y) *1000 +z >= 1003004` part is at least a bit fragile, should e.g. ever a big minor version number be used. Also you still have the check for the date, so apart from me not liking such a runtime check in general, t should be fine.

I checked it on Debian with `mawk` 1.3.4.20200120-3.1, which yields `__fzf_awk=awk`, and with 1.3.4.20230808-1, which yields `__fzf_awk=mawk`. So that works :-)

shell/key-bindings.bash
```diff
@@ -48,15 +48,15 @@ __fzf_cd__() {
   dir=$(set +o pipefail; eval "$cmd" | FZF_DEFAULT_OPTS="$opts" $(__fzfcmd)) && printf 'builtin cd -- %q' "$dir"
 }
 
-if command -v perl > /dev/null; then
+if command -v perl >&- ; then
```
junegunn, line 51: Probably not an issue here, but wouldn't it be safer to redirect to `/dev/null`? Because we can't always be sure how the program will behave when writing to a closed stream.

```sh
$ ls >&-
$ command -v ls >&-
$ echo hello >&-
-bash: echo: write error: Bad file descriptor
$ echo hello > /dev/null
$
```
```

**Comment by step-** [comment]
@calestyo thank you for testing on Debian.
@junegunn you're right, who can be sure how all target platforms and bash versions will behave when stdout is closed. Better be safe.

File: shell/key-bindings.bash [hunk 2]
```diff
@@ -48,15 +48,15 @@ __fzf_cd__() {
   dir=$(set +o pipefail; eval "$cmd" | FZF_DEFAULT_OPTS="$opts" $(__fzfcmd)) && printf 'builtin cd -- %q' "$dir"
 }
 
-if command -v perl > /dev/null; then
+if command -v perl > /dev/null ; then
```
junegunn, line 51:
```suggestion
if command -v perl > /dev/null; then
```

shell/key-bindings.bash
```diff
@@ -66,16 +66,14 @@ if command -v perl > /dev/null; then
       READLINE_POINT=0x7fffffff
     fi
   }
-else # awk
+elif command -v awk > /dev/null || command -v mawk > /dev/null ; then # awk - fallback for POSIX systems
```
junegunn, line 69:
```suggestion
elif command -v awk > /dev/null || command -v mawk > /dev/null; then # awk - fallback for POSIX systems
```

shell/key-bindings.bash
```diff
@@ -86,8 +84,8 @@ else # awk
     END { if (NR) P(b) }'
     output=$(
       set +o pipefail
-      builtin fc -lnr -2147483648 2> /dev/null |   # ( $'\t '<lines>$'\n' )* ; <lines> ::= [^\n]* ( $'\n'<lines> )*
-        $__fzf_awk "$script"                          |   # ( <counter>$'\t'<lines>$'\000' )*
+      builtin fc -lnr -2147483648 2> /dev/null   |   # ( $'\t '<lines>$'\n' )* ; <lines> ::= [^\n]* ( $'\n'<lines> )*
+        command $__fzf_awk "$script"             |   # ( <counter>$'\t'<lines>$'\000' )*
```
junegunn, line 88:
```suggestion
      builtin fc -lnr -2147483648 2> /dev/null |   # ( $'\t '<lines>$'\n' )* ; <lines> ::= [^\n]* ( $'\n'<lines> )*
        command $__fzf_awk "$script"           |   # ( <counter>$'\t'<lines>$'\000' )*
```

junegunn (changes requested):
Trying to remove unnecessary diffs.

junegunn:
> * Handling the case that `awk` isn't available is in principle fine by me, though at least POSIX wise, systems are expected to have `awk`. So assuming POSIX systems, handling this is unnecessary.

I also think it's unnecessary. The absence of awk can be a sign of a broken system, and we should probably warn the user with an error instead of silently disabling a subset of our feature, making it harder to fix the system.

shell/key-bindings.bash
```diff
@@ -66,16 +66,14 @@ if command -v perl > /dev/null; then
       READLINE_POINT=0x7fffffff
     fi
   }
-else # awk
+elif command -v awk > /dev/null || command -v mawk > /dev/null; then # awk - fallback for POSIX systems
```
junegunn, line 69:
```suggestion
else # awk - fallback for POSIX systems
```

shell/key-bindings.bash
```diff
@@ -112,20 +110,24 @@ if (( BASH_VERSINFO[0] < 4 )); then
...
-  # CTRL-R - Paste the selected command from history into the command line
-  bind -m emacs-standard '"\C-r": "\C-e \C-u\C-y\ey\C-u`__fzf_history__`\e\C-e\er"'
-  bind -m vi-command '"\C-r": "\C-z\C-r\C-z"'
-  bind -m vi-insert '"\C-r": "\C-z\C-r\C-z"'
+  if [[ $(type -t __fzf_history__) == function ]]; then
+    # CTRL-R - Paste the selected command from history into the command line
+    bind -m emacs-standard '"\C-r": "\C-e \C-u\C-y\ey\C-u`__fzf_history__`\e\C-e\er"'
+    bind -m vi-command '"\C-r": "\C-z\C-r\C-z"'
+    bind -m vi-insert '"\C-r": "\C-z\C-r\C-z"'
+  fi
```
junegunn, line 118:
```suggestion
  # CTRL-R - Paste the selected command from history into the command line
  bind -m emacs-standard '"\C-r": "\C-e \C-u\C-y\ey\C-u`__fzf_history__`\e\C-e\er"'
  bind -m vi-command '"\C-r": "\C-z\C-r\C-z"'
  bind -m vi-insert '"\C-r": "\C-z\C-r\C-z"'
```

shell/key-bindings.bash
```diff
@@ -112,20 +110,24 @@ if (( BASH_VERSINFO[0] < 4 )); then
...
-  # CTRL-R - Paste the selected command from history into the command line
-  bind -m emacs-standard -x '"\C-r": __fzf_history__'
-  bind -m vi-command -x '"\C-r": __fzf_history__'
-  bind -m vi-insert -x '"\C-r": __fzf_history__'
+  if [[ $(type -t __fzf_history__) == function ]]; then
+    # CTRL-R - Paste the selected command from history into the command line
+    bind -m emacs-standard -x '"\C-r": __fzf_history__'
+    bind -m vi-command -x '"\C-r": __fzf_history__'
+    bind -m vi-insert -x '"\C-r": __fzf_history__'
+  fi
```
junegunn, line 130:
```suggestion
  # CTRL-R - Paste the selected command from history into the command line
  bind -m emacs-standard -x '"\C-r": __fzf_history__'
  bind -m vi-command -x '"\C-r": __fzf_history__'
  bind -m vi-insert -x '"\C-r": __fzf_history__'
```

junegunn (approved):

junegunn:
Merged, thanks!

calestyo:
Looks good to me, though I probably still have placed the `command ` inside the variable, as described above.

calestyo:
Oh, one more thing that I had noticed before, but forgot to write down:
https://github.com/junegunn/fzf/blob/884856023ab53fa1eb7f8dbf158aa522daef0ae5/shell/key-bindings.bash#L75

Wouldn't it be safer to use `read -r`?

PR #3464, opened 2023-10-07
Add --builtin-filter-dirs

kelleyma49:
* fzf's builtin filesystem walker is faster then overridding via `FZF_DEFAULT_COMMAND`, especially on Windows
* If this switch is specified, fzf will read directories instead of files if `FZF_DEFAULT_COMMAND` is not specified and there's no input from tty

### DIFF
--- a/src/core.go
+++ b/src/core.go
@@ -117,7 +117,7 @@ func Run(opts *Options, version string, revision string) {
 		reader = NewReader(func(data []byte) bool {
 			return chunkList.Push(data)
 		}, eventBox, opts.ReadZero, opts.Filter == nil)
-		go reader.ReadSource()
+		go reader.ReadSource(opts.FilterDirs)
 	}
 
 	// Matcher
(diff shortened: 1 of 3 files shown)


kelleyma49:
I'm open to ideas, but using `FZF_DEFAULT_COMMAND` or passing through std-in is way slower than using the default walker in fzf.

kelleyma49:
I agree that design-wise it makes sense, but pragmatically it's just faster to walk the filesystem in native Go.

kelleyma49:
Here's an example comparing the built-in go walker vs Window's dir built-in command: 

fzfDiff

junegunn:
Thanks. Is the performance difference due to the performance limit of `dir`, or is it a fundamental limitation of the pipe mechanism in Windows? I mean, can we achieve better performance by using another program that is faster than `dir`?

junegunn:
Closed in #3649.

PR #3465, opened 2023-10-08
fix: #2094

### LINKED ISSUE #2094: /home/{user}/.fzf/shell/key-bindings.zsh:emulate:35: unknown argument -o

## Info

- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [ ] bash
    - [x] zsh
    - [ ] fish
    
I have followed the instruction to install fzf using git way.

I have selected to use key binding.

After installation and `source ~/.zshrc`, I have this error:

```
/home/{my-user-name}/.fzf/shell/key-bindings.zsh:emulate:35: unknown argument -o
```

And it keeps popping up every time I login to the host.

LangLangBart:
### Description

- fix #2094
- The 'emulate' command's ability to accept 'setopt' flags after the shell name is not available for
  Zsh versions below 5, causing an unwanted message when sourcing the file.
	- This issue was confirmed by @romkatv.
- Despite the original reporter claiming to have a more recent version of zsh, it's recommended they
  verify their version again. @fluency03

### DIFF
--- a/shell/completion.zsh
+++ b/shell/completion.zsh
@@ -67,7 +67,7 @@ fi
 # control. There are several others that could wreck havoc if they are set
 # to values we don't expect. With the following `emulate` command we
 # sidestep this issue entirely.
-'emulate' 'zsh' '-o' 'no_aliases'
+'builtin' 'emulate' 'zsh' && 'builtin' 'setopt' 'no_aliases'
 
 # This brace is the start of try-always block. The `always` part is like
 # `finally` in lesser languages. We use it to *always* restore user options.
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -32,7 +32,7 @@ else
   }
 fi
 
-'emulate' 'zsh' '-o' 'no_aliases'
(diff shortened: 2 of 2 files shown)


romkatv:
> * Despite the original reporter claiming to have a more recent version of zsh, it's recommended they
>   verify their version again. @fluency03

The bulletproof way to verify this would be to add `typeset -p ZSH_VERSION` directly to `/home/{my-user-name}/.fzf/shell/key-bindings.zsh`, preferably just before the `emulate` line.

LangLangBart:
> Your check is also fine but the echo statement is worse with its couple of tiny bugs. Not a big deal but you might just as well use the correct code if you decide to reject old zsh.

I will certainly take your advice and use the correct code. Thank you again for your help! I will proceed once the maintainer has made a decision on how to handle the issue.

junegunn:
I don't think we should put in extra effort to support zsh 4, but this patch is simple and short enough, so I'll merge it. Thanks.

PR #3467, opened 2023-10-08
[fish] exit as well when called from non-interactive shell

calestyo:
Let's see whether the test suite would pass if the old `--is-interactive` is used.

### DIFF
--- a/shell/key-bindings.fish
+++ b/shell/key-bindings.fish
@@ -11,6 +11,9 @@
 # - $FZF_ALT_C_COMMAND
 # - $FZF_ALT_C_OPTS
 
+status is-interactive; or exit 0
+
+
 # Key bindings
 # ------------
 function fzf_key_bindings


calestyo:
Nope. ^^ (Thus reverted the commit to the non-deprecated form).

Well, I don't know should we keep this until someone finds out what's wrong with the test suite on github?
I don't mind as I don't use fish (but i checked it with the commit, and it seemed to work).

junegunn:
```sh
# GitHub actions runs this version of ubuntu
docker run -it ubuntu:22.04

# Install packages
apt-get update
apt install -y git curl fish
fish --version
  # fish, version 3.3.1

git clone --depth 1 https://github.com/junegunn/fzf.git ~/.fzf
~/.fzf/install --all

cd ~/.fzf
patch -p1 < <(curl https://patch-diff.githubusercontent.com/raw/junegunn/fzf/pull/3467.diff)

fish
```


```
~/.config/fish/functions/fzf_key_bindings.fish (line 14): 'return' outside of function definition
status is-interactive; or return 0
                          ^
from sourcing file ~/.config/fish/functions/fzf_key_bindings.fish
        called on line 1 of file ~/.config/fish/functions/fish_user_key_bindings.fish
in command substitution
        called on line 2 of file ~/.config/fish/functions/fish_user_key_bindings.fish
in function 'fish_user_key_bindings'
in function '__fish_reload_key_bindings'
        called on line 189 of file /usr/share/fish/functions/__fish_config_interactive.fish
in function '__fish_config_interactive'
        called on line 138 of file /usr/share/fish/config.fish
in function '__fish_on_interactive'
in event handler: handler for generic event 'fish_prompt'
source: Error while reading file '/root/.config/fish/functions/fzf_key_bindings.fish'
fish: Unknown command: fzf_key_bindings
~/.config/fish/functions/fish_user_key_bindings.fish (line 2):
  fzf_key_bindings
  ^
in function 'fish_user_key_bindings'
in function '__fish_reload_key_bindings'
        called on line 189 of file /usr/share/fish/functions/__fish_config_interactive.fish
in function '__fish_config_interactive'
        called on line 138 of file /usr/share/fish/config.fish
in function '__fish_on_interactive'
in event handler: handler for generic event 'fish_prompt'
```

calestyo:
Thanks. That helped a lot.

It seems it's only since 3359e5d2e9bcbf19d1652636c8e448a6889302ae in fish 3.4.0, that `return` may be used outside a function (including in a `.` script).

We could now simply use `… or exit`, but while that would work fine for interactive shells, it would exit non-interactive ones (of course, the script should never be sourced there, but expectation is that it just gracefully doesn’t load the our script, not exit the sourcing one).

So I think the best way for now is to an `if` block. I've also open a 2nd PR #3475 already, which corrects that an which we can leave open for say 4 years, till all old versions of fish are out of use.

> So the question is: Are we going to drop support for fish 3.3 or below? I don't see a compelling reason to do so.

Well, I guess you know my opinion from the ancient-bash-versions topic... but in this case I'd actually agree.<br>
https://github.com/fish-shell/fish-shell/commit/3359e5d2e9bcbf19d1652636c8e448a6889302ae is from 2021... that's not so old.<br>
But I think we should re-visit this in some years... I don't see a reason to keep workarounds that just make the code more complex or error prone for completely outdated stuff - if someone really wants to keep using such things, it's IMO upon him to maintain a patchset for compatibility.

Anyway, for now, let's keep it compatible and revisit this via #3475 it in 2027 and see whether <3.4.0 is still somewhere in use. ;-)

Test runs through now. Please review&merge this one, thanks :-)

Issue #3470, opened 2023-10-09
`--help`-output and manpages are both out of sync with the actual code


calestyo:
- [x] I have the latest version of fzf
- [x] I have searched through the existing issues

## Problem / Steps to reproduce

While preparing a commit to update `fzf`’s own bash completion, I've noted that the `--help`-output and manpages are also outdated.

It seems that both, "fixed" values (like e.g. `no` for `--color`), and options (like e.g. `--algo` in the `--help`-output and `--no-black` in the manpage).

Whoever takes this work up … when being lazy and trying to "automate" this a bit:
```
$ cat ./fzf.1 | tr -d "\"'" | xargs -n 1 | sort -u | grep '^--'
```
may e.g. be used to find out all options that **are** in the manpage.

Cheers,
Chris.

junegunn:
Some of the differences are intentional. I have intentionally left out some obscure options that we may deprecate in the future in the `--help` output, which is probably more accessible for an average user than the man page.

A few examples:

* `--black` (and `--no-black`) is a legacy option that was added to work around an issue of ncurses in the old days. But  1. fzf has moved away from ncurses, 2. and the identical effect can now be achieved with `--color bg:black`. So it's no longer recommended to use this option and we only kept it for backward compatibility. To phase it out over time, we don't mention it in `--help` output. I think it's about time we should remove it from the man page as well.
* The omission of `--algo` is also intentional. It's an advanced option that few users will actually need, and I think it's better for an average user to not know/worry about it at all. We may also phase this out in the future. But that's TBD.

calestyo:
Hmm well, okay if you like it that way, then it's of course up to you.

I though it would have been easier to main if everything were always in. And also easier for users, especially also with respect to deprecation or "expert/obscure" option, if such options are clearly marked as such.

Just close if you feel so.

Issue #3457, opened 2023-09-28
should completion.* be split up?


calestyo:
- [X] I have read through the manual page (`man fzf`)
- [X] I have the latest version of fzf
- [X] I have searched through the existing issues

## Info

- OS
    - [X] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [X] bash
    - [ ] zsh
    - [ ] fish

## Problem / Steps to reproduce

I wonder whether the `completion.*` should/could be split up, at least in the bash case.

Some background:<br>
I've stumbled over [Debian bug #1013356](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1013356), where the problem is that `fzf`’s shell function for `**<TAB>` completion aren’t loaded properly.<br>
The reason is that `completion.bash` is installed into `/usr/share/bash-completion/completions/fzf`, which is then however only loaded on-demand (per shell session), which bash-completion only does, when completing `fzf` like in:
```
$ fzf <TAB>
```

Also, AFAIU, the `completion.*` ship currently code for two kinds of completions:
a. completing options for `fzf` itself (e.g. `fzf --height`)
b. completing other stuff (e.g. hostnames in `ssh **<TAB>`, etc.)
And AFAIU, there's no dependency on bash-completion.

Without any bigger changes to infrastructure (on the distro side), I'd see basically the following solutions for the Debian issue:
1. don't install the file at all, but keep it e.g. as `/usr/share/fzf/shell/completion.bash` (I, personally don't like if Debian would place it as `/usr/share/doc/fzf/examples/completion.*`, which it does right now for the other shell scripts from `fzf`)
2. install it into `/etc/profile.d/`
3. install it into `/etc/bash_completion.d/`
4. install it into `/etc/bash.bashrc.d` or something like that

(1) has the disadvantage that `fzf`’s completions (both, (a) and (b)) wouldn't work out of the box anymore. But has the advantage hat a (non-root) user can easily disable them (by not manually sourcing them) and easily use e.g. alternative solutions like [lincheney/fzf-tab-completion](https://github.com/lincheney/fzf-tab-completion).

(2) Would require an additional check in all the files, whether the right shell (e.g. bash or zsh) is active. The check for interactiveness alone wouldn't suffice. Also, it’s (at least in Debian) again not easily possible for a (non-root) user to disable loading of them.

[...]

PR #3471, opened 2023-10-09
[bash] bring fzf’s own bash completion up to date

calestyo:
Well, I tried to find out all possible options and "fixed" values as good as I could from the actual parsing code.<br>
Unfortunately both, `--help`-output and the manpage seem to be also outdated (see #3470), so these were couldn’t really be used as a reference.

Yeah,... I remember your comment in https://github.com/junegunn/fzf/issues/3457#issuecomment-1741704506 ... but it sounded rather unclear whether/when you'd go that "cobra"-way... and I don't want `fzf`’s own completion to be dropped.<br>
So here's an update.

That being said, I still think we should do #3457, but maybe we can discuss that in terms of some larger ideas.

btw:<br>
Why are you doing this:
https://github.com/junegunn/fzf/blob/a0d61b4c373f6b977314f7ebe53f6236798151e9/shell/completion.bash#L123-L126
?
Can it be dropped?

### DIFF
--- a/shell/completion.bash
+++ b/shell/completion.bash
@@ -69,55 +69,185 @@ _fzf_opts_completion() {
   cur="${COMP_WORDS[COMP_CWORD]}"
   prev="${COMP_WORDS[COMP_CWORD-1]}"
   opts="
+    -h --help
     -x --extended
     -e --exact
+    --extended-exact
+    +x --no-extended
+    +e --no-exact
+    -q --query
+    -f --filter
+    --literal
+    --no-literal
     --algo
-    -i +i
+    --scheme
+    --expect
+    --no-expect
+    --enabled --no-phony
+    --disabled --phony
+    --tiebreak
+    --bind
+    --color
+    --toggle-sort
+    -d --delimiter
     -n --nth
     --with-nth
-    -d --delimiter
+    -s --sort
     +s --no-sort
+    --track
+    --no-track
     --tac
-    --tiebreak
+    --no-tac
+    -i
+    +i
     -m --multi
+    +m --no-multi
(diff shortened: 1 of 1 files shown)


junegunn:
Thank you.

> it sounded rather unclear whether/when you'd go that "cobra"-way... 

Completion for fzf itself is not my priority right now so it's not clear *when* I'm going to work on it. But what is clear is that I don't want to maintain three different implementations to provide the same functionality across bash, zsh, and fish. So yeah, Cobra (or any equivalent library) is the way to go.

> btw: Why are you doing this:
> 
> https://github.com/junegunn/fzf/blob/a0d61b4c373f6b977314f7ebe53f6236798151e9/shell/completion.bash#L123-L126
> 
> 
> ?
> Can it be dropped?

I don't really remember. If it doesn't make any difference, we can drop it.

junegunn (approved):

calestyo:
> I don't really remember. If it doesn't make any difference, we can drop it.

Uhm... well at least I can't see any difference. Though I'm not really a completion expert and there may be some side-effects that I just don't realise.

It seems to have been added in 3b5281179639811eb9a1e5fde8aeed1264f842ed, but AFAICS there is no information on why the array is set to an empty one.

I've added a commit on top that drops the part, and if it actually breaks something,... will blame everything on you. 😜😇

calestyo:
btw: Not a serious problem, but any reason for always squashing the commits? Makes cleaning up the local branches more work ^^

junegunn:
I squash when the intermediate changes are minor and don't provide much value to be seen separately.

PR #3475, opened 2023-10-11
revisit after Oct 2027: drop support for ancient fish versions in interactiveness checks

calestyo:
This improves commit 5cb5bbdc32fa581c5e2c93552cf60a9ded4d147a by just returning
from the script in case it’s sourced from a non-interactive shell, rather then
using an `if`-block to not execute any further code (which is error prone, in
case code gets accidentally added outside the `if`, and also less readable).

This removes support for fish versions < 3.4.0, which should now no longer be
used.

### DIFF
--- a/shell/key-bindings.fish
+++ b/shell/key-bindings.fish
@@ -11,7 +11,7 @@
 # - $FZF_ALT_C_COMMAND
 # - $FZF_ALT_C_OPTS
 
-if status is-interactive
+status is-interactive; or return 0
 
 
 # Key bindings
@@ -171,8 +171,5 @@ function fzf_key_bindings
 
     echo $dir
   end
-
-end
-
 
 end
--- a/shell/key-bindings.fish
+++ b/shell/key-bindings.fish
@@ -11,7 +11,7 @@
 # - $FZF_ALT_C_COMMAND
 # - $FZF_ALT_C_OPTS
 
-status is-interactive; or return 0
+status is-interactive; or exit
 
 
 # Key bindings
--- a/shell/key-bindings.fish
+++ b/shell/key-bindings.fish
@@ -11,6 +11,9 @@
 # - $FZF_ALT_C_COMMAND
 # - $FZF_ALT_C_OPTS
 
+if status is-interactive
+
+
 # Key bindings
 # ------------
 function fzf_key_bindings
@@ -168,5 +171,8 @@ function fzf_key_bindings
 
     echo $dir
   end
+
+end
+
 
 end


shell/key-bindings.fish
```diff
@@ -11,6 +11,9 @@
 # - $FZF_ALT_C_COMMAND
 # - $FZF_ALT_C_OPTS
 
+status is-interactive; or return 0
```
junegunn, line 14: Let's see if this works.

```suggestion
status is-interactive; or exit
```

calestyo: @junegunn, I don't think this should be done, as argued in 5cb5bbdc32fa581c5e2c93552cf60a9ded4d147a.

Using `exit` defeats nearly the whole idea of the check, as it kills of any non-interactive script, that would source our script by accident.

junegunn: Welcome to the wonderful land of fish (which I don't fathom).

```sh
$ echo 'echo before-exit; exit; echo after-exit' > /tmp/x
$ fish -c 'echo before-source; source /tmp/x; echo after-source'

  before-source
  before-exit
  after-source

$ sh -c 'echo before-source; source /tmp/x; echo after-source'

  before-source
  before-exit
```

junegunn: Reference: https://github.com/NixOS/nix/pull/7105

junegunn: But I don't think this is really needed anyway. Because the file only defines a function and does not alter the state of the current shell until the user manually calls the function afterward. So I suspect not many fish users will use it in non-autoload setups.

calestyo: > Welcome to the wonderful land of fish (which I don't really understand either).

Hmm... indeed... seems to work. Weird.

From the comment in https://github.com/fish-shell/fish-shell/commit/3359e5d2e9bcbf19d1652636c8e448a6889302ae I assumed that `exit` would cause the fish shell to exit, bu looking at [fish-`exit`’s documentation](https://fishshell.com/docs/2.5/commands.html#exit) it’s even documented not to do so:
> exit causes fish to exit. If STATUS is supplied, it will be converted to an integer and used as the exit code. Otherwise, the exit code will be that of the last command executed.
> If exit is called while sourcing a file (using the source builtin) the rest of the file will be skipped, but the shell itself will not exit.

I would however suggest that we use `exit 0`... otherwise we get a exit status of `1` in the real code.

calestyo: Force pushed a commit that uses `exit 0` and has updated commit message.

calestyo: But in #3467.

calestyo:
Closing this a as proper solution found above and implemented via #3467.

Issue #3461, opened 2023-10-02
key-bindings fail if user used keyseqs have non-default assignments


calestyo:
- [x] I have read through the manual page (`man fzf`)
- [x] I have the latest version of fzf
- [x] I have searched through the existing issues

## Info

- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [x] bash
    - [ ] zsh
    - [ ] fish

## Problem / Steps to reproduce

Hey.

This is the issue that I've originally tried to solve with #3448, but failed miserably to do so at first. 😇😅<br>
See also the discussion at https://github.com/junegunn/fzf/pull/3448#pullrequestreview-1651651858 for some ideas.

To round up the problem:<br>
Any key-bindings that hat use the:
```
bind [-m keymap] keyseq:readline-command
```
form of `bind`, that is `Alt-C` for old and current bash versions:
https://github.com/junegunn/fzf/blob/4fdc08295b90485b7ee30a35d77177699c070113/shell/key-bindings.bash#L132-L134

as well as all the key-bindings for the ancient versions of bash:
https://github.com/junegunn/fzf/blob/4fdc08295b90485b7ee30a35d77177699c070113/shell/key-bindings.bash#L110-L118

are prone to fail, namely when any of the keyseqs used in the `readline-command` are already assigned in a non-default way.

Example as of current [master](e833823e15833122881a2cab333124c23fcc6ba5):
If one defines:
```
~$ bind -m emacs-standard -x '"\C-b\C-k": foo'
~$ foo() { :; }
```
and then has a readline like:
```
~$ date
```
(i.e. no `Enter` pressed yet and the cursor being at the end of the line) and now does `Alt-C` one gets e.g. something like:
```
~$ `__fzf_cd__`
>                                                                                                                                                                                             
  97/97 ──────────────────────────────────────────────────────────
> go
  go/pkg
  go/pkg/mod
```
now I select `go` and `Enter` and I get a readline like:
```
~/go$ ate
```
(i.e. the `d` swallowed and the cursor is now at the `a`), which is not just a visual glitch as pressing `Enter` causes:
```
~/go$ ate  
-bash: ate: command not found
```

calestyo:
There's a similar issue in that we override bindings for "internal purposes", which the user may however have set (or may (re-)set after our code has been sourced).
I *don't* mean bindings like:
[...]

Issue #3455, opened 2023-09-26
aborted **<TAB> completion causes supurios newline to be added


calestyo:
- [*] I have read through the manual page (`man fzf`)
- [*] I have the latest version of fzf
- [*] I have searched through the existing issues

## Info

- OS
    - [*] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [*] bash
    - [ ] zsh
    - [ ] fish

## Problem / Steps to reproduce
With current `master´ (i.e. also without any of my own commits to the hostname code), I've noticed the following:

If I do:
```
$ telnet **<TAB>
<ESC>
```
then I get
```
$ telnet **_
```
(with a trailing space, symbolised above by the `_`).

If I do:
```
$ ssh **<TAB>
<ESC>
```
then I get
```
$ ssh **_
```
(with a trailing space, symbolised above by the `_`).

If I do:
```
$ ssh foo@**<TAB>
<ESC>
```
then I get
```
$ telnet foo@_
```
(with a trailing space, symbolised above by the `_`).

1. Wouldn't it be better, if in no case a trailing ` ` (space) gets added?
2. IMO, both cases should either remove the `**` on an aborted completion attempt, or keep it (or even better, it should be configurable whether it's kept or not.

Any ideas? :-)

Thanks,
Chris.

calestyo:
I think I found at least the relevant code sections, though I don't yet understand them completely (well it's 02:36 here ... time for a nap ^^).

Will have a look at it tomorrow, maybe I can produce some fix.

I'm thinking whether it would be desirable for users to be able to choose whether or not the `**` is kept, if no selection took place.<br>
I think both modes could have their benefits, what do you think?

Also I wonder, whether there's any case, when one wants a ` ` to be added, when no selection took place. So far I couldn't think of any.<br>
If one completes `  **<TAB>`, and no selection took place, one either wants the `**` to stay (to be able to re-try) or to go away (therefore my thinking above, whether this should be user's choice) - but I see no reason why an extra space should be added in either case.

Same if one completes `foo**<TAB>`... whether the `**` stays or not, I don't think one wants a space, because obviously the `foo` is not the finished word, if so, one wouldn’t have tried the `**<TAB>` completion. So either one wants to be able to `<TAB>` again immediately (without again typing `**`) or one wants to manually finish `foo` (again, users may want to be able to do both).

PR #3476, opened 2023-10-12
[bash] adapt to bash-completion’s naming consolidation

calestyo:
Recently, bash-completion started to give their functions a more systematic naming schema by adding a `_comp`- respectively `_comp_compgen`-prefix.

bash-completion’s `_known_hosts_real()`, which we optionally use, was renamed in bash-completion commit d3119216c5de651e39e1026d0e81faf8949d46f6 to `_comp_compgen_known_hosts()`, which is now, AFAIU “official API”.

This commit adds support for and prefers the new name, but – if not found – falls back to the old name, in order to support old versions of bash-completion for the foreseeable future.

### DIFF
--- a/shell/completion.bash
+++ b/shell/completion.bash
@@ -411,8 +411,18 @@ _fzf_proc_completion_post() {
 # The function is expected to print hostnames, one per line as well as in the
 # desired sorting and with any duplicates removed, to standard output.
 if ! declare -F __fzf_list_hosts > /dev/null; then
-  if declare -F _known_hosts_real > /dev/null; then
-    # if available, use bash-completions’s _known_hosts_real() for getting the list of hosts
+  if declare -F _comp_compgen_known_hosts > /dev/null; then
+    # if available, use bash-completions’s _comp_compgen_known_hosts() for getting the list of hosts
+
+    __fzf_list_hosts() {
+      # set the local attribute for any non-local variable that is set by _comp_compgen_known_hosts()
+      local COMPREPLY=()
+
+      _comp_compgen_known_hosts ''
+      printf '%s\n' "${COMPREPLY[@]}" | command sort -u --version-sort
+    }
+  elif declare -F _known_hosts_real > /dev/null; then
+    # if available, use old bash-completions’s _known_hosts_real() for getting the list of hosts
 
     __fzf_list_hosts() {
       # set the local attribute for any non-local variable that is set by _known_hosts_real()


calestyo:
I haven't tested this yet, so please wait with merging.

junegunn:
Thanks, but I already decided to drop the dependency. f212bafe46233a207d4308ad368884c210cedc46

I can see that you're ambitious about extending the bash component in this repository, and I appreciate your efforts, but as I said earlier, I want to keep the feature set and code minimal.

calestyo:
> I can see that you're ambitious about extending the bash component in this repository, and I appreciate your efforts, but as I said earlier, I want to keep the feature set and code minimal.

Well that's really bad... cause in practise it will be quite hard to set up a "competitor" to the integration scripts shipped by fzf itself, as most people will simply use what's shipped out of the box.

That in turn is however quite limited.<br>
Or at least it could be made much better, like with a generic framework that allows customisation (without having to overwrite all functions).

For the complexity added by that, one could on the other hand simplify things by throwing much code where right now you handle completion specifically for a few cases (e.g. ssh’s hosts, or ssh’s -i).<br>

What I would have envisioned (well at least in very short and just a core part of it), was to simply complete whatever bash-completion gives back (if available) (with one exception, general file paths).

If you take a look at e.g.:<br>
https://brbsix.github.io/2015/11/29/accessing-tab-completion-programmatically-in-bash/<br>
it sees to be rather easy to determine whatever bash completion would do a given readline position.

With that, we would have simply gotten nearly all (well as sad, except for filenames, that we'd have needed to handle specially) completions for free, without any need to maintain `-i` and whatsoever on our own.

This would also have fixed all cases where right now completion is simply wrong, e.g. cases like:
```
$ ls --**<Tab>
$ ssh -o **
```
which doesn't complete options, and literally all other cases that we don't handle specifically in our `completion.*`

Another thing I'd have done is to replace `**<Tab>`, which is a bit unhandy, with e.g. `<Shift-Tab>`... simpler to type, and it can be made to not show up on the terminal, just like the normal `<Tab>` completion.

Another idea would have been to allow users to configure e.g. whether (normal) `<Tab>` causes always normal shell completion, or always the fzf completion, or even selectively per cwords (which I think would make sense, cause I think for some types of completions, fzf is better suited, and for some the plain one.

I'd have also hoped to fix buggy completions in the current code, e.g. when aborting `fzf`, sometimes it removes the `**`; sometimes not, sometimes a space is added, etc. (which is #3455).


Well... it's really disappointing that you want to keep the shell integration so limited :-(
I don't think it makes then that much sense for me to get any contributions merged, cause all would be either to fix existing bugs like #3461 or expand functionality and usability like the one with hostname completion or the ideas outlined above.

Maybe I'll ask bash-completion guys if they'd be open to a mode that allows calling out to `fzf`, though I'd guess I'll rather write it as some home-brew solution just for myself.

Cheers,
Chris.

calestyo:
> I just don't have the bandwidth to maintain the ever-increasing feature sets of the shell components.

Sure, but then I absolutely don't understand why for example you didn't like the host completion code in the end?!<br>
I mean right now you have code which is partially broken (e.g. IIRC, it doesn't consider the GlobalKnownHostsFile, or manually specified known hosts files or included ssh config snippets)... so either it's only of limited use for people or they'll come around and want it to be fixed (meaning more maintenance burden).<br>
Or they want new tools or new options be supported, again, more maintenance or the need to reject such wishes.

By outsourcing this to bash-completion, I'd have expected that one needs to set up a generic/powerful integration framework only once,... and well... apart from minor adaptions and improvements, that would be it.<br>
Especially no more handling cases like `host`, `telnet`, etc. specifically.

> Making sure everything works as expected on three different shells on different OSes, keeping bash and zsh interfaces as close to each other as possible

Well... TBH, that you probably won't get (at least not by what I've envisioned). I mean I have not much idea about zsh and fish... and bash-completion is, well, bash.<br>
But question is, is it really problem if functionality of them is out of sync? I would rather say not. People typically use either shell and just as there are features in e.g. fish which aren't in bash... there would be then features in e.g. fzf’s bash integration which aren't in it’s integration for the others. One could of course try to not intentionally break things like Making `<Alt-C>` something completely different in bash than in zsh.<br>
And if some skilled zsh/fish fellow comes along he could still try to remake features added only to the integration of one shell for others.<br>
If functionality of all of them needs to be kept absolutely in sync, it's in fact unlikely that it will ever extend by a lot.

> I'm left with the responsibility of maintaining their code

Which does not necessarily mean that you must make absolutely sure that accepted code is 100% working. I mean you should probably have a look that it doesn't contain any accidental `rm -rf /` or backdoors... but apart from that, even if you merge something which then turns out it doesn't work... well that's open source, and people can report back and fix.



> There are some third-party projects for fzf/shell integration and I think that's great. Users who find the minimal integration limiting can turn to a more ambitious project, and users who feel otherwise can just use our version.

There's also https://github.com/rockandska/fzf-obc ... but I think their common problem is that none of them is "official" so they have to compete amongst each other and even start to get packaged by distros to be of a broader use.


Anyway.... just my 2 ct.

PR #3504, opened 2023-11-06
feat(fzf.vim): Adding option `g:fzf_fullscreen_layout`

alps2006:
The behavior of  fzf `Command!` (fullscreen)  has some differences between  Vim and Neovim.  

In Neovim,  it create a new tab to cause some exceptions when using `vimspector`  (but Vim does not).

https://github.com/puremourning/vimspector/issues/771

This option `g:fzf_fullscreen_layout` can customize the  fullscreen layout as you like. Below is an example:

```vim
    if has('nvim')
        let g:fzf_layout = { 'window': { 'width': 1, 'height': 0.4, 'relative': v:true, 'xoffset': 0, 'yoffset': 1 } }
        let g:fzf_fullscreen_layout = { 'window': { 'width': 1, 'height': 1, 'relative': v:false, 'xoffset': 0, 'yoffset': 1 } }
    else
        if exists('$TMUX')
            " See `fzf-tmux --help` for available options
            let g:fzf_layout = { 'tmux': '-p99%,40% -y 60%' }
            let g:fzf_fullscreen_layout = { 'tmux': '-p99%,96% -y 4%' }
        endif
    endif

```

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -403,6 +403,9 @@ function! fzf#wrap(...)
         call remove(opts, key)
       endif
     endfor
+    if exists('g:fzf_fullscreen_layout')
+      let opts = extend(opts, s:validate_layout(get(g:, 'fzf_fullscreen_layout', {})))
+    endif
   elseif !s:has_any(opts, s:layout_keys)
     if !exists('g:fzf_layout') && exists('g:fzf_height')
       let opts.down = g:fzf_height


junegunn:
I have no experience with the plugin and I don't use Neovim, so I'm not sure what's going on, but I don't quite understand how opening a tab, which is a mundane operation, would cause a problem. Shouldn't it be addressed by the plugin?

alps2006:
I think this option is very useful in some situations. Below is an operation in Neovim that create a new tab.

junegunn:
Your primary motivation for this new configuration variable is to work around the vimspector problem, so I insist that the original problem be fixed first. After that, we can discuss whether we really need a new variable here.

I have read through the discussion you linked above (https://github.com/puremourning/vimspector/issues/771), and it doesn't sound like an unfixable problem. Do we have to rely on tab numbers? Can we set up a tab-local variable (`t:something`) on the relevant tab and loop through the existing tabs (`range(1, tabpagenr('$')`) to find the one with the variable using `gettabvar()`?

alps2006:
Thanks for your advice. I will resolve the problem first.

PR #3507, opened 2023-11-08
Fix Home, End on rxvt-unicode

liskin:
Not much else to be said about this… urxvt emits slightly different codes than xterm.


junegunn:
Thanks!

Issue #3511, opened 2023-11-19
Unable to create temporary files in /tmp on OpenBSD


lcheylus:
- [x] I have read through the manual page (`man fzf`)
- [x] I have the latest version of fzf
- [x] I have searched through the existing issues

## Info

- OS
    - [ ] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [x] OpenBSD
- Shell
    - [x] bash
    - [ ] zsh
    - [ ] fish

## Problem / Steps to reproduce

On OpenBSD current version (amd64), unable to create temporary files with fzf version 0.44.1 in `/tmp` directory.

I'm using `ytfzf` Python script (https://github.com/pystardust/ytfzf) to find and watch Youtube videos from the terminal. This tool uses fzf with temporary files to stream video with `mpv` tool.

```bash
$ ytfzf -t -T {chafa,mpv,imv}
fzf[85237]: pledge "wpath", syscall 5

This happens, because fzf is trying to open the prefiew image in /tmp

  46440 fzf      NAMI  "/tmp/fzf-preview-1776267355"
  46440 fzf      PLDG  open, "wpath", errno 1 Operation not permitted
```

This error is due to OpenBSD protector with `pledge` (`src/protector/protector_openbsd.go`). We need to add `tmppath` in pledge permissions to allow creation of temp files in `/tmp` directory.

I will submit a PR to fix this issue.

PR #3512, opened 2023-11-19
Allow files creation in /tmp on OpenBSD

lcheylus:
- src/protector/protector_openbsd.go: add tmppath for pledge permissions
  - fix junegunn/fzf#3511


junegunn:
Thanks!

PR #3542, opened 2023-12-16
Terminate simple server success response with double CRLF

### LINKED ISSUE #3541: xh reports error when sending "put(x)" to fzf
- I have read through the manual page (`man fzf`)
- I have the latest version of fzf
- I have searched through the existing issues

- OS
    - Linux

- Shell
    - bash

https://github.com/ducaale/xh/issues/341

blyxxyz:
The simple success case had only the status line plus a single CRLF, and pedantic HTTP client implementations (`hyper`) stumbled over this. The RFC [requires two](https://www.rfc-editor.org/rfc/rfc9112#section-2.1) even if there's no body.

Fixes #3541.

### DIFF
--- a/src/server.go
+++ b/src/server.go
@@ -218,7 +218,7 @@ func (server *httpServer) handleHttpRequest(conn net.Conn) string {
 	}
 
 	server.actionChannel <- actions
-	return httpOk
+	return httpOk + crlf
 }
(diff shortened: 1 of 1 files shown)


junegunn:
Merged, thanks!

PR #3546, opened 2023-12-21
zsh keybindings: ctrl-t: pass the last buffer word as query

bretello:
When using the Ctrl-T keybinding, to complete paths, I often start writing part of a path only to realize I want to use fzf. 

With this patch, Ctrl-T now uses the last word in the buffer as initial query for fzf, which improves this scenario.

Here's a demo:

[zsh-keybindings-query](https://asciinema.org/a/XG5gFyULA0XKqcby0QppOi2y4)

 This always uses the last word in the buffer, so typing something like `ls<space><ctrl-t>` will start fzf with `ls` as query

### DIFF
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -39,7 +39,7 @@ fi
 
 {
 
-# CTRL-T - Paste the selected file path(s) into the command line
+# CTRL-T - Paste the selected file path(s) into the command line using the last part of the buffer as fzf query
 __fsel() {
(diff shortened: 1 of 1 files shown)


bretello:
looking into the test failures

tmpm697:
You achieve this with shell functions.

eli-percepto:
Consider also adding 2. from here: https://github.com/junegunn/fzf/pull/3918

PR #3557, opened 2024-01-01
fix(zsh): transform in CTRL_R

### LINKED ISSUE #3556: zsh: transform does not work with CTRL+R

## Info

- OS
    - [x] Linux
    - [x] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [ ] bash
    - [x] zsh
    - [ ] fish

In latest version 0.45.0, all these options won't work when trigger via key bingding `CTRL-R` with ZSH when there's no match for the query. Prompt is not replaced with query after `enter`.
```sh
# You can make CTRL-R paste the current query when there's no match
export FZF_CTRL_R_OPTS='--bind enter:accept-or-print-query'

# 'become' is apparently more versatile but it's not available on Windows.
export FZF_CTRL_R_OPTS='--bind "enter:become:if [ -z {} ]; then echo {q}; else echo {}; fi"'

# Using the new 'transform' action
export FZF_CTRL_R_OPTS='--bind "enter:transform:[ -z {} ] && echo print-query || echo accept"'
```

But the command along works normally:
```
history | fzf
```

danztran:
fix #3556

### DIFF
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -103,7 +103,7 @@ fzf-history-widget() {
   local ret=$?
   if [ -n "$selected" ]; then
     num=$(awk '{print $1}' <<< "$selected")
-    if [[ "$num" =~ ^[0-9]+$ ]]; then
+    if [[ "$num" =~ ^[1-9][0-9]*$ ]]; then
       zle vi-fetch-history -n $num
     else # selected is a custom query, not from history
       LBUFFER="$selected"
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -103,8 +103,10 @@ fzf-history-widget() {
   local ret=$?
   if [ -n "$selected" ]; then
     num=$selected[1]
-    if [ -n "$num" ]; then
+    if [[ "$num" =~ ^-?[0-9]+$ ]]; then
       zle vi-fetch-history -n $num
+    else # selected is a custom query, not from history
+      LBUFFER="${selected[@]}"
     fi
   fi
   zle reset-prompt
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -104,7 +104,7 @@ fzf-history-widget() {
   if [ -n "$selected" ]; then
     num=$selected[1]
     if [[ "$num" =~ ^-?[0-9]+$ ]]; then
-      zle vi-fetch-history -n $num
+      LBUFFER="${selected:1}"
     else # selected is a custom query, not from history
       LBUFFER="${selected[@]}"
     fi
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -104,7 +104,7 @@ fzf-history-widget() {
   if [ -n "$selected" ]; then
     num=$selected[1]
     if [[ "$num" =~ ^-?[0-9]+$ ]]; then
-      LBUFFER="${selected:1}"
+      zle vi-fetch-history -n $num
     else # selected is a custom query, not from history
       LBUFFER="${selected[@]}"
     fi
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -103,7 +103,7 @@ fzf-history-widget() {
   local ret=$?
   if [ -n "$selected" ]; then
     num=$selected[1]
-    if [[ "$num" =~ ^-?[0-9]+$ ]]; then
+    if [[ "$num" =~ ^[0-9]+$ ]]; then
       zle vi-fetch-history -n $num
     else # selected is a custom query, not from history
       LBUFFER="${selected[@]}"
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -98,15 +98,15 @@ bindkey -M viins '\ec' fzf-cd-widget
 fzf-history-widget() {
   local selected num
   setopt localoptions noglobsubst noposixbuiltins pipefail no_aliases 2> /dev/null
-  selected=( $(fc -rl 1 | awk '{ cmd=$0; sub(/^[ \t]*[0-9]+\**[ \t]+/, "", cmd); if (!seen[cmd]++) print $0 }' |
-    FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} ${FZF_DEFAULT_OPTS-} -n2..,.. --scheme=history --bind=ctrl-r:toggle-sort,ctrl-z:ignore ${FZF_CTRL_R_OPTS-} --query=${(qqq)LBUFFER} +m" $(__fzfcmd)) )
+  selected="$(fc -rl 1 | awk '{ cmd=$0; sub(/^[ \t]*[0-9]+\**[ \t]+/, "", cmd); if (!seen[cmd]++) print $0 }' |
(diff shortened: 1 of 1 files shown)


shell/key-bindings.zsh
```diff
@@ -103,8 +103,10 @@ fzf-history-widget() {
   local ret=$?
   if [ -n "$selected" ]; then
     num=$selected[1]
-    if [ -n "$num" ]; then
+    if [[ "$num" =~ ^-?[0-9]+$ ]]; then
       zle vi-fetch-history -n $num
```
danztran, line 107: I also found that `vi-fetch-history` does not always work.
Step to reproduce: 
```
CTRL_R 
-> select entry_at_n
-> insert entry_at_n to prompt (work normally)
-> clear prompt 
-> CTRL_R again 
-> select entry_at_n
-> get empty prompt (not work anymore)
```

My guess is `vi-fetch-history` does not work properly when re-select entries that were selected before. It might also mess-up with setopt `APPEND_HISTORY`, `INC_APPEND_HISTORY` and `SHARE_HISTORY`
So i suggest to change this behavior too normal text replacement.
```suggestion
      LBUFFER="${selected:1}"
```

nguyentranbao-ct:
The CI failed because the test miss a trailing space. However, the test runs normally on my machine. Is it ok to modify the test?

danztran:
The CI failed because the test miss a trailing space. However, the test runs normally on my machine. Is it ok to modify the test?

shell/key-bindings.zsh
junegunn: The downside of LBUFFER approach is that it doesn't work nicely with multi-line commands.

```sh
echo '
  hello
  world
' &&
echo '
  foo
  bar
'
```

```
$ echo '\n hello\n world\n' &&\necho '\n foo\n bar\n'

 hello
 world

zsh: command not found: necho
```

junegunn:
> The CI failed because the test miss a trailing space. However, the test runs normally on my machine. Is it ok to modify the test?

Strange. I can't reproduce the error either. Let me look into it.

junegunn:
I can reproduce the error on the Ubuntu docker container which GitHub actions uses.

```diff
diff --git a/Dockerfile b/Dockerfile
index 308016f..aba5f22 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
-FROM --platform=linux/amd64 archlinux
-RUN pacman -Sy && pacman --noconfirm -S awk git tmux zsh fish ruby procps go make gcc
+FROM --platform=linux/amd64 ubuntu:22.04
+RUN apt-get update -y && apt install -y git make golang zsh fish ruby tmux
 RUN gem install --no-document -v 5.14.2 minitest
 RUN echo '. /usr/share/bash-completion/completions/git' >> ~/.bashrc
 RUN echo '. ~/.bashrc' >> ~/.bash_profile
@@ -8,4 +8,5 @@ RUN echo '. ~/.bashrc' >> ~/.bash_profile
 RUN rm -f /etc/bash.bashrc
 COPY . /fzf
 RUN cd /fzf && make install && ./install --all
+ENV LANG C.UTF-8
 CMD tmux new 'set -o pipefail; ruby /fzf/test/test_go.rb | tee out && touch ok' && cat out && [ -e ok ]
```

```
  1) Failure:
TestZsh#test_ctrl_r [/fzf/test/test_go.rb:3188]:
Expected: "echo 3rd"
  Actual: " echo 3rd"
```

But this is a non-issue if we're not going to use the LBUFFER approach.

shell/key-bindings.zsh
danztran: Got it. LBUFFER and echo can also preserve new lines, but not preserve raw input escape `\n`.
Can't find a better solution, so i'm gonna reverse it.

```sh
LBUFFER="echo ${selected:1}"
```
```sh
echo '
  hell\n\no
  world
' &&
echo '
  foo
  bar
'
```
```sh
$ echo '
  hell

o
  world
' &&
echo '
  foo
  bar
'
```

shell/key-bindings.zsh
```diff
@@ -103,8 +103,10 @@ fzf-history-widget() {
   local ret=$?
   if [ -n "$selected" ]; then
     num=$selected[1]
-    if [ -n "$num" ]; then
+    if [[ "$num" =~ ^-?[0-9]+$ ]]; then
```
junegunn, line 106: Do we need the `-?` part? In my case, `fc -rl 1` doesn't show any entries whose number starts with `-`.

nguyentranbao-ct: Yup, it's redundant, removed.

danztran: Yup, it's redundant, removed.

shell/key-bindings.zsh
```diff
@@ -103,8 +103,10 @@ fzf-history-widget() {
   local ret=$?
   if [ -n "$selected" ]; then
     num=$selected[1]
-    if [ -n "$num" ]; then
+    if [[ "$num" =~ ^[0-9]+$ ]]; then
       zle vi-fetch-history -n $num
+    else # selected is a custom query, not from history
+      LBUFFER="${selected[@]}"
```
junegunn, line 109: This doesn't preserve multiple whitespace characters. e.g. `foo         bar` becomes `foo bar`

nguyentranbao-ct: Fixed. I turned `$selected` into plain text instead of array but not sure there is any side effect.

danztran: Fixed. Turned $selected into plain text instead of array but i'm not sure there is any side effect.

shell/key-bindings.zsh
```diff
@@ -98,13 +98,15 @@ bindkey -M viins '\ec' fzf-cd-widget
...
-  selected=( $(fc -rl 1 | awk '{ cmd=$0; sub(/^[ \t]*[0-9]+\**[ \t]+/, "", cmd); if (!seen[cmd]++) print $0 }' |
-    FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} ${FZF_DEFAULT_OPTS-} -n2..,.. --scheme=history --bind=ctrl-r:toggle-sort,ctrl-z:ignore ${FZF_CTRL_R_OPTS-} --query=${(qqq)LBUFFER} +m" $(__fzfcmd)) )
+  selected="$(fc -rl 1 | awk '{ cmd=$0; sub(/^[ \t]*[0-9]+\**[ \t]+/, "", cmd); if (!seen[cmd]++) print $0 }' |
+    FZF_DEFAULT_OPTS="--height ${FZF_TMUX_HEIGHT:-40%} ${FZF_DEFAULT_OPTS-} -n2..,.. --scheme=history --bind=ctrl-r:toggle-sort,ctrl-z:ignore ${FZF_CTRL_R_OPTS-} --query=${(qqq)LBUFFER} +m" $(__fzfcmd))"
   local ret=$?
   if [ -n "$selected" ]; then
-    num=$selected[1]
-    if [ -n "$num" ]; then
+    num=$(awk '{print $1}' <<< "$selected")
+    if [[ "$num" =~ ^[0-9]+$ ]]; then
```
junegunn, line 106:
```suggestion
    if [[ "$num" =~ ^[1-9][0-9]*$ ]]; then
```

PR #3590, opened 2024-01-19
docs(ADVANCED.md): Update switch between rg/fzf example with new feature

LangLangBart:
### description

Make use of the `transform` action and the `$FZF_PROMPT` environment variable to
toggle between ripgrep and fzf with a single hotkey.

### DIFF
--- a/ADVANCED.md
+++ b/ADVANCED.md
@@ -1,8 +1,8 @@
 Advanced fzf examples
 ======================
 
-* *Last update: 2023/12/29*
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks.

The use of `transform` action is obviously more "advanced" and may be a little daunting for some readers. So how about keeping the current example, and adding the new one with `transform` after it as a separate section?

LangLangBart:
I understand. I will make the necessary adjustments.

junegunn:
Love it, thanks!

Issue #3582, opened 2024-01-14
Expose states to the subprocesses via environment variables


junegunn:
fzf currently uses two mechanisms to expose its internal states to its child processes.

## Via placeholder expressions

* `{}` / `{+}` / `{+f}` / `{1}` / `{1..3}` / etc
* `{q}` or `{fzf:query}`
* `{fzf:action}`
* `{fzf:prompt}`

### Pros.

* Can implement its own syntax. e.g. `{2..-1}`
* Easier to type
* Can trigger preview update even when the focus hasn't changed `fzf --preview 'echo {q}'`

### Cons.

* Harder to implement. Escaping required.
* Possible name clash in the script

## Via environment variables

* `$FZF_PORT`
* `$FZF_PREVIEW_LINES`
* `$FZF_PREVIEW_COLUMNS`
* `$FZF_PREVIEW_TOP`
* `$FZF_PREVIEW_LEFT`

### Pros.

* Easier to implement. No manual escaping.
* The values are exported to child processes

### Cons.

* fzf doesn't know if the script refers the variables, so no forced preview update possible

---

I'd like to expose more information to the subprocesses to facilitate the development of sophisticated applications. #3572

So, `$FZF_TOTAL_COUNT` or `{fzf:total-count}`?

I'm leaning towards the former as I don't see much benefit of the placeholder form in these cases. I'm also thinking of deprecating `{fzf:action}`, `{fzf:prompt}` in favor of `$FZF_ACTION` and `$FZF_PROMPT` in the next version, I think it was a mistake to add them.

PR #3593, opened 2024-01-22
docs(ADVANCED.md): replace placeholder with env variable for toggle single key binding

LangLangBart:
### description
Since `{fzf:prompt}` is being phased out, let's replace the example with `$FZF_PROMPT`.
- Ref #3582

---

The previous PR was unintentionally closed. This implements the same changes as before.


junegunn:
Thanks!

Issue #2705, opened 2022-01-05
Ctrl-T command (__fzf_select__) never finishes, consumes all RAM, by default on macos


hraftery:
- [X] I have read through the manual page (`man fzf`)
- [X] I have the latest version of fzf
- [X] I have searched through the existing issues

## Info

- OS
    - [ ] Linux
    - [X] Mac OS X
    - [ ] Windows
- Shell
    - [X] bash
    - [ ] zsh
    - [ ] fish

## Problem / Steps to reproduce

1. Fairly standard setup running macOS 11.6.2, except default shell replaced with bash 5.1.8.
1. Install with `brew install fzf`
1. Then run `/usr/local/opt/fzf/install` to install key bindings and shell integrations ([ref](https://sourabhbajaj.com/mac-setup/iTerm/fzf.html)).
1. Then hit `Ctrl-T`.

While the command can be used right away, searching and selecting the files found, the number of entries grows into the tens of millions and continues endlessly. If left for 10 minutes or so, `fzf` consumes 16GB of RAM, and attempts to continue. With only 16GB of RAM on this machine, the computer becomes increasingly unresponsive and I eventually aborted the command. No lingering effects, but the behaviour can be repeated.

## Expected Behaviour

Since it's not clear which files are found first, if I'm looking for a particular file I would expect to have to wait until the search is complete to be sure. It turns out the search does not complete before bringing the system down. I expected the search to complete and the full list of files to be searchable.

## References

I wasn't able to find many relevant reports, but the search terms are tricky to get right, so I'm littering this issue with key terms to provide a trail for posterity. This is the most relevant issue I could find:

[Find process got stuck in the background and led to high CPU usage](https://github.com/junegunn/fzf/issues/2384)

hraftery:
After some exploration, I traced the issue to `shell/key-bindings.bash`, which binds `Ctrl-T` to `fzf-file-widget()`, which runs `__fzf_select__()`. That function ultimately runs this command:

```
find -L . -mindepth 1 \( -path '*/\.*' -o -fstype 'sysfs' -o -fstype 'devfs' -o -fstype 'devtmpfs' -o -fstype 'proc' \) -prune -o -type f -print -o -type d -print -o -type l -print 2> /dev/null
```

If I run that command a redirect to a file, the command runs indefinitely as the file grows to many gigabytes. So it seems like the likely culprit.

Skimming the file I see it often contains entries like:

```
./Library/Containers/<something>/Data/...
```

[...]

PR #3647, opened 2024-02-26
Exclude macOS containers from default find commands

timhillgit:
Specifically, exclude any file with the com.apple.containermanager.uuid extended attribute from the default Ctrl-T and Alt-C commands.

See additional discussion at #2705

### DIFF
--- a/shell/key-bindings.bash
+++ b/shell/key-bindings.bash
@@ -17,8 +17,11 @@
 # Key bindings
 # ------------
 __fzf_select__() {
-  local cmd opts
-  cmd="${FZF_CTRL_T_COMMAND:-"command find -L . -mindepth 1 \\( -path '*/.*' -o -fstype 'sysfs' -o -fstype 'devfs' -o -fstype 'devtmpfs' -o -fstype 'proc' \\) -prune \
+  local cmd opts macos_exclude
+  if [[ "$(uname -s)" == "Darwin" ]]; then
+    macos_exclude="-o -xattrname 'com.apple.containermanager.uuid'"
+  fi
+  cmd="${FZF_CTRL_T_COMMAND:-"command find -L . -mindepth 1 \\( -path '*/.*' -o -fstype 'sysfs' -o -fstype 'devfs' -o -fstype 'devtmpfs' -o -fstype 'proc' $macos_exclude \\) -prune \
     -o -type f -print \
     -o -type d -print \
     -o -type l -print 2> /dev/null | command cut -b3-"}"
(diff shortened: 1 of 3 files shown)


shell/key-bindings.bash
```diff
@@ -17,8 +17,11 @@
 # Key bindings
 # ------------
 __fzf_select__() {
-  local cmd opts
-  cmd="${FZF_CTRL_T_COMMAND:-"command find -L . -mindepth 1 \\( -path '*/.*' -o -fstype 'sysfs' -o -fstype 'devfs' -o -fstype 'devtmpfs' -o -fstype 'proc' \\) -prune \
+  local cmd opts macos_exclude
+  if [[ "$(uname -s)" == "Darwin" ]]; then
+    macos_exclude="-o -xattrname 'com.apple.containermanager.uuid'"
+  fi
+  cmd="${FZF_CTRL_T_COMMAND:-"command find -L . -mindepth 1 \\( -path '*/.*' -o -fstype 'sysfs' -o -fstype 'devfs' -o -fstype 'devtmpfs' -o -fstype 'proc' $macos_exclude \\) -prune \
```
junegunn, line 24: How about if we keep it simple and just put `-xattrname 'com.apple.containermanager.uuid'` regardless of the platform? Will there be any negative consequences for users?

timhillgit: Unfortunately yes. When find encounters a predicate it doesn't recognize it bails immediately and produces no results. Which actually brings up another issue: if someone on macOS has installed GNU find and deliberately placed it earlier on their PATH this command will similarly fail. On the one hand doing this is a "buyer beware" sort of situation but it's not _that_ uncommon.

I can think of a few ways of dealing with this:
1. Ignore it. Buyer beware.
2. Instead of `uname` check if `find /dev/null -xattrname 'com.apple.containermanager.uuid' 2> /dev/null` succeeds.
3. Instead of `command find` use `/usr/bin/find`.

2 is probably the least disruptive but is still adding complexity.

junegunn: Oh I see, `-xattrname` is not available on GNU find. It's unfortunate.

timhillgit: I went ahead and did number 2. I haven't been able to test this on fish yet but otherwise I think this should work.

junegunn: Thanks. Honestly, I don't know how I feel about this. Are these conditional branches which add complexity and degrade the readability worth it?

timhillgit: Yeah, I'm no longer sure. The fact that my most recent commit had a bug is a vote against it. I'll keep it open for a bit longer though.

timhillgit:
Closed in favor of #3649

PR #3649, opened 2024-02-29
Add walker options and replace 'find' with the built-in walker

### LINKED ISSUE #3646: Image preview not working with Wezterm's implementation of iTerm2 graphics protocol

## Info

- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [ ] bash
    - [x] zsh
    - [ ] fish

Is it possible to get Wezterm's implementation of iTerm2's image protocol working in FZF preview? iTerm2 image support integration is available, however, it seems that it does not extend to Wezterm's implementation:
```bash
 fzf --preview 'wezterm imgcat {}'
```

See https://github.com/wez/wezterm/issues/5049

junegunn:
Close #3464
Close #3647
Related #2705 

Add options for the built-in directory walker and use them to replace the default 'find' commands in the shell extensions.

| Option               | Description                                       | Default              |
| ---                  | ---                                               | ---                  |
| `--walker=OPTS`      | Walker options (`[file][,dir][,follow][,hidden]`) | `file,follow,hidden` |
| `--walker-root=DIR`  | Root directory from which to start walker         | `.`                  |
| `--walker-skip=DIRS` | Comma-separated list of directory names to skip   | `.git,node_modules`               |

> [!NOTE]
> #### Ways to implement `--walker-skip`
>
> ##### Benchmark
> image
>
> We use simple nested for loop comparison because the size of the list is unlikely to exceed 5.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,17 +1,46 @@
 CHANGELOG
 =========
 
+0.48.0
+------
+- Added options for customizing the behavior of the built-in walker
+    | Option               | Description                                       | Default              |
+    | ---                  | ---                                               | ---                  |
+    | `--walker=OPTS`      | Walker options (`[file][,dir][,follow][,hidden]`) | `file,follow,hidden` |
+    | `--walker-root=DIR`  | Root directory from which to start walker         | `.`                  |
+    | `--walker-skip=DIRS` | Comma-separated list of directory names to skip   | `.git,node_modules`  |
+    - Examples
+        ```sh
+        # Built-in walker is only used by standalone fzf when $FZF_DEFAULT_COMMAND is not set
+        unset FZF_DEFAULT_COMMAND
+
+        fzf # default: --walker=file,follow,hidden --walker-root=. --walker-skip=.git,node_modules
+        fzf --walker=file,dir,hidden,follow --walker-skip=.git,node_modules,target
+
+        # Walker options in $FZF_DEFAULT_OPTS
+        export FZF_DEFAULT_OPTS="--walker=file,dir,hidden,follow --walker-skip=.git,node_modules,target"
+        fzf
+
+        # Reading from STDIN; --walker is ignored
+        seq 100 | fzf --walker=dir
+
+        # Reading from $FZF_DEFAULT_COMMAND; --walker is ignored
+        export FZF_DEFAULT_COMMAND='seq 100'
+        fzf --walker=dir
+        ```
+- The shell extensions (key bindings and fuzzy completion) have been updated to use the built-in walker with these new options and they are now much faster out of the box.
+
 0.47.0
 ------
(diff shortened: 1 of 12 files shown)


junegunn:
I'm wondering if we should add `node_modules` to the default of `--walker-skip`. This won't cover the whole other types of projects, but still. Or should we try to implement `.gitignore` support?

alex-courtis:
Alex from [nvim-tree](https://github.com/nvim-tree/nvim-tree.lua) here, please ignore me if I'm talking out of place.

> I'm wondering if we should add `node_modules` to the default of `--walker-skip`. This won't cover the whole other types of projects and not everyone works with JavaScript projects, but still. Would anyone want to look inside `node_modules`?

`node_modules` have been particularly problematic for us and we [recommend that users ignore them](https://github.com/nvim-tree/nvim-tree.lua/wiki/Troubleshooting#ignore-large-directories) and gitignore them.

We should have ignored `node_modules` by default from the start, also `.git`.

Could I suggest that you ignore `build` as well as `target`? Meson and ninja write a _lot_ in there, mostly boilerplate and repeated files.

timhillgit:
> I'm wondering if we should add `node_modules` to the default of `--walker-skip`.

I think that's safe. Lots of node developers out there and I don't think looking in node_modules would normally be useful.

> Or should we try to implement `.gitignore` support? Is it too much?

I think it's probably too much for a default, like you I do sometimes want to search for files that are gitignored. Better to recommend using `rg` for that case.

> Could I suggest that you ignore build as well as target?

Similarly, I think this is probably too broad. Without any other context they're too common of words and might appear in other places.

junegunn:
@alex-courtis Thanks a lot for the comments.

> Could I suggest that you ignore build as well as target?

I also thought about that. `target` is particularly problematic on Maven projects. There are a bunch of `.class` files you never need to browse through. However, the directory also holds the final JAR artifacts that you do care about, so I'm a bit torn on this.

@timhillgit 

> Similarly, I think this is probably too broad. Without any other context they're too common of words and might appear in other places.

I think most of the users of fzf are software engineers, so they are unlikely to create directories named `target` or `build` for other purposes. But it's a fair point.

kelleyma49:
This is awesome, @junegunn ! Thank you for doing this. This should solve the problem that I was trying solve in #3464 . I will definitely utilize this

timhillgit:
@junegunn Release 0.47.0 mentions these changes but this PR still needs to be merged. I'm guessing you probably knew but wanted to make sure.

junegunn:
@kelleyma49 Thanks, I'm going to merge this and release a new version in the coming days. Please test this on Windows and let me know if you run into any problems.

@timhillgit Thanks, I'm aware. I was in a bit of a rush to push out some important bug fixes so I prematurely released 0.47.0 without realizing that I hadn't merged this PR. However, the release note for 0.47.0 isn't incorrect in that we have already merged fastwalk patch to the master branch. The default find command has been replaced but without the options proposed here to customize the behavior.

PR #3655, opened 2024-03-04
Refactor some code

onee-only:


onee-only:
Sorry, forgot to test it.

src/server.go
```diff
@@ -236,17 +236,15 @@ func parseGetParams(query string) getParams {
...
-				if err == nil {
-					params.limit = val
-				}
+				params.limit = val
 			case "offset":
-				val, err := strconv.Atoi(parts[1])
-				if err == nil {
-					params.offset = val
-				}
+				params.offset = val
```
junegunn, line 247: I'm not sure about this. We don't want to call `Atoi` for other unsupported parameters.

This is what I'm suggesting.
```go
switch parts[0] {
case "limit", "offset":
	if val, err := strconv.Atoi(parts[1]); err == nil {
		if parts[0] == "limit" {
			params.limit = val
		} else {
			params.offset = val
		}
	}
}
```

onee-only: This looks much better! I'll update it as you suggested. Thank you for your feedback!

junegunn:
Merged, thanks!

PR #3675, opened 2024-03-12
Embed shell integration scripts in fzf binary (`--bash` / `--zsh` / `--fish`)

junegunn:
This simplifies the distribution, and the users are less likely to have problems caused by using incompatible scripts and binaries.

* bash
  ```sh
  # Set up fzf key bindings and fuzzy completion
  eval "$(fzf --bash)"
  ```
* zsh
  ```sh
  # Set up fzf key bindings and fuzzy completion
  eval "$(fzf --zsh)"
  ```
* fish
  ```fish
  # Set up fzf key bindings
  fzf --fish | source
  ```

### DIFF
--- a/main.go
+++ b/main.go
@@ -30,16 +30,21 @@ func main() {
 	protector.Protect()
 	options := fzf.ParseOptions()
 	if options.Bash {
+		fmt.Println("### key-bindings.bash ###")
 		fmt.Println(string(bashKeyBindings))
+		fmt.Println("### completion.bash ###")
 		fmt.Println(string(bashCompletion))
 		return
 	}
 	if options.Zsh {
+		fmt.Println("### key-bindings.zsh ###")
 		fmt.Println(string(zshKeyBindings))
+		fmt.Println("### completion.zsh ###")
 		fmt.Println(string(zshCompletion))
 		return
 	}
 	if options.Fish {
+		fmt.Println("### key-bindings.fish ###")
 		fmt.Println(string(fishKeyBindings))
 		fmt.Println("fzf_key_bindings")
 		return
--- a/main.go
+++ b/main.go
@@ -1,6 +1,9 @@
 package main
 
 import (
+	_ "embed"
+	"fmt"
+
 	fzf "github.com/junegunn/fzf/src"
 	"github.com/junegunn/fzf/src/protector"
 )
@@ -8,7 +11,30 @@ import (
 var version string = "0.47"
 var revision string = "devel"
 
+//go:embed shell/key-bindings.bash
+var bashKeyBindings []byte
+
+//go:embed shell/completion.bash
+var bashCompletion []byte
+
+//go:embed shell/key-bindings.zsh
+var zshKeyBindings []byte
+
+//go:embed shell/completion.zsh
+var zshCompletion []byte
+
 func main() {
 	protector.Protect()
-	fzf.Run(fzf.ParseOptions(), version, revision)
+	options := fzf.ParseOptions()
+	if options.Bash {
+		fmt.Println(string(bashKeyBindings))
+		fmt.Println(string(bashCompletion))
+		return
+	}
+	if options.Zsh {
+		fmt.Println(string(zshKeyBindings))
+		fmt.Println(string(zshCompletion))
+		return
+	}
+	fzf.Run(options, version, revision)
 }
--- a/main.go
+++ b/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	_ "embed"
 	"fmt"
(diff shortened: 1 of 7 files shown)


akinomyoga:
In Bash 3.2, the `source` builtin doesn't support reading from a pipe, so `source <(...)` fails (silently). What is the minimal Bash version that fzf supports? As far as I see `key-bindings.bash`, fzf intends to cover Bash 3. If that is the case, `source "$(fzf --bash)"` could be used.

akinomyoga:
Is there a way to turn on either `key-bindings.bash` or `completion.bash` while the other is turned off? I'm not sure about a good interface, but for example,

```bash
fzf --bash=[key-bindings | completion | all]
```

defaulted to `all`.

junegunn:
> In Bash 3.2, the `source` builtin doesn't support reading from a pipe

Ah, that's unfortunate. I wasn't aware of it. I liked `source <(...)` better than `eval "$(...)"` because it looked better (no quotes). It's a superficial cosmetic reason that doesn't justify dropping support for bash 3 (which is the built-in version of bash in macOS), so I'm going to suggest `eval` instead.

> Is there a way to turn on either key-bindings.bash or completion.bash while the other is turned off? 

I thought about it, but I decided to keep it simple. We could add a magic comment at the top of each section so that advanced users can `sed` away the unwanted part.

akinomyoga:
Thanks for your reply!

main.go
```diff
@@ -1,14 +1,53 @@
...
+
 func main() {
 	protector.Protect()
-	fzf.Run(fzf.ParseOptions(), version, revision)
+	options := fzf.ParseOptions()
+	if options.Bash {
+		fmt.Println("### key-bindings.bash ###")
+		fmt.Println(string(bashKeyBindings))
+		fmt.Println("### completion.bash ###")
+		fmt.Println(string(bashCompletion))
```
akinomyoga, line 36: I think the suggested `'/### completion/,$p'` and `'/### completion/,$d'` are not forward compatible. These sed formulae can be broken when the order of the section is swapped or when another new section is added. I think it would be more robust to have terminators like

```suggestion
		fmt.Println("### end ###")
		fmt.Println("### completion.bash ###")
		fmt.Println(string(bashCompletion))
		fmt.Println("### end ###")
```

so that users can specify
- `sed -n '/^### completion/,/^### end/p'`
- or `sed -n '/^### key-bindings/,/^### end/p'`
- or `sed -n '/^### key-bindings/,/^### end/p;/^### completion/,/^### end/p'`.

junegunn: Good point!

junegunn: 2b26315cf4342c3578fe13da52ea93dbe0b61212

akinomyoga:
@junegunn I suggested evaluating the integration shell scripts by `eval` (for the Bash compatibility), but I noticed there is a problem with the current content of the integration scripts. `[[ $- =~ i ]] || return 0` at the beginning of the integration scripts, when executed in non-interactive shells, terminates the execution of `.bashrc`, `.zshrc`, etc. *in the middle* because `eval` runs the commands in the context where `eval` is called.

One solution is to avoid using `return 0` and enclose the entire content of the integration scripts with a big `if` statement as `if [[ $- =~ i ]]; then <all the script contents>; fi`. Another solution is to enclose the script contents within a temporary function when embedding them in stdout. For example, this function

https://github.com/junegunn/fzf/blob/4cd37fc02b4491d4c9a5f6a27ad41895ad1a62cc/main.go#L30-L34

can be modified to additionally print

- `__fzf_initialize() {` and `unset -f __fzf_initialize }; __fzf_initialize` for Bash,
- print `() {` and `}` for Zsh
- and something equivalent for Fish.

junegunn:
@akinomyoga Thanks, I'll update the script files. We used to wrap the whole thing before https://github.com/junegunn/fzf/commit/f103aa4753b435f8f45c5130323effeb75583c15.

akinomyoga:
Thank you for fixing it! I think the fish integration script also needs to be adjusted.

https://github.com/junegunn/fzf/blob/4cd37fc02b4491d4c9a5f6a27ad41895ad1a62cc/shell/key-bindings.fish#L14

junegunn:
We instruct fish users to use `source` via pipe.

https://github.com/junegunn/fzf?tab=readme-ov-file#setting-up-shell-integration

And as far as I can tell, it doesn't have the same issue.

akinomyoga:
Ah, OK! I missed it. Sorry for the noise, and thank you for the clarification. I think now everything should be fine.

Thank you for your quick responses and actions.

PR #3678, opened 2024-03-15
Make it possible to disable `Ctrl+T` / `Alt+C` / completions

elibarzilay:
This makes it possible to skip one of the above key bindings or completions by setting a variable to an empty string. For example,

    FZF_CTRL_T_COMMAND= FZF_ALT_C_COMMAND= FZF_COMPLETION_TRIGGER= \
      eval "$(fzf --zsh)"

would leave only the history search. Makes it more convenient then producing the files and changing them myself to not include what I don't want.

This should be working, but for completeness it should also include at least some explaining comment at the top (since there's no docs about these vars?) and something equivalent for bash/fish.

### DIFF
--- a/shell/completion.zsh
+++ b/shell/completion.zsh
@@ -10,7 +10,7 @@
 # - $FZF_COMPLETION_OPTS    (default: empty)
 
 [[ -o interactive ]] || return 0
-
+[[ "${FZF_COMPLETION_TRIGGER-x}" != "" ]] || return 0
 
 # Both branches of the following `if` do the same thing -- define
 # __fzf_completion_options such that `eval $__fzf_completion_options` sets
--- a/shell/completion.zsh
+++ b/shell/completion.zsh
@@ -10,7 +10,6 @@
 # - $FZF_COMPLETION_OPTS    (default: empty)
 
 [[ -o interactive ]] || return 0
-[[ "${FZF_COMPLETION_TRIGGER-x}" != "" ]] || return 0
 
 # Both branches of the following `if` do the same thing -- define
 # __fzf_completion_options such that `eval $__fzf_completion_options` sets
--- a/shell/completion.zsh
+++ b/shell/completion.zsh
@@ -11,6 +11,7 @@
 
 [[ -o interactive ]] || return 0
 
+
 # Both branches of the following `if` do the same thing -- define
 # __fzf_completion_options such that `eval $__fzf_completion_options` sets
 # all options to the same values they currently have. We'll do just that at
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,22 @@
 CHANGELOG
 =========
 
+0.48.1
+------
+- CTRL-T and ALT-C bindings can be disabled by setting `FZF_CTRL_T_COMMAND` and `FZF_ALT_C_COMMAND` to empty strings respectively when sourcing the script
+    ```sh
+    # bash
+    FZF_CTRL_T_COMMAND= FZF_ALT_C_COMMAND= eval "$(fzf --bash)"
+
+    # zsh
+    FZF_CTRL_T_COMMAND= FZF_ALT_C_COMMAND= eval "$(fzf --zsh)"
+
+    # fish
+    fzf --fish | FZF_CTRL_T_COMMAND= FZF_ALT_C_COMMAND= source
+    ```
+    - Setting the variables after sourcing the script will have no effect
+- Bug fixes
+
 0.48.0
 ------
 - Shell integration scripts are now embedded in the fzf binary. This simplifies the distribution, and the users are less likely to have problems caused by using incompatible scripts and binaries.
--- a/README.md
+++ b/README.md
@@ -437,6 +437,8 @@ fish.
         --preview 'bat -n --color=always {}'
         --bind 'ctrl-/:change-preview-window(down|hidden|)'"
       ```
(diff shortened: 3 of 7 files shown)


shell/completion.zsh
```diff
@@ -10,7 +10,7 @@
 # - $FZF_COMPLETION_OPTS    (default: empty)
 
 [[ -o interactive ]] || return 0
-
+[[ "${FZF_COMPLETION_TRIGGER-x}" != "" ]] || return 0
```
junegunn, line 13: Shouldn't do this because an empty trigger is allowed in zsh.

Related: https://github.com/junegunn/fzf/wiki/Configuring-fuzzy-completion#dedicated-completion-key

```suggestion
```

shell/completion.zsh
```diff
@@ -11,7 +11,6 @@
 
 [[ -o interactive ]] || return 0
```
junegunn, line 13:
```suggestion


```

junegunn:
> something equivalent for bash/fish

Can you update those files as well?

shell/completion.zsh
elibarzilay: Well, I wasn't too happy with it anyway, since it looks a bit random to just hook on one variable per disable-able feature.

But if you think that this whole thing is a good addition, then maybe:
* Use a random non-empty string value, something like `DISABLE` etc.
* Replace the whole thing with a single `$DISABLE` that could be set to a value like `ctrl_t:alt_r:completion`.
* Or anything similar.

I'll be happy with whatever allows me to disable these without copying the file.

junegunn:
> Well, I wasn't too happy with it anyway, since it looks a bit random to just hook on one variable per disable-able feature.

Yeah, I kind of felt the same. But the idea of disabling a key binding if its COMMAND variable is intentionally set to an empty string doesn't seem too far-fetched and it's the simplest approach at hand. So I'm going to accept it.

The limitation of this approach though is that it doesn't address the need of some users to bind the feature to a different key. But it's not easy to come up with a solution that works consistently with all 3 shells because they all use different key expressions.

I'll take over this PR from here and finish it up (bash, fish, and documentation). Thanks.

junegunn:
Merged, thanks!

elibarzilay:
@junegunn Huge thanks for taking it over!

Some replies that are the least I owe you:
* Might be nice to replace the `*COMMAND` by some `*KEY` so you can both disable it or change the key that is used.
* Yes, I'm using `--zsh` to get both files.  The `sed` thing is kind of lame as a way to disable loading one file, but yeah, it's hard to find some elegant variable.
* Sidenote, for the purists that dislike jumping onto sed for whatever, you can use the shell instead.  It's also possible to avoid leaving the code in a temporary variable like this:
  ```
  eval "$(x=$(fzf --zsh); echo "${x//"### completion"*"completion.zsh ###"}")"
  ```
  This should work at least for zsh / bash.

junegunn:
So what was your motivation behind this? Which feature did you want to opt-out? I decided to make `--*sh` print both script files because I figured most of the users would want both.

> * Might be nice to replace the `*COMMAND` by some `*KEY` so you can both disable it or change the key that is used.

That would be ideal, but I wanted the variables to work consistently across three different shells to minimize the confusion, but it's not hard to achieve that with the approach because all three shells use different key expressions. `^T`, `\C-t`, `\ct`.

bluz71:
> I decided to make `--*sh` print both script files because I figured most of the users would want both.

Maybe I am not "_most users_" but I certainly wish fine grain control over the scripts. We are not quite there yet (from what I can see).

In the old mechanism there existed separate `completion.bash` and `keybindings.bash` scripts, and I only ever sourced the later (`keybindings`) and never the former (`completion`). Completion slowed my SHELL startup too much (about 20ms) whilst keybindings only added a couple ms. I don't need `**` support.

In the new mechanism `eval "$(fzf --bash)"` I can not figure out how to **not** load completion.

Ideally I would like something like:

```sh
FZF_COMPLETION= eval "$(fzf --bash)"
```

I do want `Ctrl-r`, `Ctrl-t`, `Alt-c` but I do not want fuzzy `**` completion.

bluz71:
I just saw `FZF_COMPLETION_TRIGGER= eval "$(fzf --bash)"`, my bad. However, this has poor startup performance:

``` 
% time FZF_COMPLETION_TRIGGER= eval "$(fzf --bash)"
real	0m0.029s
```

Compared with old-school `. $HOMEBREW_PREFIX/opt/fzf/shell/key-bindings.bash`:

```sh
time . $HOMEBREW_PREFIX/opt/fzf/shell/key-bindings.bash
real	0m0.004s
```

29ms vs 4ms.

It would be nice to have true opt-out of Bash fzf `**` completion.

junegunn:
@bluz71 

> I can not figure out how to **not** load completion

```sh
eval "$(fzf --bash | sed '/### completion/,$d')"
```

> 29ms vs 4ms.

You don't have to use `--bash`, you can still use the old method if you prefer it. The files are still included in the package.

bluz71:
Thanks @junegunn, `eval "$(fzf --bash | sed '/### completion/,$d')"` works.

I assume that the the scripts on-disk will eventually go away?

I am satisfied.

bluz71:
Oh, the scripts will always be available, if so, then I will stick with those.

All good, sorry for the diversion.

junegunn:
No problem, thanks for sharing your case.

PR #3684, opened 2024-03-19
Fix Ctrl-T and Alt-C in fish no longer using last token as search root

zeertzjq:
This fixes a regression from #3649.


junegunn:
Merged, thanks!

PR #3691, opened 2024-03-24
Let toggle-track disable track action

druckdev:
Currently the `toggle-track` action does not toggle off the tracking of the current selection started by the `track` action. The only way of disabling it is by changing the focus.

Change this by treating `trackCurrent` like `trackEnabled`.

### DIFF
--- a/src/terminal.go
+++ b/src/terminal.go
@@ -3772,6 +3772,8 @@ func (t *Terminal) Loop() {
 				req(reqPrompt)
 			case actToggleTrack:
 				switch t.track {
+				case trackCurrent:
+					fallthrough
 				case trackEnabled:
 					t.track = trackDisabled
 				case trackDisabled:


druckdev:
What I noticed too: The `track` action currently only switches over to `trackCurrent` if tracking is disabled at the moment. Wouldn't it make sense to do it always (i.e. also on `trackEnabled`) since the description of `track` states that the tracking will be disabled on focus change?

I do have to say that the differentiation between `Enabled` and `Current` was not very clear to me as a user before reading the source code. Maybe a name like `track-current` would suite better for the `track` action to emphasize the difference between the way it works and the tracking from `--track` and `toggle-track`?

Just my 2cts on this topic that might be slightly related to this PR.

junegunn:
I understand. There can be confusion between the global tracking state set by `--track` / `toggle-track` and a one-off `track` action.

The suggested change will make `toggle-track` affect the latter, but I'm concerned that it might cause more confusion. A user might accidentally turn on the global tracking state on the second invocation of `toggle-track`. So here's what I'm suggesting in addition to your proposal.

* Rename `track` to `track-current`, but keep the old name for backward compatibility
* Add `untrack-current` action that only affects the local tracking state
* Leave `toggle-track` unchanged. It should only affect the global tracking state

druckdev:
I see your point, yes those changes including the rename would probably make it clearer.

What about the behavior of FZF when `track-current` is executed while the global tracking is enabled? Should it stay like now that it does nothing?

junegunn:
I think `track-current` (a.k.a. `track`) and `untrack-current` should both be no-op when global tracking is enabled.

Anyway, what was your primary motivation behind this? I guess you're not using `--track` but `track` action and wanted a way to "untrack" it?

druckdev:
Makes sense.

I wanted to have automatic tracking that turns on by pressing backspace and turns off again on `backward-eof`.

Your new PR looks great, with it it should be achievable with:

```
--bind bspace:track-current+backward-delete-char,backward-eof:untrack-current
```

or something similar.

Also nice that you also included `toggle-track-current`.
Thanks!

PR #3687, opened 2024-03-20
chore: transition from markdown to YAML for issue template

LangLangBart:
### description
Usage of an issue template form written in YAML.

#### refs
[Syntax for issue forms - GitHub Docs](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms)

#### demo
https://github.com/LangLangBart/fzf/issues/new/choose

---

Pros
- Validates certain fields, such as `fzf version`, before filing a report
- Ensures consistency, as every report follows the same format
- …

Cons
- Imposes more restrictions on the user to adhere to a specific format
- …

PR #3715, opened 2024-04-06
chore: Add new option in issue checklist and modify requirements

LangLangBart:
### description

- PR #3687 removed the checkbox that asked the user to verify they have the latest version of
  fzf. This was due to the introduction of the new `input` field for the 'Output of "fzf
  --version"', which made the checkbox seem redundant.
```txt
- [ ] I have the latest version of fzf
```

- However, there were minor bug reports where simply updating to a later `fzf` version resolved the
  issue. Therefore, it might be beneficial to reintroduce this checkbox, but set the validation to
  `false`.

- We could also consider setting the validation for reading the `man` page to `false`. This would
  prevent users from being forced to tick the box to open a report, even if they haven't done what
  the box suggests. Currently, the `man` page is approximately 20-30 PDF pages long, so it's
  understandable if not everything has been read.

```bash
# open 'fzf' man page as a PDF

# macos Ventura+
mandoc -T pdf "$(man -w fzf)" | open -fa Preview
# prior to Ventura
man -t fzf | open -fa "Preview"
```

### DIFF
--- a/.github/ISSUE_TEMPLATE/issue_template.yml
+++ b/.github/ISSUE_TEMPLATE/issue_template.yml
@@ -15,6 +15,8 @@ body:
           required: true
         - label: I have searched through the existing issues
           required: true
+        - label: For bug reports, I have checked if the bug is reproducible in the latest version of fzf
+          required: false
 
   - type: input
     attributes:


junegunn:
Thanks a lot.

> * Currently, the `man` page is approximately 20-30 PDF pages long

The point is to prevent users from 1. opening feature requests for features that already exist, 2. asking questions that can easily be answered just by looking at the man page or searching for related issues. We don't owe users anything and our time is precious. By opening an issue, they are demading our limited time and attention for free, so they should at least do some research.

But yeah, even so we're probably asking for too much (especially for obvious bug reports). Can there be a happy middle ground?

junegunn:
Anyway, merging this, thanks!

LangLangBart:
> By opening an issue, they are demading our limited time and attention for free, so they should at least do some research on their own before doing so.


Absolutely. 

Humor can help nudge someone into doing the right thing.

It turns out that images can be used in an issue template. 😀

PR #3722, opened 2024-04-10
Add arguments to disable loading keybindings or completion

SuperSandro2000:
NixOS switched (https://github.com/NixOS/nixpkgs/pull/298692) to the new way of loading the completion files which was added in #3675 but that removes support for not loading the completion, so I just added options for that.

### DIFF
--- a/main.go
+++ b/main.go
@@ -37,18 +37,28 @@ func main() {
 	protector.Protect()
 	options := fzf.ParseOptions()
 	if options.Bash {
-		printScript("key-bindings.bash", bashKeyBindings)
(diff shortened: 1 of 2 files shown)


junegunn:
Thanks, but I'm going to keep it simple and not add more options. As noted in the instruction, you can still load individual script files if you need finer control.

https://github.com/junegunn/fzf?tab=readme-ov-file#setting-up-shell-integration

> If you have an older version of fzf, or want finer control, you can source individual script files in the [/shell](https://github.com/junegunn/fzf/blob/master/shell) directory.

PR #3746, opened 2024-04-21
Add --with-shell for shelling out with different command and flags

### LINKED ISSUE #3732: [Feature Request] Customizable shell command flag for command execution
### Checklist

### Output of `fzf --version`

0.50.0 (brew)

### OS

- [ ] Linux
- [X] macOS
- [ ] Windows
- [ ] Etc.

### Shell

- [ ] bash
- [X] zsh
- [ ] fish

This report was created from a discussion on a discarded PR: #3726

---

### Issue

`fzf` currently executes commands with `$SHELL` and a hardcoded `-c` flag.

https://github.com/junegunn/fzf/blob/f97d2754134607b24849fc4a2062dbfcaafddd6a/src/util/util_unix.go#L22-L29

If a user wants to use a different tool with a different flag to execute commands, they must prepend the desired command to each string, making proper quoting more cumbersome and leading to repetitive addition of the command.

```bash
# [bun] https://github.com/oven-sh/bun
fzf --bind "start:reload:bun --print \"Object.keys(Bun.env).sort().join('\n')\""
```

---

### Feature Proposal

Ability to change the flag that is passed to the `$SHELL` command.

If the flag would be subject to change, then one would only need to write the string.

```bash
# Syntax example, not actual usage
SHELL=$(which bun) FZF_SHELL_FLAG="--print" \
  fzf --bind "start:reload:Object.keys(Bun.env).sort().join('\n')"

# Different syntax example proposed by the maintainer
fzf --with-shell "bun --print" \
  --bind "start:reload:Object.keys(Bun.env).sort().join('\n')"
```

Here are a few examples of tools with their corresponding flags to illustrate the variety of flags that might be used:
- `jsc -e`: a short, single-letter option [^1]
- `bun --print`: a long, word-based option [^2]
- `osascript -l JavaScript -e`: multiple short, single-letter options and an argument [^3]

---

### Workaround
The maintainer provided a workaround [^4].

```bash
fzf2() (
  export SHELL=/tmp/fzfsh
  if ! [[ -x $SHELL ]]; then
    echo '#!/bin/sh' > "$SHELL"
    echo 'shift; echo "$@"' >> "$SHELL"
    chmod +x "$SHELL"
  fi
  command fzf "$@"
)

fzf2 --preview 'run this command: {}'
```

[^1]: [JSC – WebKit](https://trac.webkit.org/wiki/JSC)
[^2]: [Bun — A fast all-in-one JavaScript runtime](https://bun.sh/)
[^3]: [JXA - JavaScript for Automation](https://developer.apple.com/library/archive/releasenotes/InterapplicationCommunication/RN-JavaScriptForAutomation/Articles/Introduction.html)
[^4]: [PR #3726 (comment)](https://github.com/junegunn/fzf/pull/3726#issuecomment-2054007955)

junegunn:
Close #3732

This hasn't been tested on Windows.

### DIFF
--- a/src/util/util_unix.go
+++ b/src/util/util_unix.go
@@ -5,27 +5,55 @@ package util
 import (
 	"os"
 	"os/exec"
+	"strings"
 	"syscall"
 
 	"golang.org/x/sys/unix"
 )
 
-// ExecCommand executes the given command with $SHELL
-func ExecCommand(command string, setpgid bool) *exec.Cmd {
-	shell := os.Getenv("SHELL")
-	if len(shell) == 0 {
-		shell = "sh"
-	}
-	return ExecCommandWith(shell, command, setpgid)
+type Executor struct {
+	shell   string
+	args    []string
+	escaper *strings.Replacer
 }
 
-// ExecCommandWith executes the given command with the specified shell
-func ExecCommandWith(shell string, command string, setpgid bool) *exec.Cmd {
-	cmd := exec.Command(shell, "-c", command)
+func NewExecutor(withShell string) *Executor {
+	shell := os.Getenv("SHELL")
+	args := strings.Fields(withShell)
+	if len(args) > 0 {
+		shell = args[0]
+		args = args[1:]
+	} else {
+		if len(shell) == 0 {
+			shell = "sh"
+		}
+		args = []string{"-c"}
+	}
+
+	var escaper *strings.Replacer
+	tokens := strings.Split(shell, "/")
+	if tokens[len(tokens)-1] == "fish" {
+		// https://fishshell.com/docs/current/language.html#quotes
+		// > The only meaningful escape sequences in single quotes are \', which
+		// > escapes a single quote and \\, which escapes the backslash symbol.
+		escaper = strings.NewReplacer("\\", "\\\\", "'", "\\'")
+	} else {
+		escaper = strings.NewReplacer("'", "'\\''")
+	}
+	return &Executor{shell, args, escaper}
+}
+
+// ExecCommand executes the given command with $SHELL
+func (x *Executor) ExecCommand(command string, setpgid bool) *exec.Cmd {
+	cmd := exec.Command(x.shell, append(x.args, command)...)
 	if setpgid {
 		cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}
 	}
 	return cmd
+}
+
+func (x *Executor) QuoteEntry(entry string) string {
+	return "'" + x.escaper.Replace(entry) + "'"
 }
 
 // KillCommand kills the process for the given command
--- a/src/util/util_unix.go
+++ b/src/util/util_unix.go
@@ -59,8 +59,8 @@ func (x *Executor) QuoteEntry(entry string) string {
 func (x *Executor) Become(stdin *os.File, environ []string, command string) {
 	SetStdin(stdin)
 	shellPath, err := exec.LookPath(x.shell)
-	if err == nil {
-		shellPath = x.shell
+	if err != nil {
+		Exit(127)
 	}
 	args := append([]string{shellPath}, append(x.args, command)...)
 	syscall.Exec(shellPath, args, environ)
--- a/src/util/util_unix.go
+++ b/src/util/util_unix.go
@@ -3,6 +3,7 @@
 package util
 
 import (
+	"fmt"
 	"os"
 	"os/exec"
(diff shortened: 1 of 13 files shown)


LangLangBart:
Thanks for dedicating time to implementing it.

Was it a deliberate decision that you made while working on it to not also apply it to the `become` action?

https://github.com/junegunn/fzf/blob/056cb5e52c7753d303f45dc8cf15777c7d0285ff/src/terminal.go#L3396-L3398

```bash
: | fzf --with-shell 'ruby -e' --bind 'enter:become:puts "Hello world"'
```

---

The example in the man page works well, but it may cause some issues for people with a defined `FZF_DEFAULT_COMMAND`. Consider this alternative:

```bash
tail /usr/share/dict/words | fzf --with-shell 'ruby -e' --preview 'puts {}.upcase.reverse'
```

junegunn:
> Was it a deliberate decision that you made while working on it to not also apply it to the `become` action?
> 
> [fzf](https://github.com/junegunn/fzf/blob/056cb5e52c7753d303f45dc8cf15777c7d0285ff/src/terminal.go#L3396-L3398)

No, it was an oversight.

> The example in the man page works well, but it may cause some issues for people with a defined `FZF_DEFAULT_COMMAND`.

Good point, thanks!

junegunn:
@rashil2000 Hi, can you take a look at this patch when you have some time? I could use some help on the Windows part, and you're the one who last changed the relevant code in #2647.

src/util/util_unix.go
```diff
@@ -5,29 +5,67 @@ package util
...
 	return cmd
 }
 
+func (x *Executor) QuoteEntry(entry string) string {
+	return "'" + x.escaper.Replace(entry) + "'"
+}
+
+func (x *Executor) Become(stdin *os.File, environ []string, command string) {
+	SetStdin(stdin)
+	shellPath, err := exec.LookPath(x.shell)
```
LangLangBart, line 61: This works as long as the user provides the full path to their binary.

```bash
fzf --with-shell '/usr/bin/ruby -e' --bind 'enter:become:puts "Hello world"'
```
However, if the user only provides the name, then the execution stalls for the `become` part.

```bash
fzf --with-shell 'ruby -e' --bind 'enter:become:puts "Hello world"'
```

How about this ?

```suggestion
	if err != nil {
		return
	}
```

LangLangBart: or better use `os.Exit(...)` ?

```bash
fzf --with-shell 'idontexist -e' --bind 'enter:become:puts "Hello world"'
```

junegunn: Thanks, it was an apparent bug. I'll make it exit with 127.

```sh
$ sh -c 'nope'
sh: nope: command not found
$ echo $?
127
```

junegunn: Maybe we should print such an error message as well?

junegunn:
```
$ fzf --with-shell 'ruby2 -e' --bind 'start:become:puts "Hello world"'
fzf (become): exec: "ruby2": executable file not found in $PATH
```

LangLangBart: Thanks, the error message is useful.

In both `zsh`[^1] and `bash`[^2], there is a distinction between exit code `127` (command not found) and `126` (command found but not executable). It is unclear if you wish to apply this distinction here.

```bash
sh -c "$HOME"
# sh: /Users/paria: is a directory
echo $?
# 126
```

If not, simply returning error code `2` would suffice without the need to modify the man page for the exit status. 

https://github.com/junegunn/fzf/blob/bacb7d3ba301094bd7e108997193860712ec5f32/man/man1/fzf.1#L938-L945

[^1]: [zsh: 8 Command Execution](https://zsh.sourceforge.io/Doc/Release/Command-Execution.html#Command-Execution)
[^2]: [Bash Reference Manual - Exit Status](https://www.gnu.org/software/bash/manual/bash.html#Exit-Status)

rashil2000:
> @rashil2000 Hi, can you take a look at this patch when you have some time? I could use some help on the Windows part, and you're the one who last changed the relevant code in #2647.

Sure, I'm currently travelling, will have a look in a few days

junegunn:
So I did some preliminary testing of this on a Windows VM against `cmd` and `powershell`, and at least it doesn't seem to break the existing workflow.

junegunn:
`cmd.exe` execution is broken, restoring `CmdLine` parameter.

Issue #3748, opened 2024-04-21
TestOSExitNotAllowed fails when `go test` is passed `-trimpath`


Foxboron:
### Checklist

- [X] I have read through the manual page (`man fzf`)
- [X] I have searched through the existing issues
- [X] For bug reports, I have checked if the bug is reproducible in the latest version of fzf

### Output of `fzf --version`

N/A

### OS

- [X] Linux
- [ ] macOS
- [ ] Windows
- [ ] Etc.

### Shell

- [X] bash
- [ ] zsh
- [ ] fish

### Problem / Steps to reproduce

```
λ fzf master Ɇ » go test ./...
?   	github.com/junegunn/fzf/src/protector	[no test files]
ok  	github.com/junegunn/fzf	1.449s
ok  	github.com/junegunn/fzf/src	(cached)
ok  	github.com/junegunn/fzf/src/algo	(cached)
ok  	github.com/junegunn/fzf/src/tui	(cached)
ok  	github.com/junegunn/fzf/src/util	(cached)
λ fzf master Ɇ » go test -trimpath ./...
?   	github.com/junegunn/fzf/src/protector	[no test files]
--- FAIL: TestOSExitNotAllowed (0.02s)
    --- FAIL: TestOSExitNotAllowed/github.com/junegunn/fzf (0.00s)
        main_test.go:91: /home/fox/Git/prosjekter/Go/fzf/main.go:4:4: could not import embed (cannot find package "embed" in any of:
            	($GOROOT not set)
            	/home/fox/.go/src/embed (from $GOPATH))
    --- FAIL: TestOSExitNotAllowed/github.com/junegunn/fzf/src (0.01s)
        main_test.go:91: /home/fox/Git/prosjekter/Go/fzf/src/actiontype_string.go:5:8: could not import strconv (cannot find package "strconv" in any of:
            	($GOROOT not set)
            	/home/fox/.go/src/strconv (from $GOPATH))
    --- FAIL: TestOSExitNotAllowed/github.com/junegunn/fzf/src/algo (0.00s)
        main_test.go:91: /home/fox/Git/prosjekter/Go/fzf/src/algo/algo.go:81:2: could not import bytes (cannot find package "bytes" in any of:
            	($GOROOT not set)
            	/home/fox/.go/src/bytes (from $GOPATH))
    --- FAIL: TestOSExitNotAllowed/github.com/junegunn/fzf/src/tui (0.00s)
[...]

PR #3758, opened 2024-04-30
test: fix TestOSExitNotAllowed to handle empty GOROOT

charlievieth:
When tests are ran the `-trimpath` flag `runtime.GOROOT()` returns an empty string. This causes [`TestOSExitNotAllowed`](https://github.com/charlievieth/fzf/blob/2dc3fc40fc9236b249085a550ce9c722e5dcfc21/main_test.go#L141-L174) to fail because [`build.Context.ImportDir()`](https://pkg.go.dev/go/build#Context.ImportDir) cannot find packages in the standard library (without GOROOT it does not know where to look - see issue https://github.com/junegunn/fzf/issues/3748). This commit fixes that by falling back to `go env GOROOT` if `runtime.GOROOT()` returns an empty string. This works because `go env GOROOT` performs more work to identify the GOROOT than `runtime.GOROOT()`.

This should fix https://github.com/junegunn/fzf/issues/3748.

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bytes"
 	"fmt"
 	"go/ast"
 	"go/build"
@@ -10,6 +11,7 @@ import (
 	"go/types"
 	"io/fs"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"sort"
 	"strings"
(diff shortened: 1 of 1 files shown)


Foxboron:
Thanks for working on this :)

junegunn:
Merged, thanks!

PR #3766, opened 2024-05-03
feature: header-click event for reporting clicks within header

kuremu:
Sets `FZF_HEADERCLICK_ROW` and `FZF_HEADERCLICK_LINE` env vars with coordinates of last click inside and relative to the header.

`printf "header1\nheader2\nheader3\na\nb\nc" | fzf --header-lines=3 --bind "header-click:transform-prompt(echo \${FZF_HEADERCLICK_LINE}x\${FZF_HEADERCLICK_COLUMN})"`

Could be useful for eg. selecting a sort column. I think it could just as easily add a `FZF_HEADERCLICK_WORD` var as well.

[Here is a more complex demo with an interactive keyboard](https://gist.github.com/kuremu/4d1441828ebab9d2da67f1298102fc3f) (invoke like `cat /usr/share/dict/words | python keyboard.py`).

### DIFF
--- a/src/options.go
+++ b/src/options.go
@@ -708,6 +708,8 @@ func parseKeyChordsImpl(str string, message string, exit func(string)) map[tui.E
 			add(tui.Jump)
 		case "jump-cancel":
 			add(tui.JumpCancel)
+		case "header-click":
+			add(tui.HeaderClick)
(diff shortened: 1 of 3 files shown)


kuremu:
sleep-deprived typo in the commit, i'm going to submit it again

PR #3768, opened 2024-05-04
 feature: header-click event for reporting clicks within header

kuremu:
Sets `FZF_HEADERCLICK_LINE` and `FZF_HEADERCLICK_COLUMN` env vars with coordinates of last click inside and relative to the header and fires header-click event.

`printf "header1\nheader2\nheader3\na\nb\nc" | fzf --header-lines=3 --bind "header-click:transform-prompt(echo \${FZF_HEADERCLICK_LINE}x\${FZF_HEADERCLICK_COLUMN})"`

Could be useful for eg. selecting a sort column. Could set a `FZF_HEADERCLICK_WORD` var as well if that is more useful to people.

[Here is a more complex demo with an interactive keyboard](https://gist.github.com/kuremu/4d1441828ebab9d2da67f1298102fc3f) (invoke like `cat /usr/share/dict/words | python keyboard.py`).

### DIFF
--- a/man/man1/fzf.1
+++ b/man/man1/fzf.1
@@ -1278,12 +1278,12 @@ e.g.
      \fBfzf --bind space:jump,jump:accept,jump-cancel:abort\fR
 .RE
 
-\fIheader-click\fR
+\fIclick-header\fR
 .RS
-Triggered when a mouse click occurs within the header. Sets \fBFZF_HEADERCLICK_LINE\fR and \fBFZF_HEADERCLICK_COLUMN\fR environment variables.
+Triggered when a mouse click occurs within the header. Sets \fBFZF_CLICK_HEADER_LINE\fR and \fBFZF_CLICK_HEADER_COLUMN\fR environment variables.
 
 e.g.
-     \fBprintf "head1\\nhead2" | fzf --header-lines=2 --bind header-click:transform-prompt:"printf \\${FZF_HEADERCLICK_LINE}x\\${FZF_HEADERCLICK_COLUMN}"\fR
+     \fBprintf "head1\\nhead2" | fzf --header-lines=2 --bind click-header:transform-prompt:"printf \\${FZF_CLICK_HEADER_LINE}x\\${FZF_CLICK_HEADER_COLUMN}"\fR
 
 .RE
 
--- a/man/man1/fzf.1
+++ b/man/man1/fzf.1
@@ -1283,7 +1283,7 @@ e.g.
 Triggered when a mouse click occurs within the header. Sets \fBFZF_CLICK_HEADER_LINE\fR and \fBFZF_CLICK_HEADER_COLUMN\fR environment variables.
 
 e.g.
-     \fBprintf "head1\\nhead2" | fzf --header-lines=2 --bind click-header:transform-prompt:"printf \\${FZF_CLICK_HEADER_LINE}x\\${FZF_CLICK_HEADER_COLUMN}"\fR
+     \fBprintf "head1\\nhead2" | fzf --header-lines=2 --bind 'click-header:transform-prompt:printf ${FZF_CLICK_HEADER_LINE}x${FZF_CLICK_HEADER_COLUMN}'\fR
 
 .RE
 
--- a/man/man1/fzf.1
+++ b/man/man1/fzf.1
@@ -1278,6 +1278,15 @@ e.g.
      \fBfzf --bind space:jump,jump:accept,jump-cancel:abort\fR
 .RE
 
+\fIheader-click\fR
+.RS
+Triggered when a mouse click occurs within the header. Sets \fBFZF_HEADERCLICK_LINE\fR and \fBFZF_HEADERCLICK_COLUMN\fR environment variables.
+
+e.g.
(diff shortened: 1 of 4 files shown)


junegunn (commented):
Please update the man page as well.

man/man1/fzf.1
```diff
@@ -1278,6 +1278,15 @@ e.g.
      \fBfzf --bind space:jump,jump:accept,jump-cancel:abort\fR
 .RE
 
+\fIheader-click\fR
```
junegunn, line 1281: I would name it `click-header` because we may add more `click-xxx` events in the future and it would be nice for them to share the same `click-` prefix.

man/man1/fzf.1
```diff
@@ -1278,6 +1278,15 @@ e.g.
...
+.RS
+Triggered when a mouse click occurs within the header. Sets \fBFZF_HEADERCLICK_LINE\fR and \fBFZF_HEADERCLICK_COLUMN\fR environment variables.
```
junegunn, line 1283: `FZF_CLICK_HEADER_..` to respect the hyphen.

src/terminal.go
```diff
@@ -857,6 +859,8 @@ func (t *Terminal) environ() []string {
 	env = append(env, fmt.Sprintf("FZF_LINES=%d", t.areaLines))
 	env = append(env, fmt.Sprintf("FZF_COLUMNS=%d", t.areaColumns))
 	env = append(env, fmt.Sprintf("FZF_POS=%d", util.Min(t.merger.Length(), t.cy+1)))
+	env = append(env, fmt.Sprintf("FZF_HEADERCLICK_LINE=%d", t.headerClickLine))
+	env = append(env, fmt.Sprintf("FZF_HEADERCLICK_COLUMN=%d", t.headerClickColumn))
```
junegunn, line 863: I'd like to avoid exporting these on other events. To do so, we can reset the variables to zero after running the bound actions, and export the variables when they are greater than zero,

src/terminal.go
```diff
@@ -4009,6 +4013,35 @@ func (t *Terminal) Loop() {
...
+							}
+							if numItems < 0 {
+								numLines += numItems
+							}
+						}
+						my = util.Constrain(my-lineOffset, -1, numLines)
+						mx -= 2 // offset gutter
+						if my >= 0 && my < numLines && mx >= 0 {
+							t.headerClickLine = my
+							t.headerClickColumn = mx
```
junegunn, line 4041: For "LINE" and "COLUMN", it is more natural to use numbers starting from 1.

src/terminal.go
kuremu: Would this be a suitable way to handle the reset?

```
if my >= 0 && my < numLines && mx >= 0 {
    t.clickHeaderLine = my + 1
    t.clickHeaderColumn = mx + 1
    evt := tui.ClickHeader
    res := doActions(actionsFor(evt))
    t.clickHeaderLine = 0
    t.clickHeaderColumn = 0
    return res
}
```

junegunn: Yes, and,

```go
if t.headerClickLine > 0 {
  env = append(...)
  env = append(...)
}

File: man/man1/fzf.1 [hunk 5]
```diff
@@ -1278,6 +1278,15 @@ e.g.
      \fBfzf --bind space:jump,jump:accept,jump-cancel:abort\fR
 .RE
 
+\fIclick-header\fR
+.RS
+Triggered when a mouse click occurs within the header. Sets \fBFZF_CLICK_HEADER_LINE\fR and \fBFZF_CLICK_HEADER_COLUMN\fR environment variables.
+
+e.g.
+     \fBprintf "head1\\nhead2" | fzf --header-lines=2 --bind click-header:transform-prompt:"printf \\${FZF_CLICK_HEADER_LINE}x\\${FZF_CLICK_HEADER_COLUMN}"\fR
```
junegunn, line 1286:
```suggestion
     \fBprintf "head1\\nhead2" | fzf --header-lines=2 --bind 'click-header:transform-prompt:printf ${FZF_CLICK_HEADER_LINE}x${FZF_CLICK_HEADER_COLUMN}'\fR
```

src/terminal.go
junegunn: See 354b8b26dc310aaefc770c6a415e9be45a61ff9d

junegunn:
Merged, thanks!

kuremu:
Glad to contribute! Thanks

junegunn:
Hmm, I think we should reconsider how to report the line number as it can be confusing when you use multi-line `--header`.

Unlike `--header-lines`, multi-line `--header`s are always displayed from top to bottom, so in the following case,

```sh
fzf --header $'1. hello\n2. world' --header-first \
  --bind 'click-header:transform-prompt:echo ${FZF_CLICK_HEADER_LINE}x${FZF_CLICK_HEADER_COLUMN}'
```

clicking on `1. hello` reports line 2.

But when you're in `--reverse` layout, it will report line 1.

```sh
fzf --header $'1. hello\n2. world' --header-first \
  --bind 'click-header:transform-prompt:echo ${FZF_CLICK_HEADER_LINE}x${FZF_CLICK_HEADER_COLUMN}'
```

This gets even more confusing when you use both `--header` and `--header-lines`.

Should we always count from the top? Let me think about that.

PR #3781, opened 2024-05-08
feature: set FZF_SELECTION env var containing text of selected item

kuremu:
Sets `FZF_SELECTION` to the ansi-stripped text of the currently selected item in t.environ(), essentially inserting the value of `{}` into the environment. If there is nothing in the list, `FZF_SELECTION` is set to an empty string.

`fzf --bind "focus:transform-prompt(echo \${FZF_SELECTION}\>\ )"`

This would be useful to me as my reload/transform calls have become pretty complex, and I'm already capturing the other FZF env vars.

### DIFF
--- a/man/man1/fzf.1
+++ b/man/man1/fzf.1
@@ -992,6 +992,8 @@ fzf exports the following environment variables to its child processes.
 .br
 .BR FZF_POS "             Vertical position of the cursor in the list starting from 1"
 .br
+.BR FZF_SELECTION "       Text of currently selected line"
+.br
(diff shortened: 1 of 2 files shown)


junegunn:
I'm not sure about this (at least not yet). You can already do this with `current={}; rest_of_the_command`. And by doing this, we're pushing the current item to every child process, whether they need it or not. And in some cases, it can be quite lengthy (e.g. minified JS files, single line of hundreds of KBs), so it might cause some problems.

kuremu:
I see what you mean, no worries I'll close this for now

PR #3785, opened 2024-05-09
chore: Update flags to include long-form options for case

LangLangBart:
### Description

Addition of long options for the existing `-i` and `+i` flags.

The rationale behind long options is their self-descriptive nature. Currently, these flags are outliers as they do not have corresponding long options. Introducing them would improve consistency across command-line flags.


junegunn (approved):

junegunn:
Thanks!

Issue #3731, opened 2024-04-15
v0.50.0 not showing process data correctly


blaineventurine:
### Checklist

- [X] I have read through the manual page (`man fzf`)
- [X] I have searched through the existing issues
- [ ] For bug reports, I have checked if the bug is reproducible in the latest version of fzf

### Output of `fzf --version`

0.50.0 (brew)

### OS

- [ ] Linux
- [X] macOS
- [ ] Windows
- [ ] Etc.

### Shell

- [ ] bash
- [X] zsh
- [ ] fish

### Problem / Steps to reproduce

I have my completion trigger set to `,,`. After doing a `brew upgrade`, this is now the output of `kill ,,` for me:

<img width="1290" alt="image" src="https://github.com/junegunn/fzf/assets/18129692/3fc12384-9cd6-4f74-8653-2370f8fe705a">

If I select one (or more), it does not have a process ID, the command will be entered as: `kill 2^[[0m`

junegunn:
Looks like your ps commmand is producing ANSI color codes when it shouldn't. You see colored output when you do `ps -ef | cat`? And what is the output of `command ps`?

blaineventurine:
> Looks like your ps commmand is producing ANSI color codes when it shouldn't. You see colored output when you do `ps -ef | cat`? And what is the output of `command -v ps`?

No color output, and `/bin/ps`

LangLangBart:
@blaineventurine sets an alias for `cat`[^1] and has a config file for `bat` with the `--color=always` flag. [^2]

```bash
alias cat='bat'
```

Additionally, there is no `--ansi` flag in the `FZF_DEFAULT_OPTS`[^3].

---

Minimal setup to reproduce the issue:

```sh
export FZF_DEFAULT_OPTS=""
export FZF_COMPLETION_TRIGGER=',,'
alias cat='bat --color=always --number'
source /Users/paria/Developer/fzf/shell/completion.zsh
```

<img src="https://github.com/junegunn/fzf/assets/92653266/f741e687-709a-4272-9426-f3ced540fddf" width="600">

---


It seems to have come from commit 152988c17bd6c35ce55652d123db8b65a32f98e4.

Prepending `command` to the `cat` command in the function below seems to fix the issue. However,
this does not explain why the issue only became apparent after commit 152988c17bd6c35ce55652d123db8b65a32f98e4. Anyone understands this?

[...]

PR #3794, opened 2024-05-14
docs: update zsh integration instructions

LangLangBart:
### Problem

In #3736 the `alias` expansion was only tested for the old way of sourcing the `*.zsh` shell files.

```bash
export FZF_DEFAULT_OPTS=""
alias cat='bat --color=always --number'
source <path to file>
kill ** <TAB>
```

However, when attempting to use `eval`, it doesn't disable `alias expansion`, leading to the same issue as seen in #3731.

```bash
export FZF_DEFAULT_OPTS=""
alias cat='bat --color=always --number'
eval "$(fzf --zsh)"
kill ** <TAB>
```

---

### Proposal

The `eval` command affects the setup established in issue #1944. However, the following syntax appears to work effectively:

```bash
source <(fzf --zsh)
```

#### Alternative Solution

To resolve this, we could prepend `command` to the commands, such as `awk`, `cat`, `sed`, etc.

```diff
--- a/shell/completion.zsh
+++ b/shell/completion.zsh
@@ -201,5 +201,5 @@ _fzf_feed_fifo() (
   command rm -f "$1"
   mkfifo "$1"
-  cat <&0 > "$1" &
+  command cat <&0 > "$1" &
 )
```

---

### Other Projects

For comparison, consider how `kubectl` is set up on macOS:

- [Install and Set Up kubectl on macOS](https://kubernetes.io/docs/tasks/tools/install-kubectl-macos/)
```bash
source <(kubectl completion zsh)
```

---

Pinging @romkatv for your valuable input: we're trying to understand why `eval "$(fzf --zsh)"` has issues with alias expansion, while `source <(fzf --zsh)` seems to work well. Could you help us see if there are any potential problems with using the latter for sourcing the `*.zsh` files? Your expertise would be incredibly helpful.


junegunn:
Thanks, I can confirm that`source` does fix the issue though I don't understand why.

romkatv:
Aliases expanded during parsing. When using `eval`, the whole script is parsed at once, hence aliases are expanded before anything from `$script` is executed. When using `source`, commands are parsed one at a time.

Here's a demo:

```zsh
% eval '
  alias greet="echo hello"
  greet'
(eval):3: command not found: greet

% source <(print '
  alias greet="echo hello"
  greet')
hello
```

LangLangBart:
Thank you for the explanation.

```sh
alias cat='bat --color=always --number'
eval "$(fzf --zsh)"
whence -c _fzf_feed_fifo
_fzf_feed_fifo () {
  (
    command rm -f "$1"
    mkfifo "$1"
    bat --color=always --number <&0 > "$1" &
  )
}
```
```sh
alias cat='bat --color=always --number'
source <(fzf --zsh)
whence -c _fzf_feed_fifo
_fzf_feed_fifo () {
  (
    command rm -f "$1"
    mkfifo "$1"
    cat <&0 > "$1" &
  )
}
```

Issue #3777, opened 2024-05-07
Calling fzf#run with a list as source fail (n)vim is used from git bash


DanSM-5:
### Checklist

- [X] I have read through the manual page (`man fzf`)
- [X] I have searched through the existing issues
- [X] For bug reports, I have checked if the bug is reproducible in the latest version of fzf

### Output of `fzf --version`

0.51.0-devel (c5fb0c4)

### OS

- [ ] Linux
- [ ] macOS
- [X] Windows
- [ ] Etc.

### Shell

- [X] bash
- [X] zsh
- [ ] fish

### Problem / Steps to reproduce

I noticed some issues running `:Lines` or `:BLines` when using nvim under bash from git. The source of the issueis that when the `source` in the options passed to `fzf#run` is a list, the `source_command` will use the **type** command to get the content of the temporary file. This works fine if called from powershell or cmd as `type` will do the expected action but for bash (and zsh if you install it) it will fail as type is a shell built-in.

![image](https://github.com/junegunn/fzf/assets/49035812/ba7dafbe-7ad4-4e96-b3ae-29ef11f25734)

```diff
  if has_key(dict, 'source')
    let source = remove(dict, 'source')
    let type = type(source)
    if type == 1
      let source_command = source
    elseif type == 3
      let temps.input = s:fzf_tempname()
      call s:writefile(source, temps.input)
+     let source_command = (s:is_win ? 'type ' : 'cat ').fzf#shellescape(temps.input)
    else
      throw 'Invalid source type'
    endif
  else
```

I've manually changed the command from `type` to `cat` and that fixes the issue but it is hard to identify when bash/zsh is the current interactive shell. I'd suggest to provide some global variable like `g:fzf_cat` that if present, it will be used instead of choosing between `type` and `cat`.

If the detection route is preferred, checking for the environment variable `$SHELL` should work as it won't be set neither from cmd nor powershell.

junegunn:
I wonder if this patch fixes your issue:

```diff
diff --git a/plugin/fzf.vim b/plugin/fzf.vim
index b73f2aa..02dd971 100644
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -511,16 +511,16 @@ try
     let source = remove(dict, 'source')
     let type = type(source)
     if type == 1
-      let source_command = source
+      let prefix = '('.source.')|'
     elseif type == 3
       let temps.input = s:fzf_tempname()
       call s:writefile(source, temps.input)
[...]

Issue #1983, opened 2020-04-14
Disable path conversion for Git Bash


Tantalus13A98B5F:
- [x] I have read through the manual page (`man fzf`)
- [x] I have the latest version of fzf
- [x] I have searched through the existing issues

## Info

- OS
    - [ ] Linux
    - [ ] Mac OS X
    - [x] Windows
    - [ ] Etc.
- Shell
    - [x] bash (Git Bash/MSYS under mintty)
    - [ ] zsh
    - [ ] fish
    
## Problem / Steps to reproduce

It's the vim integration. 

1. Use the most recent Git for Windows (2.26.0)
2. Type `:FZF<CR>` in Vim
3. No new window of `cmd.exe` is started, and the window for vim itself would enter `cmd.exe`.

Seems the bash contained in Git for Windows is tranlating the `/C` arg for `cmd.exe` as a path (probably to `C:\`). The following patch would help (according to [this](https://github.com/git-for-windows/git/issues/577#issuecomment-166118846))

```diff
diff --git a/plugin/fzf.vim b/plugin/fzf.vim
index a84dcab..26dcaca 100644
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -587,7 +587,7 @@ function! s:execute(dict, command, use_height, temps) abort
   elseif has('win32unix') && $TERM !=# 'cygwin'
     let shellscript = s:fzf_tempname()
     call writefile([command], shellscript)
-    let command = 'cmd.exe /C '.fzf#shellescape('set "TERM=" & start /WAIT sh -c '.shellscript)
+    let command = 'MSYS_NO_PATHCONV=1 cmd.exe /C '.fzf#shellescape('set "TERM=" & start /WAIT sh -c '.shellscript)
     let a:temps.shellscript = shellscript
   endif
   if a:use_height
```

Konfekt:
While `MSYS_NO_PATHCONV=1` is one solution, why prefer it over doubling the first slash as alternatively proposed ?

junegunn:
Fixed in #3798

PR #3798, opened 2024-05-15
make :FZF work in Vim from Git Bash

Konfekt:
Despite its title 'Calling fzf#run with a list as source fail (n)vim is used from git bash' the issue in 

https://github.com/junegunn/fzf/issues/3777

of running `:FZF` in Vim in Git Bash was apparently only fixed for Neovim in Git Bash on Windows 11, but not for Vim from Git Bash.

In view of this, replacing /C by ///C might be considered a universal fix.

This PR just proposes the patch in https://github.com/junegunn/fzf/issues/1983 that still seems open.

In view of the fourth item in the most recent 2.45.0 https://github.com/git-for-windows/build-extra/blob/main/ReleaseNotes.md#known-issues little seems to have changed regarding path conversion of arguments containing forward slashes

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -711,7 +711,7 @@ function! s:execute(dict, command, use_height, temps) abort
   elseif has('win32unix') && $TERM !=# 'cygwin'
     let shellscript = s:fzf_tempname()
     call s:writefile([command], shellscript)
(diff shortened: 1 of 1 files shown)


junegunn (approved):
I just tested it on a Windows Server 2022 EC2 instance on AWS. Without the patch, cmd.exe opens when I run any command from fzf.vim, but with the patch, fzf opens in a cmd window.

Though I don't fully understand what's going on under the hood, it apparently fixes the problem, so I'm going to merge this.

Thanks a lot.

Issue #2798, opened 2022-04-22
fzf not working in git bash windows 10


ghost:
- [x ] I have read through the manual page (`man fzf`)
- [ x] I have the latest version of fzf
- [ x] I have searched through the existing issues

## Info

- OS
    - [ ] Linux
    - [ ] Mac OS X
    - [x ] Windows
    - [ ] Etc.
- Shell
    - [ x] bash
    - [ ] zsh
    - [ ] fish

## Problem / Steps to reproduce
Installation via scoop.
scoop install fzf
It's working out of box for cmd/powershell but on git bash all I can do is fzf --help
 When i use fzf in git-bash it seems to start but nothing happens.

DanSM-5:
I saw that behavior on the default git bash terminal. I'd suggest you to use a different terminal like `windows terminal`.

sindrom91:
I am also experiencing this issue.

sakirmohammad:
I found the solution, fzf not wroking on git bash
but you can use
```sh
winpty fzf
```
or in .bashrc file for git bash (idk if it exist or not)
```sh
alias fzf  ="winpty fzf"
```

knutze:
I am working on this issue in MSYS2 + mintty environment.

winpty solves the problem when stdin is a tty, but hangs when stdin is a pipe (same as #2255).
This means that the CTRL+R and CTRL+T functions provided by key-bindings.bash are not available.

I am trying the following wrappers and so far they are working fine in my environment.

```sh
fzf_winpty=~/.fzf/bin/fzf-winpty
cat >"${fzf_winpty}" <<'EOF'
#!/usr/bin/env bash
set -eo pipefail

prefix="$(basename "${BASH_SOURCE:-$0}")-${UID:-$(id -u)}"
tmpdir=$(mktemp -dp "${TMPDIR:-/tmp}" "${prefix}.XXXXX")
trap "rm -rf -- '${tmpdir}'" EXIT

args=
[[ $# -ge 1 ]] && args=$(printf ' %q' "$@")

if [[ -t 0 ]]; then
  winpty </dev/tty >/dev/tty -- bash -c \
    "command fzf${args} >'${tmpdir}'/output"
  cat "${tmpdir}"/output
else
  cat - >"${tmpdir}"/input
  winpty </dev/tty >/dev/tty -- bash -c \
    "command fzf${args} <'${tmpdir}'/input >'${tmpdir}'/output"
  cat "${tmpdir}"/output
fi
EOF
[...]

PR #3806, opened 2024-05-19
make fzf work in standard Git Bash Term

Konfekt:
Script provided by @knutze

Addresses https://github.com/junegunn/fzf/issues/2798

### DIFF
--- a/shell/completion.bash
+++ b/shell/completion.bash
@@ -42,6 +42,37 @@ __fzf_defaults() {
   echo "${FZF_DEFAULT_OPTS-} $2"
 }
 
+# make fzf work in Git Bash standard Mintty terminal by leveraging supplied winpty
+if [ "$TERM_PROGRAM" = "mintty" ]; then fzf() {
+  OLD_SHELL_OPTS=$(set +o)
+  set -eo pipefail
+  # Generate a unique temporary directory based on the script name and user ID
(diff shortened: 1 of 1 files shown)


junegunn:
The approach looks quite similar to what we've been doing in [fzf-tmux](https://github.com/junegunn/fzf/blob/master/bin/fzf-tmux), and the script was recently rewritten in [native Go code](https://github.com/junegunn/fzf/pull/3787). I'll see if we can do the same for this. If that's possible,

* it will work without extra configuration
* it will work not just on bash, but also on other shells

junegunn:
Can you test 573df524fed1c493ce7d8ea893f06ab90f2ca18a? It seems to work for me fine. Because it's using named pipes, it doesn't block even when the source command takes a while to complete.

```sh
./fzf
ls | ./fzf
(sleep 5; ls) | ./fzf
```

Attaching the binary built from the source.

[fzf.exe.zip](https://github.com/junegunn/fzf/files/15375382/fzf.exe.zip)

Konfekt:
I can confirm that this release works fine in Git Bash Term under Windows 10 without the bash script work around.
(It works also fine with the Bash work around. Removing the new exe, the pipe is broken again.)

junegunn:
Cool, I'll close this then and include the patch in the next release.

PR #3807, opened 2024-05-19
Development branch for next release (0.53)

junegunn:
Because this release will have some documentation changes, I'll be working on a non-master branch. Tests are welcome.

## Native `--tmux` integration

* https://github.com/junegunn/fzf/pull/3787

```sh
# --tmux [center|top|bottom|left|right][,SIZE[%]][,SIZE[%]]

# Center, 50% width 50% height
fzf --tmux center

# Center, 80% width and height
fzf --tmux 80%

# Center ,100% width and 50% height
fzf --tmux 100%,50%

# Left, 40% width
fzf --tmux left,40%

# Left, 40% width, 90% height
fzf --tmux left,40%,90%

# Top, 40% height
fzf --tmux top,40%

# Top, 90% height, 40% height
fzf --tmux top,90%,40%
```

## New defaults for pointer and marker

```sh
fzf --tmux 30% --multi
```
image

```sh
fzf --tmux 30% --multi --color gutter:-1
```

image

```sh
fzf --tmux 30% --multi --no-color
```

image

* The previous defaults (`>`) wouldn't look nice with multi-line display
* These Unicode bars look nice with or without the gutter and feel kind of modern

## Multi-line display of multi-line items

```sh
rg -p bash | perl -0777 -pe 's/\n\n/\n\0/gm' | fzf --read0 --ansi --multi --highlight-line --reverse --tmux 70%
```

image

## `--tail=NUM` to limit the number of items to keep in memory

Useful when you want to browse an endless stream of data (e.g. log stream) with fzf while limiting memory usage.

```sh
# Interactive filtering of a log stream
tail -f *.log | fzf --tail 100000 --tac --no-sort --exact
```

## Following `NO_COLOR` protocol

* https://no-color.org/

If `$NO_COLOR` environment variable is set to a non-empty value, fzf will default to `--no-color` mode.

## Git bash (mintty) support

fzf now works on Git bash out of the box via `winpty`. However, `--height` option doesn't work.

## Embedded man page

You can run `fzf --man` to see the man page. This simplifies the distribution.


junegunn:
image

Windows terminals don't seem to render `▏` correctly. Going to use `▌` for marker instead.

LangLangBart:
> Tests are welcome

---

> Multi-line display of multi-line items

Multiline is quite effective 😲 . Now I have a powerful tool to search through all my functions for a keyword.

```sh
# zsh
# -x2: sets tab indentation
functions -x2 | perl -0777 -pe 's/^}\n/}\0/gm' |
  fzf \
    --bind 'enter:become:echo {1}' \
    --preview "bat --color=always -pl zsh {f}" \
    --read0
```




---

> New defaults for pointer and marker

Pointer looks good; I will adopt it. As for the marker, I have not decided yet.

---

> These Unicode bars look nice with or without the gutter and feel kind of modern

Something else that looks nice is the `--scroll-off` option. With this new visual changes, it might be a good time to make this the default.

```sh
fzf --scroll-off 3
```

Hard to imagine someone doesn't like it.

LangLangBart:
So it is no longer necessary to append `>/dev/tty`. Thank you.

Do you think it should be mentioned in the `CHANGELOG` ?

```sh
# old
selection=$(command fd --type file | fzf --bind 'ctrl-a:execute:less {} >/dev/tty')

# new
selection=$(command fd --type file | fzf --bind 'ctrl-a:execute:less {}')
```

LangLangBart:
Adding `126`[^1] to the `man` page ?

https://github.com/junegunn/fzf/blob/996abb283152219732b6da5d11ccc386f9565913/man/man1/fzf.1#L981-L990

```sh
: | fzf --bind "start:become:$HOME"
# zsh:1: permission denied: /Users/paria
echo $?
# 126
: | fzf --bind 'start:become:cheese_cake'
# zsh:1: command not found: cheese_cake
echo $?
# 127
```

[^1]: [POSIX: Shell Command Language - 2.8.2 Exit Status for Commands](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_08_02)

junegunn:
@LangLangBart Thanks for the test!

> Now I have a powerful tool to search through all my functions for a keyword.

That looks super nice.

> Pointer looks good; I will adopt it. As for the marker, I have not decided yet.

The primary motivation for changing the defaults was that `>` looks awful with multi-line entries. The bars look okay with both single-line and multi-line entries.

For multi-line entries, I also thought about using different characters for the first line and the last line, so something like these are possible.

image

image

It's nice, but I didn't like that we would need to add several more options for marker

1. first line char
1. last line char
1. char for the rest of the lines
1. marker for a single-line item

```sh
--marker '∎' --marker-multi '┏┃┗'
``` 

But is it worth it? Isn't it too much? I'm not sure.

> Something else that looks nice is the --scroll-off option.

We can consider that. But the option value is based on the number of items, not lines, so 3 can mean quite many lines in some multi-line cases.

> So it is no longer necessary to append >/dev/tty. Thank you.
> Do you think it should be mentioned in the CHANGELOG ?

Yeah. I didn't like the inconsistency between `fzf --bind space:execute:vim` and `fzf --bind space:execute:vim > output`, where the latter would require `> /dev/tty` redirection. It may break some exotic use cases, but `execute` is really for launching an interactive application from within fzf, and it shouldn't affect the final output of fzf, so I believe this is the right direction. However, the problem is still there for Windows. Will add it to CHANGELOG.

> Adding 126[1](https://github.com/junegunn/fzf/pull/3807#user-content-fn-1-64cdaf87bd8872217e693b4caa5444e8) to the man page ?

Will do, thanks!

src/options.go
```diff
@@ -81,8 +92,8 @@ const Usage = `usage: fzf [options]
     --scrollbar[=C1[C2]]   Scrollbar character(s) (each for main and preview window)
     --no-scrollbar         Hide scrollbar
     --prompt=STR           Input prompt (default: '> ')
-    --pointer=STR          Pointer to the current line (default: '>')
-    --marker=STR           Multi-select marker (default: '>')
+    --pointer=STR          Pointer to the current line (default: '▌' or '>')
+    --marker=STR           Multi-select marker (default: '▏' or '>')
```
LangLangBart, line 96:
```suggestion
    --marker=STR           Multi-select marker (default: '┃' or '>')
```

LangLangBart:
>> New defaults for pointer and marker

> ... As for the marker, I have not decided yet.

An earlier version of the marker was quite thin and hence hard to see, the new version suits me better.



| Character Name          | Unicode                                                             |
| ------------------ | ------------------------------------------------------------------- |
| LEFT ONE EIGHTH BLOCK              | [U+258F](https://util.unicode.org/UnicodeJsps/character.jsp?a=258F) |
| BOX DRAWINGS HEAVY VERTICAL              | [U+2503](https://util.unicode.org/UnicodeJsps/character.jsp?a=2503) |

---

> It's nice, but I didn't like that we would need to add several more options for marker

The image of the multiline markers looks very modern. Maybe the name of the `marker-multi` flag should be `marker-multiline`, as `marker-multi` might be intuitively interpreted as an alias for `marker`. Its semantics can be argued either way.

junegunn:
How do you feel about `▍` or ` ▎`? I think they look slightly better than `┃` because there is a little more space between the bar and the text though it's not a huge difference.

```sh
fzf -m --marker ▍
fzf -m --marker ▎
fzf -m --marker ▎ --color gutter:-1

**Comment by junegunn** [comment]
Ah, never mind, they don't display correctly on Windows terminals.

**Comment by LangLangBart** [comment]
It can happen that some parts of the current `marker` remain visible after it has been toggled off.

GIF with Apple Terminal (tested for `devel` and `master` branches, same result). Notice the little squares remaining after toggling off.

```sh
pwd
# /Users/paria/Developer/fzf
FZF_DEFAULT_OPTS="" FZF_DEFAULT_COMMAND="" fzf --multi
```



---

| Character Name              | Unicode                                                             | little squares remaining after toggling |
| --------------------------- | ------------------------------------------------------------------- | --------------------------------------- |
| LEFT THREE EIGHTHS BLOCK    | [U+258D](https://util.unicode.org/UnicodeJsps/character.jsp?a=258D) | NO                                      |
| LEFT ONE QUARTER BLOCK      | [U+258E](https://util.unicode.org/UnicodeJsps/character.jsp?a=258E) | NO                                      |
| LEFT ONE EIGHTH BLOCK       | [U+258F](https://util.unicode.org/UnicodeJsps/character.jsp?a=258F) | NO                                      |
| BOX DRAWINGS HEAVY VERTICAL | [U+2503](https://util.unicode.org/UnicodeJsps/character.jsp?a=2503) | YES                                     |

#### environment info

```sh
command env | grep '^TERM_PROGRAM'
# TERM_PROGRAM=Apple_Terminal
# TERM_PROGRAM_VERSION=433
```

Font: `MesloLGS NF Regular`
- [powerlevel10k/font.md at master · romkatv/powerlevel10k · GitHub](https://github.com/romkatv/powerlevel10k/blob/master/font.md#recommended-font-meslo-nerd-font-patched-for-powerlevel10k)



PS: The issue is not reproducible with iTerm 2.

LangLangBart:
## Multiline `shell` history

In `bash` and `fish`, the `--read0` flag is already included. `bash` lists the event numbers first, and then the command. With your code changes in this `devel` branch, the line breakage looks atypical. `--read0` was added to `bash`  with #1837.

In `fish`, there are no event numbers, so there is no issue at all.

In `zsh`, `--read0` is not used, so there is no issue other than inconsistent appearance compared to `bash` and `fish`.

---

Do you think that `--with-nth 2..` should be enabled in `bash` by default?

Should `zsh` include `--read0` and `--with-nth 2..`?  

---

### bash



https://github.com/junegunn/fzf/blob/32eb32ee5e80ab4e10ead9bc6db8c6ad59cf509d/shell/key-bindings.bash#L64-L66

https://github.com/junegunn/fzf/blob/32eb32ee5e80ab4e10ead9bc6db8c6ad59cf509d/shell/key-bindings.bash#L93-L95

### fish



https://github.com/junegunn/fzf/blob/32eb32ee5e80ab4e10ead9bc6db8c6ad59cf509d/shell/key-bindings.fish#L71-L73

### zsh

junegunn:
Thanks for pointing it out.

> the line breakage looks atypical

Yes, but I don't think it's that bad, and it doesn't bother me. Actually, I prefer seeing them in multi-line form. It's the same output format you see when you run `HISTTIMEFORMAT= history`. Some indentation would have been nicer, though.

> Do you think that `--with-nth 2..` should be enabled in `bash` by default?

No, I prefer to have the numbers. They help me understand the chronological ordering among the matched entries.

image

> Should `zsh` include `--read0` and `--with-nth 2..`?

Can we make zsh also display actual new-line characters instead of `\n`

LangLangBart:
> Can we make zsh also display actual new-line characters instead of `\n`

I have not started experimenting with it; I only looked at how `bash` does it. There is always a way, but it may not be trivial.


> No, I prefer to have the numbers.

OK. Lots of other `fzf` users might have been surprised by this as well and complained.

>  I prefer seeing them in multi-line form

I see. If a user doesn't like it, they can assign `FZF_CTRL_R_OPTS` with `--no-multi-line`, I assume.

LangLangBart:
> bash
> This shows that, without numbers, it can be hard to see if the lines (`echo "multi` and `line"`) are from separate commands or not.


To wrap up the discussion on multiline shell history:

- **bash**: requires no changes, as you prefer multiline with event numbers
- **fish**: I recommend including `--no-multi-line`, because of the lack of event numbers. see image below for more clarity.



- **zsh**: requires no changes unless a `zsh` command emerges that replicates bash's output method.

Do you agree with the above ?

---

Regarding `zsh`, this command displays unique history commands, separated by a `NUL` byte, without event numbers and in multiline format.


```sh
print -rN -- ${(uv)history[@]} | fzf --read0 
```

I continue to seek a native command that presents event numbers alongside the command lines, without resorting to external tools.

junegunn:
> Do you agree with the above ?

I'm a bit torn. Yes, it can be confusing at times, but I think it's actually desirable to see multi-line commands in their original form. Let me think about it more.

---

On fish, we could use something like these.

* `history --show-time="%Y-%m-%d %H:%M  " -z | fzf --read0  --nth 3.. | awk 'NR == 1 { $1 = ""; $2 = ""; sub("  ", "") } 1'
`
    * Need to check listing performance
    * TMI?
* `history -z | perl -0 -pe 's/^/$.\t/g; s/\n/\n\t/gm' | fzf --read0`
    * Numbers are backwards
    * Requires perl

junegunn:
New fish version using perl.

image

It's going to be slower than the previous version because of `history --reverse` and `--tac`. The difference is slightly noticeable when listing 100K+ entries.

shell/key-bindings.fish
```diff
@@ -69,8 +69,13 @@ function fzf_key_bindings
       # history's -z flag was added in fish 2.4.0, so don't use it for versions
       # before 2.4.0.
       if [ "$FISH_MAJOR" -gt 2 -o \( "$FISH_MAJOR" -eq 2 -a "$FISH_MINOR" -ge 4 \) ];
-        history -z | eval (__fzfcmd) --read0 --print0 -q '(commandline)' | read -lz result
-        and commandline -- $result
+        if type -q perl
+          history -z --reverse | perl -0 -pe 's/^/$.\t/g; s/\n/\n\t/gm' | eval (__fzfcmd) --tac --read0 --print0 -q '(commandline)' | sed 's/^[0-9]*\t//' | read -lz result
```
LangLangBart, line 73: > New `fish` version using perl.



```suggestion
          history -z --reverse | perl -0 -pe 's/^/$.\t/g; s/\n/\n\t/gm' | eval (__fzfcmd) --delimiter '\t' --nth 2..,.. --tac --read0 --print0 -q '(commandline)' | sed 's/^[0-9]*\t//' | read -lz result
```

LangLangBart:
bug introduced with 04db44067dd4; to reproduce:

```sh
FZF_DEFAULT_OPTS="--bind 'scroll-down:offset-down'" printf "Eins\nZwei\0" | fzf --read0
# try to scroll down with the mouse
# panic: runtime error: index out of range [-1]  
```

shell/key-bindings.fish
LangLangBart: Thanks for the changes to `fish`. 

I worked on the `zsh` version, but had to use `perl` as well. Will test it for a day and make a PR.

```sh
printf '%1$s\t%2$s\000' "${(vk)history[@]}" | 
  perl -0 -ne 'if (!$seen{(/^\s*[0-9]+\**\s+(.*)/, $1)}++) { s/\n/\n\t/gm; print; }' |
  fzf --read0
```

LangLangBart: > USE OF '\t' IN SED SCRIPTS: For clarity in documentation, we have used
the expression '\t' to indicate a tab character (0x09) in the scripts.
However, most versions of sed do not recognize the '\t' abbreviation,
so when typing these scripts from the command line, you should press
the TAB key instead. '\t' is supported as a regular expression
metacharacter in awk, perl, and in a few implementations of sed.
> 
> https://www.unixguide.net/unix/sedoneliner.shtml

In my version of `sed`, the `\t` character is not recognized. However, `[[:blank:]]` or `[[:space:]]` works with the `-E` flag, or a literal tab character. The latter option is not my preferred choice because the tab character is invisible in scripts and could be accidentally altered when copying and pasting.


`fish` does not support  C-Quotes `$'\t'` either



I would suggest drop `sed` for a `perl` or `awk` solution.

```sh
perl -pe 's/^\d+\t//'
```

LangLangBart: I think `s/\n/\n\t/gm` makes the history look better in `fzf` but the result is an additional `tab` char, i think we should not alter the command or revert the extra tab `'s/\n\t/\n/gm'`?

junegunn: I used `[0-9]*` instead of `+` to strip leading tabs in every line.

junegunn: > I would suggest drop `sed` for a `perl` or `awk` solution.

Sure, see 3afd543 and 700bf1d

junegunn:
> bug introduced with [04db440](https://github.com/junegunn/fzf/commit/04db44067dd4d9bf7f85ae4d704e740d7420f957)

Fixed in[49c9b5f](https://github.com/junegunn/fzf/pull/3807/commits/49c9b5fb56c630c2516b295d407941e5da4144c8). Thanks for thorough testing!

shell/key-bindings.fish
LangLangBart: Thank you. I'm unclear on why `--nth 2..,..` wasn't included as well.

junegunn: It was an oversight. Now fish also needs it.

junegunn: 21f8d88f18a7327d67986f6a534b7b087e6c7fdc

shell/key-bindings.fish
```diff
@@ -59,7 +59,7 @@ function fzf_key_bindings
   function fzf-history-widget -d "Show command history"
     test -n "$FZF_TMUX_HEIGHT"; or set FZF_TMUX_HEIGHT 40%
     begin
-      set -lx FZF_DEFAULT_OPTS (__fzf_defaults "" "--scheme=history --bind=ctrl-r:toggle-sort $FZF_CTRL_R_OPTS +m")
+      set -lx FZF_DEFAULT_OPTS (__fzf_defaults "" "-n2..,.. --scheme=history --bind=ctrl-r:toggle-sort --highlight-line $FZF_CTRL_R_OPTS +m")
```
LangLangBart, line 62:
```suggestion
      set -lx FZF_DEFAULT_OPTS (__fzf_defaults "" "--scheme=history --bind=ctrl-r:toggle-sort --highlight-line $FZF_CTRL_R_OPTS +m")
```

shell/key-bindings.fish
```diff
@@ -69,8 +69,13 @@ function fzf_key_bindings
       # history's -z flag was added in fish 2.4.0, so don't use it for versions
       # before 2.4.0.
       if [ "$FISH_MAJOR" -gt 2 -o \( "$FISH_MAJOR" -eq 2 -a "$FISH_MINOR" -ge 4 \) ];
-        history -z | eval (__fzfcmd) --read0 --print0 -q '(commandline)' | read -lz result
-        and commandline -- $result
+        if type -q perl
+          history -z --reverse | perl -0 -pe 's/^/$.\t/g; s/\n/\n\t/gm' | eval (__fzfcmd) --tac --read0 --print0 -q '(commandline)' | perl -pe 's/^\d*\t//' | read -lz result
```
LangLangBart, line 73:
```suggestion
          history -z --reverse | perl -0 -pe 's/^/$.\t/g; s/\n/\n\t/gm' | eval (__fzfcmd) -n2..,.. --tac --read0 --print0 -q '(commandline)' | perl -pe 's/^\d*\t//' | read -lz result
```

LangLangBart (changes requested):
apply `-n2..,..` only when there are numbers

shell/key-bindings.fish
junegunn: Ah, okay. But in this way, we can't allow users to override `--nth` in their `$FZF_CTRL_R_OPTS`. Please see 05b4e09.

LangLangBart:
Do you think my earlier comment about the marker is an issue, or should it be left to the user to change the marker in their `FZF_DEFAULT_OPTS`?

https://github.com/junegunn/fzf/pull/3807#issuecomment-2132402770

junegunn:
The problem is only reproducible when using the font you mentioned, which has to be installed manually, right? If that's the case, I think it's fine, and we should probably report the problem to the font maintainers.

I'm just guessing here, but most people who care enough about their terminal experience to even install a custom font are probably using different terminal emulators.

LangLangBart:
> The problem is only reproducible when using the font you mentioned, which has to be installed manually, right? 

Right, but I did not test all of them, just a random sample. The one mentioned remains the only one.

Increasing the `line spacing` also _fixed_ it for me.

fzf

Issue #3809, opened 2024-05-21
On fzf.vim opening in new cmd Window in Git Bash


ykhan21:
### Checklist

- [X] I have read through the manual page (`man fzf`)
- [X] I have searched through the existing issues
- [X] For bug reports, I have checked if the bug is reproducible in the latest version of fzf

### Output of `fzf --version`

0.52.1 (6432f00)

### OS

- [ ] Linux
- [ ] macOS
- [X] Windows
- [ ] Etc.

### Shell

- [X] bash
- [ ] zsh
- [ ] fish

### Problem / Steps to reproduce

[This](https://github.com/junegunn/fzf/pull/3798) PR made by @Konfekt allows fzf.vim to work in Git Bash.
However, I found that by commenting the following part of the if-branch allows fzf.vim with the stock vim in Git Bash to run normally.
https://github.com/junegunn/fzf/blob/daa602422dc272cbec6b99bc7520262672ad7230/plugin/fzf.vim#L711-L716

In Git Bash's vim, `:echo has('unix')` returns 1, `:echo has('win32unix')` returns 1, `:echo has('win32')` returns 0, and `:echo has('win64')` returns 0.

What is has('win32unix') used for? It seems like it can be removed in a few cases.

Konfekt:
This checks for the Win32 version of Vim, using Unix files (Cygwin), according to `:help feature-list`, but Cygwin nowadays rather refers to MSYS2. 
One striking difference is that it expects Unix paths.

As `vim` in Git Bash is the default Vim, we looked no further.
By `has('win32unix')` being true, we know not only that this is most likely Git Bash vim, but also that the Git Bash executable exists, used by line 714.

So yes, one could likely check for something like ` has('win32unix') || (has('win32') && &shell =~# '\v<(ba)?sh$')` instead of has('win32unix')

Likely only its original author @janlazo can tell us why `$TERM !=# 'cygwin'` must be ensured.

ykhan21:
Why does it need to be run as a cmd.exe process?

Removing this branch lets Git Bash's vim use the unix branches and most importantly does not cause a separate cmd.exe Window to appear.

Konfekt:
> Removing this branch lets Git Bash's vim use the unix branches and most importantly does not cause a separate cmd.exe Window to appear.

Is it? What does branch refer to? If you mean spawning cmd.exe, that would be great.
[...]

PR #3811, opened 2024-05-22
Git Bash Mintty: only use cmd.exe if winpty missing

Konfekt:
Addresses https://github.com/junegunn/fzf/issues/3809

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -710,10 +710,10 @@ function! s:execute(dict, command, use_height, temps) abort
       call jobstart(cmd, fzf)
       return []
     endif
-  elseif has('win32unix') && $TERM !=# 'cygwin'
+  elseif has('win32unix') && $TERM_PROGRAM ==# 'mintty' && !executable('winpty')
     let shellscript = s:fzf_tempname()
     call s:writefile([command], shellscript)
-    let command = 'cmd.exe //C '.fzf#shellescape('set "TERM=" & start /WAIT sh -c '.shellscript)
+    let command = 'start //WAIT sh -c '.shellscript
     let a:temps.shellscript = shellscript
   endif
   if a:use_height
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -529,8 +529,9 @@ try
         \ executable('tput') && filereadable('/dev/tty')
   let has_vim8_term = has('terminal') && has('patch-8.0.995')
   let has_nvim_term = has('nvim-0.2.1') || has('nvim') && !s:is_win
-  let use_term = has_nvim_term ||
-    \ has_vim8_term && !has('win32unix') && (has('gui_running') || s:is_win || s:present(dict, 'down', 'up', 'left', 'right', 'window'))
+  let use_term = has_nvim_term || has_vim8_term 
+    \ && !(has('win32unix') && $TERM_PROGRAM ==# 'mintty' && !executable('winpty'))
+    \ && (has('gui_running') || s:is_win || s:present(dict, 'down', 'up', 'left', 'right', 'window'))
   let use_tmux = (has_key(dict, 'tmux') || (!use_height && !use_term || prefer_tmux) && !has('win32unix') && s:splittable(dict)) && s:tmux_enabled()
   if prefer_tmux && use_tmux
     let use_height = 0


junegunn:
Thanks. While testing the patch, I have noticed a few things.

* When fzf is started from Git bash, CTRL-C fails to properly terminate fzf and I get `error restoring terminal mode: Input/output error` message. `winpty -- fzf --no-winpty` doesn't have to problem. I'll look into it
* The patch works; fzf opens in full-screen mode. But we also have that `CTRL-C` problem.
* Looks like we can use built-in terminal of Vim. It works nicely, and fzf opens in the central popup window inside Vim. We can do this in addition to the patch. What do you think?
```diff
diff --git a/plugin/fzf.vim b/plugin/fzf.vim
index eee5e9c..290db45 100644
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -530,7 +530,7 @@ try
   let has_vim8_term = has('terminal') && has('patch-8.0.995')
   let has_nvim_term = has('nvim-0.2.1') || has('nvim') && !s:is_win
   let use_term = has_nvim_term ||
-    \ has_vim8_term && !has('win32unix') && (has('gui_running') || s:is_win || s:present(dict, 'down', 'up', 'left', 'right', 'window'))
+    \ has_vim8_term && (has('gui_running') || s:is_win || s:present(dict, 'down', 'up', 'left', 'right', 'window'))
   let use_tmux = (has_key(dict, 'tmux') || (!use_height && !use_term || prefer_tmux) && !has('win32unix') && s:splittable(dict)) && s:tmux_enabled()
   if prefer_tmux && use_tmux
     let use_height = 0
```
* However, the border around fzf is for some reason not visible
  image
* When I select a file from fzf, I'm getting this error.
  image
    * Seems to be caused by mixed `/` and `\`.

junegunn:
CTRL-C problem is fixed in 303c3bae7ff02996bdbad4b743cdb23c1813a4c6.

Konfekt:
> When fzf is started from Git bash, CTRL-C fails to properly terminate fzf and I get error restoring terminal mode: Input/output error message. I'll look into it

I can't reproduce it with Windows 10, Git Bash 2.42.0 and latest fzf 0.52.1dev (without `winpty`), neither in Git Bash nor inside Vim in Git Bash. However, exiting, independent of the shortcut, sends an `130` signal. That might be on purpose, though.

> We can do this in addition to the patch. What do you think?

I think that's a good idea. My slight adaption tries to keep everything working where `winpty` is not readily available, say older installs of Git on administered Window work machines, in harmony with the change at the other line.

Let me suggest an additional check for the existence of `winpty` at [`573df52` (#3807)](https://github.com/junegunn/fzf/pull/3807/commits/573df524fed1c493ce7d8ea893f06ab90f2ca18a#diff-0687569756f14eb4aea3bc7f26ca8dcbf2274db3687b628212e14fbe909ca873) to this end.

junegunn:
> Let me suggest an additional check for the existence of `winpty` at [`573df52` (#3807)](https://github.com/junegunn/fzf/pull/3807/commits/573df524fed1c493ce7d8ea893f06ab90f2ca18a#diff-0687569756f14eb4aea3bc7f26ca8dcbf2274db3687b628212e14fbe909ca873) to this end.

So what do we do when `winpty` is not available? I decided to not perform the check and let it fail because fzf wouldn't run it anyway without winpty, but according to the latest discussion in https://github.com/junegunn/fzf/issues/3809#issuecomment-2124893297 that is not necessarily true?

junegunn:
```diff
diff --git a/src/core.go b/src/core.go
index 9c5ac03..f0728f8 100644
--- a/src/core.go
+++ b/src/core.go
@@ -3,6 +3,7 @@ package fzf
 
 import (
 	"os"
+	"os/exec"
 	"sync"
 	"time"
 
@@ -25,7 +26,9 @@ func Run(opts *Options) (int, error) {
 	}
 
 	if os.Getenv("TERM_PROGRAM") == "mintty" && !opts.NoWinpty {
-		return runWinpty(os.Args, opts)
+		if _, err := exec.LookPath("winpty"); err == nil {
+			return runWinpty(os.Args, opts)
+		}
 	}
 
 	if err := postProcessOptions(opts); err != nil {
```

junegunn:
* `winpty` detection: https://github.com/junegunn/fzf/commit/b5b0d6b3ead5d4d65367b57aac5d879975714cd9
* Fix mixed forward and backward slashes: https://github.com/junegunn/fzf/commit/b5b0d6b3ead5d4d65367b57aac5d879975714cd9

Konfekt:
> fzf wouldn't run it anyway without winpty, but according to the latest discussion in https://github.com/junegunn/fzf/issues/3809#issuecomment-2124893297 that is not necessarily true?

When I removed winpty on Windows 10 with fzf 0.52.1-dev then `:FZF` in Vim in Git Bash Mintty with

```vim
  elseif has('win32unix') && $TERM_PROGRAM ==# 'mintty' " && !executable('winpty')
    let shellscript = s:fzf_tempname()
    call s:writefile([command], shellscript)
    let command = 'start //WAIT sh -c '.shellscript
    let a:temps.shellscript = shellscript
  endif
```

the pop up shows

grafik

whereas fzf 0.52.1 works fine.

So it was meant that 0.52.1-dev keeps doing in the absence of winpty on Windows 11 what 0.52.1 did

PR #3820, opened 2024-05-28
Change the full path logic for Windows to include backslash.

Konstantin-Glukhov:
On Windows paths starting with "\" are prepended with the value of "cwd", which does not make sense. This commit to address this issue and treat  paths starting with "\" the same way as "^[A-Z]:\".

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -327,7 +327,7 @@ function! s:common_sink(action, lines) abort
     " the execution (e.g. `set autochdir` or `autocmd BufEnter * lcd ...`)
     let cwd = exists('w:fzf_pushd') ? w:fzf_pushd.dir : expand('%:p:h')
     for item in a:lines
-      if item[0] != '~' && item !~ (s:is_win ? '^[A-Z]:\' : '^/')
+      if ! (item =~ '^\~' || item =~ '^/' || item =~ '^\([A-Z]:\)*\')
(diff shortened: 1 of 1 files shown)


junegunn:
What are the examples of paths starting with `\`?

Konstantin-Glukhov:
\dir1\dir2\file.txt
\\Server1\MountPoint\dir1\dir2\file.txt

I tested the change with all cases (including on Linux). It seems working fine.

plugin/fzf.vim
```diff
@@ -327,7 +327,7 @@ function! s:common_sink(action, lines) abort
     " the execution (e.g. `set autochdir` or `autocmd BufEnter * lcd ...`)
     let cwd = exists('w:fzf_pushd') ? w:fzf_pushd.dir : expand('%:p:h')
     for item in a:lines
-      if item[0] != '~' && item !~ (s:is_win ? '^[A-Z]:\' : '^/')
+      if ! (item =~ '^\~' || item =~ '^/' || item =~ '^\([A-Z]:\)*\')
```
junegunn, line 330: We should only test for `[A-Z]:` on Windows.

So how about this?

```suggestion
      if item !~ '^[/~]' && !(s:is_win && item =~ '^\([A-Z]:\)*\')
```

junegunn:
Merged, thanks!

Konstantin-Glukhov:
Thank you for a quick review!

PR #3822, opened 2024-05-29
Add support for Alacritty preview in ANSI Format

jeeftor:
## Change 

Modification to `fzf-preview.sh` to support Alacritty which doesn't have [sixel support](https://github.com/alacritty/alacritty/issues/910)

image

- **Alacritty** black bg
- **Wezterm** Blue bg

Alacritty doesn't support SIXELS but `chafa` can still work

### DIFF
--- a/bin/fzf-preview.sh
+++ b/bin/fzf-preview.sh
@@ -57,7 +57,12 @@ if [[ $KITTY_WINDOW_ID ]]; then
 
 # 2. Use chafa with Sixel output
 elif command -v chafa > /dev/null; then
-  chafa -f sixel -s "$dim" "$file"
+  # Alacritty doesn't support sixels
+  if [[ $ALACRITTY_WINDOW_ID ]]; then
+    chafa --colors 256 -s "$dim" "$file"
+  else
+    chafa -f sixel -s "$dim" "$file"
+  fi  
   # Add a new line character so that fzf can display multiple images in the preview window
   echo
 


bin/fzf-preview.sh
```diff
@@ -57,7 +57,12 @@ if [[ $KITTY_WINDOW_ID ]]; then
 
 # 2. Use chafa with Sixel output
 elif command -v chafa > /dev/null; then
-  chafa -f sixel -s "$dim" "$file"
+  # Alacritty doesn't support sixels
+  if [[ $ALACRITTY_WINDOW_ID ]]; then
+    chafa --colors 256 -s "$dim" "$file"
+  else
+    chafa -f sixel -s "$dim" "$file"
+  fi
```
junegunn, line 65: The man page of chafa states that it auto-detects the right format.

```
OUTPUT ENCODING
       -f, --format format
           Set output format; one of [iterm, kitty, sixels, symbols]. The
           default is iterm, kitty or sixels if the connected terminal
           supports one of these, falling back to symbols ("ANSI art")
           otherwise.

```

Which means we can just drop `-f sixel` instead?

```suggestion
  chafa -s "$dim" "$file"
```

---

Hmm, no, chafa fails to detect it's on iTerm2 and doesn't render image nicely on it. The issue is fixed on master, but no releases yet.

https://github.com/hpjansson/chafa/issues/196

---

I don't use alacritty, but I believe it's capable of showing more than 256 colors. So why don't we just drop `--colors 256`?

```
If left unspecified, an optimal default will be chosen based on the current environment.
```

jeeftor: I'll give this a test ... on vacation for a week but have some time today to verify

jeeftor: image

I can confirm this ... on iterm2 it shows up as ansi/

image

junegunn: Now that we have the new chafa version with the fix, I think we can just drop `-f sixel` as suggested above.

PR #3823, opened 2024-05-30
feat: zsh multiline mode

LangLangBart:
### description

enable multiline mode for `zsh`

### DIFF
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -122,7 +122,7 @@ fzf-history-widget() {
   fi
   local ret=$?
   if [ -n "$selected" ]; then
-    if num=$(command awk '{print $1; exit}' <<< "$selected" | command grep -o '^[1-9][0-9]*'); then
+    if num=$(awk '{print $1; exit}' <<< "$selected" | grep -o '^[1-9][0-9]*'); then
       zle vi-fetch-history -n $num
     else # selected is a custom query, not from history
       LBUFFER="$selected"
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -122,10 +122,7 @@ fzf-history-widget() {
   fi
   local ret=$?
   if [ -n "$selected" ]; then
-    num=$(awk '/^[[:blank:]]*[1-9][0-9]*\*?[[:blank:]]+/ {print $1}' <<< "$selected")
-    if (( ${#num} )); then
-      # zsh utilizes the 'atoi' function to convert a string into an integer
-      # https://github.com/junegunn/fzf/issues/3591#issuecomment-1902090700
+    if num=$(command awk '{print $1; exit}' <<< "$selected" | command grep -o '^[1-9][0-9]*'); then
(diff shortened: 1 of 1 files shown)


shell/key-bindings.zsh
```diff
@@ -108,14 +108,25 @@ fi
...
+  fi
   local ret=$?
   if [ -n "$selected" ]; then
-    num=$(awk '{print $1}' <<< "$selected")
-    if [[ "$num" =~ '^[1-9][0-9]*\*?$' ]]; then
-      zle vi-fetch-history -n ${num%\*}
+    num=$(awk '/^[[:blank:]]*[1-9][0-9]*\*?[[:blank:]]+/ {print $1}' <<< "$selected")
+    if (( ${#num} )); then
+      # zsh utilizes the 'atoi' function to convert a string into an integer
+      # https://github.com/junegunn/fzf/issues/3591#issuecomment-1902090700
```
junegunn, line 128: How about using grep to extract the number part from the first token of the first line?

```suggestion
    if num=$(command awk '{print $1; exit}' <<< "$selected" | command grep -o '^[1-9][0-9]*'); then
```

(I wonder if we should consistently prepend `command` to all instances of awk and perl.)

LangLangBart:
> (I wonder if we should consistently prepend command to all instances of awk and perl.)

Would you like me to do this here for the `*.zsh` shell files, or should I create a new PR?

I guess you fear some users might be using the `eval` method to source their scripts?

junegunn:
> I guess you fear some users might be using the `eval` method to source their scripts?

Oh, we made sure aliases are not expanded when properly sourced. I forgot about that bit. It's fine then.

shell/key-bindings.zsh
```diff
@@ -108,14 +108,22 @@ fi
...
+    selected="$(fc -rl 1 | awk '{ cmd=$0; sub(/^[ \t]*[0-9]+\**[ \t]+/, "", cmd); if (!seen[cmd]++) print $0 }' |
+      FZF_DEFAULT_OPTS=$(__fzf_defaults "" "-n2..,.. --scheme=history --bind=ctrl-r:toggle-sort --highlight-line ${FZF_CTRL_R_OPTS-} --query=${(qqq)LBUFFER} +m") \
+      FZF_DEFAULT_OPTS_FILE='' $(__fzfcmd))"
+  fi
   local ret=$?
   if [ -n "$selected" ]; then
-    num=$(awk '{print $1}' <<< "$selected")
-    if [[ "$num" =~ '^[1-9][0-9]*\*?$' ]]; then
-      zle vi-fetch-history -n ${num%\*}
+    if num=$(command awk '{print $1; exit}' <<< "$selected" | command grep -o '^[1-9][0-9]*'); then
```
junegunn, line 125: Removing `command`.

```suggestion
    if num=$(awk '{print $1; exit}' <<< "$selected" | grep -o '^[1-9][0-9]*'); then
```

junegunn:
Thanks!

Issue #2967, opened 2022-09-12
fish: `cd -` doesn't work after changing directory using `ALT-C`


psevdaisthisi:
- [x] I have read through the manual page (`man fzf`)
- [x] I have the latest version of fzf
- [x] I have searched through the existing issues

## Info

- OS
    - [x] Linux
    - [x] macOS 
- Shell
    - [x] fish

## Problem / Steps to reproduce

After #2799, under Fish shell, I can no longer switch back to the previous directory using `cd -` after `ALT-C`ing to a directory in the first place. It is working fine under Bash, but not under Fish.

- `fzf --version`: `0.33.0 (e03ac31)`
- `fish --version`: `fish, version 3.5.1`

To confirm that it wasn't my environment screwing things up, I did the following:
```fish
fish --no-config
source /usr/share/fish/vendor_functions.d/fzf_key_bindings.fish
fzf_key_bindings
fzf-cd-widget
# change to some directory ...
cd -
```

junegunn:
I can reproduce the problem with fish only, without fzf, like so.

```sh
builtin cd /tmp
cd -
# Not going back
pwd

builtin cd -
  # cd: The directory '-' does not exist
```

Looks like `builtin cd` of fish does not implement `-` for going back. If fish maintainers don't have a plan to implement it, I'm not sure what we should do.

psevdaisthisi:
May be we could do one of the following:
- Allow a user-defined change-directory command that defaults to the current `builtin cd`.
- Set the current directory in a session variable before changing directories, so we have a target to change back to if we want.

What do you think?
Thank you!

junegunn:
> * Set the current directory in a session variable before changing directories, so we have a target to change back to if we want

With `cd -`?

junegunn:
According to their doc, it looks like `cd -` is not going to be a built-in feature of fish.

> Fish also ships a wrapper function around the builtin cd that understands cd - 

(https://github.com/fish-shell/fish-shell/blob/master/doc_src/cmds/cd.rst#description)

So, I'm just going to remove `builtin` from fish version of `ALT-C`. It was added along with the bash/zsh version, but maybe it's not necessary for fish.

https://github.com/junegunn/fzf/commit/54841248e7a26ffe545220df2a444e11c96ec757

PR #3830, opened 2024-05-31
fix: use builtins for cd and history in fish

### LINKED ISSUE #3826: Use `builtin cd` in fzf Alt+C Key Bindings for Fish Shell
### Checklist

### Output of `fzf --version`

0.52.0 (bcda25a5)

### OS

- [X] Linux
- [ ] macOS
- [ ] Windows
- [ ] Etc.

### Shell

- [ ] bash
- [ ] zsh
- [X] fish

I'm encountering an issue with the `Alt+C` key binding in the Fish shell when using zoxide, which overrides the cd command. When I activate the Alt+C shortcut, it is not calling the builtin cd because of the alias setup by zoxide, leading to a failure in switching directories.

I think changing [this line](https://github.com/junegunn/fzf/blob/124cd70710cb7c94ef2282773494bb0d727598d8/shell/key-bindings.fish#L96) to `builtin cd -- $result` is sufficient to solve this problem.

This modification is aligned with the implementation in the [Bash script](https://github.com/junegunn/fzf/blob/124cd70710cb7c94ef2282773494bb0d727598d8/shell/key-bindings.bash#L54), where builtin cd is explicitly used to avoid similar conflicts.

LangLangBart:
#### description
- fix #3826

In addition to the `cd` command, I deemed `history` to be treated as well.

```fish
# Exemplary fish command history alias
function history
    builtin history --show-time='%F %T '
end
```

```fish
# For a list of all builtins
builtin -n
```

### DIFF
--- a/shell/key-bindings.fish
+++ b/shell/key-bindings.fish
@@ -69,16 +69,16 @@ function fzf_key_bindings
         if type -P perl > /dev/null 2>&1
           set -lx FZF_DEFAULT_OPTS (__fzf_defaults "" "-n2..,.. --scheme=history --bind=ctrl-r:toggle-sort --highlight-line $FZF_CTRL_R_OPTS +m")
           set -lx FZF_DEFAULT_OPTS_FILE ''
-          history -z --reverse | command perl -0 -pe 's/^/$.\t/g; s/\n/\n\t/gm' | eval (__fzfcmd) --tac --read0 --print0 -q '(commandline)' | command perl -pe 's/^\d*\t//' | read -lz result
+          builtin history -z --reverse | command perl -0 -pe 's/^/$.\t/g; s/\n/\n\t/gm' | eval (__fzfcmd) --tac --read0 --print0 -q '(commandline)' | command perl -pe 's/^\d*\t//' | read -lz result
           and commandline -- $result
         else
           set -lx FZF_DEFAULT_OPTS (__fzf_defaults "" "--scheme=history --bind=ctrl-r:toggle-sort --highlight-line $FZF_CTRL_R_OPTS +m")
           set -lx FZF_DEFAULT_OPTS_FILE ''
-          history -z | eval (__fzfcmd) --read0 --print0 -q '(commandline)' | read -lz result
+          builtin history -z | eval (__fzfcmd) --read0 --print0 -q '(commandline)' | read -lz result
           and commandline -- $result
         end
       else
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks!

harkabeeparolus:
I would like to add that I do not appreciate this change. It causes a bug for all vanilla fish users, who do not use zoxide.

### Broken Directory History

If you use standard vanilla fish shell, without any third-party plugins, by default the builtin cd command is [wrapped by a function](https://github.com/fish-shell/fish-shell/blob/master/share/functions/cd.fish) that provides Directory History. This function maintains the `dirprev` and `dirnext` lists, which is what enables [fish shells's Directory History](https://fishshell.com/docs/current/interactive.html#id13) (`cdh` and `dirh` commands, as well as the Alt-Left and Alt-Right hotkeys) to work as intended. When fzf bypasses this function with `builtin cd`, the previous directory is never saved to the Directory History, and this information is lost to the user.

### Solutions

If the choice is between breaking zoxide, or causing slight information loss and annoyance for all fish users, that's not a fun choice. I would prefer to investigate if this bug could be fixed in zoxide, if possible. I personally think Directory History is one of the best interactive features of fish shell, so I strongly prefer to have it working.

A possible workaround for zoxide users is that instead of putting `fzf --fish | source` into their `config.fish`, they could add the code permanently by running `fzf --fish > $__fish_config_dir/conf.d/fzf.fish`, and then they could add the `builtin cd` and `builtin history` fixes themselves. Not ideal, but possible.

junegunn:
@harkabeeparolus Thanks for the report. This one was my oversight.

Initially, the maintainer of zoxide, @ajeetdsouza sent a pull request to use `builtin cd`.
* https://github.com/junegunn/fzf/commit/54841248e7a26ffe545220df2a444e11c96ec757

However, the problem of `builtin cd` was reported in #2967, and I reverted the change in f931e538903a20d7a63162f0f10f58447f1117c8.

Fast forward 2 years, I totally forgot about that and merged this PR.

> I would prefer to investigate if this bug could be fixed in zoxide

That would be the way to go.

harkabeeparolus:
Also, I can't reproduce the bug. I installed the latest version of zoxide, and tried:

```fish
fzf --fish | source
zoxide init fish | source
```

And both `z`, fzf *Alt-C* and Directory History seems to be working properly.

PR #3875, opened 2024-06-18
Remove comment in command substitution in completion.zsh

hexium310:
After 22adb6494f4957d0bcc29a821b738b4e5d5796a5, the Zsh completion using `**` fails.

```sh
$ cat **__fzf_generic_path_completion:23: bad pattern: #
```

The cause of this bug is that a comment is in the command substitution.

### DIFF
--- a/shell/completion.zsh
+++ b/shell/completion.zsh
@@ -157,7 +157,6 @@ __fzf_generic_path_completion() {
       [ -z "$dir" ] && dir='.'
       [ "$dir" != "/" ] && dir="${dir/%\//}"
       matches=$(
-        # Declare and assign separately for older zsh versions.
         export FZF_DEFAULT_OPTS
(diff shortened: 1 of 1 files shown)


LangLangBart:
Thanks for testing it before the release. I had `interactive_comments` enabled, which prevented the error from occurring when I tested it.

Comments within a command substitution inside a function are not automatically removed when parsed, unlike other comments in a function. Therefore, moving the comment outside of the substitution should not trigger any error.

---

**Options:**
1. Remove the comment (preferred)
2. `setopt localoptions interactive_comments`
3. Move the comment outside the command substitution

Option 1 seems safe, and one can use `git blame` to check why the line was written that way.

Option 2 should work, but who knows what kind of pitfalls this might unleash.

Option 3 makes the comment one line separated from the actual affected code line, which looks odd.

junegunn:
Thanks!

PR #3882, opened 2024-06-20
fix(shell): add (?s) modifier to perl command

### LINKED ISSUE #3881: [BUG] Multi-line commands surrounded by parentheses not showing up in history
### Checklist

### Output of `fzf --version`

0.53.0 (b9d15569)

### OS

- [X] Linux
- [ ] macOS
- [ ] Windows
- [ ] Etc.

### Shell

- [ ] bash
- [X] zsh
- [ ] fish

Hello, thank you for this awesome tool!
I have many multi-line commands in my Zsh history in the following format:

```
(
echo do stuff
echo do more stuff
)
```

The surrounding parentheses are to avoid escaping every newline.

After upgrading to v0.53.0 on an Arch install, fzf's history has trouble finding and listing them, only showing the last run command in this format.

A simple way to reproduce:
Run:
```
(
echo Hello
echo World
)
```
Then:
```
ls
```
And then:
```
(
echo 1337
echo h4x
)
```

Hitting Ctrl+r will show `ls` and the third command, but not the first one - even when searching for "World".

I compared `/usr/share/fzf/key-bindings.zsh` with v0.52.1 and it looks like the problem is in the new Perl regex in `fzf-history-widget()`, but I don't know Perl and couldn't find the issue.

Tested both on current Arch package `0.53.0 (c4a9ccd6)` and current master `0.53.0 (b9d15569)` with same result.

LangLangBart:
### description

Close #3881

The syntax is described under: `man perlre | less --pattern "embedded pattern-match modifiers"`

The modifiers are explained under: `man perlre | less --pattern "Modifiers"`

Details for `.` (dot) under: `man perlrecharclass | less --pattern "The dot"`

### DIFF
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -112,7 +112,7 @@ fzf-history-widget() {
   # history lines, is loaded, and that Perl is installed for multi-line output.
   if zmodload -F zsh/parameter p:history 2>/dev/null && (( ${#commands[perl]} )); then
     selected="$(printf '%1$s\t%2$s\000' "${(vk)history[@]}" |
-      perl -0 -ne 'if (!$seen{(/^\s*[0-9]+\**\s+(.*)/, $1)}++) { s/\n/\n\t/gm; print; }' |
+      perl -0 -ne 'if (!$seen{(/^\s*[0-9]+\**\s+((?s).*)/, $1)}++) { s/\n/\n\t/gm; print; }' |
       FZF_DEFAULT_OPTS=$(__fzf_defaults "" "-n2..,.. --scheme=history --bind=ctrl-r:toggle-sort --highlight-line ${FZF_CTRL_R_OPTS-} --query=${(qqq)LBUFFER} +m --read0") \
       FZF_DEFAULT_OPTS_FILE='' $(__fzfcmd))"
   else


shell/key-bindings.zsh
```diff
@@ -112,7 +112,7 @@ fzf-history-widget() {
   # history lines, is loaded, and that Perl is installed for multi-line output.
   if zmodload -F zsh/parameter p:history 2>/dev/null && (( ${#commands[perl]} )); then
     selected="$(printf '%1$s\t%2$s\000' "${(vk)history[@]}" |
-      perl -0 -ne 'if (!$seen{(/^\s*[0-9]+\**\s+(.*)/, $1)}++) { s/\n/\n\t/gm; print; }' |
+      perl -0 -ne 'if (!$seen{(/^\s*[0-9]+\**\s+((?s).*)/, $1)}++) { s/\n/\n\t/gm; print; }' |
```
junegunn, line 115: Thanks, I realized that I had a vague understanding of these flags.

Can we put `s` flag globally? It would result in a shorter code. Also, it looks like we can drop `m` from the latter regex.

```suggestion
      perl -0 -ne 'if (!$seen{(/^\s*[0-9]+\**\s+(.*)/s, $1)}++) { s/\n/\n\t/g; print; }' |
```

LangLangBart, line 115:
```suggestion
      perl -0 -ne 'if (!$seen{(/^\s*[0-9]+\**\t(.*)/s, $1)}++) { s/\n/\n\t/g; print; }' |
```

I agree with using `//s` and removing `//m`.

---

Instead of a `\s+`, replace it with a `\t`, which we know must follow a number at least once.

---

One might have also considered replacing `[0-9]` with `\d` to save another three characters, but after reading the `man` page, I would keep the current status quo.

```sh
man perlrecharclass | less --pattern "Digits"
       Digits

       "\d" matches a single character considered to be a decimal digit.  If
       the "/a" regular expression modifier is in effect, it matches [0-9].
       Otherwise, it matches anything that is matched by "\p{Digit}", which
       includes [0-9]. …

       What this means is that unless the "/a" modifier is in effect "\d" not
       only matches the digits '0' - '9', but also Arabic, Devanagari, and
       digits from other languages.  This may cause some confusion, and some
       security issues.

       Some digits that "\d" matches look like some of the [0-9] ones, but
       have different values.  For example, BENGALI DIGIT FOUR (U+09EA) looks
       very much like an ASCII DIGIT EIGHT (U+0038).  An application that is
       expecting only the ASCII digits might be misled, or if the match is
       "\d+", the matched string might contain a mixture of digits from
       different writing systems that look like they signify a number
       different than they actually do.  "num()" in Unicode::UCD can be used
       to safely calculate the value, returning "undef" if the input string
       contains such a mixture.

       What "\p{Digit}" means (and hence "\d" except under the "/a" modifier)
       is "\p{General_Category=Decimal_Number}", or synonymously,
       "\p{General_Category=Digit}".  Starting with Unicode version 4.1, this
       is the same set of characters matched by "\p{Numeric_Type=Decimal}".
       But Unicode also has a different property with a similar name,
       "\p{Numeric_Type=Digit}", which matches a completely different set of
       characters.  These characters are things such as "CIRCLED DIGIT ONE" or
       subscripts, or are from writing systems that lack all ten digits.
       …
```

shell/key-bindings.zsh
```diff
@@ -112,7 +112,7 @@ fzf-history-widget() {
   # history lines, is loaded, and that Perl is installed for multi-line output.
   if zmodload -F zsh/parameter p:history 2>/dev/null && (( ${#commands[perl]} )); then
     selected="$(printf '%1$s\t%2$s\000' "${(vk)history[@]}" |
```
LangLangBart, line 114:
```suggestion
    selected="$(printf '%s\t%s\000' "${(@kv)history}" |
```

The numbers in the format part of the `printf` command were added by me, but they are superfluous.

```sh
man printf
…
n$          Allows reordering of the output according to argument.
```

As for switching `vk` to `kv`, it does not affect the outcome, but it provides a more logical order for the reader, where 'key' precedes 'value'. And moving the `@` inside the `()` allows us to remove the `[]` saving two characters.

```sh
man zshexpn
…
       Parameter  expansions  can  involve flags, as in `${(@kv)aliases}'
…
       @      In double quotes, array elements are put  into  separate  words.
              E.g.,   `"${(@)foo}"'   is   equivalent   to  `"${foo[@]}"'  and
              `"${(@)foo[1,2]}"' is the same as `"$foo[1]"  "$foo[2]"'.   This
              is  distinct  from field splitting by the f, s or z flags, which
              still applies within each array element.
```

shell/key-bindings.zsh
junegunn: Sounds good to me.

shell/key-bindings.zsh
junegunn: Thanks for the explanation.

junegunn: One question, though. Is this supported by the old versions of zsh?

LangLangBart: I checked with `5.0.2`.

```sh
❯ ~/.local/zsh-test/bin/zsh-5.0.2 -f
Paria% echo test
test
Paria% seq 2
1
2
Paria% printf '%s\t%s\n' "${(@kv)history}"
2	seq 2
1	echo test
Paria% 
```

LangLangBart: But checking the source code, this seems to be a very old feature, reaching back into the nineties.

junegunn: Thanks!

junegunn:
Merged, thanks!

PR #3885, opened 2024-06-22
[man] Escape hyphens to prevent conversion to Unicode hyphens

akinomyoga:
ASCII hyphens (U+002D HYPHEN-MINUS) in the option names (e.g. `-x` and `--extended`) and the code examples in the man page source should be escaped as `\-` (e.g. `\-x` and `\-\-extended`) to prevent them being converted to Unicode hyphens in some environments.

For example, in openSUSE Tumbleweed, the raw ASCII hyphens in the man-page sources are configured to become the Unicode hyphen (U+2010 HYPHEN).  This makes it impossible to search the option names on the `fzf` man page by e.g. `/--extended`<kbd>RET</kbd>.  A problem also arises in copying and pasting option names and code examples from the man page.  It appears to be a normal ASCII hyphen by appearance (in typical terminal fonts) but is not recognized as the ASCII hyphen when specified in command-line arguments of the `fzf` command.

An explanation can also be found in the corresponding section in `man man-pages`:

```
   Generating optimal glyphs
       Where a real minus character is required (e.g., for numbers such as -1,
       for man page cross references such as utf-8(7), or when writing options
       that  have a leading dash, such as in ls -l), use the following form in
       the man page source:

           \-

       This guideline applies also to code examples.

       The use of real minus signs serves the following purposes:

       o  To provide better renderings on various  targets  other  than  ASCII
          terminals, notably in PDF and on Unicode/UTF-8-capable terminals.

       o  To generate glyphs that when copied from rendered pages will produce
          real minus signs when pasted into a terminal.
```

### DIFF
--- a/man/man1/fzf.1
+++ b/man/man1/fzf.1
@@ -182,7 +182,7 @@ Comma-separated list of sort criteria to apply when the scores are tied.
 .SS Interface
 .TP
 .B "\-m, \-\-multi"
-Enable multi-select with tab/shift-tab. It optionally takes an integer argument
+Enable multi-select with tab/shift\-tab. It optionally takes an integer argument
 which denotes the maximum number of items that can be selected.
 .TP
 .B "+m, \-\-no\-multi"
--- a/man/man1/fzf.1
+++ b/man/man1/fzf.1
@@ -1656,7 +1656,7 @@ e.g.
      # The default properties given by `\-\-preview\-window` are inherited, so an empty string in the list is interpreted as the default
      fzf \-\-preview 'cat {}' \-\-preview\-window 'right,40%,border\-left' \-\-bind 'ctrl\-/:change\-preview\-window(70%|down,border\-top|hidden|)'
 
-     # This is equivalent to toggle-preview action
+     # This is equivalent to toggle\-preview action
      fzf \-\-preview 'cat {}' \-\-bind 'ctrl\-/:change\-preview\-window(hidden|)'
 
 .SH AUTHOR
--- a/man/man1/fzf.1
+++ b/man/man1/fzf.1
@@ -101,8 +101,8 @@ Fuzzy matching algorithm (default: v2)
 .BI "\-n, \-\-nth=" "N[,..]"
 Comma-separated list of field index expressions for limiting search scope.
 See \fBFIELD INDEX EXPRESSION\fR for the details. When you use this option with
-\fB--with-nth\fR, the field index expressions are calculated against the
-transformed lines (unlike in \fB--preview\fR where fields are extracted from
+\fB\-\-with\-nth\fR, the field index expressions are calculated against the
+transformed lines (unlike in \fB\-\-preview\fR where fields are extracted from
 the original lines) because fzf doesn't allow searching against the hidden
 fields.
 .TP
--- a/man/man1/fzf.1
+++ b/man/man1/fzf.1
@@ -38,27 +38,27 @@ with omitted characters and still get the results you want.
 .SH OPTIONS
 .SS Note
 .TP
-Most long options have the opposite version with \fB--no-\fR prefix.
+Most long options have the opposite version with \fB\-\-no\-\fR prefix.
 
 .SS Search mode
 .TP
-.B "-x, --extended"
+.B "\-x, \-\-extended"
 Extended-search mode. Since 0.10.9, this is enabled by default. You can disable
-it with \fB+x\fR or \fB--no-extended\fR.
+it with \fB+x\fR or \fB\-\-no\-extended\fR.
 .TP
-.B "-e, --exact"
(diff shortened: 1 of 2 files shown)


junegunn:
Thanks for the patch.

This is painful, though. So this is the only way?

man/man1/fzf.1
```diff
@@ -98,62 +98,62 @@ Fuzzy matching algorithm (default: v2)
 .br
 
 .TP
-.BI "-n, --nth=" "N[,..]"
+.BI "\-n, \-\-nth=" "N[,..]"
 Comma-separated list of field index expressions for limiting search scope.
 See \fBFIELD INDEX EXPRESSION\fR for the details. When you use this option with
 \fB--with-nth\fR, the field index expressions are calculated against the
 transformed lines (unlike in \fB--preview\fR where fields are extracted from
```
junegunn, line 105:
```suggestion
\fB\-\-with\-nth\fR, the field index expressions are calculated against the
transformed lines (unlike in \fB\-\-preview\fR where fields are extracted from
```

man/man1/fzf.1
```diff
@@ -1583,81 +1583,81 @@ expressions as \fBexecute(...)\fR.
...
-     fzf --preview 'cat {}' --bind 'ctrl-/:change-preview-window(right,70%|down,40%,border-horizontal|hidden|right)'
+     # Rotate through the options using CTRL\-/
+     fzf \-\-preview 'cat {}' \-\-bind 'ctrl\-/:change\-preview\-window(right,70%|down,40%,border\-horizontal|hidden|right)'
 
-     # The default properties given by `--preview-window` are inherited, so an empty string in the list is interpreted as the default
-     fzf --preview 'cat {}' --preview-window 'right,40%,border-left' --bind 'ctrl-/:change-preview-window(70%|down,border-top|hidden|)'
+     # The default properties given by `\-\-preview\-window` are inherited, so an empty string in the list is interpreted as the default
+     fzf \-\-preview 'cat {}' \-\-preview\-window 'right,40%,border\-left' \-\-bind 'ctrl\-/:change\-preview\-window(70%|down,border\-top|hidden|)'
 
      # This is equivalent to toggle-preview action
```
junegunn, line 1659:
```suggestion
     # This is equivalent to toggle\-preview action
```

man/man1/fzf.1
akinomyoga: Ah, thanks. I think i need to double-check my changes again.

man/man1/fzf.1
akinomyoga: Thanks!

akinomyoga:
I added another fix: 49356cb6fc726145357235b9261e3c6566c4eef9.

akinomyoga:
> This is painful, though. So this is the only way?

Yeah, I agree, but as far as I know, there is no other way. Usually the option names are escaped like `\-\-help` in man-page sources.

man/man1/fzf.1
```diff
@@ -1122,19 +1122,19 @@ The following variables are additionally exported to the preview commands.
...
-Unless specified otherwise, fzf will start in "extended-search mode". In this
+Unless specified otherwise, fzf will start in "extended\-search mode". In this
 mode, you can specify multiple patterns delimited by spaces, such as: \fB'wild
 ^music .mp3$ sbtrkt !rmx\fR
 
 You can prepend a backslash to a space (\fB\\ \fR) to match a literal space
 character.
 
-.SS Exact-match (quoted)
+.SS Exact\-match (quoted)
```
akinomyoga, line 1132: I'm now wondering whether we should escape these hyphens. Are these just the names in the explanations and not the keywords that are programmatically processed? Then, maybe we should allow the man page system to make those hyphens the real Unicode hyphens.

junegunn: I agree. Let's keep those hyphens as they are.

junegunn (approved):

junegunn:
Merged, thanks!

akinomyoga:
Thanks!

man/man1/fzf.1
akinomyoga: Actually, some were escaped as `exact\-match` and some were left as `exact-match` in the PR branch, so I wanted to normalize it to either one. If you prefer `exact\-match`, you can use

[0001-man-Escape-exact-match-and-anchored-match.patch.txt](https://github.com/user-attachments/files/15948505/0001-man-Escape-exact-match-and-anchored-match.patch.txt)

Or, if you prefer `exact-match`, you can use

[0001-man-Revert-exact-match-and-anchored-match.patch.txt](https://github.com/user-attachments/files/15948513/0001-man-Revert-exact-match-and-anchored-match.patch.txt)

You can directly edit them by yourself, or if needed I can submit a followup PR.

PR #3887, opened 2024-06-22
Add --wrap option and 'toggle-wrap' action

### LINKED ISSUE #461: Can you wrap lines so that they are fully displayed?
My most common usecase for fzf is search through my history, which often contains very long commands. When the terminal window is too narrow for the full command line, it is trimmed. Is it possible to wrap instead? ideally, I'd be able to specify the maximum number of lines that are used for the wrapping. This way I can make sure that extremely long items don't take all the screen, but also that I can see more of every item.

Thanks!

### LINKED ISSUE #577: Long commands aren't horizontally scrollable

### LINKED ISSUE #2236: Missing scroll left/right or wrap results

## Info

- OS
    - [X] Linux
- Shell
    - [X] zsh
    
When a result list item is too long there is no way to scroll the display horizontally as if in a pager. 
Instead the long line is shortened with two dots shown at the truncated end.
I've no --no-hscroll in FZF_DEFAULT_OPTS

An option to wrap long lines can also help.

### LINKED ISSUE #3619: How to enable wrap in normal window not preview window?
- [ x] I have read through the manual page (`man fzf`)
- [ x] I have the latest version of fzf
- [ x] I have searched through the existing issues

## Info

- OS
    - [x ] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [ x] bash
    - [ ] zsh
    - [ ] fish

Hello,

I'm Mr. Bacgi, a very friendly person.

How do I enable word wrap in the normal window? This is not the preview window, I think, with your terminology.

I just have a simple txt file, catting it and piping to fzf. But then all the long ilnes end with ... which is not ideal for my purposes because i need to see the end of the line.

--wrap doesn't work
--preview-window wrap doesn't work

Any advice?

junegunn:
With the multi-line support, it wasn't too hard to implement line wrap of long items.
Should we apply this by default to CTRL-R bindings?

Needs more testing.

Close #3619
Close #2236
Close #577
Close #461


LangLangBart:
> Needs more testing.

I noticed some gaps in the `border` when interacting with the listed values. See the `GIF` below for more clarity.

Terminal: iTerm2 
Font: Monaco (tested other fonts, but issue remained)



```sh
fc -rl 1 | FZF_DEFAULT_OPTS="" fzf \
  --tac \
  --wrap \
  --preview 'bat -pl zsh --color=always {f2..}'
```


> Should we apply this by default to CTRL-R bindings?

Some might prefer it, but on my small terminal, I would probably disable it again. I find the left image below more appealing.

junegunn:
Thanks for the input, really appreciated.

> Some might prefer it, but on my small terminal, I would probably disable it again. I find the left image below more appealing.

Agreed. It doesn't look quite right. Maybe some extra indentation would help, but I'm not sure.

junegunn:
> I noticed some gaps in the `border` when interacting with the listed values.

This should be fixed now.



> Maybe some extra indentation would help, but I'm not sure.

Now we can add a tab indentation using `--wrap-sign`.

```sh
export FZF_CTRL_R_OPTS=$'--wrap --wrap-sign "\t↳ "'
```

This does help a little, but I can't really say I prefer it over `--no-wrap`. Enabling wrapping via `toggle-wrap` only when necessary seems like the best option.

```sh
export FZF_CTRL_R_OPTS=$'--bind ctrl-/:toggle-wrap --wrap-sign "\t↳ "'
```

LangLangBart:
Everything works smoothly. Did you consider adding a new label when `--wrap` is enabled, similar to `--track` ?

It's just something I noticed when testing the command below:

```sh
fzf --wrap --track --multi --bind 'ctrl-s:toggle-sort'
```


Might actually not be needed; users can see it in the list when the lines are long enough.

junegunn:
Thanks for testing it!

> Did you consider adding a new label when `--wrap` is enabled

No, I hadn't thought about it. As you mentioned, `--wrap` is clearly visible, unlike "track" or "sort", so I don't think it's necessary.

PR #3893, opened 2024-06-24
fix zsh ${(@)history} syntax does not work with ksh_arrays

lincheney:
When `setopt ksh_arrays` in zsh, `${(@kv)history}` gives only the first entry rather than all.

This is breaking the zsh control-r history widget to show only the latest history entry rather than all (which isn't very useful).

I have `setopt ksh_arrays` set in my zshrc.

I believe this is caused by https://github.com/junegunn/fzf/pull/3882/files#diff-c1d0c74baf2323d5dc27290b500868528b9a80be6a0844e998eefb2f9f6ab968R114
specifically the change from `${(kv)history[@]}` to `${(@kv)history}` does not seem to work when you have `setopt ksh_arrays` (it gives back only the one latest history entry).

This fixes it for me, but a potentially better fix could be to add something like `emulate -LR zsh`  at the start of all widgets

### DIFF
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -111,7 +111,7 @@ fzf-history-widget() {
   # Ensure the associative history array, which maps event numbers to the full
   # history lines, is loaded, and that Perl is installed for multi-line output.
   if zmodload -F zsh/parameter p:history 2>/dev/null && (( ${#commands[perl]} )); then
-    selected="$(printf '%s\t%s\000' "${(@kv)history}" |
+    selected="$(printf '%s\t%s\000' "${(kv)history[@]}" |
       perl -0 -ne 'if (!$seen{(/^\s*[0-9]+\**\t(.*)/s, $1)}++) { s/\n/\n\t/g; print; }' |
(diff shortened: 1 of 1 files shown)


LangLangBart:
Thanks for the fix. 

```sh
man zshexpn
         k      If name refers to an  associative  array,  substitute  the  keys
                (element  names)  rather  than the values of the elements.  Used
                with subscripts (including ordinary arrays),  force  indices  or
                keys to be substituted even if the subscript form refers to val-
                ues.  However, this flag may  not  be  combined  with  subscript
                ranges.   With  the KSH_ARRAYS option a subscript `[*]' or `[@]'
                is needed to operate on the whole array, as usual.
```

> something like `emulate -LR zsh` at the start of all widgets

Seems to have been tried in the past[^1] but reverted. The commenter on the commit probably deleted their account, as their comment on the change of this commit is no longer visible.

[^1]: [[zsh] emulate -L zsh to avoid issues with incompatible options · junegunn/fzf@3a6af27 · GitHub](https://github.com/junegunn/fzf/commit/3a6af27586c65bb9243c5af7c43a40619503db52)

junegunn:
Thanks!

Issue #3859, opened 2024-06-13
Missing directory slashes when completion chosen


timabell:
### Checklist

- [X] I have read through the manual page (`man fzf`)
- [X] I have searched through the existing issues
- [ ] For bug reports, I have checked if the bug is reproducible in the latest version of fzf

### Output of `fzf --version`

0.53.0 (c4a9ccd)

### OS

- [ ] Linux
- [ ] macOS
- [X] Windows
- [ ] Etc.

### Shell

- [ ] bash
- [X] zsh
- [ ] fish

### Problem / Steps to reproduce

Thanks for the fab tool.

For a new windows install I'm working on I am getting the backslashes that separate directories all go missing when I hit enter to choose an entry which then means the path is corrupted.

I have done a load of searching and can't find any mention of this being an issue anywhere or what it might be. I don't know enough about how fzf works to guess further so any ideas welcome.

Begin a completion with `file **<tab>`:
![image](https://github.com/junegunn/fzf/assets/19378/7f2235c6-44c9-4c30-94a9-935495e2ee65)

Hit `enter` to choose an item in a subfolder and all the slashes are missing:
![image](https://github.com/junegunn/fzf/assets/19378/de10d73a-7ae0-472f-9b1b-bb6602cff5df)

Curiously if I call `fzf` directly and then search it works fine:

![image](https://github.com/junegunn/fzf/assets/19378/76c6ed6c-7f3a-4847-a690-76f8c50f1780)

![image](https://github.com/junegunn/fzf/assets/19378/82e5e469-66c0-40bd-8687-3dac35fcc93d)

This is my dotmatrix config for the windows machine (windows branch) at time of writing https://github.com/timabell/dotmatrix/blob/0c0de5b7107de4a43c22f1f6c727b2537ffc1056/.zshrc#L107

junegunn:
Is this WSL? Looks like you're running an fzf binary for Windows that uses backslash as the file separator. How did you install fzf?

junegunn:
You should be using Linux binary on WSL. It works for me fine.

<img width="755" alt="image" src="https://github.com/junegunn/fzf/assets/700826/71fbb9ff-b4cf-4b07-89fa-394685f8421f">

junegunn:
Please comment if you still have the problem with a linux binary.

timabell:
Hi, thanks for responding, this is not WSL, this is zsh installed directly on windows, being used in windows terminal.

Yes this is still an issue.

[...]

PR #3907, opened 2024-07-03
Update charlievieth/fastwalk to use forward-slashes on WSL and MSYS

charlievieth:
This commit changes FZF to enforce that all paths are joined with forward-slashes when running on WSL (Windows Subsystem for Linux) even when the FZF binary was compiled for Windows.

Update: github.com/charlievieth/fastwalk to v1.0.5
Fixes:  https://github.com/junegunn/fzf/issues/3859

### DIFF
--- a/src/reader.go
+++ b/src/reader.go
@@ -233,6 +233,20 @@ func isSymlinkToDir(path string, de os.DirEntry) bool {
 	return false
 }
 
+func trimPath(path string) string {
+	if len(path) == 0 {
+		return "."
+	}
+
+	bytes := stringBytes(path)
+
+	for len(bytes) > 1 && bytes[0] == '.' && (bytes[1] == '/' || bytes[1] == '\\') {
+		bytes = bytes[2:]
+	}
+
+	return byteString(bytes)
+}
+
 func (r *Reader) readFiles(root string, opts walkerOpts, ignores []string) bool {
 	r.killed = false
 	conf := fastwalk.Config{
@@ -244,7 +258,7 @@ func (r *Reader) readFiles(root string, opts walkerOpts, ignores []string) bool 
 		if err != nil {
 			return nil
 		}
-		path = filepath.Clean(path)
+		path = trimPath(path)
 		if path != "." {
 			isDir := de.IsDir()
 			if isDir || opts.follow && isSymlinkToDir(path, de) {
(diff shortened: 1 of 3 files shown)


charlievieth:
**NOTE:** This is just a draft and is incompatible with the `-walker-path-sep` option since (it only allows the option to force forward slashes, but does auto-detect when they're needed on WSL see: [DefaultToSlash](https://pkg.go.dev/github.com/charlievieth/fastwalk#DefaultToSlash)).

junegunn:
Thanks, I wish we could avoid adding `--walker-path-sep`.

Defaulting to `/` on WSL sounds great. But we may also do that on MSYS2.

Related: https://github.com/sharkdp/fd/pull/730

junegunn:
> incompatible with the `-walker-path-sep` option since

With this and #3909, we can safely remove `--walker-path-sep`.

src/reader.go
```diff
@@ -235,7 +235,11 @@ func isSymlinkToDir(path string, de os.DirEntry) bool {
 
 func (r *Reader) readFiles(root string, opts walkerOpts, ignores []string) bool {
 	r.killed = false
-	conf := fastwalk.Config{Follow: opts.follow}
+	conf := fastwalk.Config{
+		Follow: opts.follow,
+		// Use forward slashes when running a Windows binary under WSL.
```
junegunn, line 240:
```suggestion
		// Use forward slashes when running a Windows binary under WSL or MSYS
```

junegunn:
So I removed `--walker-path-sep`, updated the PR branch, applied the latest version of fastwalk (`go get github.com/charlievieth/fastwalk@master`), tested it on MSYS2, but it wasn't working as expected.

Turned out that `filepath.Clean(path)` is replacing `/` back to `\` 😢 

https://github.com/charlievieth/fzf/blob/fe9dc4c75617007ae15c205db8516fca212a6e30/src/reader.go#L247

junegunn:
We probably don't need the whole `filepath.Clean(path)`. We should be fine, even if we just remove the `./` or `.\`.

junegunn:
9b30ec559c908b83142b8f6d5ffad5524df427b3 removes `filepath.Clean`. It will not prefectly handle `fzf --walker-root ./././../././`, but I don't care.

PR #3909, opened 2024-07-04
[zsh] Fix backslash escaping

### LINKED ISSUE #3859: Missing directory slashes when completion chosen

junegunn:
Fix #3859

To test:

```zsh
FZF_CTRL_T_COMMAND="echo -E 'foo\bar\baz'; echo -E 'hello\world'"

_fzf_compgen_path() {
  eval $FZF_CTRL_T_COMMAND
}

source shell/key-bindings.zsh
source shell/completion.zsh
```

### DIFF
--- a/shell/completion.zsh
+++ b/shell/completion.zsh
@@ -171,9 +171,9 @@ __fzf_generic_path_completion() {
             rest=${FZF_COMPLETION_PATH_OPTS-}
           fi
           __fzf_comprun "$cmd" ${(Q)${(Z+n+)fzf_opts}} -q "$leftover" --walker "$walker" --walker-root="$dir" ${(Q)${(Z+n+)rest}} < /dev/tty
-        fi | while read item; do
+        fi | while read -r item; do
           item="${item%$suffix}$suffix"
-          echo -n "${(q)item} "
+          echo -n -E "${(q)item} "
         done
       )
       matches=${matches% }
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -52,8 +52,8 @@ __fzf_select() {
   local item
   FZF_DEFAULT_COMMAND=${FZF_CTRL_T_COMMAND:-} \
   FZF_DEFAULT_OPTS=$(__fzf_defaults "--reverse --walker=file,dir,follow,hidden --scheme=path" "${FZF_CTRL_T_OPTS-} -m") \
-  FZF_DEFAULT_OPTS_FILE='' $(__fzfcmd) "$@" < /dev/tty | while read item; do
-    echo -n "${(q)item} "
+  FZF_DEFAULT_OPTS_FILE='' $(__fzfcmd) "$@" < /dev/tty | while read -r item; do
+    echo -n -E "${(q)item} "
   done
   local ret=$?
   echo


LangLangBart:
I guess you have already found the desired formula.

---

`print -nr -- …` should have worked as well.


```diff
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -53,6 +53,6 @@ __fzf_select() {
   FZF_DEFAULT_COMMAND=${FZF_CTRL_T_COMMAND:-} \
   FZF_DEFAULT_OPTS=$(__fzf_defaults "--reverse --walker=file,dir,follow,hidden --scheme=path" "${FZF_CTRL_T_OPTS-} -m") \
-  FZF_DEFAULT_OPTS_FILE='' $(__fzfcmd) "$@" < /dev/tty | while read item; do
-    echo -n "${(q)item} "
+  FZF_DEFAULT_OPTS_FILE='' $(__fzfcmd) "$@" < /dev/tty | while read -r item; do
+    print -nr -- "${(q)item} "
   done
   local ret=$?
```

or `bsd_echo`

```diff
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -49,9 +49,9 @@ __fzf_defaults() {
 # CTRL-T - Paste the selected file path(s) into the command line
 __fzf_select() {
-  setopt localoptions pipefail no_aliases 2> /dev/null
+  setopt localoptions pipefail no_aliases bsd_echo 2> /dev/null
   local item
   FZF_DEFAULT_COMMAND=${FZF_CTRL_T_COMMAND:-} \
   FZF_DEFAULT_OPTS=$(__fzf_defaults "--reverse --walker=file,dir,follow,hidden --scheme=path" "${FZF_CTRL_T_OPTS-} -m") \
-  FZF_DEFAULT_OPTS_FILE='' $(__fzfcmd) "$@" < /dev/tty | while read item; do
+  FZF_DEFAULT_OPTS_FILE='' $(__fzfcmd) "$@" < /dev/tty | while read -r item; do
     echo -n "${(q)item} "
   done
```

junegunn:
Thanks. Do they have pros and cons or are they more or less the same and it doesn't matter which one we choose?

LangLangBart:
> Thanks. Do they have pros and cons or are they more or less the same and it doesn't matter which one we choose?

Difference between `echo`   and `print`.

> 3.2.1: Builtins for printing
> The commands `echo' and `print' are shell builtins; they just show what you typed, after the shell has removed all the quoting. The difference between the two is really historical: `echo' came first, and only handled a few simple options; ksh provided `print', which had more complex options and so became a different command. The difference remains between the two commands in zsh; if you want wacky effects, you should look to print. Note that there is usually also an external command called echo, which may not be identical to zsh's; there is no standard external command called print, but if someone has installed one on your system, the chances are it sends something to the printer, not the screen.
>
> Source: [A User's Guide to the Z-Shell](https://zsh.sourceforge.io/Guide/zshguide03.html#l32)

`print` comes with more flags, making it useful for special cases, but using `echo -nE` or `print -nr --` is indifferent here.

PR #3911, opened 2024-07-06
0.54.0

junegunn:


junegunn:
> [Bind CTRL-/ to toggle-wrap by default](https://github.com/junegunn/fzf/pull/3911/commits/53d4b90ec04851293ea3d12a53774e6c9a017f37)

Wondering if this is a good idea. CTRL-/ is not a standard and not all terminal emulators support it (e.g. doesn't work in mac terminal.app). Also, CTRL-/ is used to [toggle preview window in fzf.vim](https://github.com/junegunn/fzf.vim/blob/279e1ec068f526e985ee7e3f62a71f083bbe0196/autoload/fzf/vim.vim#L175).

Maybe ALT-/ is a better choice.

LangLangBart:
> doesn't work in mac terminal.app

confirmed, works only in my vscode terminal

> Maybe ALT-/ is a better choice.

works, when I switch to `meta` with <kbd>⌘ Command</kbd> + <kbd>⌥ Option</kbd> + <kbd>O</kbd>

https://github.com/junegunn/fzf/assets/92653266/244d0034-054a-4e21-9bcc-c34ba581a70d

junegunn:
Let's bind both keys to `toggle-wrap`. It wouldn't hurt.

PR #3936, opened 2024-07-21
CTRL-R: Show timestamp and syntax highlighted command

### LINKED ISSUE #1049: Feature request: Add an option to display date-time of fzf command history
- Category
    - [x] fzf binary
    - [ ] fzf-tmux script
    - [ ] Key bindings
    - [ ] Completion
    - [ ] Vim
    - [ ] Neovim
    - [ ] Etc.
- OS
    - [x] Linux
    - [x] Mac OS X
    - [x] Windows
    - [x] Etc.
- Shell
    - [x] bash
    - [x] zsh
    - [x] fish

I would love to view the date and time in the fzf command history list, if possible, much like
```
fc -li
```
does

### LINKED ISSUE #3836: [Feature Request] History syntax highlighting
### Checklist

### Output of `fzf --version`

0.52.1 (6432f00)

### OS

- [X] Linux
- [ ] macOS
- [ ] Windows
- [ ] Etc.

### Shell

- [ ] bash
- [X] zsh
- [ ] fish

I just started to use fzf and i love it. But i've got one problem about the history backsearch with key bidings.
I was using https://github.com/zdharma-continuum/history-search-multi-word to have a similar features of fuzzy search with Ctrl-R and it has syntax highlighting wich make it visualy more attractive.
So i was wondering if it would be possible to get syntax highlighting like zdharma-continuum/history-search-multi-word for history backsearch with fzf ?
Or add an option to disable only the Ctrl-R key biding so i'm able to use fzf with Ctrl-T and Alt-C and zdharma-continuum/history-search-multi-word with Ctrl-R

junegunn:
Close #1049
Close #3836

This is a proof-of-concept implementation of enhanced CTRL-R binding for bash.

## Screenshot

image

### HISTTIMEFORMAT performance

```sh
$ HISTTIMEFORMAT='%F %T: '
$ time history | wc -l
   18295

real    0m1.132s
user    0m0.190s
sys     0m0.975s

$ time history > /dev/null

real    0m0.034s
user    0m0.024s
sys     0m0.009s
```

For some reason, `$HISTTIMEFORMAT` makes piping `history` extremely slow. Strangely, there is no slowdown when redirecting the output to another file.

So we can work around this performance problem by writing it to a temp file and then reading it. But...

### Syntax-highlighting of each command

We need to syntax-highlight each command, however, it's not viable to execute an external process such as `bat` for each command.

```sh
# This is extremely slow
while read line; do
  bat --style=plain --no-pager --language=bash <<< "$line"
done < ~/.bash_history
```

It's much faster to pass the whole file to `bat` at once,

```sh
bat --style=plain --no-pager --language=bash ~/.bash_history
```

However, syntax-highlighting quickly breaks down due to malformed commands in history.

### Complexity of the code

The Perl and Awk scripts for de-duplicating the list and handling multi-line commands is already quite complex and not easy to read and maintain. Adding timestamp and syntax-highlighting to it would bring too much complexity.

## Solution

I used Ruby script to implement the enhanced CTRL-R binding. [rouge](https://github.com/rouge-ruby/rouge) gem was used for syntax-highlighting. The code isn't short, but I'd argue that it's still more readable.

## Drawbacks

* Unavailability of Ruby on some systems.
* Yet another dependency: `rouge` gem required for syntax-highlighting.
* Performance. Slow to syntax-highlight many commands.

## Alternatives

* Should we write an external tool for formatting the history file? It would yield better performance, but it's a hard dependency users have to install.
* Embed the functionality in fzf. This avoids an extra dependency, but this is a bad design choice.

## TODO

* How about other shells?

### DIFF
--- a/shell/key-bindings.bash
+++ b/shell/key-bindings.bash
@@ -54,7 +54,45 @@ __fzf_cd__() {
   ) && printf 'builtin cd -- %q' "$(builtin unset CDPATH && builtin cd -- "$dir" && builtin pwd)"
 }
 
-if command -v perl > /dev/null; then
+if command -v ruby > /dev/null; then
+  __fzf_history__() {
+    local n output
+    builtin history -w /tmp/fzf-bash-history
+    output=$(
+      ruby -e '
+        fmt = begin
+          require "rouge"
+          formatter = Rouge::Formatters::Terminal256.new(Rouge::Themes::Monokai.new)
+          lexer = Rouge::Lexers::Shell.new
+          lambda { |c| formatter.format(lexer.lex(c)) }
+        rescue LoadError
+          lambda { |c| c }
+        end
+
+        h = {}
+        i = 0
+        File.read("/tmp/fzf-bash-history").scan(/^#([0-9]+)$\n(.*?)\n(?=^#[0-9]+$|\z)/m) do |t, c|
+          next if c.empty?
+          h.delete(c)
+          h[c] = [i += 1, t]
+        end
+        h.to_a.reverse.each do |c, it|
+          i, t = it
+          print "\x1b[33m#{i}\t\x1b[32m#{Time.at(t.to_i).strftime(%[%F %T])}\x1b[m  "
+          print fmt[c.gsub("\n", "\n\t")]
+          print "\0"
+        end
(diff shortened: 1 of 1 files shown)


hedgieinsocks:
Hi, is there any (option or hack?) way to get syntax highlighting only for the cursor line? I would be perfectly content with something like `--syntax-highlight-matched-line=bash`

FilipeBento:
What I do is have a preview that highlights the line with bat.

--preview='echo -n {2..} | bat -l bash -p --color=always --theme 1337'

it's not perfect, but helps when I need the highlighting. Highlighting also affects readability, since fzf's search match colors are harder to visualize.

I also have the following set, but it doesn't work as expected :)
export FZF_CTRL_R_COMMAND="history | bat -l bash -p --color=always"

PR #3942, opened 2024-07-25
chore: keep it real by adding Cons

samjcombs:
While its always nice to have Pros, the masses love when products keep it real and own their Cons.


junegunn:
Thanks :)

PR #3951, opened 2024-08-04
Set console to UTF-8 encoding on Windows

### LINKED ISSUE #3799: Can't input non ascii characters
### Checklist

### Output of `fzf --version`

0.52.1 (6432f00)

### OS

- [ ] Linux
- [ ] macOS
- [X] Windows
- [ ] Etc.

### Shell

- [X] bash
- [ ] zsh
- [ ] fish

Specifically I use bash from git for windows.
OS: Win 10
Terminal:
1. Windows Terminal (1.20.11271.0)
2. mintty 3.7.0 (shipped with git for windows)

When I hit CTRL+T and try to input greek characters they appear in the prompt either transliterated to their equivalent ascii (eg `α` becomes `a`) or printed as the `?` character.
This happens with both terminals.

Both terminals are able to print greek characters (eg `ls` a folder with greek filenames) and are able to take greek input in their prompts. So the issue is with fzf's prompt.

kelleyma49:
- Should resolve #3799
- Here's an example with my change: 
utf8example

### DIFF
--- a/src/tui/light_windows.go
+++ b/src/tui/light_windows.go
@@ -20,6 +20,25 @@ var (
 	consoleFlagsOutput = uint32(windows.ENABLE_VIRTUAL_TERMINAL_PROCESSING | windows.ENABLE_PROCESSED_OUTPUT | windows.DISABLE_NEWLINE_AUTO_RETURN)
 )
 
+func SetConsoleUTF8() error {
+	var err error
+	var result uintptr = 0
+
+	const utf8CodePage = 65001
+
+	k32 := syscall.MustLoadDLL("kernel32.dll")
+	result, _, err = k32.MustFindProc("SetConsoleCP").Call(utf8CodePage)
+	if result == 0 {
+		return err
+	}
+	result, _, err = k32.MustFindProc("SetConsoleOutputCP").Call(utf8CodePage)
+	if result == 0 {
+		return err
+	}
+
+	return nil
+}
+
 // IsLightRendererSupported checks to see if the Light renderer is supported
 func IsLightRendererSupported() bool {
 	var oldState uint32
@@ -60,6 +79,10 @@ func (r *LightRenderer) initPlatform() error {
 		return err
 	}
 	r.inHandle = uintptr(inHandle)
+
+	if err := SetConsoleUTF8(); err != nil {
+		return err
+	}
 
 	r.setupTerminal()
 


junegunn:
Thank you very much for the patch.

I've tested the patch on cmd.exe on my Windows VM, but when I start fzf with `--height` option and paste `한글` to the prompt, fzf just terminates. It works fine without `--height`. Any idea?

https://github.com/user-attachments/assets/98913a07-eb80-482f-9e10-7ad8cf232e8d

kelleyma49:
I did all of my testing in Windows Terminal, so let me test some more. It could be because of this: 

> API Differences
> The major known difference between legacy and current is the implementation of UTF-8. The legacy host has extremely rudimentary and often incorrect support of UTF-8 with [code page 65001](https://learn.microsoft.com/en-us/windows/win32/intl/code-pages). The current console host contains incremental improvements release-over-release of Windows 10 to improve this support. Applications that are attempting to rely on predicting "known incorrect" interpretations of UTF-8 from the legacy console will find themselves receiving different answers as support is improved.
> 
> Other differences experienced with APIs should be reported to the [microsoft/terminal GitHub repository](https://github.com/microsoft/terminal/) or via the [Feedback Hub](https://learn.microsoft.com/en-us/windows-insider/feedback-hub/feedback-hub-app) for triage and possible remediation.
> 

https://learn.microsoft.com/en-us/windows/console/legacymode#api-differences

junegunn:
I've finally had a chance to test this on Windows Terminal. It works, however, on Windows Terminal, I don't seem to have any problem entering non-ASCII characters even without the patch (binary built from master branch).

junegunn:
Sorry, I was wrong. This does fix the problem on Windows Terminal. Before the patch, I had to manually run `chcp 65001` to enable non-ASCII input.

DanSM-5:
I would advice to avoid changing the code page permanently as it may lead to side effects with other tools that expect the default code page.

I think a simple solution is to query the initial code page with [GetConsoleCP](https://learn.microsoft.com/en-us/windows/console/getconsolecp) and restore it at the end.

> Or should we just tell the users to run chcp 65001 if they want to input non-ASCII query instead of doing it for them?

It is a solution but users need to be aware of the meaning of changing chcp. Because of some issues I've had in the past, I think it is usually better to always restore the code page to whatever it was before. I even created a powershell script which only purpose is to run things with a code page change and revert it back once it is done.

```powershell
# With-UTF8.ps1
# Usage
# ```powershell
# With-UTF8 { command }
# ```

[CmdletBinding()]
Param (
  [scriptblock] $block
)

try {
  # Ref: https://stackoverflow.com/questions/49476326/displaying-unicode-in-powershell
  # Save the current settings and temporarily switch to UTF-8.
  $oldOutputEncoding = $OutputEncoding;
  $oldConsoleEncoding = [Console]::OutputEncoding
  $OutputEncoding = [Console]::OutputEncoding = New-Object System.Text.Utf8Encoding

  # Execute block with utf-8 encoding
  return & $block
} finally {
  # Restore the previous settings.
  $OutputEncoding = $oldOutputEncoding;
  [Console]::OutputEncoding = $oldConsoleEncoding
}
```

I use this script very frequently on some custom functions that use fzf to handle the encoding of file names correctly (for display rather than for input) and I've never had an issue with fzf itself, so I'd say that changing the code page internally temporarily should be fine.

Issue #3924, opened 2024-07-15
Cygwin reports ps: unknown option -- o for wildcard completion


JaroslawPokropinski:
### Checklist

- [X] I have read through the manual page (`man fzf`)
- [X] I have searched through the existing issues
- [X] For bug reports, I have checked if the bug is reproducible in the latest version of fzf

### Output of `fzf --version`

0.53.0 (c4a9ccd)

### OS

- [ ] Linux
- [ ] macOS
- [X] Windows
- [ ] Etc.

### Shell

- [X] bash
- [X] zsh
- [ ] fish

### Problem / Steps to reproduce

When trying to complete command using `kill -9 **` and pressing TAB i get the following error
```
kill **ps: unknown option -- o
```
This seems to be caused by incompatibility / lack of the -o feature in ps command. For context I use Windows Cygiwin and I tried with bot bash and zsh. The output of `ps --help` is the following:
```
Usage: ps [-aefls] [-u UID] [-p PID]

Report process status

 -a, --all       show processes of all users
 -e, --everyone  show processes of all users
 -f, --full      show process uids, ppids
 -h, --help      output usage information and exit
 -l, --long      show process uids, ppids, pgids, winpids
 -p, --process   show information for specified PID
 -s, --summary   show process summary
 -u, --user      list processes owned by UID
 -V, --version   output version information and exit
 -W, --windows   show windows as well as cygwin processes

With no options, ps outputs the long format by default
```

Are you planning to add support for such case (eg. fallback some combination of ps -p $$ and awk command)?

JaroslawPokropinski:
I get the same error on the MSYS2
![image](https://github.com/user-attachments/assets/50cefaa3-4c7c-48a3-8e38-8d3957199b7d)

I think the issue with those is that cygwin version of ps does not support -o option (and it seems that msys2 also uses the same ps binary)
![image](https://github.com/user-attachments/assets/04cdc63a-9b2a-41e6-aa81-da9b0f500ec4)

I should specify that most of the functionalities of fzf work fine this is specifically competition of kill command.

JaroslawPokropinski:
Also WSL does not really meet my needs and this issue is not dealbreaker for me, I will just keep using fzf for ctrl+R reverse search.

DanSM-5:
I can confirm that the issue happens in bash from git for windows as well (which I understand it is based on MSYS2). It uses the same ps as cygwin:

```bash
ps --version
ps (cygwin) 3.4.10
Show process statistics
[...]

PR #3955, opened 2024-08-04
fix: Add fallback for cygwin ps

DanSM-5:
Description
=======

Add fallback for cygwin version of `ps` command. This should fix the `ps: unknown option -- o` error displayed currently when completing `kill **<tab>`.

WindowsTerminal_B8BnDksiwt

Fixes: https://github.com/junegunn/fzf/issues/3924

## Notes

I tried to test the change in zsh completion and the error is indeed gone however I noticed that it doesn't show any content in fzf. I dig a bit and it seems like all of the completions that use `_fzf_complete` have the same issue on my machine, this includes `export **<tab>`, `unalias **<tab>`, `unset **<tab>`, etc.

A cause I can think of is that `mkfifo` is not working the same in windows and that seems to be the mayor difference I see between the bash and zsh completions.


junegunn:
Thanks!

PR #3960, opened 2024-08-09
fix: support comments in FZF_DEFAULT_OPTS_FILE by sanitizing content

daephx:
Added a sanitization step to the `FZF_DEFAULT_OPTS_FILE` processing, allowing users to include comments in their configuration.

The sanitizer skips empty lines and lines starting with a `#`, inserting the remaining options into a buffer that will be passed to the parser.

Inline comments are not supported. Additionally, any argument that includes a `#` should remain unaffected by this change.

### DIFF
--- a/src/options.go
+++ b/src/options.go
@@ -2928,7 +2928,16 @@ func ParseOptions(useDefaults bool, args []string) (*Options, error) {
 				return nil, errors.New("$FZF_DEFAULT_OPTS_FILE: " + err.Error())
 			}
 
-			words, parseErr := shellwords.Parse(string(bytes))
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks, but I'm afraid this won't be able to handle cases like this:

```sh
FZF_DEFAULT_OPTS='--header "# world

# peace"'
```

I wonder if it's possible to extend [mattn/go-shellwords](https://github.com/mattn/go-shellwords) to trim comments from the input.

daephx:
Ahh, that is rather obscure but understandable. Perhaps I'll go look over there when I get some more time, thanks anyway!

PR #3967, opened 2024-08-13
Implement exact-boundary match type

### LINKED ISSUE #3963: Matching with word boundaries
### Checklist

### Output of `fzf --version`

0.54.3 (c423c496)

### OS

- [X] Linux
- [ ] macOS
- [ ] Windows
- [ ] Etc.

### Shell

- [X] bash
- [ ] zsh
- [ ] fish

Hello,

recently I started using fzf (mainly with vim which is my primary development editor) and it is really a game-changer for me. Unfortunately, there is a major weakness which I face quite often in my daily work: inability to make fzf query more strict by specifying that its part(s) should be matched against whole word(s), i.e. with word boundary check.

The sequence of actions is usually as follows:
- in vim: feeding fzf in fuzzy mode with results from 'ag' to find where a term is used in source code
- finding hundreds or even thousands of lines
- trying to reduce the number of results by prefixing the pattern with '
- there are still too many results (e.g. if I look for a variable "Window" there may be many results like "bottom_window" or "InitWindow()")
- switching to other methods like running 'ag -w' in shell

It would be very helpful to have a way to toggle word boundaries check for a part of query, for example by:
- prefixing with "
- surrounding with '
- introducing regular expressions of some kind

Also it would be very nice to keep fuzzy mode still possible for other parts of the same query.

I know there are a number of similar issues requesting features like toggling "case + non-fuzzy" mode or something. It seems the closest one is #2394. Sadly, it looks like such requests are often rejected, so here are some argument in favor.

In my opinion, now fzf is not just a fuzzy finder, its primary goal is to allow a **quick** selection of one or few items among many others. It provides different modes and ways for doing that. It is a selector as such, it can switch quickly between fuzzy and non-fuzzy mode, it can stick the search pattern to beginning or end of line and so on. Thus, it is very disappointing if none of those modes and switches help and fzf still fails to limit the number of items in the list enough.

Switching to a different source of data (e.g. running ag with '-w') is too long and inconvenient and it may require several iterations of refining the query pipeline.

As a programmer I have had a quick look at the sources of fzf (though, I haven't used Golang before) and it seems the implementation should not be difficult. Please, consider it.

junegunn:
Close #3963

This implements the `exact-boundary` match type, which finds exact matches for a search term where both ends are at word boundaries. `Exact-boundary` term should start with `'` and end with `'`.

* I specifically didn't choose `"TERM"` because I often use such patterns to search for string literals in the source code.
    * We couldn't do the same with `'TERM'` before, so we're not breaking a use case here.
* In `--exact` mode, `'` prefix enables fuzzy matching, so `exact-boundary` term should be in `TERM'` form.
   * This is not ideal, considering cases where you type `it's`.
* Should we allow partial boundary matches? i.e. only one end at a word boundary.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -36,7 +36,7 @@ jobs:
       run: sudo apt-get install --yes zsh fish tmux
 
     - name: Install Ruby gems
-      run: sudo gem install --no-document minitest:5.17.0 rubocop:1.43.0 rubocop-minitest:0.25.1 rubocop-performance:1.15.2
+      run: sudo gem install --no-document minitest:5.25.1 rubocop:1.65.0 rubocop-minitest:0.35.1 rubocop-performance:1.21.1
 
     - name: Rubocop
       run: rubocop --require rubocop-minitest --require rubocop-performance
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -6,7 +6,7 @@ Lint/ShadowingOuterLocalVariable:
   Enabled: false
 Style/MethodCallWithArgsParentheses:
   Enabled: true
-  IgnoredMethods:
+  AllowedMethods:
     - assert
     - exit
     - paste
@@ -15,7 +15,7 @@ Style/MethodCallWithArgsParentheses:
     - refute
     - require
     - send_keys
-  IgnoredPatterns:
+  AllowedPatterns:
     - ^assert_
     - ^refute_
 Style/NumericPredicate:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,13 @@ CHANGELOG
 
 0.55.0
 ------
+- Added `exact-boundary-match` type to the search syntax. When a search term is single-quoted, fzf will search for the exact occurrences of the string with both ends at word boundaries.
+  ```sh
+  fzf --query "'here'" << EOF
+  come here
+  not there
+  EOF
+  ```
 - [bash] Fuzzy path completion is enabled for all commands
     - 1. If the default completion is not already set
     - 2. And if the current bash supports `complete -D` option
--- a/Makefile
+++ b/Makefile
@@ -77,7 +77,6 @@ endif
 all: target/$(BINARY)
 
 test: $(SOURCES)
-	[ -z "$$(gofmt -s -d src)" ] || (gofmt -s -d src; exit 1)
 	SHELL=/bin/sh GOOS= $(GO) test -v -tags "$(TAGS)" \
 				github.com/junegunn/fzf/src \
 				github.com/junegunn/fzf/src/algo \
@@ -86,6 +85,10 @@ test: $(SOURCES)
 
 bench:
 	cd src && SHELL=/bin/sh GOOS= $(GO) test -v -tags "$(TAGS)" -run=Bench -bench=. -benchmem
+
+lint: $(SOURCES) test/test_go.rb
+	[ -z "$$(gofmt -s -d src)" ] || (gofmt -s -d src; exit 1)
+	rubocop --require rubocop-minitest --require rubocop-performance
 
 install: bin/fzf
 
@@ -184,4 +187,4 @@ update:
 	$(GO) get -u
 	$(GO) mod tidy
 
-.PHONY: all generate build release test bench install clean docker docker-test update
+.PHONY: all generate build release test bench lint install clean docker docker-test update
--- a/README.md
+++ b/README.md
@@ -376,15 +376,16 @@ Unless otherwise specified, fzf starts in "extended-search mode" where you can
 type in multiple search terms delimited by spaces. e.g. `^music .mp3$ sbtrkt
 !fire`
 
-| Token     | Match type                 | Description                          |
-| --------- | -------------------------- | ------------------------------------ |
-| `sbtrkt`  | fuzzy-match                | Items that match `sbtrkt`            |
-| `'wild`   | exact-match (quoted)       | Items that include `wild`            |
-| `^music`  | prefix-exact-match         | Items that start with `music`        |
-| `.mp3$`   | suffix-exact-match         | Items that end with `.mp3`           |
-| `!fire`   | inverse-exact-match        | Items that do not include `fire`     |
-| `!^music` | inverse-prefix-exact-match | Items that do not start with `music` |
-| `!.mp3$`  | inverse-suffix-exact-match | Items that do not end with `.mp3`    |
+| Token     | Match type                              | Description                                  |
+| --------- | --------------------------------------  | ------------------------------------------   |
+| `sbtrkt`  | fuzzy-match                             | Items that match `sbtrkt`                    |
+| `'wild`   | exact-match (quoted)                    | Items that include `wild`                    |
+| `'wild'`  | exact-boundary-match (quoted both ends) | Items that include `wild` at word boundaries |
+| `^music`  | prefix-exact-match                      | Items that start with `music`                |
+| `.mp3$`   | suffix-exact-match                      | Items that end with `.mp3`                   |
+| `!fire`   | inverse-exact-match                     | Items that do not include `fire`             |
+| `!^music` | inverse-prefix-exact-match              | Items that do not start with `music`         |
+| `!.mp3$`  | inverse-suffix-exact-match              | Items that do not end with `.mp3`            |
 
 If you don't prefer fuzzy matching and do not wish to "quote" every word,
 start fzf with `-e` or `--exact` option. Note that when  `--exact` is set,
--- a/man/man1/fzf.1
+++ b/man/man1/fzf.1
@@ -1146,6 +1146,22 @@ A term can be prefixed by \fB^\fR, or suffixed by \fB$\fR to become an
 anchored-match term. Then fzf will search for the lines that start with or end
 with the given string. An anchored-match term is also an exact-match term.
 
+.SS Exact\-boundary\-match (quoted both ends)
+A single-quoted term is interpreted as an "exact\-boundary\-match". fzf will
+search for the exact occurrences of the string with both ends at the word
+boundaries. Unlike in regular expressions, this also sees an underscore as
+a word boundary. But the words around underscores are ranked lower and appear
+later in the result than the other words around the other types of word
+boundaries.
+
+1. xxx foo xxx (highest score)
+.br
+2. xxx foo_xxx
+.br
+3. xxx_foo xxx
+.br
+4. xxx_foo_xxx (lowest score)
+
 .SS Negation
 If a term is prefixed by \fB!\fR, fzf will exclude the lines that satisfy the
 term from the result. In this case, fzf performs exact match by default.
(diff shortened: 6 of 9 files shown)


junegunn:
> Could you please describe this in more detail?

I was worried about match type being changed multiple times against the user's will.

* `it` - exact match
* `it'` - exact boundary match
* `it's` - exact match

The transition can surprise the user, although if the final query is `it's`, the result will be the same, so it's probably okay. But the user will no longer match literal `it'`.


> Also I suspect that work of '-suffix in `--exact` mode is not so obvious from the documentation and, by the way, will the man page be changed as well?

This is just a draft, no documentation, no tests yet.


> I would like to propose few alternatives:

The first option is probably the least confusing one, but the users will experience unintended mode changes in `--exact` mode while typing.

* `'it` - fuzzy match (many results)
* `'it'` - exact boundary match (few results)
* `'it's` - fuzzy match (many results again)


> The second option would also allow quicker enabling of exact-boundary-match.

`'`-prefix syntax was borrowed from [Lisp](https://stackoverflow.com/questions/134887/when-to-use-or-quote-in-lisp), so users with Lisp experience can see the reasoning behind it. `'`-suffix enabling boundary match feels a little awkward and arbitrary. On the other hand, I think fully-quoted syntax shows the intention of the user better, i.e. I want 'stricter' matching.

maxaykin:
> \bfront, back\b, \bboth\b, à la regular expression.

I don't know if this is really needed but it could be a good option. Though, if implemented it may become a precedent for other requests for regexp-like modifiers. :wink:  BTW, I noticed that some people asked for a toggle of case sensitiveness. It is what I would like to have as well (e.g. in the way it is done in vim with modifiers "\C", "\c").

Regarding your implementation of word boundary check (based on bonus), now words cannot include underscore. As far as I understand, this is uncommon (see https://stackoverflow.com/questions/1324676/what-is-a-word-boundary-in-regex). And this is also not what I would expect.

junegunn:
> Though, if implemented it may become a precedent for other requests for regexp-like modifiers.

Right. I don't want that.

> Regarding your implementation of word boundary check (based on bonus), now words cannot include underscore. As far as I understand, this is uncommon (see https://stackoverflow.com/questions/1324676/what-is-a-word-boundary-in-regex). And this is also not what I would expect.

Thanks for pointing it out. But I don't know, expectations are subjective. When I see `src/options_no_pprof.go`, underscores do look like word separator, and I would be surprised if I don't get it when I type in `'no'`, but you may expect different.

Expectations also depend on context. If a project of your interest uses the `snake_case` convention in its code and file paths, it may be better to not regard `_` as a word separator. But some languages use `-` as a word separator (Lisp, CSS), and I personally work with some projects that use `-` as a word separator in the file paths. e.g. `roles/hadoop/tasks/format-namenode.yml`. Should we also not regard `-` as a word separator for cases like that? Vim provides options like `iskeyword` to deal with different conventions, but we don't want to go down that rabbit hole. Acknowledging the fact that there's no one-size-fits-all solution, I think this match type should be more inclusive than exclusive.

That said, I'm not going to make the decision right away, I'm going to take some time to use it myself to see how I like it better.

junegunn:
After experimenting with this for a while, one thing I can agree is that `..._foo_...` should be ranked lower than `...-foo-...`, `...[foo]...`, `...(foo)...`, `... $foo ...`, etc. We can still include those underscore-boundary results but lower their ranks.

maxaykin:
> Expectations also depend on context... there's no one-size-fits-all solution

I understand your point and I agree with it partially but this approach does not solve the problem for me (yes, most of the projects I work with use the `snake_case` convention, and thus, there are still too many results after entering a query in fzf).

> When you're looking for win, but not _win, you can add !_w to the query to further narrow down the results. 'win' !_w

This does not help. Consider the following examples which are quite common, I think (at least, I see them very often and it was one of the reasons to request this feature) :
```
new_win_id = get_id(win)
win = get_win_by_id(win_id)
```

The query you suggested will not show them at all while these results may be very important if I search for all places where term 'win' is used. Use of logical OR (i.e. `'win' | !_w`) will not help as well since it will include everything else.

Could you please consider possibility of making a configurable list of characters that would be treated as part of words? For example, storing it in an environment variable would allow different setups for projects based on different languages. I believe, it would satisfy most of people.

If not, is there a proper and simple way of hard-coding it in a local copy of fzf repository (I mean something better than what I  implemented in my fork)?

Also I noticed something weird in behavior of the current logic: using the above examples as the input data query `'_win'` shows both lines but queries `'win_'` and `'_win_'` show nothing. This is quite confusing.

junegunn:
Have you tried getting used to the `change:reload` approach? The more I hear from you, the more I feel it's the solution you're looking for; only getting the exact matches using precise regular expressions. fzf is a fuzzy finder and it is expected by its nature that it generates many irrelevant matches. It focuses on implementing a good scoring mechanism so that relevant matches appear first, not on completely excluding less relevant matches from the result.

maxaykin:
> Have you tried getting used to the change:reload approach?

Yes but typing a regexp every time is inconvenient. So, now this is my backup option. Besides, I would like to have the same way of working with any source of data, not only for what is produced by grep-like tools. And inventing various wrappers/workarounds for different situations is not what anyone would like to do with fzf.

> It focuses on implementing a good scoring mechanism so that relevant matches appear first, not on completely excluding less relevant matches from the result.

This contradicts with the behavior of fzf extended search mode, doesn't it? I mean that switches like ^-prefix hide irrelevant results.

junegunn:
> Yes but typing a regexp every time is inconvenient.

You can find examples in https://github.com/junegunn/fzf/blob/master/ADVANCED.md where you can toggle between two modes.

> And inventing various wrappers/workarounds for different situations

Anything you have in mind in particular?

> I mean that switches like ^-prefix hide irrelevant results.

My point is that it's not the main focus of fzf. Yes, you can filter out some stuff using basic prefix, suffix patterns, but that's about it. They are most of the time incapable of completely removing irrelevant matches, and fzf was never designed to do that. If that was the point, fzf would have supported proper regular expressions for precise filtering. If you need it, there are other interactive filter programs that support it, such as [peco](https://github.com/peco/peco?tab=readme-ov-file#select-filters), but fzf chose not to.

Instead, fzf has focused on improving the scoring mechanism to put more relevant matches appear first. It's an interactive filter program unlike traditional `grep`, so once you met an irrelevant match, you can stop browsing there. And in this case, I believe fzf should match both `x_win_x` and `x win x`, but present the latter first. If you're not interested in `_win_` and the likes, you can stop there, but if you're looking for them, you can have them. It meets both expectations nicely.

> Also I noticed something weird in behavior of the current logic: using the above examples as the input data query `'_win'` shows both lines but queries `'win_'` and `'_win_'` show nothing. This is quite confusing.

That is not expected. Let me look into it.

maxaykin:
I believe that the only purpose of scoring is to show more suitable results first but all results should be relevant matches. And I don't understand why you say that fzf was not designed to remove irrelevant matches completely. Based on my experience with its modern version, this is what it does most of the time, even if one uses just fuzzy matching mode. Otherwise we would see all lines of source data every time, they would be just sorted in a curtain way.

So, what we are really discussing here is the set of instruments for filtering out irrelevant input data. I see that you personally are satisfied with the current set and trying to avoid its further extending. Ok, you are the author and you decide how to develop and support the tool. I'm not going to continue debating this.

Regarding the suggestions of using regexp (with other tools) for more precise matching, this way looks not so attractive. I really like how it is done in fzf extended mode: you just enter words or their parts that you remember or think of and then you can make just few very quick and simple modifications to reduce number of results if needed. Construction of regular expressions is not what I want to switch back.

maxaykin:
It is clear form me that scoring helps to present the most relevant (to the intention) matches at the top of the resulting list but it has nothing to do with filtering out irrelevant (to the query) parts of input data. If I understand it correctly, you confirmed that the list of results does always meet the query criteria. Thus, scoring cannot make including irrelevant to the query results acceptable. According to the current behavior of fzf they just should be hidden.

When I requested this feature, I had not anticipated that "word boundary" might have different meaning for different projects/people. Later your changes and our discussion raised this problem. While it is quite fine for you to have underscore as a word delimiter, for me results like `bottom_window` (if I typed `'window'`) look irrelevant to the query and must be hidden.

Now the question is how to deal with term "word boundary". If you state in the documentation that it is a specific version of boundary matching, probably  that will be formally ok but it will not help people like me.

I'm testing the following local modification to suit my needs:

```
diff --git a/src/algo/algo.go b/src/algo/algo.go
index f056918..607c741 100644
--- a/src/algo/algo.go
+++ b/src/algo/algo.go
@@ -167,6 +167,7 @@ const (
        charWhite charClass = iota
        charNonWord
        charDelimiter
+       charWord
        charLower
        charUpper
        charLetter
@@ -202,6 +203,8 @@ func Init(scheme string) bool {
                        c = charUpper
                } else if char >= '0' && char <= '9' {
                        c = charNumber
+               } else if char == '_' {
+                       c = charWord
                } else if strings.ContainsRune(whiteChars, char) {
                        c = charWhite
                } else if strings.ContainsRune(delimiterChars, char) {
```

junegunn:
> but it will not help people like me.

I disagree. I mentioned above how a user with a different expectation can still benefit from it because matches around `_` are ranked lower than the rest. 

If we decide not to see `_` as a word boundary, it will better meet your expectations, but then the implementation will become unusable for the other group of users. As the maintainer of this widely popular program, I strive to find a solution that will serve a larger number of users, even if it means that it will be not "ideal" for some. Adding an option is a cop-out, more options around the core functionality will generally make the program harder to understand and use, so I'd like to avoid that.

maxaykin:
> I mentioned above how a user with a different expectation can still benefit from it because matches around _ are ranked lower than the rest.

Yes, but this makes the behaviour a bit inconsistent and confusing because considering underscore as a possible part of words is more common. At least, grep and ag do not see it as a word splitter. Similarly people complaints about "exact match" being not fully exact (i.e. matching the case is not enabled by quoting which is not obvious from the documentation). So, I suggest stating these peculiarities in the man page explicitly.

> then the implementation will become unusable for the other group of users. As the maintainer of this widely popular program, I strive to find a solution that will serve a larger number of users, even if it means that it will be not "ideal" for some. Adding an option is a cop-out, more options around the core functionality will generally make the program harder to understand and use

In my turn I cannot agree with this. :smile:  A larger number of users would be served better if a greater number of their needs were satisfied. An option or a setting or a configuration/environment variable would make it possible to provide a different behaviour for those who need it and not to affect others. You could develop it so that it would not consider underscore as a word splitter by default. Excuse me, please, but not adding it with a reference to possible complexity is precisely a cop-out. I don't think that an additional option would break the compromise between complexity and quality of the tool, look at vim which is much more popular: it is complex but the number of features and its flexibility is its power.

junegunn:
Well, it's a philosophical difference in software product design so we'll have to agree to disagree, but thanks for sharing your thoughts.

maxaykin:
Thank you for your efforts! :+1:

mentonin:
This feature is very useful to me, thanks for implementing it. However, I miss the "partial match" feature. Since I did not find a reasoning for dropping it in the thread, but did see it mentioned, I am posting this comment instead of opening a new issue.

For me, I would like some 4 syntaxes: boundary at start, at end, both, either. If we can only have `'query'`, I would rather have it match with either, with higher scoring for "both".

Tangent to this, I wish the boundary scoring would also recognize camelcase, scoring it lower than underlines.

Finally, searching for boundaries combined with `$` and `^` would be great too - search for full words at line start/length seems well within fzf's scope

PR #3969, opened 2024-08-15
Make previewer script posix compliant

unrealapex:
The current implementation of the previewer script could benefit from being POSIX compliant. I have tweaked it to improve portability.

### DIFF
--- a/bin/fzf-preview.sh
+++ b/bin/fzf-preview.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/sh
 #
 # The purpose of this script is to demonstrate how to preview a file or an
 # image in the preview window of fzf.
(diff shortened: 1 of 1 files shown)


junegunn:
Care to explain the motivation? We are sacrificing the readability and the performance of the script (replacing `=~` with `grep` process, etc) and I don't see how someone using fzf, bat, and stuff wouldn't have bash installed on their system.

unrealapex:
The motivation for this change is for two reasons:
1. To make the script more portable. One of the pros stated in the README is portability and I think this change would further promote that.

In addition this, making the script POSIX complaint would allow the user to use whatever POSIX compliant shell they might have set for `/bin/sh` instead of Bash. 

  For example, I have `/bin/sh` linked to [the Dash shell](http://gondor.apana.org.au/~herbert/dash/) because [Dash is several times faster than Bash](https://unix.stackexchange.com/questions/148035/is-dash-or-some-other-shell-faster-than-bash).  The script uses `env` to find Bash regardless so I am unable to take advantage of Dash's performance.

2. This script is simple enough that it doesn't require any Bashisms.


As for the possible performance degradation for using `grep` in my implementation, I will fix the conditional so it doesn't use any external processes. AFAIK POSIX [defines pattern matching](https://www.shellcheck.net/wiki/SC3015).

junegunn:
Thanks for the clarification. But I've decided not to merge this.

* The performance benefit of `/bin/sh` over `bash` is unlikely to be noticeable for this small script. And it is not even clear if this version is more performant overall because the use of external programs over builtins may offset the performance advantage of `/bin/sh`.
* As noted in the comment, the purpose of the script is to demonstrate how to preview a file or an image in the preview window of fzf. Therefore, the code should be easy to read and understand. And I think the original bash version is easier to follow for an average user.
* I'll be the one maintaining the script, and I'm more comfortable with bash than POSIX-compliant shell. I'd like to avoid the extra burden if the benefit is not significant.
* Your version doesn't seem to run on my Macbook.
  ```
  sed: 1: "s/^~//Users/jg/": bad flag in substitute command: 'U'
  expr: syntax error
  ```
  I haven't had time to investigate the issue, but I suspect it's because of the differences between GNU and BSD utilities. But we don't have to worry if we just use with bash builtins.

PR #4002, opened 2024-09-19
Add `refresh` as `--preview-window` option

### LINKED ISSUE #2248: Reload preview window on SIGWINCH (resize of window)

## Info

- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [x] bash
    - [ ] zsh
    - [ ] fish
    
Run the following : 

I run the following : 

```
printf "%s\n" "ls" "vim" | fzf --preview 'man {} 2>/dev/null' --preview-window 'right:50%:nohidden'
```

I can see the man page show up in the right.

However, when I resize the window having fzf, the resize event does not cause a redraw in the preview window, meaning that the preview window shows only half of its capacity.

I don't know whether this should be made automatic or optional with a flag.

record_2020-11-13_103939

gmou3:
When set, automatically reruns the preview command on terminal resize (SIGWINCH).

Default behaviour remains the same (`norefresh`).

Resolves #2248.

### DIFF
--- a/shell/completion.bash
+++ b/shell/completion.bash
@@ -211,6 +211,8 @@ _fzf_opts_completion() {
       nowrap
       cycle
       nocycle
+      refresh
+      norefresh
       up top
       down bottom
       left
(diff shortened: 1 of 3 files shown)


junegunn:
Thanks, but it's currently possible via `--bind resize:refresh-preview`. I'm going to close the linked issue.

```sh
fzf --bind resize:refresh-preview --preview date
```

PR #4020, opened 2024-09-30
Leverage existing bash completion

junegunn:
This is a PoC implementation for leveraging existing bash completion

## Example: kubectl

image

However, when we select a command, fzf pastes the whole line to the command, which is definitely not what we want.

image

## Example: git

image

### DIFF
--- a/shell/completion.bash
+++ b/shell/completion.bash
@@ -283,7 +283,7 @@ _fzf_handle_dynamic_completion() {
 }
 
 __fzf_generic_path_completion() {
-  local cur base dir leftover matches trigger cmd
+  local cur base dir leftover matches trigger cmd rest
   cmd="${COMP_WORDS[0]}"
   if [[ $cmd == \\* ]]; then
     cmd="${cmd:1}"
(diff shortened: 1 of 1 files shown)


lincheney:
I think you're running into something related to this https://github.com/spf13/cobra/issues/1508, https://github.com/spf13/cobra/pull/1509
I believe kubectl uses cobra which is adding those extra descriptions.
If you set an appropriate `$COMP_TYPE` cobra will stop doing it, but you will lose the descriptions e.g. https://github.com/lincheney/fzf-tab-completion/commit/d88bf5b61c3be95be2d917b8bdf5a99a71637645
or something like what ble.sh does where they detect and specially handle cobra completions https://github.com/akinomyoga/ble.sh/commit/71d073627185258c8116248293d706761b8310fc

junegunn:
@lincheney Thank you for the tip. Your project looks very nice.

I have two problems with this:
* I really like having the description and some extra information in the list. But because the formatting is not standardized, we can't safely extract the actual entry from each line. This is unfortunate.
* If we just use the candidates from the existing bash completion, we lose the recursive traversal of the files under the current directory.

Issue #4027, opened 2024-10-04
--walker-root not used


skarasov:
### Checklist

- [X] I have read through the manual page (`man fzf`)
- [X] I have searched through the existing issues
- [X] For bug reports, I have checked if the bug is reproducible in the latest version of fzf

### Output of `fzf --version`

0.55.0 (fc69308)

### OS

- [ ] Linux
- [ ] macOS
- [X] Windows
- [ ] Etc.

### Shell

- [ ] bash
- [ ] zsh
- [ ] fish

### Problem / Steps to reproduce

shell ps 7.4.5

- run command from drive c:
```
cd "$(fzf --walker=dir --walker-root='e:\\')"
```
- cd to any dir in drive e:
- run again 
```
cd "$(fzf --walker=dir --walker-root='e:\\')"
```
- listed current dir but root of 'e:\'

junegunn:
@charlievieth I'm also seeing this unexpected behavior of "root" argument of fastwalk.Walk. Any idea?

```sh
# When run in c:\, this recursively traverses all files under c:\ as expected
cd c:\
fzf --walker-root 'c:\'

# But when not in c:\, this lists hidden files right in the current directory incorrectly prefixed by 'c:\'
cd Users
fzf --walker-root 'c:\'
```

<img width="586" alt="image" src="https://github.com/user-attachments/assets/50ff66ea-2492-4106-8036-3ec8f60873b9">

junegunn:
If I add `.` after `:`, it works as expected.

```sh
fzf --walker-root 'c:\.'
```

skarasov:
Non of this works (from my home dir):
```
cd $(fzf --walker-root='c:.' --walker=dir)
cd $(fzf --walker-root='c:\.' --walker=dir)
cd $(fzf --walker-root='c:\\.' --walker=dir)
```
blank selection ^

skarasov:
As a workaround:
```
cd "e:\$(ls -path 'e:\' -r -ad -n -d 5 | fzf)"
```
but it slooooooooow... even cutted

skarasov:
Maybe its because somewhere walker path/root trimmed from last `\`
`cd c:\` - cd to root of c:
`cd c:` - cd to current dir on c:

junegunn:
It looks like the problem is that fastwalk trims the trailing path separators from the root argument.

[...]

PR #4063, opened 2024-10-25
Update fastwalk to v1.0.9 to fix handling of disk root paths on Windows

charlievieth:
Update fastwalk to [v1.0.9](https://github.com/charlievieth/fastwalk/releases/tag/v1.0.9).

This fixes an issue on Windows when the argument to `--walker-root` was the root of a disk drive (e.g. `C:\`). Instead of walking the disk root `C:\` it walked the current directory. This occurred because we were transforming the fully qualified path `C:\` to the relative path `C:` which caused the current directory to be walked ([docs](https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file#fully-qualified-vs-relative-paths)).

Fixes: https://github.com/junegunn/fzf/issues/4027

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,7 +1,7 @@
 module github.com/junegunn/fzf
 
 require (
-	github.com/charlievieth/fastwalk v1.0.8
+	github.com/charlievieth/fastwalk v1.0.9
 	github.com/gdamore/tcell/v2 v2.7.4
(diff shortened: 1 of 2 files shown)


charlievieth:
Example output:
Screenshot 2024-10-25 at 10 49 45 AM

junegunn:
Thank you very much for the fix!

PR #4067, opened 2024-10-28
Update fzf.vim - support nushell

### LINKED ISSUE #4066: vim plugin: support nushell
### Checklist

### Output of `fzf --version`

0.56.0 (brew)

### OS

- [ ] Linux
- [X] macOS
- [ ] Windows
- [ ] Etc.

### Shell

- [ ] bash
- [ ] zsh
- [ ] fish

I am using [nushell](https://www.nushell.sh), which is not working properly with the `fzf` vim plugin.

Specifically, the vim `systemlist` call `fzf --version --no-height` in is [this file](https://github.com/junegunn/fzf/blob/ff1687744db737c935db0f8e4d29373db9f84d8c/plugin/fzf.vim#L174) is giving an error because the fzf command is passed to `shellescape` before being passed into `systemlist`.

#### Setup:
I have `$SHELL == 'User/jasha/.cargo/bin/nu` and have set `'shell'` and `'shellxquote'` to appropriate values.

Jasha10:
This closes #4066, adding support for nushell in the fzf plugin.

I've taken care to ensure that behavior for other shells will not be modified.

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -171,7 +171,14 @@ function s:get_version(bin)
   if has_key(s:versions, a:bin)
     return s:versions[a:bin]
   end
-  let command = (&shell =~ 'powershell\|pwsh' ? '&' : '') . s:fzf_call('shellescape', a:bin) . ' --version --no-height'
+  if &shell =~ 'powershell\|pwsh'
+    let command = '& ' . s:fzf_call('shellescape', a:bin) . ' --version --no-height'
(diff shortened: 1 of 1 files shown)


junegunn:
You can't pass a list to `system` and `systemlist` on regular Vim. Also, when `&shell` is set to `nu`, `system` doesn't work at all on Vim.

```sh
set shell=nu
echo system('ls')
  " E282: Cannot read from "..."
```

It's a known issue: https://github.com/nushell/nushell/issues/8981

PR #4070, opened 2024-10-30
reader: Fixed race condition

alex-huff:
reader: Fixed race condition where reader would be terminated before it was properly initialized.

Currently, when reloading input very quickly (happens when updating the prompt quickly with 'reload' bound to 'change') it is possible for the reader's terminate method to be called before readFromCommand properly spawns the new reader process. This means the process is never killed and input is not reloaded until the process completes naturally or another 'EvtSearchNew' is triggered.

An example where this becomes a problem is when using fzf with rg integration and searching a large directory. Occasionally, the query being processed by rg is not the same as the query displayed by fzf and only after further modifying the query or waiting until rg finishes will the input reload to the correct state. This commit introduces a 'started' condition that the terminate method will wait for before trying to kill the process. The 'started' condition is signalled only when the reader reaches a state where it can be safely terminated.

### DIFF
--- a/src/core.go
+++ b/src/core.go
@@ -299,6 +299,7 @@ func Run(opts *Options) (int, error) {
 		itemIndex = 0
 		inputRevision.bumpMajor()
 		header = make([]string, 0, opts.HeaderLines)
+		reader.reset()
 		go reader.restart(command, environ)
 	}
 
--- a/src/reader.go
+++ b/src/reader.go
@@ -18,23 +18,27 @@ import (
 
 // Reader reads from command or standard input
 type Reader struct {
-	pusher   func([]byte) bool
-	executor *util.Executor
-	eventBox *util.EventBox
-	delimNil bool
-	event    int32
-	finChan  chan bool
-	mutex    sync.Mutex
-	exec     *exec.Cmd
-	execOut  io.ReadCloser
-	command  *string
-	killed   bool
-	wait     bool
+	pusher     func([]byte) bool
+	executor   *util.Executor
+	eventBox   *util.EventBox
+	delimNil   bool
+	event      int32
+	finChan    chan bool
+	mutex      *sync.Mutex
+	exec       *exec.Cmd
+	execOut    io.ReadCloser
+	command    *string
+	killed     bool
+	started    bool
+	startedCon *sync.Cond
+	wait       bool
 }
 
 // NewReader returns new Reader object
 func NewReader(pusher func([]byte) bool, eventBox *util.EventBox, executor *util.Executor, delimNil bool, wait bool) *Reader {
-	return &Reader{pusher, executor, eventBox, delimNil, int32(EvtReady), make(chan bool, 1), sync.Mutex{}, nil, nil, nil, false, wait}
+	mutex := sync.Mutex{}
+	startedCon := sync.NewCond(&mutex)
+	return &Reader{pusher, executor, eventBox, delimNil, int32(EvtReady), make(chan bool, 1), &mutex, nil, nil, nil, false, false, startedCon, wait}
 }
 
 func (r *Reader) startEventPoller() {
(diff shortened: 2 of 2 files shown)


junegunn:
Is there any way I can easily reproduce the problem without having to rely on ripgrep and a large directory? Maybe using a contrived command with` sleep` for an artificial delay.

alex-huff:
```sh
fzf --disabled --bind 'change:reload:echo Current query: {q} && sleep 1d'
```

When running this command, sometimes when you stop typing a query you may see that the item list does not update for the latest version of the query. You either have to wait 1 day or modify the query to get any new results.

https://github.com/user-attachments/assets/a81a0d94-84ff-404c-83d6-8a11b1c170ab

junegunn:
Please see dcb4c3d84a5bd7985dd70dd7de7a68992693c20c and let me know if it doesn't fix the problem, or you find any flaw in the patch. Thanks.

alex-huff:
Thank you! https://github.com/junegunn/fzf/commit/dcb4c3d84a5bd7985dd70dd7de7a68992693c20c does fix the problem I was encountering.

I did notice 3 minor issues when reading through `reader.go`/`core.go`.

1. In `restart`, if `startCommand` returns `nil, nil` then `fin` is never called. This means the coordinator routine never receives a `EvtReadFin` event. So, if `exec.Cmd.Start` fails, no future reload commands will ever be run. I verified this in a contrived way by entering a query, moving my shell, attempting to modify the query (input did not reload as expected), moving my shell back, and then further modifying the query (still the input did not reload even with valid shell). `exec.Cmd.Start` is unlikely to fail but in the case that it does I think that `fin` should be called to properly shutdown the event poller and allow future input reloading.

2. There is still a race condition with `ReadSource` when `!streamingFilter`.
https://github.com/junegunn/fzf/blob/d938fdc496ccecfe5d747500927b675d31e1835a/src/core.go#L170-L176
This race condition is much less likely to cause problems then the original one, but still exists. It can be simulated by adding a `time.Sleep(time.Second)` to ReadSource and then entering a query when fzf launches. When `ReadSource` wakes up, the reader's `termFunc` will have already been set to `nil`. fzf is then in a state where the item list will not be updated until either the initial command completes or the query is modified.

In action:
```sh
fzf --disabled --bind "start:reload:echo start command; sleep 1d" --bind "change:reload:echo reloaded input with query {q}"
```
https://github.com/user-attachments/assets/10ae13cf-67bd-4d06-a1f7-ffd4f2b280d6

3. When piping data to fzf and then trying to reload the input, the `Read` call in `feed` will not return until more data is sent to stdin. This means that when piping data to fzf from a long running program that only occasionally (or never) produces output, attempting to reload input will fail for potentially long periods of time.

Here is a contrived example:
```sh
(while true; do echo stdin data; sleep 10; done) | fzf --disabled --bind "change:reload:echo reloaded with: {q}"
```
https://github.com/user-attachments/assets/b247a46e-710c-47bb-80ca-6c4e81840c08

This is because the stdin file is in blocking mode. Calling `util.SetNonblock(os.Stdin, true)` before `readFromStdin` and then appropriately handling `syscall.EAGAIN` errors from `Read` seemed to fix the issue **on my system**.

junegunn:
Thank you very much for the detailed analysis. I'll address the issues and let you know of the progress.

junegunn:
I believe 19495eb addresses the first two issues you mentioned. It restored the previous code structure, but added a barrier using a channel to ensure that the process has started.

I'm not going to think too much about the third issue for now, as I'm afraid trying to fix it might cause other side effects (e.g. affecting the ingestion performance) and we can work around it using a `start:reload` binding.

```sh
fzf --disabled \
    --bind 'change:reload:echo reloaded with: {q}' \
    --bind 'start:reload:while true; do echo stdin data; sleep 10; done'
```

Thanks for your feedback!

junegunn:
Hmm, the last commit has a bug that can freeze fzf when a process failed to start. Let me see..

alex-huff:
Thank you!

PR #4073, opened 2024-11-01
fix(zsh): move 'fc -RI' inside command substitution

LangLangBart:
### description
- hotfix for #4071

When the selection of <kbd>⌃</kbd><kbd>R</kbd> is aborted with <kbd>Escape</kbd>, the widget assigned to the <kbd>↑</kbd> (up arrow key), which is by default `up-line-or-history`, ignores any input.

I don't have an explanation for why this occurs, and why it only happens in combination with the `-I` flag on the `fc` builtin command.

Moving it inside the command substitution alleviates the issue.

Related widget with `fc -RI` usage:
https://zsh.org/mla/users/2021/msg00186.html

---

### to reproduce the issue

1. Start a new session and paste the following:

```bash
test-widget() {
	fc -RI
}
zle -N test-widget

bindkey '^p' test-widget
bindkey '^[[A' up-line-or-history
```

2. Press <kbd>⌃</kbd><kbd>P</kbd>.
3. Press <kbd>↑</kbd>; notice that nothing happens.

The issue doesn't occur in an environment where the `HISTFILE` is undefined or the assigned file is empty.

### DIFF
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -111,10 +111,13 @@ fzf-history-widget() {
   # Ensure the module is loaded if not already, and the required features, such
   # as the associative 'history' array, which maps event numbers to full history
   # lines, are set. Also, make sure Perl is installed for multi-line output.
-  if zmodload -F zsh/parameter p:{commands,history,options} 2>/dev/null && (( ${#commands[perl]} )); then
+  if zmodload -F zsh/parameter p:{commands,history} 2>/dev/null && (( ${#commands[perl]} )); then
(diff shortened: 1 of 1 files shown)


LangLangBart:
> I don't have an explanation for why this occurs, and why it only happens in combination with the `-I` flag on the `fc` builtin command.

> The issue doesn't occur in an environment where the HISTFILE is undefined or the assigned file is empty.

No satisfying explanation found! Giving up for now to find one.

junegunn:
That's strange, probably a zsh bug. Anyway, thanks!

PR #4077, opened 2024-11-03
POC: [zsh] Use currently typed word as query for file-widget

druckdev:


druckdev:
This is a proof-of-concept. Before implementing this for the other shells or putting more work into it, I wanted to check if this is a wanted feature. Let me know what you think :)

junegunn:
We recommend users to use fuzzy completion for this. It's more powerful that you can specify the search root along with the initial query. 

```sh
ls **<tab>
ls foo**<tab>
ls ~/github/**<tab>
ls ~/github/foo**<tab>
```

druckdev:
That makes sense, thanks!

PR #4078, opened 2024-11-04
Add 'FZF_GLOBAL_POS' environment variable

jaydee-coder:
This env variable can be helpful in a variety of cases for example, for highlighting the original line as part of the preview panel.  

The current `FZF_POS` variable only tracks the index of the selected item _as part of the filtered-down list_ instead of the index in the original list/file. That's why `FZF_GLOBAL_POS` is added, which adds the ability to track the original index.

Changes: 
- Added support to the Terminal source file
- Added tests and man page snippets.

Please let me know if you have any coding/style feedback especially regarding the `global_pos` string variable in `terminal.go`

### DIFF
--- a/man/man1/fzf.1
+++ b/man/man1/fzf.1
@@ -1106,7 +1106,9 @@ fzf exports the following environment variables to its child processes.
 .br
 .BR FZF_SELECT_COUNT "    Number of selected items"
 .br
-.BR FZF_POS "             Vertical position of the cursor in the list starting from 1"
+.BR FZF_POS "             Vertical position of the cursor in the filtered list starting from 1"
+.br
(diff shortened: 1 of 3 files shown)


junegunn:
I believe you can use the existing `{n}` for that purpose.

```sh
seq 10000 | fzf --preview 'echo {n}'
```

jaydee-coder:
> I believe you can use the existing `{n}` for that purpose.
> 
> ```shell
> seq 10000 | fzf --preview 'echo {n}'
> ```

dang, I'm such a dummy, you're right, lol!
How did I miss it when I used it in the subsequent pull request lol

Thanks for the hint, and closing this useless PR :D

junegunn:
No problem, thanks for your interest in the project!

PR #4079, opened 2024-11-04
Allow specifying '{n}' as the OFFSET in the preview-window flag

jaydee-coder:
This feature can be useful in scrolling to the selected line in the preview window.

For example, it can be used to use fzf as a quick filtering tool for large files while keeping an eye on the selected line as it shows up in the original file.

Example usage:
```bash
man fzf | tee /tmp/tmp.log | fzf \
  --reverse \
  --exact \
  --query 'position' \
  --preview 'bat --force-colorization /tmp/tmp.log --highlight-line $(({n}+1))' \
  --preview-window '+{n}/2'
```

### DIFF
--- a/src/options.go
+++ b/src/options.go
@@ -1726,7 +1726,7 @@ func parsePreviewWindowImpl(opts *previewOpts, input string) error {
 	var err error
 	tokenRegex := regexp.MustCompile(`[:,]*(<([1-9][0-9]*)\(([^)<]+)\)|[^,:]+)`)
 	sizeRegex := regexp.MustCompile("^[0-9]+%?$")
-	offsetRegex := regexp.MustCompile(`^(\+{-?[0-9]+})?([+-][0-9]+)*(-?/[1-9][0-9]*)?$`)
+	offsetRegex := regexp.MustCompile(`^(\+{(-?[0-9]+|n)})?([+-][0-9]+)*(-?/[1-9][0-9]*)?$`)
 	headerRegex := regexp.MustCompile("^~(0|[1-9][0-9]*)$")
 	tokens := tokenRegex.FindAllStringSubmatch(input, -1)
 	var alternative string
--- a/man/man1/fzf.1
+++ b/man/man1/fzf.1
@@ -805,7 +805,7 @@ e.g. \fBborder\-rounded\fR (border with rounded edges, default),
 * \fB[:+SCROLL[OFFSETS][/DENOM]]\fR determines the initial scroll offset of the
 preview window.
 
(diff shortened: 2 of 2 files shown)


src/options.go
```diff
@@ -1726,7 +1726,7 @@ func parsePreviewWindowImpl(opts *previewOpts, input string) error {
 	var err error
 	tokenRegex := regexp.MustCompile(`[:,]*(<([1-9][0-9]*)\(([^)<]+)\)|[^,:]+)`)
 	sizeRegex := regexp.MustCompile("^[0-9]+%?$")
-	offsetRegex := regexp.MustCompile(`^(\+{-?[0-9]+})?([+-][0-9]+)*(-?/[1-9][0-9]*)?$`)
+	offsetRegex := regexp.MustCompile(`^(\+{-?[0-9n]+})?([+-][0-9]+)*(-?/[1-9][0-9]*)?$`)
```
junegunn, line 1729: We need to refine the expression not to allow things like `{-nnn123nnn}`.

```suggestion
	offsetRegex := regexp.MustCompile(`^(\+{(-?[0-9]+|n)})?([+-][0-9]+)*(-?/[1-9][0-9]*)?$`)
```

jaydee-coder: Thanks for the suggestion! I was thinking of the same thing but didn't want to risk introducing new matching groups in case it breaks the code that depends on this regex somehow.

Let me give your suggestion a try and re-upload, but yes agreed it looks much nicer and more focused.

jaydee-coder: Tried it and all tests are passing! Rejects all other illegal values like `{-n}`, `{12n}`, or `{-54n1}`!

jaydee-coder: Updated and re-uploaded the PR! Didn't want to click 'Commit suggestion' since it would've created another commit on top of the existing 2 and polluted the git tree.

Please let me know if it's all good to merge!

junegunn:
Merged, thanks!

amariusz:
This is great! For really large text files, when `bat` fails to complete in reasonable time, I found such modified invocation especially useful:

```
man fzf | tee /tmp/tmp.log | fzf \
    --reverse \
    --exact \
    --query 'position' \
    --preview  'sed -n "$(( {n} - 150 )),$(( {n} + 150 ))p" /tmp/tmp.log | bat --force-colorization  --highlight-line 152 --language man' \
    --preview-window '+150/2'

```
This highlights and loads only a small relevant sample (300 lines) into preview window.
Of course swap `man fzf` with `cat very_large_text_file | .. ` and use language of your choice.

Thanks!

PR #4082, opened 2024-11-08
refactor: simplify command extraction in zsh completion

### LINKED ISSUE #1992: Completion in zsh is always based on first command

## Info

- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [ ] bash
    - [x] zsh
    - [ ] fish
    
In bash, you can invoke different completions in one line, e.g.:

    ssh **<Tab>; ls **<Tab>; unalias **<Tab>;

But in zsh, the ssh menu would be displayed everytime.

LangLangBart:
### Description

- Resolve #1992

The idea is to gain acces to the `words` array.

```bash
man 1 zshcompwid | less --pattern="^\s+words"
#   words  This  array  contains the words present on the command line currently being edited.
```

The great thing about this array is that the first element contains the name of
the command. The only downside is that the `words` array is available only in
`user-defined completion widgets`, which are created with `zle -C`, as opposed
to `user-defined widgets` created with `zle -N`.

In a command like `ls; foo=bar kill **[TAB]`, `words[1]` will be `kill`.

To understand how the `words` array is constructed, one must read the `zsh`
source code[^1].

> […] the completion system has already done the hard job (and that's not an
> overstatement, I can tell you) of deciding what makes up a word on the command
> line, taking into account quoting and special characters. The array of words
> is stored, unsurprisingly, in the array $words.[…]
>
> **Source:** A User's Guide to the Z-Shell - Chapter 6: Completion[^2]

---

### Notes

Does that mean the user has to run `autoload -Uz compinit && compinit` in their
setup to use `fzf`'s completion from now on?

**NO.**

The **new zsh completion system** basically relies on shell functions, and all
of them become available as soon as the user calls `compinit`. Since we don't
use any of these functions here, nothing in the setup changes.

The `fzf` completion system is designed to be quite minimal for a couple of
commands, like `kill`, and to be easily extended with a shell function.

> Zsh's system has become more and more sophisticated, and in version 3.1.6 a
> new completion system appeared which is supposed to do everything for you: you
> simply call a function, compinit, from an initialization file, after which zsh
> knows, for example, that gunzip should be followed by files ending in .gz. The
> new system is based on shell functions, an added bonus since they are
> extremely flexible and you already know the syntax. However, given the
> complexity it's quite difficult to get started writing your own completions
> now, and hard enough to know what to do to change the settings the way you
> like.
>
> **Source:** A User's Guide to the Z-Shell - Chapter 6: Completion[^2]

There is a very nice debug hotkey one can press, `control-X + ?`, and it will
list the function calls in a debug file.

```bash
man 1 zshcompsys | less --pattern _complete_debug
```

In the debug example below, we actually see that `zsh` also relies on the
`words` array in their `_set_command`[^3] function to extract the command that
shall be completed, and this is where the entire idea for this patch came from.

```bash
command env -i HOME=$HOME TERM=$TERM USER=$USER PATH=$PATH PAGER=$PAGER zsh -f
autoload -Uz compinit && compinit

# Enter a command, for example 'ls' and press 'control-X + ?'
ls [^X?]
# Trace output left in /tmp/zsh58725ls1 (up-history to view)
# ...

less /tmp/zsh58725ls1 ;: 'ls '
# ...
#   +_normal:37> _set_command
#    +_set_command:6> local command
#    +_set_command:8> command=ls
# ...
```

[^1]: [zsh/Code - Src/Zle/complete.c](https://sourceforge.net/p/zsh/code/ci/master/tree/Src/Zle/complete.c#l1249)
[^2]: [A User's Guide to the Z-Shell - Chapter 6: Completion, old and new](https://zsh.sourceforge.io/Guide/zshguide06.html)
[^3]: [zsh/Code - Completion/Base/Utility/_set_command](https://sourceforge.net/p/zsh/code/ci/master/tree/Completion/Base/Utility/_set_command)

### DIFF
--- a/shell/completion.zsh
+++ b/shell/completion.zsh
@@ -308,8 +308,10 @@ fzf-completion() {
   setopt localoptions noshwordsplit noksh_arrays noposixbuiltins
 
   # Check if at least one completion system (old or new) is active
-  if ! zmodload -e zsh/compctl || ! (( $+functions[compdef] )); then
-    zmodload -i zsh/compctl
+  if ! zmodload -F zsh/parameter p:functions 2>/dev/null || ! (( $+functions[compdef] )); then
+    if ! zmodload -e zsh/compctl; then
+      zmodload -i zsh/compctl
+    fi
   fi
   # http://zsh.sourceforge.net/FAQ/zshfaq03.html
   # http://zsh.sourceforge.net/Doc/Release/Expansion.html#Parameter-Expansion-Flags
--- a/shell/completion.zsh
+++ b/shell/completion.zsh
@@ -307,10 +307,9 @@ fzf-completion() {
   local tokens prefix trigger tail matches lbuf d_cmds
   setopt localoptions noshwordsplit noksh_arrays noposixbuiltins
 
-  # TODO: To successfully complete 'test_dir_completion', add an explanation!
-  # It has to do with the presence of 'zle -C …' 
-  if ! zmodload -Fe zsh/{complete,compctl}; then
-    zmodload -i zsh/{complete,compctl}
+  # Check if at least one completion system (old or new) is active
+  if ! zmodload -e zsh/compctl || ! (( $+functions[compdef] )); then
+    zmodload -i zsh/compctl
   fi
   # http://zsh.sourceforge.net/FAQ/zshfaq03.html
   # http://zsh.sourceforge.net/Doc/Release/Expansion.html#Parameter-Expansion-Flags
--- a/shell/completion.zsh
+++ b/shell/completion.zsh
@@ -307,6 +307,11 @@ fzf-completion() {
   local tokens prefix trigger tail matches lbuf d_cmds
   setopt localoptions noshwordsplit noksh_arrays noposixbuiltins
 
+  # TODO: To successfully complete 'test_dir_completion', add an explanation!
+  # It has to do with the presence of 'zle -C …' 
+  if ! zmodload -Fe zsh/{complete,compctl}; then
+    zmodload -i zsh/{complete,compctl}
+  fi
   # http://zsh.sourceforge.net/FAQ/zshfaq03.html
   # http://zsh.sourceforge.net/Doc/Release/Expansion.html#Parameter-Expansion-Flags
   tokens=(${(z)LBUFFER})
--- a/shell/completion.zsh
+++ b/shell/completion.zsh
@@ -120,25 +120,19 @@ __fzf_comprun() {
   fi
 }
 
-# Extract the name of the command. e.g. foo=1 bar baz**<tab>
+# Extract the name of the command. e.g. ls; foo=1 ssh **<tab>
 __fzf_extract_command() {
-  local token tokens
-  tokens=(${(z)1})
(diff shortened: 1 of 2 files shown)


LangLangBart:
### why the test failed

```txt
  2) Failure:
TestZsh#test_dir_completion [test/test_go.rb:3684]:
Expected: "cd /tmp/fzf-test/d55/"
  Actual: "cd /tmp/fzf-test/d55"

254 runs, 2627 assertions, 1 failures, 0 errors, 1 skips
```

The test seems to fail due to a `zsh` oddity, explained below:

> As it turns out, if you attempt to use completion without first defining at
> least one widget, zsh automatically loads zsh/compctl just so that something
> sensible will happen.  The illusion of predefined completion widgets is
> maintained, but it really is an illusion.
>
> Source: https://zsh.org/mla/workers/2003/msg00703.html

Running the Docker command below and entering `cd /tmp` followed by pressing
`TAB` will not add a `/` (slash).

```bash
docker run --rm --tty --interactive ubuntu:24.10 sh -uec '
apt-get update && apt-get install -y zsh
rm -rf /etc/zsh
echo "testing() { : }; zle -C testing .complete-word testing" > ~/.zshenv
/usr/bin/zsh'
```

Running the same command without the `~/.zshenv` line will cause `cd /tmp[TAB]`
to add the `/` (slash).

Widgets can be listed with `zle -l`.

---

Since starting `zsh` with no widgets is rare but possible, the following test
has been added to prevent future issues:

```bash
# Check if at least one completion system (old or new) is active
if ! zmodload -e zsh/compctl || ! (( $+functions[compdef] )); then
  zmodload -i zsh/compctl
fi
```

LangLangBart:
> The following test has been added to prevent future issues:
> 
> ```shell
> # Check if at least one completion system (old or new) is active
> if ! zmodload -e zsh/compctl || ! (( $+functions[compdef] )); then
>   zmodload -i zsh/compctl
> fi
> ```

I would suggest adjusting the `test_go.rb` file instead of adding this `if` case
to the `fzf-completion` widget.


```diff
--- a/test/test_go.rb
+++ b/test/test_go.rb
@@ -3681,4 +3681,9 @@ module CompletionTest
     tmux.send_keys :BSpace, :BSpace, :BSpace
     tmux.until { |lines| assert_equal 'cd /tmp/fzf-test/d55', lines[-1] }
+    if instance_of?(TestZsh)
+      tmux.send_keys 'C-c'
+      tmux.send_keys 'zmodload -i zsh/compctl', :Enter
+      tmux.send_keys 'cd /tmp/fzf-test/d55'
+    end
     tmux.send_keys :Tab
     tmux.until { |lines| assert_equal 'cd /tmp/fzf-test/d55/', lines[-1] }
```

Mostly, I believe that there isn't an actual person with such a bizarre setup,
except when testing an edge case like this. (And for the use of `functions`, one
would need to add another check to see if the `zsh/parameter` module was loaded.)

Should I adjust the test file instead of the `fzf-completion` widget?

junegunn:
Thanks! it works great as expected.

PR #4084, opened 2024-11-11
fix zsh $+commands[perl] syntax does not work with setopt ksh_arrays

lincheney:
When `setopt ksh_arrays` in zsh, the syntax `$+commands[perl]` doesn't work properly and gives a syntax warning saying `fzf-history-widget:6: bad output format specification`

I think the everything else still actually works, just the warning is a bit annoying.

I have setopt ksh_arrays set in my zshrc.

### DIFF
--- a/shell/completion.zsh
+++ b/shell/completion.zsh
@@ -308,7 +308,7 @@ fzf-completion() {
   setopt localoptions noshwordsplit noksh_arrays noposixbuiltins
 
   # Check if at least one completion system (old or new) is active
-  if ! zmodload -F zsh/parameter p:functions 2>/dev/null || ! (( $+functions[compdef] )); then
+  if ! zmodload -F zsh/parameter p:functions 2>/dev/null || ! (( ${+functions[compdef]} )); then
     if ! zmodload -e zsh/compctl; then
       zmodload -i zsh/compctl
     fi
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -111,7 +111,7 @@ fzf-history-widget() {
   # Ensure the module is loaded if not already, and the required features, such
   # as the associative 'history' array, which maps event numbers to full history
   # lines, are set. Also, make sure Perl is installed for multi-line output.
-  if zmodload -F zsh/parameter p:{commands,history} 2>/dev/null && (( $+commands[perl] )); then
+  if zmodload -F zsh/parameter p:{commands,history} 2>/dev/null && (( ${+commands[perl]} )); then
     # Import commands from other shells if SHARE_HISTORY is enabled, as the
     # 'history' array only updates after executing a non-empty command.
     selected="$(


LangLangBart:
Thank you again[^1].

```man
DESCRIPTION OF OPTIONS
  In the following list, options set by default in all emulations are marked <D>; those set by
  default  only  in csh, ksh, sh, or zsh emulations are marked <C>, <K>, <S>, <Z> as appropri-
  ate. 

KSH_ARRAYS <K> <S>
  Emulate ksh array handling as closely as possible.  If this option is set, array ele-
  ments  are  numbered  from  zero,  an array parameter without subscript refers to the
  first element instead of the whole array, and braces are required to delimit  a  sub-
  script (`${path[2]}' rather than just `$path[2]') or to apply modifiers to any param-
  eter (`${PWD:h}' rather than `$PWD:h').
```

I always feel bad about such mistakes because they are preventable.

Given that the original message was deleted and we no longer have the details
about why `emulate -LR zsh` failed for a user[^2], should we reintroduce the
command and find out?





[^1]: [fix zsh ${(@)history} syntax does not work with ksh_arrays by lincheney · Pull Request #3893 · junegunn/fzf · GitHub](https://github.com/junegunn/fzf/pull/3893)
[^2]: [[zsh] emulate -L zsh to avoid issues with incompatible options · junegunn/fzf@3a6af27 · GitHub](https://github.com/junegunn/fzf/commit/3a6af27586c65bb9243c5af7c43a40619503db52)

shell/completion.zsh
```diff
@@ -308,7 +308,7 @@ fzf-completion() {
   setopt localoptions noshwordsplit noksh_arrays noposixbuiltins
 
   # Check if at least one completion system (old or new) is active
-  if ! zmodload -F zsh/parameter p:functions 2>/dev/null || ! (( $+functions[compdef] )); then
+  if ! zmodload -F zsh/parameter p:functions 2>/dev/null || ! (( ${+functions[compdef]} )); then
```
LangLangBart, line 311: The `{}` (curly brackets) aren't needed since we already disable `ksh_arrays` in the local options. We should also disable it in the `fzf-history-widget` to avoid `ksh_arrays` issues in future commits for these widgets.

junegunn: You mean curlies here can be omitted? If it wouldn't hurt to have them here, I'd suggest keeping them for consistency.

junegunn:
> Given that the original message was deleted and we no longer have the details
> about why `emulate -LR zsh` failed for a user[2](#user-content-fn-2-c95872dfc4c3cc1f78d976393a1d7da2), should we reintroduce the
> command and find out?

It was a long time ago and I don't really remember the details. I'd like to keep the change small for now and release a new patch version soon because this can be seen as a regression in 0.56.1.

LangLangBart:
Approved.

shell/completion.zsh
LangLangBart: > You mean curlies here can be omitted?

For this specific widget, they could because of `noksh_arrays`, but it wouldn't hurt to have them.

junegunn:
Thanks @lincheney for the patch. And thanks @LangLangBart for the review.

PR #4090, opened 2024-11-12
fix(zsh): handle backtick trigger edge case

### LINKED ISSUE #4088: 0.56.1/2 breaks my `FZF_COMPLETION_TRIGGER`
### Checklist

### Output of `fzf --version`

0.56.2

### OS

- [ ] Linux
- [X] macOS
- [ ] Windows
- [ ] Etc.

### Shell

- [ ] bash
- [X] zsh
- [ ] fish

Thank you for the great work!

In my `.zshrc`, I set ``FZF_COMPLETION_TRIGGER='`'``. With up to 0.56.0 installed, this works well (and has for years). 

Starting in 0.56.1 and still existing in 0.56.2, this no longer works. When I type for example `` cd `<TAB> ``, nothing happens. If I set the env var to nearly anything else, it works again.

If I downgrade to 0.56.0, it works again.

The fzf portion of my `.zshrc`: https://github.com/jefferickson/dotfiles/blob/d552e8eae75edd1ce4d14b11ae960a9aeccb370a/.zshrc#L96-L106

LangLangBart:
### Description
- Resolve #4088

If a single backtick is used as a trigger, the `words` array fails to populate
with the command elements.

~~The idea here is to temporarily alter the `LBUFFER` before calling
`__fzf_extract_command` by escaping the backtick.~~

The idea is to move the `CURSOR` before the trigger.

This approach seems a bit hacky, which is why this PR is marked as a draft.

### Alternative A

Create a list of invalid triggers, since other characters have also been found
to cause the same issue:

Testing covered only ASCII values 32-126; others were excluded for simplicity.

| Symbol  | Common Name  | ASCII | Occurrences in Public Git Repos |
| ------- | ------------ | ----- | --------------------------------------- |
| `&`     | Ampersand    | 38    | 1                                       |
| `;`     | Semicolon    | 59    | ~25                                     |
| `[`     | Left Bracket | 91    | 1                                       |
| `` ` `` | Back-tick    | 96    | ~20                                     |
| `\|`    | Pipe         | 124   | 0                                       |

If the user attempts to use one, display a message using `zle -M [string]` to
indicate the error. This places the string below the command line, prompting the
user to choose a different trigger character.

### Alternative B
Fall back to the old extraction method if the words array is empty and the
`LBUFFER` is not empty, with the downside that it would only work for the first
command.

```diff
--- a/shell/completion.zsh
+++ b/shell/completion.zsh
@@ -123,9 +123,20 @@ __fzf_comprun() {
 # Extract the name of the command. e.g. ls; foo=1 ssh **<tab>
 __fzf_extract_command() {
+  local token tokens
   setopt localoptions noksh_arrays
-  # Control completion with the "compstate" parameter, insert and list noting
+  # Control completion with the "compstate" parameter, insert and list nothing
   compstate[insert]=
   compstate[list]=
   cmd_word="${words[1]}"
+  [[ -n $cmd_word ]] && return
+  tokens=(${(z)LBUFFER})
+  for token in $tokens; do
+    token=${(Q)token}
+    if [[ "$token" =~ [[:alnum:]] && ! "$token" =~ "=" ]]; then
+      cmd_word="$token"
+      return
+    fi
+  done
+  cmd_word="${tokens[1]}"
 }

@@ -341,5 +352,4 @@ fzf-completion() {
     # Make the 'cmd_word' global
     zle __fzf_extract_command || :
-    [[ -z "$cmd_word" ]] && return

     [ -z "$trigger"      ] && prefix=${tokens[-1]} || prefix=${tokens[-1]:0:-${#trigger}}
```

---

Alternative B might be the best option. If a user encounters an issue like the
one described in #1992, simply suggest using a different trigger?

### DIFF
--- a/shell/completion.zsh
+++ b/shell/completion.zsh
@@ -126,7 +126,7 @@ __fzf_extract_command() {
   # Control completion with the "compstate" parameter, insert and list nothing
   compstate[insert]=
   compstate[list]=
-  cmd_word="${words[1]}"
+  cmd_word="${(Q)words[1]}"
 }
 
 __fzf_generic_path_completion() {
@@ -338,8 +338,8 @@ fzf-completion() {
   if [ ${#tokens} -gt 1 -a "$tail" = "$trigger" ]; then
     d_cmds=(${=FZF_COMPLETION_DIR_COMMANDS-cd pushd rmdir})
 
-    # Move the cursor before the trigger to maintain word array elements when
-    # trigger characters like ';' or '`' reset the array.
+    # Move the cursor before the trigger to preserve word array elements when
+    # characters like ';' or '`' would otherwise reset the 'words' array.
     cursor_pos=$CURSOR
     CURSOR=$((cursor_pos - ${#trigger} - 1))
     # Make the 'cmd_word' global
--- a/shell/completion.zsh
+++ b/shell/completion.zsh
@@ -303,6 +303,7 @@ _fzf_complete_kill_post() {
 }
 
 fzf-completion() {
+  typeset -g cmd_word
   trap 'unset cmd_word' EXIT
   local tokens prefix trigger tail matches lbuf d_cmds cursor_pos
   setopt localoptions noshwordsplit noksh_arrays noposixbuiltins
@@ -343,7 +344,7 @@ fzf-completion() {
       # Move the cursor before the trigger to preserve word array elements when
       # trigger chars like ';' or '`' would otherwise reset the 'words' array.
       CURSOR=$((cursor_pos - ${#trigger} - 1))
-      # Make the 'cmd_word' global
+      # Assign the extracted command to the global variable 'cmd_word'
       zle __fzf_extract_command
     } always {
       CURSOR=$cursor_pos
--- a/shell/completion.zsh
+++ b/shell/completion.zsh
@@ -338,16 +338,13 @@ fzf-completion() {
   if [ ${#tokens} -gt 1 -a "$tail" = "$trigger" ]; then
     d_cmds=(${=FZF_COMPLETION_DIR_COMMANDS-cd pushd rmdir})
 
+    # Move the cursor before the trigger to maintain word array elements when
+    # trigger characters like ';' or '`' reset the array.
+    cursor_pos=$CURSOR
+    CURSOR=$((cursor_pos - ${#trigger} - 1))
     # Make the 'cmd_word' global
-    if [[ $trigger =~ '[&;[`|]' ]]; then
-      # Move cursor before trigger to get word array elements of the command.
-      cursor_pos=$CURSOR
-      CURSOR=$((cursor_pos - ${#trigger} - 1))
-      zle __fzf_extract_command || :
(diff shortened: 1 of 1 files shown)


jefferickson:
I'd advocate for Alternative B as I've used the backtick as my trigger for 6+ years. The key is right next to the <TAB> on an American layout and quite convenient.

Thank you!

LangLangBart:
For the current version, the `CURSOR` is moved before the trigger, as the
`words` array likely interprets any of the listed characters in the table above
as the start of a new command. 

The approach would still extract the correct command.

More testing of these changes is needed.

LangLangBart:
> I'd advocate for Alternative B

That was also my intention, but temporarily moving the cursor before calling the
`__fzf_extract_command` completion widget seems to work well and ensures that
the user still gets the correct list of matches displayed in commands like below:

```bash
ls ; kill -9 `[TAB]
```

---

Please test the latest patch. The zsh version affects many users, and it's
important to avoid releasing with errors. Feedback will help ensure the fix
works well. Thanks!

```bash
source <(curl -fsSL https://raw.githubusercontent.com/LangLangBart/fzf/refs/heads/zsh_backtick/shell/completion.zsh)
```

jefferickson:
> Please test the latest patch.

The patch works for me and my backtick trigger! Thank you!

junegunn:
I'm testing the latest revision, and I'm seeing this warning message.
image

Turns out I have `setopt WARN_CREATE_GLOBAL` in my `.zshrc`. Can we fix it? Or am I supposed to remove the setting from my config? From a confused bash user.

LangLangBart:
> Turns out I have `setopt WARN_CREATE_GLOBAL` in my `.zshrc`. Can we fix it? Or am I supposed to remove the setting from my config? From a confused bash user.

Thanks for testing. I will add `typeset -g cmd_word`.

```bash
man zshoptions | less --pattern 'WARN_CREATE_GLOBAL'
```

```txt
WARN_CREATE_GLOBAL
  Print a warning message when a global parameter is created in  a
  function  by an assignment or in math context.  This often indi-
  cates that a parameter has  not  been  declared  local  when  it
  should  have  been.   Parameters explicitly declared global from
  within a function using typeset -g do not cause a warning.  Note
  that  there  is no warning when a local parameter is assigned to
  in a nested function, which may also indicate an error.
```

```bash
# see 'zshbuiltins'  for 'typeset'
man zshbuiltins
```

junegunn (approved):

junegunn:
Thanks, it all looks good to me, is there anything else you want to address?

LangLangBart:
> Thanks, it all looks good to me, is there anything else you want to address?

I am finished. During testing, I found no other issues. I hope `fzf` users won't find any either.

The bugs (missing enhancement) reported from my `zsh` patches weren't created intentionally, and I am sorry you had to deal with so many releases.

junegunn:
Merged thanks!

jefferickson:
Thank you @LangLangBart ! Really appreciate the fast turn around and response to my report.

PR #4093, opened 2024-11-13
revert(zsh): remove 'fc -RI' call in the history widget

### LINKED ISSUE #4091: 0.56.1 broke fzf-history-widget completion in zsh 5.9 when share_history is enabled
### Checklist

### Output of `fzf --version`

0.56.2 (brew)

### OS

- [ ] Linux
- [X] macOS
- [ ] Windows
- [ ] Etc.

### Shell

- [ ] bash
- [X] zsh
- [ ] fish

This seems related to #4061:

I am using default `source <(fzf --zsh)` from macOS Homebrew with zsh 5.9 (x86_64-apple-darwin23.0) with `setopt share_history` enabled, but as of fzf version [0.56.1](https://github.com/junegunn/fzf/releases/tag/v0.56.1) the history completion with `^R` bound to `fzf-history-widget` is broken (last working version was 0.56.0). Previous behaviour in zsh was that `foo^r^m` would return fzf’s default matching result for `foo` (e.g. `found-history`) but now it returns my original input string `foo`. In the simplest case of `^r^m` the widget simply returns nothing (previously, it would have returned the fzf’s default selected item). Intermittently it does return a result, but rarely. Despite having `share_history` enabled, I do not have a second shell using the history file; this issue is reproducible with a single shell on its own. (And unlike https://github.com/junegunn/fzf/issues/4088 I’m not using `FZF_COMPLETION_TRIGGER`.) If I downgrade to 0.56.0 or disable `share_history` the previous expected behaviour is restored.

LangLangBart:
### Description
- Resolve #4091
- Reopen #4061

Intentionally added to solve an issue with updating the `history` array, which
only updated after the user ran a non-empty command[^1][^2].

There were a couple of issues:

One issue was that executing `fc -RI` seemed to prevent access to recent history
when canceling the `fzf` window with `esc` or `ctrl-c`. Pressing `up` did not
retrieve the expected history. This was mitigated by placing it inside the
command substitution[^3]. However, this led to the issue described in the
referenced ticket.

In the past[^4], it was even reported that there was an unbearable delay when
using `fc -RI`.

[^1]: [ctrl+r no longer pulling from shared history until a command is run succesfully · Issue #4061 · junegunn/fzf · GitHub](https://github.com/junegunn/fzf/issues/4061)
[^2]: [fix(zsh): history loading with shared option by LangLangBart · Pull Request #4071 · junegunn/fzf · GitHub](https://github.com/junegunn/fzf/pull/4071)
[^3]: [fix(zsh): move 'fc -RI' inside command substitution by LangLangBart · Pull Request #4073 · junegunn/fzf · GitHub](https://github.com/junegunn/fzf/pull/4073)
[^4]: [[zsh] Reload shared history before searching by mpolden · Pull Request #2251 · junegunn/fzf · GitHub](https://github.com/junegunn/fzf/pull/2251)

### DIFF
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -112,16 +112,10 @@ fzf-history-widget() {
   # as the associative 'history' array, which maps event numbers to full history
   # lines, are set. Also, make sure Perl is installed for multi-line output.
   if zmodload -F zsh/parameter p:{commands,history} 2>/dev/null && (( ${+commands[perl]} )); then
-    # Import commands from other shells if SHARE_HISTORY is enabled, as the
-    # 'history' array only updates after executing a non-empty command.
-    selected="$(
-      if [[ -o sharehistory ]]; then
-        fc -RI
-      fi
-      printf '%s\t%s\000' "${(kv)history[@]}" |
-        perl -0 -ne 'if (!$seen{(/^\s*[0-9]+\**\t(.*)/s, $1)}++) { s/\n/\n\t/g; print; }' |
-        FZF_DEFAULT_OPTS=$(__fzf_defaults "" "-n2..,.. --scheme=history --bind=ctrl-r:toggle-sort --wrap-sign '\t↳ ' --highlight-line ${FZF_CTRL_R_OPTS-} --query=${(qqq)LBUFFER} +m --read0") \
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks again, I'll reopen #4061.

PR #4109, opened 2024-11-22
Allow walking multiple root directories

msabathier:
Hi,
I have been trying to search a file or directory with fzf on Windows on multiple drives.
I managed to do it by piping the output of `Get-ChildItem` cmdlet (slow) and then the output of `fd` (faster but not ideal).
It was not a satisfying solution to me so i started to look at the code of fzf and found a solution to this problem by using a comma separated list in the `--walker-root` argument.
I am not proficient in go so the implementation might not be what you are looking for, don't hesitate to guide me towards a solution that suits you.
Best regards,
Martin

### DIFF
--- a/man/man1/fzf.1
+++ b/man/man1/fzf.1
@@ -1005,8 +1005,8 @@ Determines the behavior of the built-in directory walker that is used when
 .br
 
 .TP
-.B "\-\-walker\-root=DIR"
-The root directory from which to start the built-in directory walker.
+.B "\-\-walker\-root=DIRS"
+Comma-separated list of directory names to start the built-in directory walker.
 The default value is the current working directory.
 
 .TP
--- a/man/man1/fzf.1
+++ b/man/man1/fzf.1
@@ -1005,8 +1005,8 @@ Determines the behavior of the built-in directory walker that is used when
 .br
 
 .TP
-.B "\-\-walker\-root=DIRS"
-Comma-separated list of directory names to start the built-in directory walker.
+.B "\-\-walker\-root=DIR [...]"
+List of directory names to start the built-in directory walker.
 The default value is the current working directory.
 
 .TP
--- a/src/options.go
+++ b/src/options.go
@@ -145,7 +145,7 @@ Usage: fzf [options]
 
   Directory traversal       (Only used when $FZF_DEFAULT_COMMAND is not set)
     --walker=OPTS           [file][,dir][,follow][,hidden] (default: file,follow,hidden)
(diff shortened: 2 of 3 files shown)


junegunn:
> then the output of `fd` (faster but not ideal)

Can you explain why it felt not ideal? fd is a dedicated directory traversal program with much more versatility (notably .gitignore support). The built-in walker of fzf can be a good default for many users, but it isn't supposed to be, or tries to be, as good as dedicated external programs.

msabathier:
That was my initial thought, but after a bit of testing, I found that the built-in walker was actually quite faster than `fd`, at least on Windows (maybe the piping is introducing some overhead ?).

Also when using `fd`, the pipe crashes if I select a result before `fd` has finished executing. This can probably be fixed but it led me to test other solutions first, one of which was this PR that I found faster and more practical than using `fd`.

junegunn:
@charlievieth Thanks for the comment.

> That said, `,` is a valid (though unusual) file path element

I was also worried about the same thing and thought that we should take multiple arguments instead of a single comma-separated string.

See cff9d83

Some counter-points:
* One may argue that it's inconsistent with `--walker-skip`
* Although highly unlikely, if you have a directory whose name matches the name of an fzf option, it can be mistakenly recognized as an argument to `--walker-root`. e.g. `--walker-root foo bar --reverse` (and you have a directory named `--reverse`)

msabathier:
Thank you both for the informative comments !

I wasn't aware that `,` was a valid character in UNIX file paths and I tried to mimic the behaviour of other list-like arguments.

I remain at your disposal if you need anything.

junegunn:
@msabathier Please review and test the last two commits I made on your branch.

msabathier:
I just reviewed and tested your changes and everything looks and works fine to me.

On a side note, i did a bit of performance testing and i found that `fzf` was on par with `fd` except when looking only for directories on Windows :
- `fzf` took 6.2s on average.
  `Measure-Command {fzf -f "" --walker=dir,hidden --walker-root=C:\}`
- `fd` took 10.0s on average.
  `Measure-Command {fd -H -t d "" C:\}`

If we pipe `fd` into `fzf` in interactive picker mode then `fzf` always wins because of the piping overhead.

charlievieth:
> I was also worried about the same thing and thought that we should take multiple arguments instead of a single comma-separated string.

Makes sense to me.

> On a side note, i did a bit of performance testing and i found that fzf was on par with fd except when looking only for directories on Windows

Nice and and thank you for taking the time to dig into this!

Looks good to me and thanks for the ping.

PR #4115, opened 2024-12-02
chore: completion test command sequence

LangLangBart:
### description

Add a completion test for command sequences with different triggers.

Clean up global scope pollution in Zsh:
- `cmd_word` doesn't need to be global since the `zle __fzf_extract_command` call doesn't actually
  create a new scope.
- The completion widget is deleted after its usage; its sole purpose is to provide access to the `words` array.

---

~~Draft reason:~~
- ~~Unable to test semicolon (`;`) and backtick (`` ` ``)~~
	- ~~Using them in the `test_go.rb` file always ends in failure.~~

```diff
--- a/test/test_go.rb
+++ b/test/test_go.rb
@@ -3744,5 +3744,5 @@ module CompletionTest
     triggers = ['**', '~~', '++', 'ff', '/']
     # TODO: Find a way to test semicolon (';') and backtick ('`')
-    triggers.concat(['&', '[']) if instance_of?(TestZsh)
+    triggers.concat([';', '`', '&', '[']) if instance_of?(TestZsh)

     triggers.each do |trigger|
```

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -3,7 +3,7 @@ name: Test fzf on Linux
 
 on:
   push:
-    branches: [ master, devel ]
+    branches: [ master, devel, zsh_completion]
   pull_request:
     branches: [ master ]
   workflow_dispatch:
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -3,7 +3,7 @@ name: Test fzf on Linux
 
 on:
   push:
-    branches: [ master, devel, zsh_completion]
+    branches: [ master, devel ]
   pull_request:
     branches: [ master ]
   workflow_dispatch:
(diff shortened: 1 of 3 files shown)


junegunn:
> Unable to test semicolon (`;`) and backtick (`` ` ``)

tmux seems to treat a trailing semicolon differently.

```sh
$ tmux send-keys 'ha;ha'; echo
ha;ha
$ tmux send-keys 'ha;ha;'; echo
ha;ha
$ tmux send-keys 'ha;ha\;'; echo
ha;ha;
```

Which means we would have to escape it to `\;` or `;;`.

```diff
diff --git a/test/test_go.rb b/test/test_go.rb
index 98c2f25..11dc7a0 100755
--- a/test/test_go.rb
+++ b/test/test_go.rb
@@ -164,6 +164,7 @@ class Tmux
   private
 
   def go(args)
+    args = args.map { |a| a.sub(/;$/, '\;') }
     IO.popen(%w[tmux] + args) { |io| io.readlines(chomp: true) }
   end
 end
```

Not sure about backticks yet.

junegunn (approved):
LGTM, thanks!

PR #4116, opened 2024-12-03
Make the preview window resizable by mouse drag

druckdev:
This PR enables resizing the preview window by dragging its border with the mouse. This works with all border styles except for `none`. Counter-intuitively, having the border only on the opposite side of the window works too - dragging from it will first decrease the preview size to its minimum.

I tested this with quite a lot of different border configurations, although only on Linux. For this I used:

```
./bin/fzf --bind 'ctrl-space:change-preview-window(right,border-horizontal|right,border-left|right,border-right|down,border-rounded|down,border-vertical|down,border-down|down,border-top|left,border-rounded|left,border-horizontal|left,border-right|left,border-left|up,border-rounded|up,border-vertical|up,border-down|up,border-top)' --preview='cat {}'
```

with and without `--padding 2 --margin 3 --border` as well as `--no-scrollbar`.

---

While implementing this I found some issues, present already prior to my changes:

- The minimum window height decreases when no extra line for the horizontal separator is used (e.g. with `--info=inline --no-separator`). In this case the preview window should be able to occupy this extra line.
- When the chosen preview border shape has no left and/or right border, the minimum total preview window size decreases. But due to the hardcoded value for the size it could not be decreased further than 5.
- `--preview-window=99%,left --no-border` lets the preview window grow one cell too big
- Sometimes when the border style changes, some cells are not properly redrawn:
	- Changing the border shape let's the preview scrollbar disappear sometimes and leave parts of the old border:

	```
	fzf --preview='cat {}' --preview-window=border-horizontal --bind 'ctrl-space:change-preview-window(border-rounded|border-horizontal)'
	```

	- When changing the size and the border-shape of the preview, characters remain:

	```
	fzf --preview='cat {}' --preview-window=30%,left,border-rounded --bind 'ctrl-space:change-preview-window(50%,border-horizontal|30%)'
	```

This PR suffers a bit from the second and last one. The second results in an offset between the cursor and the border when dragging, since the preview window size after a resize is not the one "requested". The last one makes the scrollbar disappear when resizing and acts as if the preview window was one cell bigger (until one starts to scroll).

Since they are loosely related, I addressed the first two in separate commits, but will open issues for the rest. I'll happily split the commits in their own PR or discard them though, if you prefer it that way or have a different solution in mind.

In addition I de-duplicated and refactored a bit of code that could use the `Has{Top,Right,Bottom,Left}()` functions instead of "manual" switch cases. Here again, feel free to discard this, if not wanted.

## Related out-of-scope note:

At the moment `change-preview-window(...)` is the only way of resizing the preview window at "run-time", which is rather limited. This PR starts to fill this gap, but further implementing `resize-window(...)` or `{grow,shrink}(...)` actions that allow relative resizes via keystrokes, would probably make a lot of sense.

### DIFF
--- a/src/options.go
+++ b/src/options.go
@@ -381,14 +381,46 @@ func (a previewOpts) aboveOrBelow() bool {
 	return a.size.size > 0 && (a.position == posUp || a.position == posDown)
 }
 
-func (a previewOpts) sameLayout(b previewOpts) bool {
-	return a.size == b.size && a.position == b.position && a.border == b.border && a.hidden == b.hidden && a.threshold == b.threshold &&
-		(a.alternative != nil && b.alternative != nil && a.alternative.sameLayout(*b.alternative) ||
-			a.alternative == nil && b.alternative == nil)
-}
+type previewOptsCompare int
 
-func (a previewOpts) sameContentLayout(b previewOpts) bool {
-	return a.wrap == b.wrap && a.headerLines == b.headerLines && a.info == b.info
+const (
+	previewOptsSame previewOptsCompare = iota
+	previewOptsDifferentContentLayout
+	previewOptsDifferentLayout
+)
+
+func (o *previewOpts) compare(active *previewOpts, b *previewOpts) previewOptsCompare {
+	a := o
+
+	sameThreshold := o.position == b.position && o.threshold == b.threshold
+	// Alternative layout is being used
+	if o.alternative == active {
+		a = active
+
+		// If the other also has an alternative layout,
+		if b.alternative != nil {
+			// and if the same condition is the same, compare alt vs. alt.
+			if sameThreshold {
+				b = b.alternative
+			} else {
+				// If not, we pessimistically decide that the layouts may not be the same
+				return previewOptsDifferentLayout
+			}
+		}
+	} else if b.alternative != nil && !sameThreshold {
+		// We may choose the other's alternative layout, so let's be conservative.
+		return previewOptsDifferentLayout
+	}
+
+	if !(a.size == b.size && a.position == b.position && a.border == b.border && a.hidden == b.hidden) {
+		return previewOptsDifferentLayout
+	}
+
+	if a.wrap == b.wrap && a.headerLines == b.headerLines && a.info == b.info && a.scroll == b.scroll {
+		return previewOptsSame
+	}
+
+	return previewOptsDifferentContentLayout
 }
 
 func firstLine(s string) string {
--- a/src/terminal.go
+++ b/src/terminal.go
@@ -1679,7 +1679,13 @@ func (t *Terminal) resizeWindows(forcePreview bool) {
 					createPreviewWindow(marginInt[0]+height-pheight, marginInt[3], width, pheight)
 				}
 			case posLeft, posRight:
-				pwidth := calculateSize(width, previewOpts.size, minWidth, 5, 4)
+				pad := borderColumns(previewOpts.border, t.borderWidth)
+				if len(t.scrollbar) > 0 && !previewOpts.border.HasRight() {
+					// Need a column to show scrollbar
+					pad += 1
+				}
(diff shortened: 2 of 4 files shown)


druckdev:
What I forgot to mention: The code still includes the TODO "should this allow a size of zero" before the lines that inforce a minimum size of `1` [^1].

Since a size of zero differs from no preview, if I understand it correctly, I thought that it would be better to disallow shrinking it that far. It could also create the issue that one could accidentally hide the preview window with no means of re-displaying it. But it still seemed like it had potential for discussion, thus I left the TODO.

[^1]: https://github.com/junegunn/fzf/pull/4116/files#diff-519ee16445a8adb05632a100d7ae9de2cd49afa0805e94be2457a0751e83d04bR4766-R4769

junegunn:
Thanks for taking the initiative. I'll review the code when I get some time.

> Since a size of zero differs from no preview, if I understand it correctly

Correct.

> it would be better to disallow shrinking it that far. It would also create the issue that one could accidentally hide the preview window with no means of re-displaying it. 

I agree.

druckdev:
I only ran `make test` locally. I will try to look at the failing Integration tests. I would think that 50f7ebe4666f ("Fix minimum preview width without left/ right borders") is the commit that broke something.

junegunn:
The failure was caused by fzf using the alternative layout (`t.previewOpts.alternative` or `t.activePreviewOpts`) while the new code was looking at the default layout (`t.previewOpts`). And it looks like there are many other places with the same problem. Let me take this over and finish it up.

junegunn:
> * Sometimes when the border style changes, some cells are not properly redrawn:
>   
>   * Changing the border shape let's the preview scrollbar disappear sometimes and leave parts of the old border:
>   
>   ```
>    fzf --preview='cat {}' --preview-window=border-horizontal --bind 'ctrl-space:change-preview-window(border-rounded|border-horizontal)'
>   ```
>   
>   * When changing the size and the border-shape of the preview, characters remain:
>   
>   ```
>    fzf --preview='cat {}' --preview-window=30%,left,border-rounded --bind 'ctrl-space:change-preview-window(50%,border-horizontal|30%)'
>   ```

Those problems can be fixed by completely erasing the windows before recreating them.

```diff
diff --git a/src/terminal.go b/src/terminal.go
index 28b2829..907e5cc 100644
--- a/src/terminal.go
+++ b/src/terminal.go
@@ -1541,10 +1541,12 @@ func (t *Terminal) resizeWindows(forcePreview bool) {
 		t.border.Close()
 	}
 	if t.window != nil {
+		t.window.Erase()
 		t.window.Close()
 		t.window = nil
 	}
 	if t.pborder != nil {
+		t.pborder.Erase()
 		t.pborder.Close()
 		t.pborder = nil
 	}
```

However, this causes noticeable flickering on every resize, which I don't think is unacceptable. Unfortunately, no easy fix comes to my mind.

junegunn:
> Those problems can be fixed by completely erasing the windows before recreating them.

It looks like [ca72bb0](https://github.com/junegunn/fzf/pull/4116/commits/ca72bb09fa89d3f9afccca6b817094515c711180) effectively fixes them.

junegunn:
d6c79780a05a5529d4814872594de89e42e16301 changes the way we calculate the desired width/height of the preview window. I think it's easier to understand and the code is shorter, but let me know if I'm missing something.

junegunn:
I'm going to merge this here, as all seems to work well as expected. Please let me know if you find any problems. Thanks for the contribution!

LangLangBart:
Looks fantastic and works so smooth.

druckdev:
Thanks for taking this over, finishing it and merging! I'm really happy to see this feature in the new version! And sorry for getting back to you so late, I didn't find the time earlier.

Everything looks good and also really nice that you fixed the issue with the artifacts! I only found the remaining issue that the scrollbar disappears as soon as I reach the maximum width of the preview window.

PR #4136, opened 2024-12-19
[fish] Leading characters: don't strip `.`, expand `~/`

bitraid:
A couple of more changes to fish key bindings: A fix for a bug that removed the leading dot of the query, and an enhancement that enables the expansion of `~/` (more info in commit messages). There is also a small syntax modification to some commands (the position of `--` that marks the end of switches), for consistency with the rest of the code (no actual change).


junegunn:
Thanks!

PR #4137, opened 2024-12-19
[fish] Partly revert change of 0167691

bitraid:
I remembered why I didn't use the `-f` switch of `string split` in my first PR: it was added in fish version 3.2.0, so it's not the right option for extracting the major/minor version number.


junegunn:
Thank you!

PR #4138, opened 2024-12-20
Replace bash to sh in Makefile

kimminss0:
Some operating systems do not ship with bash by default, e.g. BSDs, which breaks the build.

This PR switches to old-fashioned `sh` without introducing breaking changes.


junegunn:
Merged, thanks!

PR #4143, opened 2024-12-25
Don't include the file name in type information

xuhdev:
Reduce the changes of misjudging the type, e.g., when file is under an `image/`
directory.


junegunn (approved):

junegunn:
Thanks!

PR #4145, opened 2024-12-26
Support zellij floating pane via --popup (new name for --tmux)

junegunn:


noahbald:
Love this, it works super well, pretty much how I'd expect!

junegunn:
Thanks for testing it. Here's why this is still marked as a draft.

It currently doesn't work well with non-percent width and height values. Zellij lacks options to open floating windows at the center of the screen, unlike tmux (i.e. `-xC -yC`), and it makes it difficult, if not impossible, to get it right in a portable way.

Also, as I mentioned in the original discussion, [the lack of an option to run a program in a blocking mode](https://github.com/zellij-org/zellij/issues/3018) forces us to rely on another wrapper script using a temporary named pipe, and it complicates the implementation, which I find unfortunate.

Another problem I noticed is that Zellij somehow fails to process `0` and `0%` in the same way as reported in https://github.com/zellij-org/zellij/issues/3447.

```
zellij run --name fzf --floating --close-on-exit -x0 -- fzf --no-tmux
zellij run --name fzf --floating --close-on-exit -x0% -- fzf --no-tmux
```

And it makes us have silly code branches in our code, and I'm not so happy with it.

https://github.com/junegunn/fzf/pull/4145/files#diff-930cebf5a5e26b62d6004e9d0dca81108461c4fb37caa06877abfce271603719R46-R50

noahbald:
I did see that TODO so I hadn't bothered testing that part in particular.
I've seen that blocking mode problem before too which is unfortunate. Though looking at the wrapper script I don't find it too worrisome since it's at least contained within `zellij.go` -- though I've never worked with Go so I don't know what issues it may bring.
As for the other issues, are you going to wait until they're fixed in Zellij before continuing?

I appreciate you putting time into a fairly niche request anyway, especially given the time of year.

junegunn:
I'm going to ask if they're willing to make the floating pane start at the center of the screen by default. It's a little strange that it no longer starts at the center once a custom width or height is given.

Given that https://github.com/zellij-org/zellij/issues/3447 was reported 6 months ago, but no progress has been made, I don't think we can wait for all the issues to be resolved.

junegunn:
https://github.com/zellij-org/zellij/issues/3901

junegunn:
Another thing I would like to see from Zellij is an option to disable the border around the floating pane, because fzf has its own `--border` which is much more flexible and configurable. i.e. `--border-label`, `--color`, etc. Related #4155

imsnif:
Hey @junegunn - greetings from Zellij. I hope you don't mind the ping.

I have recently implemented all of your requests for integration. Specifically:
1. Specifying `0%` now works correctly when opening floating panes (thanks for the bug report).
2. Floating panes are now properly centered regardless of their fixed/percentage coordinates (no need for a special argument, this is now the default on both horizontal and vertical planes)
3. Floating panes can now optionally block until their command exited (by specifying `--block`, or optionally `--block-until-exit-failure`, `--block-until-exit-success`, etc.) The exit status from this command is also sent back to the caller.
4. It's now possible to open a floating pane (or indeed any pane) without drawing the pane frame (by specifying `--borderless`), allowing `fzf` to draw its own frame as you requested.

These changes have not been released yet, but have been merged into `main` and will be in the next release. I wanted to reach out before this happens to see if there's anything else you would need.

Finally - I apologize this took so long. I only recently became aware of these requests and made a note to address them before the next release. I take integrations with other tools and requests for such seriously and it's often hard for me to distinguish them from all other traffic an OSS project receives. In the future if you need anything similar, you are more than welcome to `@` me.

junegunn:
Hi @imsnif, that's great news! I'll revisit this with the latest Zellij and let you know how it goes.

imsnif:
> Hi @imsnif, that's great news! I'll revisit this with the latest Zellij and let you know how it goes.

Amazing. I'll ping here when the version is released.

imsnif:
Hey @junegunn - the [latest Zellij version](https://zellij.dev/news/remote-sessions-windows-cli/) now contains all the desired changes to support fzf.

There are some new application-centric doc pages that might make things easier for you: [cli recipes](https://zellij.dev/documentation/cli-recipes.html) and [programmatic control](https://zellij.dev/documentation/programmatic-control.html).

If you can't find how to do something or are missing functionality, please feel free to ping me. Thanks!

junegunn:
Thanks, now the code is much simpler than before. Please build and test it!

```sh
# Popup in the center with 70% width and height
fzf --popup 70%

# Popup on the left with 40% width and 100% height
fzf --popup right,40%

# Popup on the bottom with 100% width and 30% height
fzf --popup bottom,30%

# Popup on the top with 80% width and 40% height
fzf --popup top,80%,40%

# Popup with a native tmux or Zellij border in the center with 80% width and height
fzf --popup center,80%,border-native
```

imsnif:
Looks great!

image

PR #4155, opened 2025-01-02
Add support for --tmux-title option

afh:
Personally I see value in adding support for a `--tmux-title` option, with which a title for the tmux popup can be set (see `-T` option of `display-popup` command in [tmux(1)](https://man.openbsd.org/OpenBSD-current/man1/tmux.1#display-popup)). This option would only work in combination with the `--tmux` option and look better if the `--no-border` option is set as to avoid rendering two borders: one from fzf and one from tmux, rendering the tmux border is required in order to display the title.

Please note that this is just a minimal working proof-of-concept. If the maintainers and folks using fzf find this useful and would consider this for inclusion, I'd be happy to elaborate on this PR; ideally with some guidance from the maintainers given that I'm rather unfamiliar with fzf's codebase.

### DIFF
--- a/src/options.go
+++ b/src/options.go
@@ -78,6 +78,7 @@ Usage: fzf [options]
     --tmux[=OPTS]            Start fzf in a tmux popup (requires tmux 3.3+)
                              [center|top|bottom|left|right][,SIZE[%]][,SIZE[%]]
                              (default: center,50%)
+    --tmux-title=STR         Add given title to tmux popup (requires tmux 3.3+)
     --layout=LAYOUT          Choose layout: [default|reverse|reverse-list]
     --border[=STYLE]         Draw border around the finder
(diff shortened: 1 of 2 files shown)


junegunn:
I would suggest that you use `--border-label` instead. You can configure the position of the label using `--border-label-pos` and it also supports ANSI color codes. Do you think you still need to use `-T` option?

```sh
fzf --tmux --border-label ' Hello '

fzf --tmux --border-label ' Hello ' --border-label-pos 2

fzf --tmux --border-label $' \x1b[33mHello '
```

afh:
Thank you for the quick and helpful response, @junegunn, very much appreciated! Exactly what I'm looking for; not sure how I missed this when skimming through fzf(1) while looking for title.

PR #4157, opened 2025-01-03
add `border-native` option to `--tmux` flag

### LINKED ISSUE #4156: [feature request] add an option to enable tmux popup window borders
### Checklist

### Output of `fzf --version`

0.57.0

### OS

- [ ] Linux
- [X] macOS
- [ ] Windows
- [ ] Etc.

### Shell

- [ ] bash
- [ ] zsh
- [X] fish

Currently, the tmux popup window is launched with:
https://github.com/junegunn/fzf/blob/fb3bf6c9841d849ec459fc6b251b4aa0f16d8038/src/tmux.go#L36

The `-B` flag disables the popup window border. This seems to be a good default. However, there are situations where it would be nice if the popup window had a border. For example, I have a fzf command listing git-log entries and a key binding that executes a pager for viewing the commit message or entire commit (including diff).

Without a popup border, the pager output isn't visually separated from the background and harder to read.

Screenshot 2025-01-03 at 13 18 40

vs.

Screenshot 2025-01-03 at 13 19 41

If customizing this should be supported, I see two ways to do this:
- Allow tmux flags in the list of elements passed via `--tmux[=[center|top|bottom|left|right][,SIZE[%]][,SIZE[%]]]`
   For example, `--tmux[=[center|top|bottom|left|right][,SIZE[%]][,SIZE[%]],[-B],[-E],[-d DIR]]`. If no flags are
   specified, the current list of default flags is applied.
- Add a new flag `--tmux-flags` that contains a list of flags passed to `tmux popup`. Again, if not specified, the current
   list of default flags is applied.

aead:
This commit adds the `border-native` resulting in
the following:
```
--tmux[=[center|top|bottom|left|right][,SIZE[%]][,SIZE[%]][,border-native]]
```

By default, when not specified, the `-B` flag is passed to the
`tmux popup-window` command such that no border is drawn around
the tmux popup window.

When the `border-native` option is present the `-B` flag is omitted
and the popup window is drawn using the border style configured in
the tmux config file.

Fixes #4156

### DIFF
--- a/src/options.go
+++ b/src/options.go
@@ -310,8 +310,7 @@ func defaultTmuxOptions(index int) *tmuxOptions {
 		position: posCenter,
 		width:    sizeSpec{50, true},
 		height:   sizeSpec{50, true},
-		index:    index,
-	}
+		index:    index}
 }
 
 func parseTmuxOptions(arg string, index int) (*tmuxOptions, error) {
@@ -660,8 +659,7 @@ func defaultOptions() *Options {
 		WalkerRoot:   []string{"."},
 		WalkerSkip:   []string{".git", "node_modules"},
 		Help:         false,
-		Version:      false,
-	}
+		Version:      false}
 }
 
 func optString(arg string, prefixes ...string) (bool, string) {
@@ -1736,7 +1734,7 @@ func strLines(str string) []string {
 }
 
 func parseSize(str string, maxPercent float64, label string) (sizeSpec, error) {
-	spec := sizeSpec{}
+	var spec = sizeSpec{}
 	var val float64
 	var err error
 	percent := strings.HasSuffix(str, "%")
@@ -1822,8 +1820,7 @@ func parseInfoStyle(str string) (infoStyle, string, error) {
 	}
 	for _, spec := range []infoSpec{
 		{"inline", infoInline},
-		{"inline-right", infoInlineRight},
-	} {
+		{"inline-right", infoInlineRight}} {
 		if strings.HasPrefix(str, spec.name+":") {
 			return spec.style, strings.ReplaceAll(str[len(spec.name)+1:], "\n", " "), nil
 		}
--- a/src/options.go
+++ b/src/options.go
@@ -77,7 +77,7 @@ Usage: fzf [options]
                              (default: 10)
     --tmux[=OPTS]            Start fzf in a tmux popup (requires tmux 3.3+)
                              [center|top|bottom|left|right][,SIZE[%]][,SIZE[%]]
-                             (default: center,50%)
+                             [,border-native] (default: center,50%)
     --layout=LAYOUT          Choose layout: [default|reverse|reverse-list]
     --border[=STYLE]         Draw border around the finder
                              [rounded|sharp|bold|block|thinblock|double|horizontal|vertical|
--- a/src/options.go
+++ b/src/options.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"os"
 	"regexp"
-	"slices"
 	"strconv"
 	"strings"
 	"unicode"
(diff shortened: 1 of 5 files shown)


src/options.go
```diff
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"regexp"
+	"slices"
```
junegunn, line 8: https://github.com/junegunn/fzf/commit/a34e8dcdc99d1509667f74a5e24a41ace172757e

src/options.go
```diff
@@ -340,6 +343,15 @@ func parseTmuxOptions(arg string, index int) (*tmuxOptions, error) {
 		tokens = append([]string{"center"}, tokens...)
 	}
 
+	if i := slices.Index(tokens, "border-native"); i >= 0 {
+		// The border option, if present, is the last one
```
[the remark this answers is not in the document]
aead: The current code allows e.g.
```
--tmux 50%,border-native
```
or 
```
--tmux 50%,80%,border-native
```
but not
```
--tmux 50%,border-native,80%,
```

I'm fine dropping the order constraint - just want to make sure this is the behavior we want.

src/tmux.go
```diff
@@ -33,7 +33,10 @@ func runTmux(args []string, opts *Options) (int, error) {
 	// M        Both    The mouse position
 	// W        Both    The window position on the status line
 	// S        -y      The line above or below the status line
-	tmuxArgs := []string{"display-popup", "-E", "-B", "-d", dir}
+	tmuxArgs := []string{"display-popup", "-E", "-d", dir}
+	if !opts.Tmux.border {
+		tmuxArgs = append(tmuxArgs, "-B")
```
junegunn, line 38: We should also not add `--border` by default if native border is used.

src/options.go
```diff
@@ -650,7 +660,8 @@ func defaultOptions() *Options {
 		WalkerRoot:   []string{"."},
 		WalkerSkip:   []string{".git", "node_modules"},
 		Help:         false,
-		Version:      false}
+		Version:      false,
```
junegunn, line 663: Can you exclude unrelated style changes?

aead: Sry, sure. Editor auto-format. Reverted.

junegunn:
I updated the branch to prepend `--margin 0,1` so that it looks similar to when `--border` is used. What do you think? [fzf-tmux used to do that before tmux 3.3](https://github.com/aead/fzf/commit/b1619f675ff35c9bf3545d3ef3e3c396f9b6d164#diff-29747b0804ea04e1d230bd8651260a6b8687374cc1ae4c2d70ec26f05c4f719b) when there was no way to disable the native border.

```sh
fzf --tmux
fzf --tmux border-native
fzf --tmux --list-border
fzf --tmux border-native --list-border
```

One gotcha is that the extra margin is not colored when you use a non-default background color, but I don't think it's a critical problem.

```sh
fzf --tmux --color bg:#ff0000

fzf --tmux border-native --color bg:#ff0000

fzf --tmux border-native --color bg:#ff0000 --margin 0
```

aead:
The added margin does make it look nicer! For example,
Screenshot 2025-01-04 at 09 01 31
Screenshot 2025-01-04 at 09 00 01

And with `set -g popup-border-lines rounded` and coloring one can make both borders identical.


> One gotcha is that the extra margin is not colored when you use a non-default background color, but I don't think it's a critical problem.

Yeah, and with `--margin 0` you can turn it off if you really want to.

Since the native border is styled by tmux, FZF could forward its `--border` and `--color` styles to tmux to adjust things dynamically. However, even if, this may require a separate PR.

junegunn (approved):

junegunn:
Merged, thanks!

PR #4158, opened 2025-01-04
fix: make header align with list

phanen:
Currently, header is padded by hardcode length now.

When `pointer`/`marker` set to empty with, it would be better to align header with the list.
```sh
FZF_DEFAULT_OPTS= ./fzf --info=inline --pointer= --header-lines=3
```

fzf.vim `:Buffers`


junegunn:
Thanks. FWIW, I have this exact change locally as I'm working on #4159 :)

PR #4171, opened 2025-01-11
[fish] Optimize history formatting without perl

bitraid:
The previous bug fix made the operation 2x slower. With this change, it becomes almost as fast as before.

I have another change I've been testing locally, which begins to display the results instantly (but it's a bigger change and it depends on this code, so it's better to be in another PR). Currently, because the fzf command is being constructed using eval, it becomes blocking (fish blocks when executing builtins). Consider for example:
```fish
for i in (seq 5); echo $i; sleep 1; end | eval (echo fzf)
```
Versus:
```fish
for i in (seq 5); echo $i; sleep 1; end | fzf
```

### DIFF
--- a/shell/key-bindings.fish
+++ b/shell/key-bindings.fish
@@ -73,11 +73,10 @@ function fzf_key_bindings
           builtin history -z --reverse | command perl -0 -pe 's/^/$.\t/g; s/\n/\n\t/gm' | eval (__fzfcmd) --tac --read0 --print0 -q '(commandline)' | string replace -r '^\d*\t' '' | read -lz result
           and commandline -- $result
         else
-          set -l line 0
-          for i in (builtin history -z --reverse | string split0)
-            set line (math $line + 1)
-            string escape -n -- $line\t(string replace -a -- \n \n\t $i | string collect)
-          end | string join0 | string unescape -n | eval (__fzfcmd) --tac --read0 --print0 -q '(commandline)' | string replace -r '^\d*\t' '' | read -lz result
+          set -l h (builtin history -z | string split0)
+          for i in (seq (count $h) -1 1)
+            string join0 -- $i\t(string replace -a -- \n \n\t $h[$i] | string collect)
+          end | eval (__fzfcmd) --read0 --print0 -q '(commandline)' | string replace -r '^\d*\t' '' | read -lz result
           and commandline -- $result
         end
       else


junegunn:
Thanks!

> because the fzf command is being constructed using eval, it becomes blocking (fish blocks when executing builtins)

Interesting, good to know.

junegunn:
> because the fzf command is being constructed using eval, it becomes blocking (fish blocks when executing builtins)

Does this make sense? It seems to work fine.

```diff
diff --git a/shell/key-bindings.fish b/shell/key-bindings.fish
index 3155235..3d5c36e 100644
--- a/shell/key-bindings.fish
+++ b/shell/key-bindings.fish
@@ -70,13 +70,15 @@ function fzf_key_bindings
         set -lx FZF_DEFAULT_OPTS (__fzf_defaults "" "-n2..,.. --scheme=history --bind=ctrl-r:toggle-sort --wrap-sign '"\t"↳ ' --highlight-line $FZF_CTRL_R_OPTS +m")
         set -lx FZF_DEFAULT_OPTS_FILE ''
         if type -q perl
-          builtin history -z --reverse | command perl -0 -pe 's/^/$.\t/g; s/\n/\n\t/gm' | eval (__fzfcmd) --tac --read0 --print0 -q '(commandline)' | string replace -r '^\d*\t' '' | read -lz result
+          set -l fzf_command (__fzfcmd) --tac --read0 --print0 -q (commandline)
+          builtin history -z --reverse | command perl -0 -pe 's/^/$.\t/g; s/\n/\n\t/gm' | $fzf_command | string replace -r '^\d*\t' '' | read -lz result
           and commandline -- $result
         else
           set -l h (builtin history -z | string split0)
+          set -l fzf_command (__fzfcmd) --read0 --print0 -q (commandline)
           for i in (seq (count $h) -1 1)
             string join0 -- $i\t(string replace -a -- \n \n\t $h[$i] | string collect)
-          end | eval (__fzfcmd) --read0 --print0 -q '(commandline)' | string replace -r '^\d*\t' '' | read -lz result
+          end | $fzf_command | string replace -r '^\d*\t' '' | read -lz result
           and commandline -- $result
         end
       else
```

bitraid:
Yes sure, that will work. I was just hoping for the solution in https://github.com/junegunn/fzf/pull/4179. Too bad the test don't pass. What to you think if we added a check and pass the command in the pipe instead of using the variable, when running under the tests (stdin is not a TTY)? If you think it's not a good idea, its no problem - I can always use a modified version of the script, or just pass the full reload command through `FZF_CTRL_R_OPTS` (I want to be able to reload the list).

junegunn:
> when running under the tests (stdin is not a TTY)

The test runs on an interactive tmux window where you can interfere with your keyboard, so I don't think checking STDIN is a TTY device won't suffice. Anyway, I'm going to apply the above patch first, because not blocking is important.

bitraid:
> > when running under the tests (stdin is not a TTY)
> 
> The test runs on an interactive tmux window where you can interfere with your keyboard, so I don't think checking STDIN is a TTY device won't suffice. Anyway, I'm going to apply the above patch first, because not blocking is important.

We would set a global variable when the function is sourced. I already have it working.

PR #4189, opened 2025-01-19
Fix a non-constant format string

QuLogic:
Go 1.24 now has a vet check about this that causes `go test` to fail: https://github.com/golang/go/issues/60529


junegunn:
Thanks!

QuLogic:
I forgot to mention that this also affects your `go-shellwords` fork, but I didn't get a chance to write a fix.

QuLogic:
https://github.com/junegunn/go-shellwords/pull/1

PR #4200, opened 2025-01-23
[fish] Fix compatibility with v3.1.2 - v3.3.1

### LINKED ISSUE #4196: Fish shell integration broken in fzf 0.58.0
### Checklist

### Output of `fzf --version`

0.58.0 (brew)

### OS

- [x] Linux
- [ ] macOS
- [ ] Windows
- [ ] Etc.

### Shell

- [ ] bash
- [ ] zsh
- [x] fish

Running `fzf --fish | source` to set up fzf integration in the fish shell no longer works as it did in earlier versions but fails with an error due to the use of an `$(` construct:

```
 $ fzf --fish | source
- (line 79): $(...) is not supported. In fish, please use '(__fzfcmd)'.
      set -l result (eval "$FZF_DEFAULT_COMMAND | $(__fzfcmd) --read0 --print0 -q (commandline) --bind='enter:become:string replace -a -- \n\t \n {2..} | string collect'")
                                                  ^
from sourcing file -
source: Error while reading file “<stdin>”
```

I suspect that this change came in in v0.58.0

bitraid:
Don't use the command substitution syntax: $(cmd)

Fix #4196

---

I tested this change with versions 3.3.1, 3.1.2, and 3.0.2.

The first two work but the last has issues: The `string` command doesn't have the `collect` subcommand (was added in v3.1b1), and `cd` doesn't have the `--` switch (couldn't find the the version that was added).

Version 3.1.1 had broken `eval`. The remaining versions 3.1.0 and 3.1b1, may or may not work.

So, the oldest tested working version is 3.1.2.

### DIFF
--- a/shell/key-bindings.fish
+++ b/shell/key-bindings.fish
@@ -75,7 +75,7 @@ function fzf_key_bindings
           'string join0 -- $i\t(string replace -a -- \n \n\t $h[$i] | string collect);' \
           'end'
       end
-      set -l result (eval "$FZF_DEFAULT_COMMAND | $(__fzfcmd) --read0 --print0 -q (commandline) --bind='enter:become:string replace -a -- \n\t \n {2..} | string collect'")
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks a lot!

PR #4208, opened 2025-01-28
[fish] Improve fish binary path detection

bitraid:
Instead of exporting a local `$SHELL` containing the location of fish in `$PATH` when global `$SHELL` is not fish, always set `--with-shell` with the actual binary path of fish that the function is running from.

This change ensures that fzf is executing the shell commands using the same fish binary that the function is running with. Can be useful now that the upcoming version of fish can be self-installable (https://github.com/fish-shell/fish-shell/pull/10367), coexisting with a different system installed version, or when testing different builds of fish in general. Tested to work with fish v3.1.2.


junegunn:
Thanks!

PR #4213, opened 2025-01-30
Prepand PATH expansion instead of append

Rec0gnice:
This improves the evaluation of fzf when there is another version globally installed. So the custom downloaded/installed version takes precendence over the version installed by a packager manager for example.

### DIFF
--- a/install
+++ b/install
@@ -259,7 +259,7 @@ for shell in $shells; do
 # Setup fzf
 # ---------
 if [[ ! "\$PATH" == *$fzf_base_esc/bin* ]]; then
-  PATH="\${PATH:+\${PATH}:}$fzf_base/bin"
+  PATH="$fzf_base/bin\${PATH:+:\${PATH}}"
 fi
 
 EOF
--- a/test/lib/common.sh
+++ b/test/lib/common.sh
@@ -9,7 +9,7 @@ export FZF_DEFAULT_OPTS="--no-scrollbar --pointer '>' --marker '>'"
 # Setup fzf
 # ---------
 if [[ ! "$PATH" == *<%= BASE %>/bin* ]]; then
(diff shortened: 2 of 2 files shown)


junegunn:
Thanks, but I deliberately made the decision to append in fdaa4e9b18a761dd0c60a1d4429c9f6c3eb58c96. The reason is explained in the commit message. If I recall correctly, I made the change after fzf formula landed in Homebrew.

Since e74b1251c0f579335e03b3e7182cd7a9f88dbe37, I don't think the install script matters much. You only need to add a single line to your shell configuration file, and using a package manager like Homebrew is preferable — manual installation via the script makes it easy to miss updates.

Can you tell me more about your set-up? How do you install and update fzf?

Rec0gnice:
Thanks for the explanation, didn't know this was a deliberate design decision.
The reason for my PR is that I working with some setups that aren't managed by myself, i.e. I do not have the permissions to use a package manager.
In my latest company-managed system there was fzf already installed via apt on debian, but that was 20 minor versions behind. In this exact circumstance it resulted in the bash autocompletion not being supported.

Feel free to close the PR :)

junegunn:
Ah, that's unfortunate. Then you would have to
1. `./install --bin` to download the binary in `bin` directory
1. copy or symlink it to your local `bin` directory that has precedence
1. then manually [add a line to your shell configuration file](https://github.com/junegunn/fzf?tab=readme-ov-file#setting-up-shell-integration)

I can't think of a better solution now. But I'll let you know if something comes to my mind. Anyway, thanks for the suggestion.

bitraid:
Another issue that arises from appending the path, is when the install script is used to set up a profile for running the tests of some build, when another version is globally installed. This, of course, is not a big deal, since users that test changes should be able to easily make the adjustment manually. The only odd thing, is that this is only done for bash, while for zsh and fish the path is prepended.

PR #4230, opened 2025-02-08
[fish] Fix for directories with special characters

bitraid:
If the current command line token contained a directory with special characters, the script would fail to detect it.

### DIFF
--- a/shell/key-bindings.fish
+++ b/shell/key-bindings.fish
@@ -171,18 +171,20 @@ function fzf_key_bindings
         # if $dir is "." but commandline is not a relative path, this means no file path found
         set fzf_query $commandline
       else
+        # Special characters must be double escaped.
+        set -l re_dir (string escape -n -- $dir)
         # Also remove trailing slash after dir, to "split" input properly
-        set fzf_query (string replace -r -- "^$dir/?" '' $commandline)
+        set fzf_query (string replace -r -- "^$re_dir/?" '' $commandline)
       end
     end
 
-    echo (string escape -- $dir)
-    echo (string escape -- $fzf_query)
-    echo $prefix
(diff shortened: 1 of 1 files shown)


junegunn:
Can you mention some test cases in the commit message?

bitraid:
I added a commit message. Let me know if it needs more info.

junegunn:
Thanks, I've just tested the code and it does fix the issue.

```sh
mkdir "it's a test"
touch "it's a test"/foo
ls it\'s\ a\ test/<CTRL-T>
  # Works as expected
```

But I noticed a slight difference with bash and zsh's fuzzy completion.

```sh
ls it\'s\ a\ t<CTRL-T>
  # Starts fzf with "it\'s\ a\ t"
  # In bash and zsh, `ls it\'s\ a\ t**<TAB>`, starts fzf with "it's a t"
```

I don't think this is necessarily a problem, because CTRL-T of fish is not supposed to behave identical to them.  I'm just wondering how you feel about it.

bitraid:
This is not a regression from this change. Currently, in fish, the query from the command line token is processed exactly as if it was typed in fzf prompt. So, in this example, only the spaces should be escaped. This allows for less typing and not worrying about escaping and open quotes, but I understand that it can also be confusing, especially if the token for the query was created by the shell's auto-completion. We could probably change that, but I think we would have to use the `--is-valid` switch of `commandline` (and, similar to bash and zsh, not bring up fzf on open quotes), which was added in version 3.4.0 of fish. Let me know what you think.

junegunn:
Considering #4196 and this is not a regression, I think we should merge this as it is now.

junegunn:
Merged, thanks again!

PR #4231, opened 2025-02-08
Add 'exclude' action for excluding current/selected items from the result

### LINKED ISSUE #4185: [Feature Request] Action to remove/pop entries in the list
### Checklist

### Output of `fzf --version`

0.57.0 (0476a65)

### OS

- [x] Linux
- [ ] macOS
- [x] Windows
- [ ] Etc.

### Shell

- [x] bash
- [x] zsh
- [ ] fish

I have encounter the use case in which I would like to remove 1 or more items in the list displayed in fzf.

For simplicity take the following example:

```bash
printf 'top\nmiddle\nbottom' | fzf --bind 'ctrl-a:execute(operation that makes "bottom" invalid for selection)'
```

Where the options `top`, `middle` and `bottom` are all valid options. However if you press `ctrl-a` to run some extra processing, then the `bottom` option stops being valid and should not be selected anymore.

My current solution for this is to add a reload action and filtering out the invalid option(s). However depending on the case reloading and filtering could be expensive/slow or add a lot of extra complexity. My assumption is that removing/filtering out options in fzf is probably faster and easier as the search already does that when fuzzy matching to narrow options.

The proposed feature is to include a `pop` action (name used for example) that could be used to remove items from the list.

- Without arguments (`pop`) e.g. `--bind 'ctrl-a:execute(foo {1})+pop+execute(something {+f})`
  - When multiple entries are selected, they all get removed from the list.
  - When there is no selection (e.g. --no-multi), currently focused item is removed.
  - Template values (`{1}`, `{n}`, `{+f}`, etc.) should remain unchanged from the start of the --bind flag until the last action in the same --bind flag even if a `pop` action was used.
- With arguments (`pop(N)`) e.g. `--bind 'ctrl-a:execute(foo {1})+pop(5)+execute(something {+f})`
  - Argument `N` represents a number, range of numbers `1..2` or space separated list `1 2 3 4..6` that represent the index of the entries to remove.
  - Templates that represent indexes are valid `{n}` and `{+n}`

This is just a high level description of how the feature could work but it is open for discussion if the feature is considered.

Thanks.

junegunn:
Close #4185 

## Summary

Add `exclude` action to make the current item or the selected items excluded from the result.

## Discussion / Issues

* `exclude` "excludes" the selected items (`--multi`) at once if there's any, instead of excluding the "current" item on the cursor. Is this acceptable, or should we have a separate action? i.e. `exclude-current` + `exclude-selected`. I prefer the current implementation.
* The total item count does not decrease on `exclude`. I think this has a value of indicating that some items are excluded, as opposed to the list being fully "reloaded."
* To avoid the overhead of locking, we copy the exclusion list before passing it to the matcher. This is fine for small lists. But it does have performance problem if the exclusion list becomes excessively large.
  ```sh
  # But realistically, who would do this?
  seq 10000000 | fzf --sync --multi --bind start:select-all+exclude
  ```


junegunn:
Renaming it to `drop`.

junegunn:
On second thought, I think we need `exclude-current` that doesn't affect the selection.

* You are given a list, and you want to select some items.
* You carefully go through the list, select items one by one.
* You sometimes want to remove the items that you're never going to choose.
    * But you don't want to lose the previous selections!

junegunn:
On third thought, I think we should have `exclude` and `exclude-multi` instead of `exclude-current` and `exclude`.

PR #4232, opened 2025-02-09
Fix bin/fzf-preview.sh to make image preview work on ghostty terminal

Coko7:
Fix bin/fzf-preview.sh so that image preview works on [ghostty](https://github.com/ghostty-org/ghostty) terminal.

# Issue

Despite [ghostty supporting the Kitten graphics protocol](https://ghostty.org/docs/features), the image preview does not work even if `kitten` is installed on the system because the script only checks for the `KITTY_WINDOW_ID` environment variable, which is unset in ghostty terminal sessions.
So, when trying to preview images in ghostty, it merely prints the file type:
```console
coko7@example:~$ ./fzf-preview.sh arch_rainbow.png
arch_rainbow.png: PNG image data, 3840 x 2160, 8-bit/color RGBA, non-interlaced
```

# Solution

Check if `kitten` is installed on the system using `command -v kitten`.
See: https://sw.kovidgoyal.net/kitty/kittens_intro/

### DIFF
--- a/bin/fzf-preview.sh
+++ b/bin/fzf-preview.sh
@@ -58,7 +58,7 @@ elif ! [[ $KITTY_WINDOW_ID ]] && (( FZF_PREVIEW_TOP + FZF_PREVIEW_LINES == $(stt
 fi
 
 # 1. Use icat (from Kitty) if kitten is installed
-if command -v kitten > /dev/null; then
+if [[ $KITTY_WINDOW_ID ]] || [[ $GHOSTTY_RESOURCES_DIR ]] && command -v kitten > /dev/null; then
   # 1. 'memory' is the fastest option but if you want the image to be scrollable,
   #    you have to use 'stream'.
   #
--- a/bin/fzf-preview.sh
+++ b/bin/fzf-preview.sh
@@ -57,15 +57,15 @@ elif ! [[ $KITTY_WINDOW_ID ]] && (( FZF_PREVIEW_TOP + FZF_PREVIEW_LINES == $(stt
   dim=${FZF_PREVIEW_COLUMNS}x$((FZF_PREVIEW_LINES - 1))
 fi
 
-# 1. Use kitty icat on kitty terminal and ghostty terminal
-if [[ $KITTY_WINDOW_ID || $GHOSTTY_RESOURCES_DIR ]]; then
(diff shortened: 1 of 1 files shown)


junegunn:
Could you provide more details?

Coko7:
Hi, sorry for the lack of details. I have edited to include more. I hope it's better

bin/fzf-preview.sh
```diff
@@ -57,8 +57,8 @@ elif ! [[ $KITTY_WINDOW_ID ]] && (( FZF_PREVIEW_TOP + FZF_PREVIEW_LINES == $(stt
   dim=${FZF_PREVIEW_COLUMNS}x$((FZF_PREVIEW_LINES - 1))
 fi
 
-# 1. Use kitty icat on kitty terminal
-if [[ $KITTY_WINDOW_ID ]]; then
+# 1. Use kitty icat on kitty terminal and ghostty terminal
+if [[ $KITTY_WINDOW_ID || $GHOSTTY_RESOURCES_DIR ]]; then
```
junegunn, line 61: Does ghostty come with `kitty` command? If not, we should check if the command is available on the system.

Coko7: Good point.
I assumed it did, but I was wrong. It merely supports it.
In this case, I think it would be better to check with `command -v kitten` rather than env variables. I have pushed new changes.

bin/fzf-preview.sh
```diff
@@ -57,15 +57,15 @@ elif ! [[ $KITTY_WINDOW_ID ]] && (( FZF_PREVIEW_TOP + FZF_PREVIEW_LINES == $(stt
   dim=${FZF_PREVIEW_COLUMNS}x$((FZF_PREVIEW_LINES - 1))
 fi
 
-# 1. Use kitty icat on kitty terminal
-if [[ $KITTY_WINDOW_ID ]]; then
+# 1. Use icat (from Kitty) if kitten is installed
+if command -v kitten > /dev/null; then
```
junegunn, line 61: No, we should still check the environment variable. I have kitty installed so kitten is in the $PATH. So this passes regardless I use kitty or not.

junegunn: 231876a

junegunn:
Merged, thanks!

Coko7:
You are welcome!

PR #4262, opened 2025-02-20
[zsh/completion] don't unescape FZF_DEFAULT_OPTS

Lompik:
With zsh, binding `ctrl-\` in `FZF_DEFAULT_OPTS` breaks fzf completion trigger (but fzf-history-widget `ctrl-r` works). Turns out `__fzf_defaults` unescapes the backslash. This PR intends to fix that.

to reproduce: use config below and trigger fzf completions with `$FZF_COMPLETION_TRIGGER`:
```
export FZF_DEFAULT_OPTS='--bind=ctrl-\\:toggle-track'
```


junegunn:
Merged, thanks!

Issue #3188, opened 2023-02-27
fzf#shellescape() not working with powershell, causing downstream issues


evilnose:
- [ ] I have read through the manual page (`man fzf`)
- [x ] I have the latest version of fzf
- [x ] I have searched through the existing issues

## Info

- OS
    - [ ] Linux
    - [ ] Mac OS X
    - [x] Windows
    - [ ] Etc.
- Shell
    - [ ] bash
    - [ ] zsh
    - [ ] fish
    - [x] PowerShell

## Problem / Steps to reproduce
A previously related issue #2888 saw `fzf#shellescape()` not working correctly for PowerShell. That specific case was fixed by replacing the call with just `shellescape()`. However, all other places that are still using `fzf#shellescape()` still don't work for PowerShell.

For example, running `:GFiles` or `:Commits` on PowerShell will fail with `Not in git repository`, because [get_git_root()](https://github.com/junegunn/fzf.vim/blob/master/autoload/fzf/vim.vim#L649), which uses `fzf#shellescape()`, fails.

The root cause is the `^` escape added in `shellesc_cmd()` ([link](https://github.com/Kamholtz/fzf/blob/master/plugin/fzf.vim#L86)) which is incompatble with powershell. I think we should fully resolve this as proposed by the original reporter, by using `"` to quote the escaped string instead of `^"` ([link](https://github.com/junegunn/fzf/issues/2888#issuecomment-1193109689)).

junegunn:
The code has changed a lot since this, can you check if the current version works as expected now?

sim590:
I'm on the latest version 0.60.1 and this issue is still there. I added `echoerr fzf#shellescape(dir)` inside the function `s:get_git_root` and I saw that for my repository under `C:\myRepo`, the error message displayed as `^"C:\myRepo^"` which causes `GFiles` to display `Not in git repo`. If I remove the call to `fzf#shellescape` and just pass `dir`:
```
let root = systemlist('git -C ' . dir . ' rev-parse --show-toplevel')[0]
```
then the `GFiles` command works as expected.

There is still something wrong. I use Neovim on windows. Everything worked correctly until I added this in my Vimrc:

```
let &shell = executable('pwsh') ? 'pwsh' : 'powershell'
[...]

PR #4264, opened 2025-02-21
plugin/fzf: avoid calling shellesc_cmd when shell=pwsh

sim590:
This is an attempt at fixing #3188.

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -99,10 +99,11 @@ function! s:shellesc_cmd(arg)
 endfunction
 
 function! fzf#shellescape(arg, ...)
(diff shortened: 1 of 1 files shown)


sim590:
Actually, I'm thinking about it and I guess I should provide an implementation of `shellesc_pwsh` and call that function instead... I'll update my branch in a bit.

sim590:
My change seems to have broken the `:Files` command even though it fixed `:GFiles`...

sim590:
I'm thinking of abandoning this PR in favor of this one:

https://github.com/junegunn/fzf.vim/pull/1582

Issue #4265, opened 2025-02-21
last field is trimed when `--delimeter` is regex


phanen:
### Checklist

- [x] I have read through the manual page (`man fzf`)
- [x] I have searched through the existing issues
- [x] For bug reports, I have checked if the bug is reproducible in the latest version of fzf

### Output of `fzf --version`

0.60.0

### OS

- [x] Linux
- [ ] macOS
- [ ] Windows
- [ ] Etc.

### Shell

- [x] bash
- [ ] zsh
- [x] fish

### Problem / Steps to reproduce

```sh
echo a:b:c | fzf --delimiter [:] --accept-nth=2.. --bind 'load:accept'
```
print `b`, expected `b:c`

```sh
echo a:b:c | ./fzf --delimiter [:] --query c --nth=2.. --bind 'load:accept'
```
print nothing, expected `a:b:c`.

Seems cause by https://github.com/junegunn/fzf/commit/9abf2c8c9ca625f30ff0775316f51ad798a922f3.

PR #4266, opened 2025-02-21
Don't trim last field when delimiter is regex

phanen:
Fix https://github.com/junegunn/fzf/issues/4265.

### DIFF
--- a/test/test_core.rb
+++ b/test/test_core.rb
@@ -1771,6 +1771,13 @@ class TestCore < TestInteractive
       # Last delimiter and the whitespaces are removed
       assert_equal ['bar,bar,foo  :,:bazfoo'], File.readlines(tempname, chomp: true)
     end
+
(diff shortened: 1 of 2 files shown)


junegunn (approved):

test/test_core.rb
```diff
@@ -1773,6 +1773,15 @@ def test_accept_nth_regex_delimiter
     end
   end
 
+  def test_accept_nth_regex_delimiter_strip_last
+    tmux.send_keys %((echo "foo:,bar:,baz"; echo "foo:,bar:,baz:,qux:,") | #{FZF} --multi --delimiter='[:,]+' --accept-nth 2.. --sync --bind 'load:select-all+accept' > #{tempname}), :Enter
```
junegunn, line 1777: I added a test case where we should strip the last trailing delimiter.

junegunn:
Merged, thanks!

PR #4267, opened 2025-02-21
plugin/fzf: introduce fzf#with_sh for conveniently using s:use_sh

### LINKED ISSUE #3188: fzf#shellescape() not working with powershell, causing downstream issues

sim590:
As a dependency for the idea expressed in [this comment](https://github.com/junegunn/fzf.vim/pull/1582#issuecomment-2675058627) which in turn, if it is applied in `fzf.vim`, will fix #3188.

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -452,7 +452,7 @@ function! fzf#wrap(...)
   return opts
 endfunction
 
-function! fzf#use_sh()
+function! s:use_sh()
   let [shell, shellslash, shellcmdflag, shellxquote] = [&shell, &shellslash, &shellcmdflag, &shellxquote]
   if s:is_win
     set shell=cmd.exe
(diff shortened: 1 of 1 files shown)


sim590:
I'm closing this PR as it is not needed anymore due to the clarification that was brought in the discussion of [this other PR](https://github.com/junegunn/fzf.vim/pull/1582).

PR #4280, opened 2025-02-24
[fish] Enable multiple history commands insertion

bitraid:
I recently realized that the `become` action leaves behind temporary files, so I made some modifications to overcome this. While doing so, I thought it would be a good idea to allow inserting multiple commands when `--multi` is set. If for some reason you don't like this, I can remove this functionality and just adjust the code to not use `become`. If on the other hand you think that it should be enabled by default, I can make this change, too. Tested to work with the oldest supported fish version.


junegunn:
Thanks, I think this is very nice. I didn't do it in other shells, only because it was difficult to get it right.

shell/key-bindings.fish
```diff
@@ -116,10 +116,9 @@ function fzf_key_bindings
     set -l fzf_query (commandline | string escape)
 
     set -lx FZF_DEFAULT_OPTS (__fzf_defaults '' \
-      '--nth=2..,.. --scheme=history --bind=ctrl-r:toggle-sort --wrap-sign="\t↳ "' \
-      "--highlight-line --no-multi $FZF_CTRL_R_OPTS --read0 --print0" \
-      "--bind='enter:become:string replace -a -- \n\t \n {2..} | string collect'" \
-      '--with-shell='(status fish-path)\\ -c)
+      '--accept-nth=2.. --nth=2..,.. --scheme=history --wrap-sign="\t↳ "' \
+      "--bind=ctrl-r:toggle-sort --highlight-line --no-multi $FZF_CTRL_R_OPTS" \
```
junegunn, line 120:
```suggestion
      "--bind=ctrl-r:toggle-sort --highlight-line --multi $FZF_CTRL_R_OPTS" \
```

junegunn:
> I recently realized that the `become` action leaves behind temporary files

Could you clarify which temporary files you mean? I wonder if it's a bug of fzf.

bitraid:
> > I recently realized that the `become` action leaves behind temporary files
> 
> Could you clarify which temporary files you mean? I wonder if it's a bug of fzf.

The `/tmp/fzf-{temp,proxy)-*` files. I also thought that it might be bug of fzf, but after looking briefly at the code, I saw the comment below and figured that it's probably not, so I didn't look more into it:
https://github.com/junegunn/fzf/blob/e4489dcbc1e544ea15a9c89028ed29602ef7e941/src/terminal.go#L4919-L4922

shell/key-bindings.fish
bitraid: OK, done!

junegunn:
Thanks! Merged with a test case.

Issue #4282, opened 2025-02-24
Trailing newlines not getting stripped from items


alex-huff:
### Checklist

- [x] I have read through the manual page (`man fzf`)
- [x] I have searched through the existing issues
- [x] For bug reports, I have checked if the bug is reproducible in the latest version of fzf

### Output of `fzf --version`

0.60 (devel)

### OS

- [x] Linux
- [ ] macOS
- [ ] Windows
- [ ] Etc.

### Shell

- [ ] bash
- [x] zsh
- [ ] fish

### Problem / Steps to reproduce

Since commit https://github.com/junegunn/fzf/commit/84e2262ad63df2112f16b2a80fc661294c3da45e#diff-37113d7d967fa0fd21c0250890d9cda61c7d2591eeedd43017e9e7bc471b432aL138, newlines are no longer stripped from the end of each item when there are extra bytes (such as ANSI escape sequences) after the newline.

Before 84e2262:

https://github.com/user-attachments/assets/b3d2cf4e-f79e-4a32-a3d3-3e5a7aa5c663

After:

https://github.com/user-attachments/assets/9fd833be-d3b3-4ba9-8523-ae206501d260

Also, if you look at the after video I am moving back and forth between the first and second item and when I am focused on the second item the first item is no longer shown.

Here is a simple reproducing command:
```sh
printf "foo\n\e[0m\0bar\n\e[0m\0baz\n\e[0m\0" | ./fzf --read0 --ansi --with-nth ..
```

Before 84e2262:

![Image](https://github.com/user-attachments/assets/47f214ee-8a24-4760-9d37-366b8b9e83d4)

After:

![Image](https://github.com/user-attachments/assets/23e98eec-525c-4f15-8a9c-f7a4bdd5b2f5)

I will also try and find a reproducing command for the first item not showing.

alex-huff:
> There was a regression. Does [e4489dc](https://github.com/junegunn/fzf/commit/e4489dcbc1e544ea15a9c89028ed29602ef7e941) fix the problem?

No, I am currently on e4489dc

alex-huff:
https://github.com/user-attachments/assets/e9ed8e54-eba0-4992-ac34-06e7518dca99

Here is a reproducing command for the first item being truncated:
```sh
for _ in $(seq 19); do printf "foo\nbar\0"; done | ./fzf --read0
```
Change 19 so that the items no longer fit on screen.

junegunn:
[...]

PR #4285, opened 2025-02-25
Fix condition for using item numlines cache

alex-huff:
This fixes the bug where the first item was being truncated as discussed in #4282.


junegunn:
You're right. This was my oversight. Thanks for identifying the problem and fixing it.

PR #4314, opened 2025-03-17
Support setting preview window's size to achieve a specific items list size

komar007:
This PR implements negative preview window sizes which result in adjusting the preview window's size such that the items list has a specific size.

I didn't find a way to achieve this specific behavior where I want a full-size window with a specific small size of the list and a preview that takes the rest of the space.

### DIFF
--- a/man/man1/fzf.1
+++ b/man/man1/fzf.1
@@ -878,6 +878,9 @@ default until \fBtoggle\-preview\fR action is triggered.
 * If size is given as 0, preview window will not be visible, but fzf will still
 execute the command in the background.
 
+* If size is -N (negative and % is not used), the preview window will occupy the
+amount of space for which the entry list's size is N.
+
 * Long lines are truncated by default. Line wrap can be enabled with
 \fBwrap\fR flag.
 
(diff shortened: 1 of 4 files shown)


komar007:
Thanks, @junegunn,

I didn't know about adaptive height, I'll look into this and the other extra borders problem. I just tested with variations of `--border` and `--preview-window ...,border-none`

komar007:
I'm not sure I want to provide the extra dependencies to `calculateSize` to calculate this properly, because it doesn't seem like this was the intention. I guess setting item list height via a preview window setting is not the best idea after all.

Maybe it would be cleaner to implement another option (`--list-height`/`--max-list-height`) to explicitly set the number of items.

What I'm trying to achieve here is a full-size window in tmux mode that shows all provided items (there are always rather few - it is a tmux session switcher) with potentially some limit just in case, and with the preview window taking up the rest of the screen.

junegunn:
> Maybe it would be cleaner to implement another option (`--list-height`/`--max-list-height`) to explicitly set the number of items.

I guess so. But let's not rush into a decision. In this case, we also need to design how things should look when the preview window is not visible, or positioned on the right or left.

PR #4329, opened 2025-03-26
Add flag to disable the automatic return of the currently selected item.

stubbedev:
I am trying to use the fzf multiselect feature to stage files from git. This works great, except for one corner case.

```zsh
  function ga {
    local untracked_files=$(git status --porcelain | grep "??" | awk '{print $2}')
    local tracked_files=$(git status --porcelain | awk '/^.M|^MM|^R.M/ {print $2}' | sort -u)

    if [[ -n "$tracked_files" ]]; then
      local tracked_selected=$(printf "%s\n" $tracked_files | fzf --multi \
        --preview='git diff --color=always {} | bat --style=numbers --color=always' \
        --preview-window='up:90%:wrap:nohidden' \
        --prompt="(Tab: toggle file, Enter: submit selection, Ctrl+C: cancel selection): " \
        --header="Add tracked files" \
        --bind "tab:toggle" \
        --bind 'enter:accept' \
        --no-info \
        --exit-0)

      if [[ -n "$tracked_selected" ]]; then
        local tracked_selection=($(echo "$tracked_selected" | awk '{for(i=1; i<=NF; i++) print $i}'))
        for file in "${tracked_selection[@]}"; do
          git add "$file" -p
        done
      fi
    fi

    if [[ -n "$untracked_files" ]]; then
      local untracked_selected=$(printf "%s\n" $untracked_files |
        fzf --multi \
          --preview='bat --style=numbers --color=always {}' \
          --preview-window='up:90%:wrap:nohidden' \
          --prompt="(Tab: toggle file, Enter: submit selection, Ctrl+C: cancel selection): " \
          --header="Add untracked files" \
          --bind "tab:toggle" \
          --bind 'enter:accept' \
          --no-info \
          --exit-0)

      if [[ -n "$untracked_selected" ]]; then
        local untracked_selection=($(echo "$untracked_selected" | awk '{for(i=1; i<=NF; i++) print $i}'))
        for file in "${untracked_selection[@]}"; do
          git add "$file"
        done
      fi
    fi
  }
```

The issue I have is that pressing enter will add the single entry which is currently highlighted as the output. I want only the things which have been selected with the toggle or select function to be returned.

In order to allow me to do this easily I have added a flag: `--multi-no-auto-select`. If this flag is set, the current() item will not be returned as the result, unless it is in the selected map.

I think others may have similar use cases where an empty return from a populated input may be desirable.

### DIFF
--- a/src/options.go
+++ b/src/options.go
@@ -369,7 +369,8 @@ func defaultTmuxOptions(index int) *tmuxOptions {
 		position: posCenter,
 		width:    sizeSpec{50, true},
 		height:   sizeSpec{50, true},
-		index:    index}
+		index:    index,
+	}
 }
 
 func parseTmuxOptions(arg string, index int) (*tmuxOptions, error) {
@@ -566,6 +567,7 @@ type Options struct {
 	Wrap              bool
 	WrapSign          *string
 	MultiLine         bool
+	MultiNoAutoSelect bool
 	CursorLine        bool
 	KeepRight         bool
 	Hscroll           bool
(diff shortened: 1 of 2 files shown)


stubbedev:
@junegunn Thanks man, that is exactly what I was trying to do. :heart:

Issue #4242, opened 2025-02-14
automatic tty redirection should not statically target /dev/tty, but the current pty/pts


andreaskremser:
### Checklist

- [x] I have read through the manual page (`man fzf`)
- [x] I have searched through the existing issues
- [ ] For bug reports, I have checked if the bug is reproducible in the latest version of fzf

### Output of `fzf --version`

0.59.0 (bbe1721)

### OS

- [x] Linux
- [ ] macOS
- [ ] Windows
- [ ] Etc.

### Shell

- [x] bash
- [x] zsh
- [x] fish

### Problem / Steps to reproduce

Since 0.53.0 fzf will automatically redirect execute actions to /dev/tty,

so that the usual hack

```sh
ls | fzf --bind 'space:execute:vim {} < /dev/tty >/dev/tty'  > selected
```

could be simplified, without worrying about any redirection

```sh
ls | fzf --bind `space:execute:vim {}' > selected
```

but the static /dev/tty is not universally understood/not always a working solution.
A more robust approach would be to query the current pty (e.g. with /usr/bin/tty), which would return something like /dev/pts/4

The more general solution, that also works with programs like emacsclient, should probably be equivalent to something like this:

```sh
CURRENT_TTY=$(tty)
ls | fzf --bind "space:execute:emacsclient -t -a \"\" -c {} < $CURRENT_TTY > $CURRENT_TTY" > selected
```

junegunn:
Have you actually had the problem while using `emacsclient`? If so, can you describe what happens in that case?

It's odd because fzf writes directly to `/dev/tty`, so if it's not available, fzf shouldn't run at all, let alone `execute`.

https://github.com/junegunn/fzf/blob/ac32fbb3b2d58a7ba61a9a8b1c09420dc4870199/src/tui/light.go#L30

andreaskremser:
Yes emacsclient will always fail with simple /dev/tty redirection.  I guess the problem is, that it connects to the emacsdaemon, which no longer has access to a controlling terminal, but still can make use of the concrete pty device node you can obtain by asking $(tty). This is probably also the reason why the non-deamon version of emacs will work in this case.

This is not specific to emacsclient though, it's just the most likely candidate to observe this problem. 
The point is, a controlling terminal is typically available, that's why /dev/tty almost typically works, but in some cases it will fail. On the other hand, accessing the correct unix.Ttyname(os.Stdin.Fd()) device node should work reliably.

junegunn:
[...]

PR #4349, opened 2025-04-17
Use tty from stderr or fallback to /dev/tty 

tlxpro:
Fix https://github.com/junegunn/fzf/issues/4242

### DIFF
--- a/src/tui/light_unix.go
+++ b/src/tui/light_unix.go
@@ -43,14 +43,14 @@ func (r *LightRenderer) closePlatform() {
 }
 
 func openTty(mode int) (*os.File, error) {
+	tty := ttyname()
+	if len(tty) > 0 {
+		if in, err := os.OpenFile(tty, mode, 0); err == nil && util.IsTty(in) {
+			return in, nil
+		}
+	}
 	in, err := os.OpenFile(consoleDevice, mode, 0)
 	if err != nil {
-		tty := ttyname()
-		if len(tty) > 0 {
-			if in, err := os.OpenFile(tty, mode, 0); err == nil {
-				return in, nil
-			}
-		}
 		return nil, errors.New("failed to open " + consoleDevice)
 	}
 	return in, nil
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -45,4 +45,7 @@ jobs:
       run: make test
 
     - name: Integration test
+      # GitHub Actions run without a TTY device. This is a workaround to get one,
+      # https://github.com/actions/runner/issues/241
+      shell: 'script --return --quiet --log-out /dev/null --command "bash -e {0}"'
       run: make install && ./install --all && tmux new-session -d && ruby test/runner.rb --verbose
--- a/test/test_exec.rb
+++ b/test/test_exec.rb
@@ -380,7 +380,7 @@ class TestExec < TestInteractive
   end
 
   def test_become_tty
-    tmux.send_keys "sleep 0.5 | #{FZF} --bind 'start:reload:ls' --bind 'load:become:tty'", :Enter
+    tmux.send_keys "sleep 0.5 | #{FZF} --bind 'start:reload:ls' --bind 'load:become:tty' 2>/dev/null", :Enter
     tmux.until { |lines| assert_includes lines, '/dev/tty' }
   end
 


junegunn:
Not sure why the test is failing, it passes locally.

```sh
$ ruby test/runner.rb --name test_exit_0_exit_code
Run options: --name test_exit_0_exit_code --seed 29069

# Running:

.

Finished in 0.042943s, 23.2867 runs/s, 23.2867 assertions/s.

1 runs, 1 assertions, 0 failures, 0 errors, 0 skips
```

src/tui/light_unix.go
```diff
@@ -43,14 +43,14 @@ func (r *LightRenderer) closePlatform() {
 }
 
 func openTty(mode int) (*os.File, error) {
+	tty := ttyname()
+	if len(tty) > 0 {
+		if in, err := os.OpenFile(tty, mode, 0); err == nil && util.IsTty(in) {
```
junegunn, line 48: Is there a case where `util.IsTty(in)` returns false here, or are we just being more cautious?

tlxpro: I believe redirection to a file 2>err.log will succeed and not provide a tty.

tlxpro:
Not really sure what is going on here. The test environment might restrict IOCTL calls used in isTty. 

We could try to add;
```go
if tty != /dev/tty  {}
```
But this does not address the underlying issue.

tlxpro:
After thinking about it, the issue seem to be that /dev/tty does not exist in test environment. With the 2>/dev/null it now cannot fallback to proper pty.

junegunn:
Thanks.

The reason I'm not particularly happy with calling `ttyname()` first is that it's not really efficient. It takes around 0.4ms to 0.5ms on my macbook, and it may take longer on other devices with many entries under `/dev/pts` and `/dev`. While 0.5ms is barely noticeable, it's still an unnecessary cost in most common cases. I mean, for someone like me who doesn’t use Emacs, this just makes every fzf invocation slightly slower with no real benefit.

```diff
diff --git a/src/tui/ttyname_unix.go b/src/tui/ttyname_unix.go
index d0350a0b..7fb85545 100644
--- a/src/tui/ttyname_unix.go
+++ b/src/tui/ttyname_unix.go
@@ -14,7 +14,7 @@ var tty atomic.Value
 
 func ttyname() string {
 	if cached := tty.Load(); cached != nil {
-		return cached.(string)
+		// return cached.(string)
 	}
 
 	var stderr syscall.Stat_t
diff --git a/src/tui/tui_test.go b/src/tui/tui_test.go
index 3ba9bf35..a66c874a 100644
--- a/src/tui/tui_test.go
+++ b/src/tui/tui_test.go
@@ -18,3 +18,9 @@ func TestHexToColor(t *testing.T) {
 	assert("#102030", 16, 32, 48)
 	assert("#ffffff", 255, 255, 255)
 }
+
+func BenchmarkTtyname(b *testing.B) {
+	for i := 0; i < b.N; i++ {
+		ttyname()
+	}
+}
```

```
$ (cd src/tui && go test -run=Bench -bench=. -benchmem)
goos: darwin
goarch: arm64
pkg: github.com/junegunn/fzf/src/tui
cpu: Apple M3
BenchmarkTtyname-8          2907            410719 ns/op          114217 B/op       1594 allocs/op
PASS
ok      github.com/junegunn/fzf/src/tui 2.490s
```

So I'm wondering if we should make this behavior optional rather than the default. e.g. Add `--tty-default /dev/tty` option, and allow user to unset it with `--no-tty-default` and make fzf run `ttyname()`.

junegunn:
Here's an alternative approach: #4352

tlxpro:
I, myself, will not be needing this, as I also do not use emacs. I was having trouble with another tty allocation, and took this opportunity to learn the mechanics of how it was done in fzf. This alternative might also be useful for someone who wants to easily redirect to another tty.

As for the efficiency of ttyname, I do concur. It currently takes >1ms on my system.
```
/dev/ttys003
goos: darwin
goarch: amd64
pkg: github.com/junegunn/fzf/src/tui
cpu: Intel(R) Core(TM) i7-6700HQ CPU @ 2.60GHz
BenchmarkTtyname-8
    1014           1123381 ns/op          143853 B/op       1555 allocs/op
PASS
ok      github.com/junegunn/fzf/src/tui 1.753s
```

Although, I was able to get it down with an dirent optimization.
```/dev/ttys003
goos: darwin
goarch: amd64
pkg: github.com/junegunn/fzf/src/tui
cpu: Intel(R) Core(TM) i7-6700HQ CPU @ 2.60GHz
BenchmarkTtyname-8
    6963            165042 ns/op           32838 B/op          7 allocs/op
PASS
ok      github.com/junegunn/fzf/src/tui 1.644s
```

junegunn:
Thanks for the explanation. The optimization looks great, would you mind sharing what changes you made?

Anyway, I'd like to go with the alternative approach of adding `--no-tty-default`, because this is a very niche requirement and we’d rather avoid any additional overhead, however small.

tlxpro:
Here is the optimization I made. Inspired from https://github.com/ericlagergren/go-gnulib/blob/master/ttyname/ttyname_linux.go

```diff
diff --git a/src/tui/ttyname_unix.go b/src/tui/ttyname_unix.go
index d0350a0b..1b7b7cbe 100644
--- a/src/tui/ttyname_unix.go
+++ b/src/tui/ttyname_unix.go
@@ -3,9 +3,11 @@
 package tui
 
 import (
+	"bytes"
+	"golang.org/x/sys/unix"
 	"os"
 	"sync/atomic"
-	"syscall"
+	"unsafe"
 )
 
 var devPrefixes = [...]string{"/dev/pts/", "/dev/"}
@@ -17,25 +19,47 @@ func ttyname() string {
 		return cached.(string)
 	}
 
-	var stderr syscall.Stat_t
-	if syscall.Fstat(2, &stderr) != nil {
+	var stderr unix.Stat_t
+	if unix.Fstat(2, &stderr) != nil || (os.FileMode(stderr.Mode)&os.ModeCharDevice) != 0 {
 		return ""
 	}
 
+	ino := stderr.Ino
+
+	var dirent unix.Dirent
+	dbuf := bytes.Repeat([]byte{0xCD}, 1024*32)
+
 	for _, prefix := range devPrefixes {
-		files, err := os.ReadDir(prefix)
+		fd, err := unix.Open(prefix, unix.O_RDONLY, 0)
 		if err != nil {
 			continue
 		}
-
-		for _, file := range files {
-			info, err := file.Info()
-			if err != nil {
-				continue
+		defer unix.Close(fd)
+		for {
+			n, err := unix.ReadDirent(fd, dbuf)
+			if err != nil || n == 0 {
+				break
 			}
-			if stat, ok := info.Sys().(*syscall.Stat_t); ok && stat.Rdev == stderr.Rdev {
-				value := prefix + file.Name()
+
+			buf := dbuf[:n]
+			for len(buf) > 0 {
+				copy((*[unsafe.Sizeof(dirent)]byte)(unsafe.Pointer(&dirent))[:], buf)
+
+				buf = buf[dirent.Reclen:]
+
+				if dirent.Ino != ino {
+					continue
+				}
+
+				name := make([]byte, dirent.Namlen)
+				for i := 0; i < int(dirent.Namlen); i++ {
+					name[i] = byte(dirent.Name[i])
+				}
+				value := prefix + string(name)
 				tty.Store(value)
+				if value == "/dev/null" || dirent.Type != unix.DT_CHR {
+					return ""
+				}
 				return value
 			}
 		}

```

junegunn:
Thanks. I think I should keep the current implementation for now to keep things simple and easy to understand.

PR #4351, opened 2025-04-18
Disable tmux popup when already running inside one

peikk0:
`tmux popup` doesn't work and immediately returns instead when spawn inside one (`tmux popup tmux popup echo foo` results in an empty popup). For that reason, `fzf --tmux` doesn't work when executed inside a `tmux popup`, for example:

```shell
FZF_DEFAULT_OPTS=--tmux tmux popup 'find . | fzf'
```

And executing `fzf --tmux` inside `fzf --tmux` (as [`forgit`](https://github.com/wfxr/forgit) does, which was how I ran into this issue) also doesn't work, for example:

```shell
find . -type d | FZF_DEFAULT_OPTS=--tmux fzf --bind="enter:execute(find {} -type f | fzf | xargs less)"
```

A popup can be detected by the absence of the `TMUX_PANE` environment variable, so we can simply disable the `--tmux` mode when it is absent. And it also must be excluded for the exports in `runProxy()` so that the popup can also be detected properly by child processes.

### DIFF
--- a/src/core.go
+++ b/src/core.go
@@ -39,7 +39,7 @@ func (r revision) compatible(other revision) bool {
 // Run starts fzf
 func Run(opts *Options) (int, error) {
 	if opts.Filter == nil {
-		if opts.Tmux != nil && len(os.Getenv("TMUX")) > 0 && opts.Tmux.index >= opts.Height.index {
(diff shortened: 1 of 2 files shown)


junegunn (approved):

junegunn:
Thanks!

dbzix:
Yesterday, I spent hours guessing which update in my system broke my beautiful popups in `tmux`.
Thanks for [reverting this](https://github.com/junegunn/fzf/releases/tag/v0.61.3)!

PR #4356, opened 2025-04-20
Added some missing variables

RafaelDominiquini:
I was trying to create a complex command and needed some variables to restore the state of application, but I saw that I could not find some variables because they were not implemented!
Mainly the "FZF_POINTER" variable! I believe it is important that it is accessible! I exported the others to maintain consistency only.

Other thing: The variable 'FZF_LIST_LABEL' was already being exported, but not documented in the manpage!

Thanks.

### DIFF
--- a/man/man1/fzf.1
+++ b/man/man1/fzf.1
@@ -1279,9 +1279,19 @@ fzf exports the following environment variables to its child processes.
 .br
 .BR FZF_PROMPT "          Prompt string"
 .br
+.BR FZF_GHOST "           Ghost string"
+.br
+.BR FZF_POINTER "         Pointer string"
+.br
 .BR FZF_PREVIEW_LABEL "   Preview label string"
 .br
 .BR FZF_BORDER_LABEL "    Border label string"
+.br
+.BR FZF_LIST_LABEL "      List label string"
+.br
+.BR FZF_INPUT_LABEL "     Input label string"
+.br
+.BR FZF_HEADER_LABEL "    Header label string"
 .br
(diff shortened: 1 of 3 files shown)


junegunn:
> Mainly the "FZF_POINTER" variable! I believe it is important that it is accessible!

Would you mind sharing your use case?

Dominiquini:
> > Mainly the "FZF_POINTER" variable! I believe it is important that it is accessible!
> 
> Would you mind sharing your use case?

I was playing with this application today and wanted to make a more sophisticated script that lists and kills applications in the system.
For this, I had to use this variable (FZF_POINTER) to store the state of the application (if it should list all applications running or just user applications).
I was already using the variable FZF_PROMPT to hold another state (if I show a detailed list or a simplified one).

The other variables (FZF _*_ ​​LABEL) I exported because I thought it was missing, since some of them were already being exported and others did not. I couldn't think of any problem exporting more variables, so I did it!

If you think this is a bad idea, I ask you at least consider the "FZF_POINTER" variable!

Thanks.

junegunn:
I was a little concerned exporting too many variables would make process execution a little heavier. But I don't think the difference will be ever noticeable.

Not sure why the test is failing on CI as it passes locally.

Dominiquini:
In my opinion, any attribute that is possible to change during execution using actions (change-*, transform-*), must have some way to read during execution!

Thank you for considering this suggestion and looking at it!

junegunn:
Merged, thanks!

PR #4361, opened 2025-04-22
[fish] History improvements and small changes

bitraid:
The main change of this PR (first commit) is inspired by fish v4: `<ctrl-r>` operates only on the line at cursor, allowing to insert history entries when creating functions in the command line, writing begin/end blocks or constructing multiline commands with `<alt-enter>`.

A few other small changes are included:
- Make `<ctrl-r>` work when both perl and coreutils are missing.
- Simplify the `commandline` call in `fzf-file-widget`.
- Add fish version and dependencies check.

One feature that is present in the default history pager of fish, but not directly available in `fzf-history-widget`, is the ability to delete a history item (`<shift-delete>`). Of course, users can add their own binding in fzf, but the command is quite complex (in order to first restore the history command in its original form, and handle correctly multiline commands and trailing newlines). I think many users are missing out on this very useful feature. If you agree, I can add a commit that adds this functionality. The binding will be added before the user options, so it won't override any existing user bindings.

### DIFF
--- a/shell/key-bindings.fish
+++ b/shell/key-bindings.fish
@@ -147,7 +147,10 @@ function fzf_key_bindings
   end
 
   function fzf-history-widget -d "Show command history"
-    set -l fzf_query (commandline | string escape)
+    set -l -- command_line (commandline)
+    set -l -- current_line (commandline -L)
+    set -l -- total_lines (count $command_line)
+    set -l -- fzf_query (string escape -- $command_line[$current_line])
 
     set -lx FZF_DEFAULT_OPTS (__fzf_defaults '' \
       '--nth=2..,.. --scheme=history --multi --wrap-sign="\t↳ "' \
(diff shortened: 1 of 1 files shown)


shell/key-bindings.fish
```diff
@@ -160,21 +171,33 @@ function fzf_key_bindings
     if type -q perl
       set -a FZF_DEFAULT_OPTS '--tac'
       set FZF_DEFAULT_COMMAND 'builtin history -z --reverse | command perl -0 -pe \'s/^/$.\t/g; s/\n/\n\t/gm\''
-    else
+    else if type -q seq
```
junegunn, line 174: Are there any practical systems without `seq` but with fzf and fish? Just curious what made you think of that case.

junegunn:
> allowing to insert history entries when creating functions in the command line

I can confirm this works well, thanks!

> If you agree, I can add a commit that adds this functionality.

I think that's great. I've been hesitant about making the bindings more complex, because 1. it would increase the maintenance burden, 2. and because there wasn't a clear consensus on what extra features should be included. As for 1, please just don't disappear. Seriously, I can't maintain the code without your expertise. And for 2, we don't have to worry about it because it's a built-in feature of fish.

shell/key-bindings.fish
bitraid: Originally, I just wanted to replace `seq` with builtins. But after some benchmarks, I was surprised to find out that the new variation was about 40% slower, so I ended up adding it as fallback instead. But I agree that it probably won't have any practical use case. So, I better remove it then, I wasn't sure about adding it anyway.

bitraid:
>>If you agree, I can add a commit that adds this functionality.
>>
>I think that's great.

Awesome! I'll do some tests to make sure the command works on old fish versions, and I'll update the PR.

>please just don't disappear

Don't worry, I wont! Feel free to ping me anytime.

bitraid:
I've made the changes. I expect the command to work well, I've been using it for a long time. It also works fine with old fish versions.

junegunn:
Works well, thanks!

PR #4370, opened 2025-05-02
Add 'alt-bg' color for striped lines

### LINKED ISSUE #4354: Alternate list color
It would be nice to have the ability to alternate the list-bg/list-fg color.

Rofi has this feature:

Image

For multiline items or items that wrap, this would make it much easier to see where each item begins and ends.

junegunn:
### 'jump' should show alternating background colors even when 'alt-bg' is not defined as before.

  ```sh
  go run main.go --bind load:jump
  ```

Two differences:
  * The alternating lines will not be in bold (was a bug)
  * The marker column will not be rendered with alternating background color

### Use alternating background color when 'alt-bg' is set

  ```sh
  go run main.go --color bg:238,alt-bg:237
  go run main.go --color bg:238,alt-bg:237 --highlight-line
  ```

### 'selected-bg' should take precedence

  ```sh
  go run main.go --color bg:238,alt-bg:237,selected-bg:232 \
                 --highlight-line --multi --bind 'load:select+up+select+up'
  ```

---

Close #4354

### DIFF
--- a/src/terminal.go
+++ b/src/terminal.go
@@ -1258,7 +1258,7 @@ func (t *Terminal) ansiLabelPrinter(str string, color *tui.ColorPair, fill bool)
 	printFn := func(window tui.Window, limit int) {
 		if offsets == nil {
 			// tui.Col* are not initialized until renderer.Init()
-			offsets = result.colorOffsets(nil, nil, t.theme, *color, *color, t.nthAttr, false)
+			offsets = result.colorOffsets(nil, nil, t.theme, *color, *color, t.nthAttr)
 		}
 		for limit > 0 {
 			if length > limit {
@@ -2791,17 +2791,28 @@ func (t *Terminal) printItem(result Result, line int, maxLine int, index int, cu
 	_, selected := t.selected[item.Index()]
 	label := ""
 	extraWidth := 0
+	alt := false
+	altBg := t.theme.AltBg
+	selectedBg := selected && t.theme.SelectedBg != t.theme.ListBg
 	if t.jumping != jumpDisabled {
 		if index < len(t.jumpLabels) {
 			// Striped
-			current = index%2 == 0
+			if !altBg.IsColorDefined() {
+				altBg = t.theme.DarkBg
+				alt = index%2 == 0
+			} else {
+				alt = index%2 == 1
+			}
 			label = t.jumpLabels[index:index+1] + strings.Repeat(" ", util.Max(0, t.pointerLen-1))
 			if t.pointerLen == 0 {
 				extraWidth = 1
 			}
 		}
-	} else if current {
-		label = t.pointer
+	} else {
+		if current {
+			label = t.pointer
+		}
+		alt = !selectedBg && altBg.IsColorDefined() && index%2 == 1
 	}
 
 	// Avoid unnecessary redraw
@@ -2828,10 +2839,12 @@ func (t *Terminal) printItem(result Result, line int, maxLine int, index int, cu
 	maxWidth := t.window.Width() - (t.pointerLen + t.markerLen + 1)
 	postTask := func(lineNum int, width int, wrapped bool, forceRedraw bool) {
 		width += extraWidth
-		if (current || selected) && t.highlightLine {
+		if (current || selected || alt) && t.highlightLine {
 			color := tui.ColSelected
 			if current {
 				color = tui.ColCurrent
+			} else if alt {
+				color = color.WithBg(altBg)
 			}
 			fillSpaces := maxWidth - width
 			if wrapped {
@@ -2928,6 +2941,10 @@ func (t *Terminal) printItem(result Result, line int, maxLine int, index int, cu
 		} else {
 			base = tui.ColNormal
 			match = tui.ColMatch
+		}
+		if alt {
+			base = base.WithBg(altBg)
+			match = match.WithBg(altBg)
 		}
 		finalLineNum = t.printHighlighted(result, base, match, false, true, line, maxLine, forceRedraw, preTask, postTask)
 	}
@@ -3027,7 +3044,7 @@ func (t *Terminal) printHighlighted(result Result, colBase tui.ColorPair, colMat
 			sort.Sort(ByOrder(nthOffsets))
 		}
 	}
-	allOffsets := result.colorOffsets(charOffsets, nthOffsets, t.theme, colBase, colMatch, t.nthAttr, current)
+	allOffsets := result.colorOffsets(charOffsets, nthOffsets, t.theme, colBase, colMatch, t.nthAttr)
 
 	maxLines := 1
 	if t.canSpanMultiLines() {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,23 @@
 CHANGELOG
 =========
 
+0.62.0
+------
+- Added `alt-bg` color to create striped lines to visually separate rows
+  ```sh
+  fzf --color bg:237,alt-bg:238,current-bg:236 --highlight-line
+
+  declare -f | perl -0777 -pe 's/^}\n/}\0/gm' |
+    bat --plain --language bash --color always |
+    fzf --read0 --ansi --reverse --multi \
+        --color bg:237,alt-bg:238,current-bg:236 --highlight-line
+  ```
+- [fish] Improvements in CTRL-R binding (@bitraid)
+    - You can trigger CTRL-R in the middle of a command to insert the selected item
+    - You can delete history items with SHIFT-DEL
+- Bug fixes and improvements
+    - Fixed unnecessary 100ms delay after `reload`
+
 0.61.3
 ------
 - Reverted #4351 as it caused `tmux run-shell 'fzf --tmux'` to fail (#4559 #4560)
--- a/man/man1/fzf.1
+++ b/man/man1/fzf.1
@@ -270,6 +270,7 @@ color mappings.
     \fBcurrent\-bg (bg+)      \fRBackground (current line)
       \fBgutter              \fRGutter on the left
     \fBcurrent\-hl (hl+)      \fRHighlighted substrings (current line)
+    \fBalt\-bg                \fRAlternate background color to create striped lines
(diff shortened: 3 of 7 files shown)


src/terminal.go
```diff
@@ -2899,7 +2910,7 @@ func (t *Terminal) printItem(result Result, line int, maxLine int, index int, cu
 			}
 			return indentSize
 		}
-		finalLineNum = t.printHighlighted(result, tui.ColCurrent, tui.ColCurrentMatch, true, true, line, maxLine, forceRedraw, preTask, postTask)
```
alex-huff, line 2902: nitpick: `current` is always true here so this change wasn't meaningful. If it was done for readability then line 2947 should probably be updated similarly.

junegunn: That's correct. This was changed during the implementation process, but it is no longer needed.

junegunn:
@alex-huff How do you feel about

```diff
diff --git a/src/terminal.go b/src/terminal.go
index be4973d7..54897773 100644
--- a/src/terminal.go
+++ b/src/terminal.go
@@ -2800,7 +2800,7 @@ func (t *Terminal) printItem(result Result, line int, maxLine int, index int, cu
 		}
 		if index < len(t.jumpLabels) {
 			// Striped
-			alt = index%2 == 0
+			alt = index%2 == 1
 			label = t.jumpLabels[index:index+1] + strings.Repeat(" ", util.Max(0, t.pointerLen-1))
 			if t.pointerLen == 0 {
 				extraWidth = 1
@@ -2810,7 +2810,7 @@ func (t *Terminal) printItem(result Result, line int, maxLine int, index int, cu
 		if current {
 			label = t.pointer
 		}
-		alt = !selectedBg && altBg.IsColorDefined() && index%2 == 0
+		alt = !selectedBg && altBg.IsColorDefined() && index%2 == 1
 	}
 
 	// Avoid unnecessary redraw
```

I can't quite decide which one I prefer.

junegunn:
There are (subjective) pros and cons.

* `go run main.go --bind load:jump --highlight-line` doesn't look nice, because the first two consecutive lines are in the same background color.
    * That's because, in this case, `alt-bg` is undefined and we fall back to `current-bg`.
    * `go run main.go --bind load:jump --highlight-line --color alt-bg:237` looks acceptable
* Logically, it makes more sense to apply `alt-bg` to the second line, not the first.
* On the other hand, the input section (prompt and info) uses the default background color, so using `alt-bg` first can look more balanced to some eyes

alex-huff:
> There are (subjective) pros and cons.
> 
> * `go run main.go --bind load:jump --highlight-line` doesn't look nice, because the first two consecutive lines are in the same background color.

Yeah, I imagine binding jump to load is quite common. Since the current item will almost always still be the first item after loading it probably makes sense to use the alternate background on odd items.

alex-huff:
One thing I have noticed is that it is quite hard to mentally keep track of an item when scrolling through the list since its background color changes constantly.

https://github.com/user-attachments/assets/64ef8707-ddb9-4536-b289-452b794c0fc8

Try to keep track of `go.mod` when I am scrolling through the item list. The first time I scroll through the list each item preserves its background color when scrolling. The second time I scroll through the list the background color of items changes (which is the current behavior).

junegunn:
> It is quite a bit easier to visualize scrolling when items keep their background color.

I find the behavior distracting and not easy on the eyes. The whole list keeps flashing.

alex-huff:
> > It is quite a bit easier to visualize scrolling when items keep their background color.
> 
> I find the behavior distracting and not easy on the eyes. The whole list keeps flashing.

If your eyes are fixated on single spot on the screen than yes, I agree. If your eyes are following an item in the list as it scrolls than the opposite it true. In this case it is actually the current behavior that creates a flashing effect.

junegunn:
I understand your point. But when you're scrolling, you'll likely be looking at the "current" item under the cursor with `current-bg`, so the color is not changing there. Anyway, I tend to see the alternating colors as the colors of the fixed "slots" on screen rather than the colors of the "items".

> Sadly you can't just use `item.Index()` like I did since it breaks when items are filtered.

Trying to keep the background color of an item fixed defeats the initial purpose of making it easier to separate consecutive items on screen.

alex-huff:
> Trying to keep the background color of an item fixed defeats the initial purpose of making it easier to separate consecutive items on screen.

Yeah if items preserving their background was desired than I think you could use something like `merger.FindIndex(item.Index())` for the index so that consecutive items on the screen would always have different background colors but each item's background color would stay the same while scrolling (at least until the list is updated). I was just using `item.Index()` when I was testing what it would "feel" like.

> I understand your point. But when you're scrolling, you'll likely be looking at the "current" item under the cursor with `current-bg`, so the color is not changing there.

Sure this makes sense. When I initially noticed the difficulty in tracking items while scrolling I was trying to scroll the list until all the items where "fresh". I do this a lot while using fzf. I lock my eyes on whenever item is at the top of the screen and then scroll until that item leaves the bottom of the screen. However, I should just be using the `pg-up` action for this so this really is my fault.

> Anyway, I tend to see the alternating colors as the colors of the fixed "slots" on screen rather than the colors of the "items".

Makes sense.

junegunn:
> However, I should just be using the `pg-up` action for this so this really is my fault.

Nah, no one's wrong here. We all have different workflows and expectations. We just have to make some choices because a program can't be infinitely customizable.

---

Another thing to note. This `alt-bg` currently doesn't apply to gap lines.

image

I tried applying it to the gap lines, but it looked a little off.

image

---

> > There are (subjective) pros and cons.
> > 
> > * `go run main.go --bind load:jump --highlight-line` doesn't look nice, because the first two consecutive lines are in the same background color.
> 
> Yeah, I imagine binding jump to load is quite common. Since the current item will almost always still be the first item after loading it probably makes sense to use the alternate background on odd items.

One thing we can try is to do it differently on `jump` when no `alt-bg` is defined like so

```diff
diff --git a/src/terminal.go b/src/terminal.go
index f3a9a616..760a8a73 100644
--- a/src/terminal.go
+++ b/src/terminal.go
@@ -2795,12 +2795,14 @@ func (t *Terminal) printItem(result Result, line int, maxLine int, index int, cu
 	altBg := t.theme.AltBg
 	selectedBg := selected && t.theme.SelectedBg != t.theme.ListBg
 	if t.jumping != jumpDisabled {
-		if !altBg.IsColorDefined() {
-			altBg = t.theme.DarkBg
-		}
 		if index < len(t.jumpLabels) {
 			// Striped
-			alt = index%2 == 0
+			if !altBg.IsColorDefined() {
+				altBg = t.theme.DarkBg
+				alt = index%2 == 0
+			} else {
+				alt = index%2 == 1
+			}
 			label = t.jumpLabels[index:index+1] + strings.Repeat(" ", util.Max(0, t.pointerLen-1))
 			if t.pointerLen == 0 {
 				extraWidth = 1
@@ -2810,7 +2812,7 @@ func (t *Terminal) printItem(result Result, line int, maxLine int, index int, cu
 		if current {
 			label = t.pointer
 		}
-		alt = !selectedBg && altBg.IsColorDefined() && index%2 == 0
+		alt = !selectedBg && altBg.IsColorDefined() && index%2 == 1
 	}
 
 	// Avoid unnecessary redraw
```

But then again, I still can't decide which version I prefer 🫠

alex-huff:
I prefer when `alt-bg` doesn't apply to gap lines. It feels rather asymmetrical otherwise. Also, When `alt-bg` is applied to the gap line, the line looks like part of the item (almost like an underline) when logically it should separate items as the name 'gap' implies.

I think for most people `alt-bg` and `--gap` are accomplishing the same thing: making it easier to distinguish between different items when they are potentially multi-line/wrapped. For this reason it seems unlikely there would be a compelling reason to use both `alt-bg` and `--gap` together anyway.

alex-huff:
> One thing we can try is to do it differently on `jump` when no `alt-bg` is defined like so
> 
> ```diff
> diff --git a/src/terminal.go b/src/terminal.go
> index f3a9a616..760a8a73 100644
> --- a/src/terminal.go
> +++ b/src/terminal.go
> @@ -2795,12 +2795,14 @@ func (t *Terminal) printItem(result Result, line int, maxLine int, index int, cu
>  	altBg := t.theme.AltBg
>  	selectedBg := selected && t.theme.SelectedBg != t.theme.ListBg
>  	if t.jumping != jumpDisabled {
> -		if !altBg.IsColorDefined() {
> -			altBg = t.theme.DarkBg
> -		}
>  		if index < len(t.jumpLabels) {
>  			// Striped
> -			alt = index%2 == 0
> +			if !altBg.IsColorDefined() {
> +				altBg = t.theme.DarkBg
> +				alt = index%2 == 0
> +			} else {
> +				alt = index%2 == 1
> +			}
>  			label = t.jumpLabels[index:index+1] + strings.Repeat(" ", util.Max(0, t.pointerLen-1))
>  			if t.pointerLen == 0 {
>  				extraWidth = 1
> @@ -2810,7 +2812,7 @@ func (t *Terminal) printItem(result Result, line int, maxLine int, index int, cu
>  		if current {
>  			label = t.pointer
>  		}
> -		alt = !selectedBg && altBg.IsColorDefined() && index%2 == 0
> +		alt = !selectedBg && altBg.IsColorDefined() && index%2 == 1
>  	}
>  
>  	// Avoid unnecessary redraw
> ```
> 
> But then again, I still can't decide which version I prefer 🫠
> 
> I think I wanted to avoid the first two lines being darker together at start, but that really depends on the brightness of `alt-bg`.
> 
> ```shell
> # When alt-bg is darker
> go run main.go --color bg:238,current-bg:236,alt-bg:237 --highlight-line
> 
> # When alt-bg is brighter
> go run main.go --color bg:238,current-bg:236,alt-bg:239 --highlight-line
> ```

I think when the user chooses a reasonable `alt-bg` color (one that contrasts with both `list-bg` and `current-bg`) it isn't really a big deal (functionally or aesthetically) whether `index%2 == 0` or `index%2 == 1` is used. However when the alternate background color is the same as the current background color, which is generally the case when no `alt-bg` is specified, `index%2 == 0` is better.

PR #4377, opened 2025-05-09
bash: set keybinding right before printing special character

ajeetdsouza:
Currently, fzf sets the `\e[0n` keybinding upon initialization, which is sadly the only way to call readline functions in Bash. Unfortunately, this method is also used by tools like [zoxide](https://github.com/ajeetdsouza/zoxide), causing the keybinding to be overridden:

https://github.com/ajeetdsouza/zoxide/blob/095b270fea283ed0a9d147c2b154a80030f19a4c/templates/bash.txt#L160

Since the only purpose of this particular keybinding appears to be to hack around this particular limitation of Bash, the way I've handled this issue in zoxide is to set the keybinding _right_ before calling it via `printf '\e[5n'`. This PR does the same for fzf.

Context: I'm the author of zoxide, just trying to ensure our plugins don't clash :)

### DIFF
--- a/shell/completion.bash
+++ b/shell/completion.bash
@@ -31,9 +31,6 @@ if [[ $- =~ i ]]; then
 
 ###########################################################
 
-# To redraw line after fzf closes (printf '\e[5n')
-bind '"\e[0n": redraw-current-line' 2> /dev/null
-
 __fzf_defaults() {
   # $1: Prepend to FZF_DEFAULT_OPTS_FILE and FZF_DEFAULT_OPTS
   # $2: Append to FZF_DEFAULT_OPTS_FILE and FZF_DEFAULT_OPTS
@@ -328,6 +325,7 @@ __fzf_generic_path_completion() {
         else
           COMPREPLY=( "$cur" )
         fi
+        bind '"\e[0n": redraw-current-line' 2> /dev/null
         printf '\e[5n'
         return 0
       fi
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks. So this is to recover after other plugins have overridden the binding, right?

ajeetdsouza:
That is correct.

shell/completion.bash
```diff
@@ -31,9 +31,6 @@ if [[ $- =~ i ]]; then
 
 ###########################################################
 
-# To redraw line after fzf closes (printf '\e[5n')
```
junegunn, line 34: Can we keep this comment? It's not the most obvious part of the code. Let's put it above the first one.

ajeetdsouza: Done!

junegunn (approved):

junegunn:
Merged, thanks!

gnarendran:
I had an issue that bash login shell on console wouldn't honour (non-default) $INPUTRC for readline. Found that the early bind in /etc/bash_completion.d/fzf (version 0.62.0 completion.bash) via /etc/profile, was causing readline to initialize even before $INPUTRC was set in ~/.bash_profile, and caused the issue (as once initialized, readline wouldn't re-init unless forced to). Happily this pull request fixes that as well by moving the bind to where needed, thanks!

Issue #4382, opened 2025-05-12
Test driving on Cygwin is broken


clarity20:
### Checklist

- [x] I have read through the manual page (`man fzf`)
- [x] I have searched through the existing issues
- [x] For bug reports, I have checked if the bug is reproducible in the latest version of fzf

### Output of `fzf --version`

0.62.0 (d226d841)

### OS

- [ ] Linux
- [ ] macOS
- [x] Windows
- [ ] Etc.

### Shell

- [x] bash
- [ ] zsh
- [ ] fish

### Problem / Steps to reproduce

Hello,

I've just downloaded and installed the latest binary as per the **Using git** subsection of the README:

  ```
  git clone --depth 1 https://github.com/junegunn/fzf.git ~/.fzf
  ~/.fzf/install
  ```
  
Sanity-checking of the `--help` and `--version` options went fine:

![Image](https://github.com/user-attachments/assets/63cba9da-a144-467b-a5b9-4500bf8fe237)

However, "firing it up" with no command-line arguments gave the following message:

```
C:\cygwin64\tmp\fzf-temp-**3211332915**: line 4: C:\cygwin64\home\mawoo\.vim\bundle\fzf\target\fzf-windows_amd64.msys2_built: command not found
```

FWIW, the numeric codes in that temp file name change with every rerun.

xrstf:
I am affected by the same issue. For some reason, fzf won't work in a plain zsh shell in mintty, but when I run fzf in a zsh shell _in a tmux window_, it works just fine.

eabase:
@clarity20 
You're doing something wrong or very weird. Let me explain.

Works fine with **Cygwin**, and **MSYS** (in both `MSYS` and `MINGW64` environments.) However, there is no installable package for MSYS, but you can just soft link it to your `/usr/bin`. (In MINGW64 the package will install to `/mingw64/bin/`.

For **Cygwin**, you need to make sure it's available in your PATH. You can't just use a temp and expect it to work, as it uses sub-processes which will never get your locally defined variables. 

> [!IMPORTANT] 
> If you have other windows versions of Fzf in your Windows path, 
> **they have to be removed** before you can use the Cygwin ones.

Put your binaries in your (Cygwin) PATH, or in your Cugwin `$HOME` directory, under `/home/USER/.fzf/bin`.

Then use this in your `.bashrc`:

```bash
# Setup fzf:
if [[ ! "$PATH" == */home/USER/.fzf/bin* ]]; then
[...]

PR #4383, opened 2025-05-12
Fix fzf with no cmdline arguments on Cygwin

clarity20:
This addresses issue #4382.

In module `src/winpty_windows.go`, function `runWinpty()` is handing off the executable's full pathname to `runProxy()` as the first argument. The path was coming over Windows-style and causing Cygwin to choke. So I'm checking `uname -s` for Cygwin and calling `cygpath` to reformat if needed. Very basic test runs now looking good:

fzf test drive

### DIFF
--- a/src/winpty_windows.go
+++ b/src/winpty_windows.go
@@ -44,10 +44,26 @@ func needWinpty(opts *Options) bool {
 }
 
 func runWinpty(args []string, opts *Options) (int, error) {
-	argStr := escapeSingleQuote(args[0])
+	_argStr := ""
+
+	osname, _ := exec.Command("uname", "-s").Output()
+	if strings.Contains(strings.ToLower(string(osname)), "cygwin") {
+		out, err := exec.Command("cygpath", args[0]).Output()
+		if err == nil {
+			_argStr += strings.Trim(string(out), "\n")
+		} else {
+			_argStr += args[0]
+		}
+	} else {
+		_argStr += args[0]
+	}
+
+	argStr := escapeSingleQuote(_argStr)
+
 	for _, arg := range args[1:] {
 		argStr += " " + escapeSingleQuote(arg)
 	}
+
 	argStr += ` --no-winpty`
 
 	if isMintty345() {


xrstf:
This PR makes fzf work in my Cygwin64 installation on Windows 10.

src/winpty_windows.go
```diff
@@ -44,10 +44,26 @@ func needWinpty(opts *Options) bool {
 }
 
 func runWinpty(args []string, opts *Options) (int, error) {
-	argStr := escapeSingleQuote(args[0])
+	_argStr := ""
+
+	osname, _ := exec.Command("uname", "-s").Output()
```
junegunn, line 49: So we would have to start a process to determine if we're running Cygwin, but this has a cost. Can there be a lighter way to do it, like checking if an environment variable exists?

clarity20: You have a good point. Based on their respective man pages, in bash and zsh we can check that `os.Getenv("OSTYPE") == "cygwin"`, but fish makes no mention of OSTYPE and I don't see an option besides uname in that case. What are your thoughts?

clarity20: `os.Getenv()` picks up on `OSTYPE` but again that leaves the fish shell in the lurch. 
However, bash also has a `BASH_VERSINFO` and I see that `BASH_VERSINFO[5]="x86_64-pc-cygwin"` in my setup. I am looking for something analogous on fish. If I find it I will keep you updated.

xrstf: And telling folks to setup an environment variable on their system would not be an option? In case we cannot find a common pre-existing environment variable, I mean.

clarity20: I like your idea of having users set up their own envs. In my opinion they should also be reminded to set this value separately in each environment where they want to run `fzf`.

FWIW, my research inclines me to think there really isn't a "grand unified" approach through canonical environment or shell variables. See for instance [this discussion](https://stackoverflow.com/questions/3466166/how-to-check-ifrunning-in-cygwin-mac-or-linux) on Stack Overflow.

junegunn: So there is no special environment variable that is guaranteed to be present in Cygwin, right? That's unfortunate.

Just to be clear, IIUC, this issue is caused by the use of `winpty`, so what happens when you tell fzf not to use it?

```sh
fzf --no-winpty
```

DanSM-5: Maybe this can help. In windows it is almost warrantee that the `OS` environment variable will exist with the value `Windows_NT`. It seems to have been there since windows XP and from my experience it is very useful to detect windows https://stackoverflow.com/questions/2788720/environment-variable-to-determine-the-os-type-windows-xp-windows-7.

Then you can check if `BASH_VERSION`, `ZSH_VERSION` or `FISH_VERSION` exists. Each one of them should be available when running the respective shell in interactive mode.

Now, I do not now of any cygwin specific variable but you can check for the existence of  `MSYSTEM` which is set by msys or git bash. If it is not present, it should be very likely that you are running cygwin.

akinomyoga: You can simply check `PATH`.

When `fzf` is called from Cygwin, `PATH` should contain the `bin` directory of Cygwin for (almost) 100%. I think you also want to check MSYS2 paths because there should be the same problem in MSYS2, which is a fork of Cygwin. MSYS2 is the base system of MinGW and Git for Windows, so its user base isn't negligible. If you care about the cases where Cygwin/MSYS2 is installed in a strange path, you can instead check if `cygwin1.dll` / `msys-2.0.dll` / `msys-1.0.dll` exists in the `PATH`.

pjeby: fzf already has cygwin/msys detection and cygpath for the shell when running other commands, based on checking `SHELL` (or the first word of `--with-shell`) for a `/`.  Using the same approach for proxy execution would fix the problem for bash, and for any other shells if the user explicitly exports a `SHELL`.   (It would also be consistent in the sense that if you're having problems with fzf detecting your shell, you can just export `SHELL`.)

pjeby: A further update - it turns out you can just substitute all `\\` with `/` in `args[0]` before quoting, and it works fine without needing cygpath.  This is a reasonable thing to do on Windows since 1) Windows APIs allow forward slash, and 2) on this code path we're invoking sh or bash anyhow, so it's absolutely gonna be cygwin or msys and they want forward slashes.  (That is, Cygwin's bash will invoke `C:/whatever/fzf.exe` just fine, it's `C:\whatever\fzf.exe` it has a problem with.)

PR #4389, opened 2025-05-13
fix: respect winborder which introduced since neovim 0.11

jiz4oh:
Hi @junegunn, thanks for creating so awesome plugin. I love it and I use it with neovim everyday.

Neovim 0.11 had introduced a new global option named [`winborder`](https://github.com/neovim/neovim/pull/31074). And fzf had draw the border itself now, it's duplicated if I set `winborder` as `rounded`

PixPin_2025-05-13_17-51-03

This pr fix this by respect the global `winborder` option if it's a valid value and ignore the border style which passed to fzf.

PixPin_2025-05-13_18-07-00

### DIFF
--- a/plugin/fzf.vim
+++ b/plugin/fzf.vim
@@ -553,8 +553,12 @@ try
     let height = s:calc_size(&lines, dict.down, dict)
     let optstr .= ' --no-tmux --height='.height
   endif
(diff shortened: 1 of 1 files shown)


jiz4oh:
it looks amazing, I like it

junegunn:
Merged, thanks!

PR #4400, opened 2025-05-24
Reincorporate user settings (FZF_ALT_C_COMMAND, etc.) in shell-integration behavior

clarity20:
The `fzf` keyboard shortcuts (ALT-C, etc.) were not using my `FZF_*` environment settings until I made the following code change. This change ensures that once the env is parsed, the arguments specified in the env variables are forwarded to the winpty child process along with the commandline arguments.

### DIFF
--- a/main.go
+++ b/main.go
@@ -48,7 +48,7 @@ func exit(code int, err error) {
 func main() {
 	protector.Protect()
 
-	options, err := fzf.ParseOptions(true, os.Args[1:])
+	options, envArgs, err := fzf.ParseOptions(true, os.Args[1:])
 	if err != nil {
 		exit(fzf.ExitError, err)
 		return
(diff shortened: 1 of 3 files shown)


junegunn:
`FZF_ALT_C_COMMAND` is not used by the Go program, but by the shell script.

https://github.com/junegunn/fzf/blob/ffb6e28ca7ea6dbaaf5bdf3e0f71b45565cb7b67/shell/key-bindings.bash#L51

So I wouldn't modify the Go source code to make it work.

Maybe you're thinking of the other ones like `FZF_DEFAULT_OPTS`? winpty doesn't pass down the exported variables? Is it a known limitation of winpty?

junegunn:
I believe this was a bug introduced by https://github.com/junegunn/fzf/commit/aefb9a5bc41f92227e4bffa050caca0270b450ba, and 0a8ff7899c7d09457d3896fcce2d4dcffcc13e17 should fix it.

Please test it and let me know if it works for you.

PR #4405, opened 2025-06-03
[bash] Set shell options for pathname expansion `~/.ssh/config.d/*`

akinomyoga:
The current implementation of the shell function `__fzf_list_hosts` does not function as expected when the user sets options related to the pathname expansions.

The current implementation uses the pathname expansion `~/.ssh/config.d/*` to check the contents of the files in the directory `~/.ssh/config.d`.
- However, when `set -f` (aka `set -o noglob`) is set, the pathname expansion does not happen, and it would fail to find the files even if they exist.
- When `GLOBIGNORE` is set, some of the files are unexpectedly excluded from the result of the pathname expansion.
- When `dotglob` or `nocaseglob` is set, unwanted files (such as dotfiles) may match.
- When `failglob` is set and no files match the pattern `~/.ssh/config.d/*`, an error message would be printed to the terminal on an attempt of completion. It should be noted that the pathname expansion is performed before the redirection `2> /dev/null`, so the error message for the pathname-expansion failure is not suppressed by `2> /dev/null`.
- When `nullglob` is set and no files match the pattern `~/.ssh/config.d/*`, the literal path `'/home/<user>/.ssh/config.d/*'` would be passed to the `tail` command. This doesn't cause an issue because the error message from the `tail` command is suppressed by the redirection `2> /dev/null`, but it would be better not to pass the unmatched path from the beginning.

This PR fixes the problem by setting the proper options before attempting the pathname expansion `~/.ssh/config.d/*`. The options are set inside a subshell so as not to affect the user's options in the main shell.


akinomyoga:
The CI failure is unrelated to the changes in this PR:

```
error: `trunction` should be `truncation`
  --> ./test/test_layout.rb:995:18
    |
995 |   def test_label_trunction
    |                  ^^^^^^^^^
    |
Error: Process completed with exit code 2.
```

akinomyoga:
Thanks for merging PR #4406! I have rebased this PR on top of the latest master. Hope the CI tests pass.

junegunn:
Merged, thanks!

akinomyoga:
Thank you!

PR #4419, opened 2025-06-10
Implement asynchronous transform actions

### LINKED ISSUE #4418: Asynchronous transformation
## Summary

fzf offers a bunch of `transform-*` actions for transforming its internal state with the output of an external process.

```sh
fzf --bind 'focus:transform-header:file {}' --header-border
```

However, these transformations run synchronously and block the rendering during the execution, so you're not supposed to run non-trivial processes with them. This can be quite limiting.

```sh
fzf --bind 'focus:transform-header:sleep 3; file {}' --header-border
```

How about we add asynchronous versions of the actions to allow asynchronous execution of the commands.

## Things to be determined

* What should be the name of an asynchronous action?
    * `transform-something-async(...)` is clear but feels a bit too wordy.
    * The same goes for `async-transform-something(...)`.
    * `transform-something!(...)` ?
    * `xform-something(...)` ?
    * Ideas?
* The new naming scheme might contradict with the existing `reload` and `reload-sync` pair, though the meaning of the synchronicity there is a bit different
* How about making everything asynchronous?
    * This might break existing use cases where you run multiple transformations that depend on the state that is being mutated.
      ```sh
      fzf --bind 'space:transform-prompt(sleep 1; date)+transform-header(echo "Prompt is $FZF_PROMPT")'
      ```
    * We have to be careful to process the results in the specified order.

junegunn:
Close #4418

Example:

```sh
fzf --bind 'focus:&transform-header(sleep 2; date; echo {})'
```

```sh
fzf --style full --border --preview date \
    --bind 'focus:bg-transform-header(sleep 0.1; date)' \
    --bind 'focus:+bg-transform-footer(sleep 0.2; date)' \
    --bind 'focus:+bg-transform-border-label(sleep 0.3; date)' \
    --bind 'focus:+bg-transform-preview-label(sleep 0.4; date)' \
    --bind 'focus:+bg-transform-input-label(sleep 0.5; date)' \
    --bind 'focus:+bg-transform-list-label(sleep 0.6; date)' \
    --bind 'focus:+bg-transform-header-label(sleep 0.7; date)' \
    --bind 'focus:+bg-transform-footer-label(sleep 0.8; date)' \
    --bind 'focus:+bg-transform-prompt(sleep 0.9; echo "$(date)> ")' \
    --bind 'focus:+bg-transform-ghost(sleep 1.0; date)'
```

### DIFF
--- a/src/terminal.go
+++ b/src/terminal.go
@@ -2631,7 +2631,9 @@ func (t *Terminal) printPrompt() {
 
 	before, after := t.updatePromptOffset()
 	if len(before) == 0 && len(after) == 0 && len(t.ghost) > 0 {
-		w.CPrint(tui.ColGhost, t.ghost)
+		maxWidth := util.Max(1, w.Width()-t.promptLen-1)
+		runes, _ := t.trimRight([]rune(t.ghost), maxWidth)
+		w.CPrint(tui.ColGhost, string(runes))
 		return
 	}
 
@@ -5384,7 +5386,7 @@ func (t *Terminal) Loop() error {
 			currentInput := t.input
 			capture := func(firstLineOnly bool, callback func(string)) {
 				if a.t >= actBgTransform {
-					// &transform-*
+					// bg-transform-*
 					t.captureAsync(*a, firstLineOnly, callback)
 				} else if a.t >= actTransform {
 					// transform-*


junegunn:
Limiting the number of concurrent background processes to 200 was not enough, fzf can easily become a fork bomb. I think we need to limit the number of background processes per action type.

PR #4433, opened 2025-06-24
Clarify Search Syntax

Sneethe:
I just thought the example `.mp3$` wasn't great as I think it implies that `.` is a regex character in this context.

Plus I saw how someone could interpret it this way from: https://github.com/Aloxaf/fzf-tab/issues/403

 Feel free to close if you don't agree. 

P.S. Thank you very much for making fzf.


junegunn:
Thanks for the suggestion, but I'd like to keep it as is. For many users, the primary use case of fzf is file search, and the example shows how you can search by the file extension. I believe without `.`, the intention becomes less clear.

PR #4436, opened 2025-06-25
feat: history search

phanen:
A feature like readline's `history-search-{forward,backward}`.


src/terminal.go
```diff
@@ -6610,6 +6637,13 @@ func (t *Terminal) Loop() error {
...
 			}
 
+			// if t.history != nil &&
+			// 	a.t != actPrevHistory &&
+			// 	a.t != actNextHistory &&
+			// 	a.t != actPrevHistorySearch &&
+			// 	a.t != actNextHistorySearch {
+			// 	t.history.lastSearch = ""
+			// }
 			if !processExecution(a.t) {
```
phanen, line 6647: I'm not sure when/how to reset current search...

* Readline reset current search and reset history pointer when cursor moved or cmdline changed.
* Vim's `:h c_<up>` reset current search but don't reset history pointer. (And it's actually search prefix like rather then substr).

phanen:
Close as I find it not very useful for me.

PR #4442, opened 2025-07-01
Will that fix the crash? I made a test build on my Linux and it worked after this change

nugged:
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x70 pc=0x470c6a2]

goroutine 9 [running]:
github.com/junegunn/fzf/src.(*Terminal).printFooter(0xc00020f508)
    github.com/junegunn/fzf/src/terminal.go:2936 +0x122
github.com/junegunn/fzf/src.(*Terminal).Loop.func8.2(0xc00011c918)
    github.com/junegunn/fzf/src/terminal.go:5161 +0x3b6
github.com/junegunn/fzf/src/util.(*EventBox).Wait(0xc00011c918, 0xc000207fa0)
    github.com/junegunn/fzf/src/util/eventbox.go:34 +0x5d
github.com/junegunn/fzf/src.(*Terminal).Loop.func8()
    github.com/junegunn/fzf/src/terminal.go:5090 +0xd7
created by github.com/junegunn/fzf/src.(*Terminal).Loop in goroutine 22
    github.com/junegunn/fzf/src/terminal.go:5064 +0x61c

### DIFF
--- a/src/terminal.go
+++ b/src/terminal.go
@@ -2928,7 +2928,7 @@ func (t *Terminal) printHeader() {
 }
 
 func (t *Terminal) printFooter() {
-	if len(t.footer) == 0 {
+	if len(t.footer) == 0 || t.footerWindow == nil {
 		return
 	}
 	indentSize := t.headerIndent(t.footerBorderShape)


junegunn:
Hi @nugged, thanks for the patch. I think it's enough to just check if `t.footerWindow` is null, because footer was not designed to be printed inside the list window. What do you think?

```diff
diff --git a/src/terminal.go b/src/terminal.go
index 9c627e12..77e3cd6e 100644
--- a/src/terminal.go
+++ b/src/terminal.go
@@ -2928,7 +2928,7 @@ func (t *Terminal) printHeader() {
 }
 
 func (t *Terminal) printFooter() {
-	if len(t.footer) == 0 {
+	if len(t.footer) == 0 || t.footerWindow == nil {
 		return
 	}
 	indentSize := t.headerIndent(t.footerBorderShape)
```

nugged:
testing the build...

nugged:
It works in my Linux ❤️

nugged:
@junegunn what if we to check that size of window is zero too?

```diff
 func (t *Terminal) printFooter() {
-	if len(t.footer) == 0 {
+	if len(t.footer) == 0 || t.footerWindow == nil || t.footerWindow.Height() {
 		return
 	}
 	indentSize := t.headerIndent(t.footerBorderShape)
```

yet it depends from your general practice how it is build the whole app, dunno yet :) - maybe we need to run logic below even if 0 height or that "never happens" - didn't learned the whole codebase and rules yet.

junegunn:
The height can become zero if you shrink the window all the way down. But the code runs fine even then, and I prefer to have fewer branching conditions.

That said, I'm looking into why this happened in the first place, and I'll probably fix it using a different approach.

junegunn:
d34675d3c9c0f11e42d29c705e261eb4df794f25 fixes the fundamental problem that caused the panic. Really appreciate you taking the time to dig into it and report it!

Issue #4443, opened 2025-07-03
Cross-platform alternative for fzf --bind 'ctrl-g:become(printf {q})'


shaobosong:
### Checklist

- [x] I have read through the manual page (`man fzf`)
- [x] I have searched through the existing issues
- [ ] For bug reports, I have checked if the bug is reproducible in the latest version of fzf

### Output of `fzf --version`

0.62.0 (d226d841)

### OS

- [x] Linux
- [x] macOS
- [x] Windows
- [x] Etc.

### Shell

- [x] bash
- [x] zsh
- [x] fish

### Problem / Steps to reproduce

I'm currently using the following fzf command:
```shell
fzf --bind "ctrl-g:become(printf {q})"
```
When Ctrl-G is pressed, outputs the current query and exits
I'm looking for a cross-platform alternative solution that can achieve the same functionality on all major platforms (Linux, macOS, Windows).

junegunn:
~~`--bind ctrl-g:print-query`~~

PR #4444, opened 2025-07-03
Add 'print-query' action to fzf man page

shaobosong:
[Issue 4443](https://github.com/junegunn/fzf/issues/4443#issuecomment-3030731083)

### DIFF
--- a/man/man1/fzf.1
+++ b/man/man1/fzf.1
@@ -1759,6 +1759,7 @@ A key or an event can be bound to one or more of the following actions.
     \fBpreview\-bottom\fR
     \fBpreview\-top\fR
     \fBprint(...)\fR                   (add string to the output queue and print on normal exit)
+    \fBprint\-query\fR                  (print query string and normal exit)
     \fBput\fR                          (put the character to the prompt)
     \fBput(...)\fR                     (put the given string to the prompt)
     \fBrefresh\-preview\fR


junegunn:
I remembered I intentionally removed that from the man page in the past.

https://github.com/junegunn/fzf/commit/a90426b7ca59dfe9f97110caf533c450b6855a77#diff-519ee16445a8adb05632a100d7ae9de2cd49afa0805e94be2457a0751e83d04bR443

Because I thought it could be confusing and there are alternative ways.

https://github.com/junegunn/fzf/discussions/4188#discussioncomment-11878896

So I thought of phasing it out, but it stayed. Do you think we should keep it?

shaobosong:
> I remembered I intentionally removed that from the man page in the past.
> 
> [a90426b#diff-519ee16445a8adb05632a100d7ae9de2cd49afa0805e94be2457a0751e83d04bR443](https://github.com/junegunn/fzf/commit/a90426b7ca59dfe9f97110caf533c450b6855a77#diff-519ee16445a8adb05632a100d7ae9de2cd49afa0805e94be2457a0751e83d04bR443)
> 
> Because I thought it could be confusing and there are alternative ways.
> 
> [#4188 (comment)](https://github.com/junegunn/fzf/discussions/4188#discussioncomment-11878896)
> 
> So I thought of phasing it out, but it stayed. Do you think we should keep it?

The functionality of 
```sh
fzf --expect ctrl-g --print-query
```
is more powerful as it encompasses what `fzf --bind=ctrl-g:print-query` offers, while additionally providing the capability to output key bindings. This enables backend systems to map different functionalities based on the output. I think it's wonderful!

Removing the redundant option might indeed be the right decision.

junegunn:
Glad to hear that you found the right solution for you. Yeah, `print-query` action can be confusing because:
* There's no telling if the output is from `print-query` or from `accept`
* The behavior is quite different from `print(...)`.

PR #4451, opened 2025-07-10
Fix: Inverted match logic in extendedMatch

Mirza-Samad-Ahmed-Baig:
This pull request addresses a bug in the extendedMatch function where inverted search terms were not being handled correctly. Previously, if an inverted term in a term set resulted in a match, the code would simply continue to the next term. This could lead to incorrect search results, as the entire term set would be considered a match if any subsequent non-inverted term also matched. The fix ensures that if an inverted term matches, the extendedMatch function immediately returns, correctly indicating that the term set is not a match. This change ensures that inverted search queries behave as expected.

### DIFF
--- a/src/pattern.go
+++ b/src/pattern.go
@@ -394,7 +394,8 @@ func (p *Pattern) extendedMatch(item *Item, withPos bool, slab *util.Slab) ([]Of
 			off, score, pos := p.iter(pfun, input, term.caseSensitive, term.normalize, p.forward, term.text, withPos, slab)
 			if sidx := off[0]; sidx >= 0 {
 				if term.inv {
-					continue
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks for your interest in the project.

> This could lead to incorrect search results

It does not lead to an incorrect result, thanks to this condition:

https://github.com/junegunn/fzf/blob/069d71a8401530028e2ea407904462ef8c435d88/src/pattern.go#L347

But please let me know if you do find a query that generates incorrect results.

---

Your patch unfortunately can break queries with OR-operators:

```sh
# This should print 1000
echo 1000 | fzf --filter '!100 | 100'
```

PR #4461, opened 2025-07-16
Work even when typing in the wrong language in a multilingual keyboard

shaiay:
As a user of multilingual keyboard, I sometimes search using fzf but forget to switch the keyboard language to English

This patch enables a mode where non-English characters are mapped to they keyboard equivalent English characters -- the English character sitting on the same key in the standard QWERTY keyboard, so even though you did not switch keyboard, you would get search results as though you were typing in English

A bit esoteric, but helps a bit ...

Includes implementation for Hebrew and Russian, easy to add more
90% written by gemini-cli

### DIFF
--- a/src/options.go
+++ b/src/options.go
@@ -45,6 +45,8 @@ Usage: fzf [options]
     -d, --delimiter=STR      Field delimiter regex (default: AWK-style)
     +s, --no-sort            Do not sort the result
     --literal                Do not normalize latin script letters
+    --keymap-convert, +C     Enable keyboard layout conversion (e.g. Hebrew to QWERTY)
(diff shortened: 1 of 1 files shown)


nenahp:
maybe this be achieved by define a `start:execute-silent:switch-cmd` action?

shaiay:
I ma not sure I understand, but this implementation is activated by a command line switch

nenahp:
that's it. i wonder if it's technically possible to have a command to switch? ok.. in your case maybe it's a bit tricky since you have to talk with the keyboard rather than the os.

(btw there are software level keyboards like kanata. off topic here but it's great if fzf is not the only case)

PR #4468, opened 2025-07-24
fix(shell): check for mawk existence before version check

### LINKED ISSUE #4463: Take 4 seconds show command line
### Checklist

### Output of `fzf --version`

Last version 

### OS

- [x] Linux
- [ ] macOS
- [ ] Windows
- [ ] Etc.

### Shell

- [ ] bash
- [x] zsh
- [ ] fish

I set fzf  search history Ctrl+r .
 press Ctrl+r and search history when enter select command history The problem is that it is 4 seconds late to be added or show command line

LangLangBart:
fix #4463

### problem

The `__fzf_exec_awk` helper function calls `mawk -W version` without checking if `mawk` exists. If `mawk` is not installed, this can trigger a slow `command_not_found_handler` (e.g., from
`ftc.zsh`), causing a noticeable delay.

### to reproduce

1. Run this Docker command to simulate a slow `command_not_found_handler`[^1] on `alpine:edge`[^2]:

```bash
docker run --rm --interactive --tty -e TERM alpine:edge sh -uec '
apk add fzf perl zsh
{
  echo "source <(fzf --zsh)"
  echo "command_not_found_handler() { sleep 4; return 127;}"
  echo "typeset -fT fzf-history-widget __fzf_exec_awk"
  echo "PS4='\''%B%F{0}+ %D{%T:%3.} %2N:%i%f%b '\''"
} > ~/.zshrc
/bin/zsh'
```

2. Press `CTRL-R`, and `Enter`.

**Trace showing the delay:**
```bash
# ... ~4s  ...
+ 17:16:15:072 fzf-history-widget:18 [[ $(__fzf_exec_awk '{print $1; exit}' <<< "$selected") -regex-match ^[1-9][0-9]*+ 17:16:15:077 fzf-history-widget:18 __fzf_exec_awk '{print $1; exit}'
+ 17:16:15:077 __fzf_exec_awk:1 [[ -z '' ]]
+ 17:16:15:077 __fzf_exec_awk:2 __fzf_awk=awk
+ 17:16:15:077 __fzf_exec_awk:3 [[ linux-musl == solaris* ]]
+ 17:16:15:078 __fzf_exec_awk:6 local n x y z d
+ 17:16:15:078 __fzf_exec_awk:7 mawk -W version
+ 17:16:19:084 __fzf_exec_awk:7 IFS=' .' + 17:16:19:084 __fzf_exec_awk:7 read n x y z d
+ 17:16:19:084 __fzf_exec_awk:8 [[ '' == mawk ]]
+ 17:16:19:084 __fzf_exec_awk:11 LC_ALL=C awk '{print $1; exit}'
+ 17:16:15:072 fzf-history-widget:18 [[ $(__fzf_exec_awk '{print $1; exit}' <<< "$selected") -regex-match ^[1-9][0-9]* ]]
+ 17:16:19:086 fzf-history-widget:21 LBUFFER=$'\t'
```

### solution

check for the existence of the executable before checking its version

changed once in `shell/common.sh` and run `shell/update-common.sh`, nice 

### related

#4412

[^1]: see `man zshmisc | less --pattern 'command_not_found_handler'` for more infos
[^2]: alpine:edge is minimal and fast, and has a recent 'fzf' version for the package manager

### DIFF
--- a/shell/common.sh
+++ b/shell/common.sh
@@ -22,12 +22,12 @@ __fzf_exec_awk() {
       # modern point of view.  To use a standard-conforming version in Solaris,
       # one needs to explicitly use /usr/xpg4/bin/awk.
       __fzf_awk=/usr/xpg4/bin/awk
-    else
+    elif command -v mawk >/dev/null 2>&1; then
       # choose the faster mawk if: it's installed && build date >= 20230322 &&
       # version >= 1.3.4
-      local n x y z d
-      IFS=' .' read n x y z d <<< $(command mawk -W version 2> /dev/null)
-      [[ $n == mawk ]] && (( d >= 20230302 && (x * 1000 + y) * 1000 + z >= 1003004 )) && __fzf_awk=mawk
+        local n x y z d
+        IFS=' .' read n x y z d <<< $(command mawk -W version 2> /dev/null)
+        [[ $n == mawk ]] && (( d >= 20230302 && (x * 1000 + y) * 1000 + z >= 1003004 )) && __fzf_awk=mawk
     fi
   fi
   # Note: macOS awk has a quirk that it stops processing at all when it sees
--- a/shell/completion.bash
+++ b/shell/completion.bash
@@ -47,10 +47,10 @@ __fzf_exec_awk() {
     __fzf_awk=awk
     if [[ $OSTYPE == solaris* && -x /usr/xpg4/bin/awk ]]; then
       __fzf_awk=/usr/xpg4/bin/awk
-    else
-      local n x y z d
-      IFS=' .' read n x y z d <<< $(command mawk -W version 2> /dev/null)
-      [[ $n == mawk ]] && (( d >= 20230302 && (x * 1000 + y) * 1000 + z >= 1003004 )) && __fzf_awk=mawk
+    elif command -v mawk >/dev/null 2>&1; then
+        local n x y z d
+        IFS=' .' read n x y z d <<< $(command mawk -W version 2> /dev/null)
+        [[ $n == mawk ]] && (( d >= 20230302 && (x * 1000 + y) * 1000 + z >= 1003004 )) && __fzf_awk=mawk
     fi
   fi
   LC_ALL=C exec "$__fzf_awk" "$@"
@@ -524,7 +524,7 @@ if ! declare -F __fzf_list_hosts > /dev/null; then
               if ($i != "0.0.0.0")
                 print $i
           }
-        ' /etc/hosts 2> /dev/null 
+        ' /etc/hosts 2> /dev/null
       )
   }
 fi
--- a/shell/completion.zsh
+++ b/shell/completion.zsh
@@ -112,10 +112,10 @@ __fzf_exec_awk() {
     __fzf_awk=awk
     if [[ $OSTYPE == solaris* && -x /usr/xpg4/bin/awk ]]; then
       __fzf_awk=/usr/xpg4/bin/awk
-    else
-      local n x y z d
-      IFS=' .' read n x y z d <<< $(command mawk -W version 2> /dev/null)
-      [[ $n == mawk ]] && (( d >= 20230302 && (x * 1000 + y) * 1000 + z >= 1003004 )) && __fzf_awk=mawk
+    elif command -v mawk >/dev/null 2>&1; then
+        local n x y z d
+        IFS=' .' read n x y z d <<< $(command mawk -W version 2> /dev/null)
+        [[ $n == mawk ]] && (( d >= 20230302 && (x * 1000 + y) * 1000 + z >= 1003004 )) && __fzf_awk=mawk
     fi
   fi
   LC_ALL=C exec "$__fzf_awk" "$@"
--- a/shell/key-bindings.bash
+++ b/shell/key-bindings.bash
@@ -33,10 +33,10 @@ __fzf_exec_awk() {
     __fzf_awk=awk
     if [[ $OSTYPE == solaris* && -x /usr/xpg4/bin/awk ]]; then
       __fzf_awk=/usr/xpg4/bin/awk
-    else
-      local n x y z d
-      IFS=' .' read n x y z d <<< $(command mawk -W version 2> /dev/null)
-      [[ $n == mawk ]] && (( d >= 20230302 && (x * 1000 + y) * 1000 + z >= 1003004 )) && __fzf_awk=mawk
+    elif command -v mawk >/dev/null 2>&1; then
+        local n x y z d
+        IFS=' .' read n x y z d <<< $(command mawk -W version 2> /dev/null)
+        [[ $n == mawk ]] && (( d >= 20230302 && (x * 1000 + y) * 1000 + z >= 1003004 )) && __fzf_awk=mawk
     fi
   fi
   LC_ALL=C exec "$__fzf_awk" "$@"
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -54,10 +54,10 @@ __fzf_exec_awk() {
     __fzf_awk=awk
     if [[ $OSTYPE == solaris* && -x /usr/xpg4/bin/awk ]]; then
       __fzf_awk=/usr/xpg4/bin/awk
-    else
-      local n x y z d
-      IFS=' .' read n x y z d <<< $(command mawk -W version 2> /dev/null)
-      [[ $n == mawk ]] && (( d >= 20230302 && (x * 1000 + y) * 1000 + z >= 1003004 )) && __fzf_awk=mawk
+    elif command -v mawk >/dev/null 2>&1; then
+        local n x y z d
+        IFS=' .' read n x y z d <<< $(command mawk -W version 2> /dev/null)
+        [[ $n == mawk ]] && (( d >= 20230302 && (x * 1000 + y) * 1000 + z >= 1003004 )) && __fzf_awk=mawk
     fi
   fi
   LC_ALL=C exec "$__fzf_awk" "$@"


shell/common.sh
```diff
@@ -22,12 +22,12 @@ __fzf_exec_awk() {
...
-    else
+    elif command -v mawk >/dev/null 2>&1; then
       # choose the faster mawk if: it's installed && build date >= 20230322 &&
       # version >= 1.3.4
-      local n x y z d
-      IFS=' .' read n x y z d <<< $(command mawk -W version 2> /dev/null)
-      [[ $n == mawk ]] && (( d >= 20230302 && (x * 1000 + y) * 1000 + z >= 1003004 )) && __fzf_awk=mawk
+        local n x y z d
+        IFS=' .' read n x y z d <<< $(command mawk -W version 2> /dev/null)
+        [[ $n == mawk ]] && (( d >= 20230302 && (x * 1000 + y) * 1000 + z >= 1003004 )) && __fzf_awk=mawk
```
akinomyoga: The indentation is broken.

akinomyoga (commented):
The check for the command existence is supposed to be performed by the following line (together with the extraction of the version):

https://github.com/junegunn/fzf/blob/4efcc344c35e8bb7e6ba7bb23e5885051420b361/shell/common.sh#L29

However, I agree that `command_not_found_handler` causes troubles with it. `bash-completion` also had the same problem, and `bash-completion` requested to include the check for `COMP_LINE` in the `command_not_found_handle`.

https://github.com/scop/bash-completion/blob/a4fae7a8e39613eee8a859e2febb1bf634cb7daf/README.md?plain=1#L310-L318

We can add the explicit check for the `mawk` existence, but I think we should also report it to `command-not-found` if they don't check `COMP_POINT`, `READLINE_POINT`, etc. (for Bash) or their equivalents in Zsh. `command-not-found` shouldn't be triggered inside completions or keybindings in general.

akinomyoga:
@LangLangBart What are the results of the following commands?

In the case of Bash,

```console
$ (shopt -s extdebug; declare -F command_not_found_handle)
$ type command_not_found_handle
```

**edit**: In the case of Zsh,

```console
$ echo "${functions_source[command_not_found_handle]}"
$ declare -f command_not_found_handle
```

shell/common.sh
LangLangBart: adjusted, and included the `-r` flag for 'read' 

Ref: https://www.shellcheck.net/wiki/SC2046

I decided to keep `2> /dev/null`, as `mawk` may send some information to `stderr` (though this has not been checked). It can be removed if you agree; otherwise, it does no harm.

LangLangBart:
> I think we should also report it to command-not-found

likely unmaintained, based on the discussions, but i can make a report on the repo after this is merged

> their equivalents in Zsh, command-not-found shouldn't be triggered

have to study the zsh manual and grep through the zsh mailing list if this is possible in zsh

a list of other often used 'handlers' in zsh

- https://github.com/romkatv/zsh4humans/blob/5935e9a08f4894ca60ceceb8e75a90e5b19a0acc/fn/-z4h-init-zle#L62-L81

> @LangLangBart What are the results of the following commands?

Tried with docker, its not the recommend setup, see the repo for more details:

- https://aur.archlinux.org/packages/find-the-command
- https://github.com/pkasemir/find-the-command


<details>

<summary><b>BASH/ ZSH command_not_found_handler</b></summary>

```bash
docker run --platform linux/amd64 --rm --tty --interactive archlinux sh -c '
  pacman -Syu --noconfirm
  pacman -S --noconfirm git fzf zsh
  git clone https://github.com/pkasemir/find-the-command.git /ftc
  {
    echo "source /ftc/usr/share/doc/find-the-command/ftc.zsh"
    echo "source <(fzf --zsh)"
  } > ~/.zshrc
  {
    echo "source /ftc/usr/share/doc/find-the-command/ftc.bash"
    echo "eval \"\$(fzf --bash)\""
  } > ~/.bashrc
  echo
  echo ░█▀▄░█▀█░█▀▀░█░█
  echo ░█▀▄░█▀█░▀▀█░█▀█
  echo ░▀▀░░▀░▀░▀▀▀░▀░▀
  /bin/bash -ic "(shopt -s extdebug; declare -F command_not_found_handle)
    type command_not_found_handle"
  echo
  echo ░▀▀█░█▀▀░█░█
  echo ░▄▀░░▀▀█░█▀█
  echo ░▀▀▀░▀▀▀░▀░▀
  /bin/zsh -ic "echo \"\${functions_source[command_not_found_handle]}\"
  declare -f command_not_found_handle"
  '
```

```bash
░█▀▄░█▀█░█▀▀░█░█
░█▀▄░█▀█░▀▀█░█▀█
░▀▀░░▀░▀░▀▀▀░▀░▀
command_not_found_handle 197 /ftc/usr/share/doc/find-the-command/ftc.bash
command_not_found_handle is a function
command_not_found_handle ()
{
    local cmd=$1;
    local scroll_header="Shift up or down to scroll the preview";
    _cnf_pre_search_warn "$cmd" || return 127;
    local packages=$(_cnf_command_packages "$cmd");
    case $(echo $packages | wc -w) in
        0)
            _cnf_cmd_not_found "$cmd"
        ;;
        1)
            function _cnf_prompt_install ()
            {
                local packages=$1;
                if _cnf_prompt_yn "Would you like to install '$packages'?"; then
                    _cnf_asroot pacman -S "$packages";
                else
                    return 127;
                fi
            };
            local action;
            if test -z "$_cnf_action"; then
                local may_be_found="\"$cmd\" may be found in package \"$packages\"";
                echo -e "find-the-command: $may_be_found" 1>&2;
                if which fzf > /dev/null 2> /dev/null; then
                    local package_files=$(_cnf_package_files "$packages");
                    local package_info=$(pacman -Si "$packages");
                    action=$(printf "%s\n" "${_cnf_actions[@]}" | fzf --preview "echo {} | grep -q '^list' && echo '$package_files'                                     || echo '$package_info'" --prompt "Action (\"esc\" to abort):" --header "$may_be_found
$scroll_header");
                else
                    echo -e "find-the-command: What would you like to do? " 1>&2;
                    local PS3="$(echo -en "\nAction (0 to abort): ")";
                    select action in "${_cnf_actions[@]}";
                    do
                        break;
                    done;
                fi;
            else
                action="$_cnf_action";
            fi;
            case "$action" in
                install)
                    _cnf_asroot pacman -S "$packages"
                ;;
                info)
                    pacman -Si "$packages";
                    _cnf_prompt_install "$packages"
                ;;
                'list files')
                    _cnf_package_files "$packages";
                    _cnf_prompt_install "$packages"
                ;;
                'list files (paged)')
                    test -z "$PAGER" && local PAGER=less;
                    _cnf_package_files "$packages" | "$PAGER";
                    _cnf_prompt_install "$packages"
                ;;
                *)
                    return 127
                ;;
            esac
        ;;
        *)
            local package;
            echo -e "find-the-command: \"$cmd\" may be found in the following packages:" 1>&2;
            if which fzf > /dev/null 2> /dev/null; then
                for package in $(echo $packages);
                do
                    echo -e "\t$package" 1>&2;
                done;
                package=$(printf "%s\n" $packages | fzf --bind="tab:preview(type pkgfile >/dev/null 2>/dev/null &&                                 pkgfile --list {} | sed 's/[^[:space:]]*[[:space:]]*//' ||                                 pacman -Flq {})" --preview "pacman -Si {}" --header "Press \"tab\" to view files
$scroll_header" --prompt "Select a package to install (\"esc\" to abort):");
            else
                local PS3="$(echo -en "\nSelect a number of package to install (0 to abort): ")";
                select package in $(echo $packages);
                do
                    break;
                done;
            fi;
            if test -n "$package"; then
                _cnf_asroot pacman -S "$package";
            else
                return 127;
            fi
        ;;
    esac
}

░▀▀█░█▀▀░█░█
░▄▀░░▀▀█░█▀█
░▀▀▀░▀▀▀░▀░▀
/ftc/usr/share/doc/find-the-command/ftc.bash
command_not_found_handle () {
  local cmd=$1
  local scroll_header="Shift up or down to scroll the preview"
  _cnf_pre_search_warn "$cmd" || return 127
  local packages=$(_cnf_command_packages "$cmd")
  case $(echo $packages | wc -w) in
    (0) _cnf_cmd_not_found "$cmd" ;;
    (1) _cnf_prompt_install () {
        local packages=$1
        if _cnf_prompt_yn "Would you like to install '$packages'?"
        then
          _cnf_asroot pacman -S "$packages"
        else
          return 127
        fi
      }
      local action
      if test -z "$_cnf_action"
      then
        local may_be_found="\"$cmd\" may be found in package \"$packages\""
        echo -e "find-the-command: $may_be_found" >&2
        if which fzf > /dev/null 2> /dev/null
        then
          local package_files=$(_cnf_package_files "$packages")
          local package_info=$(pacman -Si "$packages")
          action=$(printf "%s\n" "${_cnf_actions[@]}" | \
                            fzf --preview "echo {} | grep -q '^list' && echo '$package_files' \
                                    || echo '$package_info'" \
                                --prompt "Action (\"esc\" to abort):" \
                                --header "$may_be_found
$scroll_header")
        else
          echo -e "find-the-command: What would you like to do? " >&2
          local PS3="$(echo -en "\nAction (0 to abort): ")"
          select action in "${_cnf_actions[@]}"
          do
          break
          done
        fi
      else
        action="$_cnf_action"
      fi
      case "$action" in
        (install) _cnf_asroot pacman -S "$packages" ;;
        (info) pacman -Si "$packages"
          _cnf_prompt_install "$packages" ;;
        ('list files') _cnf_package_files "$packages"
          _cnf_prompt_install "$packages" ;;
        ('list files (paged)') test -z "$PAGER" && local PAGER=less
          _cnf_package_files "$packages" | "$PAGER"
          _cnf_prompt_install "$packages" ;;
        (*) return 127 ;;
      esac ;;
    (*) local package
      echo -e "find-the-command: \"$cmd\" may be found in the following packages:" >&2
      if which fzf > /dev/null 2> /dev/null
      then
        for package in $(echo $packages)
        do
          echo -e "\t$package" >&2
        done
        package=$(printf "%s\n" $packages | \
                        fzf --bind="tab:preview(type pkgfile >/dev/null 2>/dev/null && \
                                pkgfile --list {} | sed 's/[^[:space:]]*[[:space:]]*//' || \
                                pacman -Flq {})" \
                            --preview "pacman -Si {}" \
                            --header "Press \"tab\" to view files
$scroll_header" \
                            --prompt "Select a package to install (\"esc\" to abort):")
      else
        local PS3="$(echo -en "\nSelect a number of package to install (0 to abort): ")"
        select package in $(echo $packages)
        do
        break
        done
      fi
      if test -n "$package"
      then
        _cnf_asroot pacman -S "$package"
      else
        return 127
      fi ;;
  esac
}
```

</details>

akinomyoga:
> > their equivalents in Zsh, command-not-found shouldn't be triggered
> 
> have to study the zsh manual and grep through the zsh mailing list if this is possible in zsh

I think `if zle; then ... fi` is a typical way to run a command only inside the interactive operations. I haven't tested, but  [`command_not_found_handler`](https://github.com/pkasemir/find-the-command/blob/fecba1a6e84a431baabe41cc23ea2068e85b46af/usr/share/doc/find-the-command/ftc.bash#L26C22-L26C84) can be `command_not_found_handler() { zle || command_not_found_handle "$@"; }`.

> Tried with docker, its not the recommend setup, see the repo for more details:
> 
> * https://aur.archlinux.org/packages/find-the-command
> * https://github.com/pkasemir/find-the-command

Thanks.

shell/common.sh
akinomyoga: > adjusted, and included the `-r` flag for 'read'

:+1:

We expect `mawk -W version` not to output backslashes, but it is in general a good habit to always use `read -r`.

> 
> Ref: https://www.shellcheck.net/wiki/SC2046
> 
> I decided to keep `2> /dev/null`, as `mawk` may send some information to `stderr` (though this has not been checked). It can be removed if you agree; otherwise, it does no harm.

I think we should keep the redirection. Some distributions (such as Debian/Ubuntu) provide different awk names as symbolic links to a particular implementation of awk. For example, `nawk` can actually be `gawk` or `mawk` in Ubuntu/Debian. This means that `mawk` may not be `mawk` but can be `nawk` or `gawk` depending on the setup. Then, `-W version` may cause the command-line argument error.

akinomyoga: (please close this conversation)

LangLangBart:
> I think `if zle; then ... fi` is a typical way

I thought you might be suggesting an option could be set within fzf's Zsh files to prevent command_not_found_handler from being triggered during execution of the fzf widgets.

Something similar to a 'try-always' block that covers all situations, no matter how the user wrote their handler.
```bash
{
  # Disable command_not_found_handler
  [[ ! -v functions[command_not_found_handler] ]] || unfunction command_not_found_handler
  # ...
} always {
  # Re-enable command_not_found_handler
  # ???
}
```

junegunn:
@LangLangBart Merged, thanks a lot!

@akinomyoga Thanks for the review and suggestions!

akinomyoga:
@junegunn Thanks!!

akinomyoga:
> > I think `if zle; then ... fi` is a typical way
> 
> I thought you might be suggesting an option could be set within fzf's Zsh files to prevent command_not_found_handler from being triggered during execution of the fzf widgets.

nah, I was talking about the general design on the side of a `command_not_found_handle(r)` implementation.

There are two parties here, the affected sides (such as `fzf`, `bash-completion`, etc.) and the affecting sides implementing `command_not_found_handle(r)` (such as `find-the-command`, `command-not-found`, etc.).  The countermeasures should be taken at both sides. In this PR, we introduced a countermeasure at the fzf side, so it's fine for this repository.

However, other configurations calling external commands (even when we are not sure if the command exists), such as bash-completion, are still affected by naive implementations of `command_not_found_handler`. In particular, `bash-completion` considers that the `command_not_found_handle` implementation should be responsible for this problem. So, I suggested the following in https://github.com/junegunn/fzf/pull/4468#pullrequestreview-3053546492:

> [...], but I think we should also report it to `command-not-found` if they don't check `COMP_POINT`, `READLINE_POINT`, etc. (for Bash) or their equivalents in Zsh.

Then, as a countermeasure, the `command_not_found_handle(r)` can skip their main processing when they detect that it is a part of the interactive operations such as completion and keybindings. As one way to detect it in Zsh, I suggested `zle` (without argument). What I suggest is unrelated to the external settings that are affected by `command_not_found_handle(r)`. I'm suggesting modifying `find-the-command`.

LangLangBart:
> I'm suggesting modifying `find-the-command`.

Understood, thank you for reviewing so quickly. I have reported the issue as well.

thanks for the warp-speed merge, @junegunn

PR #4492, opened 2025-08-16
Support old uname in macOS

yous:
`./install --all` exit with error code because of `uname -smo` in macOS 10.15.7, 12.7.6 (yes, they're old).

```console
$ ./install --all
uname: illegal option -- o
usage: uname [-amnprsv]
No prebuilt binary for  ...
go executable not found. Installation failed.
```

`man uname` doesn't contain `-o` option in macOS 10.15.7, 12.7.6. I can confirm that `man uname` contain `-o` option in macOS 15.6.

So we can fallback to `uname -sm` after trying `uname -smo`.

### DIFF
--- a/install
+++ b/install
@@ -164,7 +164,7 @@ download() {
 }
 
 # Try to download binary executable
-archi=$(uname -smo)
+archi=$(uname -smo 2>/dev/null || uname -sm)
 binary_available=1
 binary_error=""
 case "$archi" in


junegunn:
Thanks!

junegunn:
Hi @yous, a quick question. You still running Catalina? I recently [decided to upgrade the Go toolchain](https://github.com/junegunn/fzf/pull/4508#issuecomment-3247505980) to update the dependencies, so the new versions probably will not build or run on Catalina.

https://go.dev/wiki/MinimumRequirements#macos-ne-os-x-aka-darwindarwin

yous:
Hi, thanks for the notice. The link says Go 1.21 and later support 10.15 Catalina or newer, so my machine can run more future releases. 10.13 High Sierra and 10.14 Mojave may not be able to run the new versions, but I don't have actual machines running on that macOS versions.

- https://go.dev/doc/go1.20#darwin

junegunn:
Unfortunately, we recently updated to Go 1.23 due to tcell depencency.

https://github.com/junegunn/fzf/commit/d22812e91790e1f7cec00319ff1b68776cc62aad

But I don't think tcell definitely needs Go 1.23. I can attempt to fork it and maintain a version compatible with Go 1.21.

yous:
I guess Go's [release policy](https://go.dev/doc/devel/release#policy) is quite aggressive. Only the last two major releases are supported.

Because of this, dependencies like golang.org/x/text [already require 1.24.0](https://github.com/golang/text/commit/60c9786d9e6cc83e1900ce976fdba2e1c327d220). But I don't think <https://go.dev/wiki/MinimumRequirements> is aggresive, Big Sur is released in 2020 and the support is ended.

I don't have preference if fzf supports Catalina or not, as macOS Catalina is so outdated and not supported version. I will pin fzf version for that machine when it drops Go 1.21.

But providing bug fixes for older versions may require maintaining multiple branches, so I think it's easier to make fzf more compatible and maintain only one branch. Still I think Go 1.23 is a reasonable starting point.

junegunn:
Thanks for the feedback. I've opened a pull request to downgrade to Go 1.21, but I don't know if it's the right thing to do. 

https://github.com/junegunn/fzf/pull/4515

PR #4501, opened 2025-08-24
Adjust default themes

junegunn:
Motivation:

`--color 16` can be a better default than `dark` or `light`, since it uses the colors defined by the current theme. This usually blends in more naturally and works well in both light and dark modes.

However, some elements were previously hard-coded with white or black foreground colors, which can cause rendering issues in certain terminal themes.

* 16
  * Avoid using black or white foreground colors, so it works better with both dark and light themes
  * ~~Display 'info' in italic to better separate it from the other parts~~
  * `info` is now yellow
* dark / light
  * ~~Display 'info' in italic for consistency~~
  * No change


junegunn:
I think italic info is nice, but I'm not sure how others will feel about it.

Maybe not a good idea, because it may not render properly.

https://unix.stackexchange.com/questions/745710/getting-italics-working-correctly-in-tmux

WhyNotHugo:
I still see no borders/separators in dark mode with this patch.

Light:

image

Dark:

image

junegunn:
Are you sure you're running a binary with this patch? Match counters (info) are not displayed in italic.

arkriny:
> I still see no borders/separators in dark mode with this patch.

@WhyNotHugo It's probably related to your terminal or theme not supporting dimmed colors. I can confirm that it works in Alacritty with the Vague theme.

@junegunn Are there any plans to make the base16 theme the default? It has better compatibility than the hardcoded theme, and since most tools already follow it, that behavior would be more expected.

WhyNotHugo:
Regarding dimmed colours in foot:

>  Dimmed colors do not have an entry in the color palette. Applications emit them by combining a color value, and a "dim" attribute.
>
> By default, foot implements this by blending the current color with black. This is a generic approach that applies to both colors from the 256-color palette, as well as 24-bit RGB colors.

Looks like themes themselves cannot support dim, and foot does, but the blending yields black-on-black in dark mode.

junegunn:
> @junegunn Are there any plans to make the base16 theme the default? It has better compatibility than the hardcoded theme, and since most tools already follow it, that behavior would be more expected.

Maybe. If I were starting fzf today, I would do that, but since we've been using the current default theme for over 10 years, I'm reluctant to suddenly change it, which might surprise the users, maybe in a good way? I don't know.

WhyNotHugo:
Ultimately, there's only one way to know. But I can understand not wanting to receive lots of angry emails because the default theme changed.

WhyNotHugo:
I'd say that more compatible/accessible defaults are preferible, but I'm obviously biased here.

src/tui/tui.go
```diff
@@ -925,16 +925,16 @@ func init() {
 		DarkBg:           ColorAttr{colBlack, AttrUndefined},
 		Prompt:           ColorAttr{colBlue, AttrUndefined},
 		Match:            ColorAttr{colGreen, AttrUndefined},
-		Current:          ColorAttr{colYellow, AttrUndefined},
+		Current:          ColorAttr{colDefault, AttrUndefined},
```
junegunn: This doesn't work well on light terminals, due to `colBlack` used for `DarkBg`.

junegunn: Changed it to `white`, so that at least it's readable.

junegunn: Now it's bright white over grey.

WhyNotHugo:
Thanks for the efforts on this!

I can confirm that both dark mode and light mode are perfectly legible and have clear borders too!

screenshot in dark mode

screenshot in light mode

PR #4507, opened 2025-08-28
refactor: replace []byte(fmt.Sprintf) with fmt.Appendf

zhedazijingang:
Optimize code using a more modern writing style. Official support from Go, for more details visit https://pkg.go.dev/golang.org/x/tools/gopls/internal/analysis/modernize.


junegunn:
Thanks!

Issue #61, opened 2014-06-19
Allow C-V, C-S, C-T, etc. in vim plugin


glittershark:
The only thing really stopping me from replacing ctrlp entirely with fzf is support for `<C-V>` to vsplit, `<C-S>` to hsplit, `<C-T>` to open in a new tab, etc.

PR #4508, opened 2025-09-02
mod: update charlievieth/fastwalk to v1.0.13 and min Go version to 1.21

charlievieth:
This commit updates github.com/charlievieth/fastwalk to v1.0.13 which addresses fastwalk issue [#61](https://github.com/charlievieth/fastwalk/issues/61). It also updates the minimum supported Go version to 1.21 (up from 1.20) since that is now the minimum version supported by fastwalk (since we now use `binary. NativeEndian`).

@junegunn If there are any issues with updating to Go 1.21 then I can fix fastwalk in a way that allows us to use version 1.20.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -25,7 +25,7 @@ jobs:
     - name: Set up Go
       uses: actions/setup-go@v5
       with:
-        go-version: "1.20"
+        go-version: "1.21"
 
     - name: Setup Ruby
       uses: ruby/setup-ruby@v1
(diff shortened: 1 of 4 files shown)


junegunn:
Thanks @charlievieth, we recently decided to upgrade to the latest Go and finally drop support for Windows 7.

https://github.com/junegunn/fzf/pull/3996#issuecomment-3240131039

So I think this is fine. But let me wait a bit to see if the community has any feedback on the decision.

junegunn:
Merged, thanks!

PR #4515, opened 2025-09-10
Build with Go 1.21

junegunn:
Upgrading to Go 1.23 means dropping support for some old OS versions like macOS Catalina.

* https://go.dev/wiki/MinimumRequirements

tcell 2.9.0 was the reason we had to upgrade to 1.23, but we can still build tcell with Go 1.21 if we downgrade some of its dependencies.

Is this worth it? Or should we just move on?


LangLangBart:
> Is this worth it? Or should we just move on?

I also moved on from 10.15 (Catalina), using 15.7 (Sequoia) now. I would just advice whatever is easier for the maintainer. 

macos

Homebrew's analytics data (last 30 days), for getting a feeling on how many macOS users still use Catalina:

```bash
curl -s 'https://formulae.brew.sh/api/analytics/os-version/30d.json' |
  jq -r '
    .items | map(select(.os_version | startswith("macOS"))) |
    (map(.count | gsub(",";"") | tonumber) | add) as $total | .[] |
    "\(.os_version):\t\(.count) (\(.count | gsub(",";"") | tonumber / $total *1e4 | round / 100)%)"' |
  column -ts $'\t'

# macOS Sequoia (15):         10,958,747 (66.56%)
# macOS Sonoma (14):          2,929,707 (17.79%)
# macOS Ventura (13):         1,195,642 (7.26%)
# macOS Tahoe (26):           729,406 (4.43%)
# macOS Monterey (12):        462,519 (2.81%)
# macOS Big Sur (11):         94,023 (0.57%)
# macOS Catalina (10.15):     64,083 (0.39%)
# macOS High Sierra (10.13):  18,746 (0.11%)
# macOS Mojave (10.14):       10,149 (0.06%)
# macOS Sierra (10.12):       1,965 (0.01%)
```

junegunn:
@LangLangBart Thanks! That's really helpful info. So, unless there are vocal users, I'd like to move on, but keeping this branch alive for a while, backporting some important fixes.

PR #4518, opened 2025-09-11
refactor: use maps.Copy and maps.Clone

mickychang9:
There is a [new function](https://pkg.go.dev/maps@go1.21.1#Copy) added in the go1.21 standard library, which can make the code more concise and easy to read.

### DIFF
--- a/src/core.go
+++ b/src/core.go
@@ -2,6 +2,7 @@
 package fzf
 
 import (
+	"maps"
 	"os"
 	"sync"
 	"time"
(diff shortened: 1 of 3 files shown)


src/core.go
```diff
@@ -226,9 +227,7 @@ func Run(opts *Options) (int, error) {
 	patternBuilder := func(runes []rune) *Pattern {
 		denyMutex.Lock()
 		denylistCopy := make(map[int32]struct{})
-		for k, v := range denylist {
-			denylistCopy[k] = v
-		}
+		maps.Copy(denylistCopy, denylist)
```
junegunn: Thanks, would `Clone` be a better fit here?

mickychang9: Sorry for the delay.

Modified. Please review again.

junegunn (approved):

junegunn:
Merged, thanks!

PR #4525, opened 2025-09-21
style: add format rules for bash files

xieyonn:
Thank you very much for developing and maintaining fzf!

## problem
There is no unified format defined for the `bash` scripts in the repository, so I thought of adding `shfmt` as a unified formatting tool for `bash` files.

Additionally, I use `neovim` with `treesitter` in my daily work. When reading the code of `fzf --bash`, the inconsistent syntax parsing occurs due to some non-uniform coding styles.

For instance, when comparing strings, some parts use `=` while others use `==`, which causes all the subsequent content to be incorrectly recognized as string. 

image

Using `==` is equivalent to using `=`, but the display becomes normal after changing to `==`. 

image

Formatting with `shfmt` will resolve this.

## solution
1. Add a `.editorconfig` file, add rules for `*.sh` and `*.bash` and other bash files, including `/bin/fzf-preview.sh` `/bin/fzf-tmux` `/install` `/uninstall`
> of course, .editorconfig also supports other programming languages, leave them for now.
2. Add target `make fmt` in `Makefile` for formatting go, bash files.
3. Modify `make lint` to add formatting checks for `bash` files.
4. Since I haven't installed Ruby, I modified the check for the Ruby executable `bundle` in `make fmt` to skip Ruby files.

---
I'm not sure if this is an appropriate action, since formatting with `shfmt` will modify a large amount of code and disrupt the original commit history.

I split the changes into 2 commits:
1. add `.editorconfig`, modify `Makefile`, install shfmt in actions.
2. early return in `shell/completion.bash`, `shell/key-bindings.bash` reduces nesting, which will reduce change lines after format.


xieyonn:
I think it would be more reasonable to have all the changes from `make fmt` as a separate pr: https://github.com/junegunn/fzf/pull/4526

xieyonn:
Hi, `build` failed becase make lint failed (cause by shfmt check), it will be ok when https://github.com/junegunn/fzf/pull/4526 merged

shell/completion.bash
```diff
@@ -11,8 +11,9 @@
 # - $FZF_COMPLETION_PATH_OPTS (default: empty)
 # - $FZF_COMPLETION_DIR_OPTS  (default: empty)
 
-if [[ $- =~ i ]]; then
-
+if [[ ! $- =~ i ]]; then
+  return
+fi
```
junegunn: We tried this in the past but had to revert it because there was an issue.

Please take a look at https://github.com/junegunn/fzf/commit/152988c17bd6c35ce55652d123db8b65a32f98e4

.editorconfig
```diff
@@ -0,0 +1,20 @@
+root = true
+
+[*.{sh,bash}]
+indent_style = space
+indent_size = 2
+simplify           = true
+binary_next_line   = true
+switch_case_indent = true
+space_redirects    = true
+function_next_line = false
```
junegunn: Any reason you decided to not align the first two?

shell/completion.bash
xieyonn: Sure, I will revert it.

junegunn: But wouldn't it cause the entire file to be indented? Can we avoid the problem?

.editorconfig
xieyonn: got it!

shell/completion.bash
xieyonn: I actually think we might not need this check, because when people use fzf, they typically do so in an interactive shell environment—that is, they put `eval "$(fzf --bash)"` in their `~/.bashrc`. 
Typically, we add this check in `~/.bashrc`, for example:

```bash
case $- in
    *i*) ;;
    *) return ;;
esac
```
 So this check seems somewhat redundant.

xieyonn: It is the user who should avoid executing eval "$(fzf --bash)" in non-interactive bash, as using it this way has no meaning.

junegunn: Not everyone has a clear understanding of what they're doing. That's not something we can expect. Also, I can imagine users sourcing their `~/.bashrc` in a non-interactive session for certain environment variables or functions, without expecting it to interfere with their script execution.

But above all, this is a style change, and style changes should not affect any user scenario.

One solution would be to post-process the output of `shfmt` to remove two leading spaces.

```sh
shfmt file | sed 's/^  //' > file
```

junegunn:
> Additionally, I use `neovim` with `treesitter` in my daily work. When reading the code of `fzf --bash`, the inconsistent syntax parsing occurs due to some non-uniform coding styles.

So this is a bug of treesitter I suppose? Inconsist code style is of course not ideal, but it shouldn't affect the code highlighting.

shell/completion.bash
xieyonn: Yeah, got your point!
I've read through the entire process of adding/removing this check.

consider adding an entry file, such as `interactive_protection.bash`, put all interactive check here, then eval rest files.

```bash
# exec bash scripts only in interactive sessions
case $- in
    *i*) ;;
    *) return ;;
esac

# something like this:
eval "completion.bash"
eval "key-bindings.bash"
```

xieyonn:
> > Additionally, I use `neovim` with `treesitter` in my daily work. When reading the code of `fzf --bash`, the inconsistent syntax parsing occurs due to some non-uniform coding styles.
> 
> So this is a bug of treesitter I suppose? Inconsist code style is of course not ideal, but it shouldn't affect the code highlighting.

It might be a bug, or there could just be a conflict in the syntax tree parsing—I haven't delved deeply into this issue yet.

---
I added an `interactive_protection.bash` file. Anyway, if we use eval '$(fzf --bash)', these 3 files will be merged directly, so we can remove the checks in `key-bindings.bash` and `completion.bash`.

xieyonn:
Wait a minute, I should put the following 2 files inside the if statement. Let me think about how to do this.

junegunn:
> I added an `interactive_protection.bash` file.

Thanks, but I'm not in favor of that approach. fzf has a very long history, and `fzf --bash` is a relatively recent addition. Because of this, there are still many users out there sourcing those files individually.

xieyonn:
Emm.. Then there's no other way—it seems we can only let shfmt add indentation.

junegunn:
Have you considered my suggestion above? Removing leading 2-spaces from the shfmt output.

xieyonn:
Sure, it work. but handling it during the lint stage becomes a bit difficult, so I removed the changes made to the lint stage.

xieyonn:
`shfmt file | sed 's/^  //' ` make format not an idempotent operation.

Some code will indent to the left every time it is executed, until it reaches the first line.

eg:
multi line string definition:
image

junegunn:
> `shfmt file | sed 's/^ //' ` make format not an idempotent operation.

That's unfortunate.

I just tested `make fmt` locally and it looks like we should only apply the `sed` trick to the files under `shell/`.

xieyonn:
Agree, files under `shell/` have quite a bit of historical burden.

Add a variable `BASH_SCRIPTS` to track additional Bash files.

junegunn:
> Agree, files under `shell/` have quite a bit of historical burden.

We should trim the leading spaces only in two files:

- [ ] shell/common.sh
- [x] shell/completion.bash
- [x] shell/key-bindings.bash
- [ ] shell/update-common.sh

xieyonn:
Done.

How are you planning to deal with parts indented every time `make fmt` runs?

Maybe we could change it like this:
1. For multi-line strings, just align all lines with the first line. including awk scripts, opts definition.
2. For the `transformer` definition inside the `_fzf_proc_completion` function, using `function` and export it.

```bash
_fzf_proc_completion_transformer() {
    # definition here
}

export -f _fzf_proc_completion_transformer

_fzf_proc_completion() {
  _fzf_complete -m --header-lines=1 --no-preview --wrap --color fg:dim,nth:regular \
    --bind "click-header:transform:_fzf_proc_completion_transformer" -- #...
}
```

--- 

Or just leave everything to `shfmt`. After all, doing it this way feels too hacky.

junegunn:
So I decided to put BEGIN/END markers for partially formatting the file.

Thanks for the contribution!

PR #4528, opened 2025-09-23
Fix 7-byte CSI sequence leftover prints

masmu:
Some CSI key sequences like `^[[5;10~` were only partially consumed by the parser, leaving trailing bytes (e.g. `10~`) in the input buffer which could then be printed as stray characters in the terminal.

In `escSequence()` (in `light.go`) the branches handling certain 7‑byte CSI forms returned the correct event but failed to set the consumed size pointer *sz = 7, so the caller removed too few bytes from the buffer.
After setting `*sz = 7` in the affected branches the entire 7‑byte sequence is consumed and no leftover bytes are interpreted/printed.


junegunn:
Merged, thanks!

PR #4531, opened 2025-09-26
merger: fix cache never getting cleared

### LINKED ISSUE #4529: Memory leak when reloading after having searched
### Checklist

### Output of `fzf --version`

0.65.2 (416aff86)

### OS

- [x] Linux
- [ ] macOS
- [ ] Windows
- [ ] Etc.

### Shell

- [x] bash
- [ ] zsh
- [ ] fish

1. Run
```bash
seq 1000000 | FZF_DEFAULT_OPTS= fzf --bind="ctrl-r:reload(seq 1000000)" --preview='ps -eo comm,%mem -p "$PPID"'
```
2. Reload repeatedly with Ctrl-R.  Memory usage goes up to about 0.4 % (of 32 GB) and stays there, sometimes dropping to 0.3 % or 0.2 %.
3. Search for `666`.
4. Reload repeatedly with Ctrl-R.  Memory usage goes up indefinitely, roughly by increments of 0.2 %.
5. Clear the query.
6. Reload repeatedly with Ctrl-R.  Memory usage keeps going indefinitely.

alex-huff:
Fixes #4529

### DIFF
--- a/src/matcher.go
+++ b/src/matcher.go
@@ -85,11 +85,11 @@ func (m *Matcher) Loop() {
 		cacheCleared := false
 		if request.sort != m.sort || request.revision != m.revision {
 			m.sort = request.sort
-			m.revision = request.revision
 			m.mergerCache = make(map[string]*Merger)
 			if !request.revision.compatible(m.revision) {
 				m.cache.Clear()
 			}
+			m.revision = request.revision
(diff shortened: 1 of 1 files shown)


junegunn:
Great catch, thanks so much!

PR #4534, opened 2025-09-28
Introduce 'raw' mode

### LINKED ISSUE #2515: Feature request: search without filtering the list

## Info

- OS
    - [x] Linux
    - [x] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [ ] bash
    - [x] zsh
    - [ ] fish
    
I'm finding `fzf` extremely helpful for searching git history in log graph: https://git.tsundere.moe/Frederick888/frederick-settings/blob/9d7e1286a3616648eedea56e08160a84752b8cfc/.gitconfig#L20-68

However often after finding the commits, I'd also like to know their positions in the log graph. Currently what I have to do is copying the commit hash then open a `git lg` (also in the linked `.gitconfig` above) to locate it.

So I'm wondering if it's possible to search without filtering the list, where it:

- still does searching & highlighting and jumps to the first hit automatically
- but the list itself doesn't get filtered, somewhat like a pager with fuzzy search
- updates the counter as usual to show number of hits
- e.g. <kbd>Ctrl</kbd>-<kbd>j</kbd>/<kbd>k</kbd> to jump to next/previous hit

I understand that I can use gv.vim + fzf buffer lines + quick fix window to achieve this, but it'd be nice to achieve this without vim.

### LINKED ISSUE #4318: Feature request: an option like `--query` but only affects selection, without filtering
### Checklist

### Output of `fzf --version`

0.57.0 (0476a65)

### OS

- [ ] Linux
- [ ] macOS
- [ ] Windows
- [ ] Etc.

### Shell

- [ ] bash
- [ ] zsh
- [ ] fish

When invoking FZF from Vim, like in `GFiles`, I would like to pass the current file as `--query` so that the selection is positioned on the current file and I can go to see its neighbors. However if I pass `--query` it also filters the list. I am hoping for a new option .e.g '--find' that will only put the selection at the right place. 

Doesn't add a new mode and simple to implement I think.

junegunn:
Introduce 'raw' mode, which displays non-matching items in a dimmed color. You can dynamically toggle this mode with the `toggle-raw` action to see the surrounding items of a match to better understand the context. Useful for command history, commit history, file system tree, etc.

Fix #2515
Fix #4318

More tests needed.


LangLangBart:
Thanks for working on new features to enhance fzf.

---

> * `CTRL-X` was chosen because enabling the raw mode effectively "e**x**pands" the list.

makes sense

---


Have you played around with the idea of stripping original colors to apply 'hidden' style uniformly ?

```bash
HOME= git log --oneline --format="%C(auto) %cd %h %d %s" -10 --color=always |
  FZF_DEFAULT_OPTS= fzf --raw --color hidden:244 --ansi --query C
```


com

**Ref:** https://github.com/gitui-org/gitui

In GitUI, when searching, non-matching results are colored uniformly gray.

SCR 2025-09-29 at 22 00 41


---

When I run the command below and use the cursor the issue in the image below can be seen

```bash
seq 12 | FZF_DEFAULT_OPTS= fzf --raw --gutter-raw ''
```

SCR 2025-09-29 at 21 57 49

junegunn:
> stripping original colors to apply 'hidden' style uniformly ?

I can see the appeal, though I prefer keeping the colors. One option would be to add support for a special attribute named `strip` or `nocolor` which removes the original colors.

> When I run the command below and use the cursor the issue in the image below can be seen

Ah, thanks for catching it. I think I'm just going to require that the display width of a gutter character is 1.

LangLangBart:
> I can see the appeal, though I prefer keeping the colors

ok. 

> Ah, thanks for catching it. I think I'm just going to require that the display width of a gutter character is 1.

thx. fixed.

junegunn:
@LangLangBart Please take a look at d3485a06a159d07c4a73e8a78d2cc05f6a65d635 and let me know what you think.

LangLangBart:
> @LangLangBart Please take a look at [d3485a0](https://github.com/junegunn/fzf/commit/d3485a06a159d07c4a73e8a78d2cc05f6a65d635) and let me know what you think.

> [!NOTE]
> All commands below are run on the fzf 'raw' branch.

While experimenting with it, I found it can be unintuitive to configure a
uniformly colored, dimmed line for entries that don't match.

My **first** attempt was to try the command below. I noticed that a strikethrough
was only applied to parts of the lines that used color, such as the hash.

```bash
HOME='' git log --oneline --format="%C(auto) %cd %h %d %s" -10 --color=always |
  FZF_DEFAULT_OPTS='' fzf --ansi --color hidden:strip:dim:strikethrough --raw --query nn
```

strike


**Next**, I tried using a longer configuration, which more closely achieved a
uniform dim or strikethrough effect on non-matching lines.

```bash
HOME='' git log --oneline --format="%C(auto) %cd %h %d %s" -10 --color=always |
  FZF_DEFAULT_OPTS='' fzf --ansi --delimiter / --nth -1 \
    --color fg:dim:strip:strikethrough,nth:regular,hidden:strip:dim:strikethrough \
    --raw --query nn
```

default
vscode

I noticed some issues - for example, strikethrough is not displayed in the
default Apple Terminal, and after deleting the query, the dim or strikethrough
effect remains partially applied to some lines.

SCR 2025-10-02 at 01 49 30

---

The `strip` option might help customize how non-matching lines are displayed,
but since it requires more complicated configuration, I would probably just
stick with the default, and not use it.

---

Run the command below, some entries remain "dimmed" after the query is cleared.
This seems to occur more reliably reproducible with the `change:first` binding. 

```bash
fd --color always |
  FZF_DEFAULT_OPTS='' fzf --ansi --bind 'change:first' \
    --bind 'load:clear-query' --query nn --raw
```

When you move over the "dimmed" lines, notice they become properly colored again.

junegunn:
> I found it can be unintuitive to configure

Thanks for testing it. The unintuitiveness you noticed is not an intended behavior but a bug in the implementation. I'll let you know after fixing it.

junegunn:
> My **first** attempt was to try the command below. I noticed that a strikethrough was only applied to parts of the lines that used color, such as the hash.
> 
> ```shell
> HOME='' git log --oneline --format="%C(auto) %cd %h %d %s" -10 --color=always |
>   FZF_DEFAULT_OPTS='' fzf --ansi --color hidden:strip:dim:strikethrough --raw --query nn
> ```

This should work as expected now.

> strikethrough is not displayed in the default Apple Terminal

I can confirm. I didn't expect it to be supported by all terminal emulators.

> the dim or strikethrough effect remains partially applied to some lines.

Actually, it's the expected behavior. The dim and strikethrough are from `fg`,  not from `hidden`. It's because you have `--delimiter / --nth -1`, which is not the right option for this input.

junegunn:
> Run the command below, some entries remain "dimmed" after the query is cleared. This seems to occur more reliably reproducible with the `change:first` binding.

So it looks like a case where fzf doesn't refresh the list promptly. If you press CTRL-L, you can see the list is properly updated. Let me look into this.

junegunn:
e7e7bc3315724d1cad89030af5675530dad9d857 should fix the problem. I really appreciate your reports.

maxaykin:
I've just had a quick look at the feature and started playing with it. It is impressive, very nice behavior!

The only thing which seems still missing for me is navigation without change of selected item or preserving the selection somehow. Here is the scenario:
- hundreds/thousands of input lines
- entering a query to filter out unwanted items / to locate those variants which are being looked for
- selecting one of the matched lines which is of most interest
- switching to raw mode to see the context (e.g. to ensure this is the needed one)
- scrolling up/down to see more context lines
- the choice is lost, no simple way to return to it even after disabling raw mode

As a workaround one may mark interested items with tab (when --multi is used). Is there something better possible? Maybe an option that would make raw mode have it own cursor or would it be overcomplicated?

LangLangBart:
> ```shell
> HOME='' git log --oneline --format="%C(auto) %cd %h %d %s" -10 --color=always |
>   FZF_DEFAULT_OPTS='' fzf --ansi --color hidden:strip:dim:strikethrough --raw --query nn
> ```
> This should work as expected now.

✅ The ability to have uniformly colored non-matching lines in gray works. Thank you.

> > the dim or strikethrough effect remains partially applied to some lines.
> 
> Actually, it's the expected behavior. The dim and strikethrough are from `fg`, not from `hidden`. It's because you have `--delimiter / --nth -1`, which is not the right option for this input.

✅ I see. My bad.

> e7e7bc3 should fix the problem. I really appreciate your reports.

✅ Perfect. I'm glad reports were a bit helpful

---

> The only thing which seems still missing for me is navigation without change of selected item or preserving the selection somehow. Here is the scenario:

@maxaykin preserving a selected line does work for me, I tested it with the command below:

- generates many input lines
- searches for "bar$"
- switches to raw mode once loading is complete
- keeps my selected line selected
- pressing 'ctrl-x' to toggle-raw mode lets me see only matching lines again

```bash
printf "%s\n" {a..z}{a..z}{a..z}{a..z}{a..z} |
  FZF_DEFAULT_OPTS_FILE= FZF_DEFAULT_OPTS= fzf \
    --bind 'ctrl-x:toggle-raw' --query 'bar$' --bind 'load:enable-raw'
```

alex-huff:
> @maxaykin preserving a selected line does work for me, I tested it with the command below:

I believe @maxaykin wants the ability to:
1. select an item, let's call it `foo`
2. switch to raw mode
3. navigate around to see the context using scroll or pageup/pagedown
4. switch back to normal mode
5. `foo` is selected

If you accidentally select another matched item while navigating around in raw mode that item will be selected when you switch back to normal mode.

LangLangBart:
> If you accidentally select another matched item while navigating around in raw mode that item will be selected when you switch back to normal mode.

@alex-huff 

One approach would be to save the position before switching to raw mode, for
example as FZF_PROMPT (or to some other FZF env variable/file), and then restore
the position when switching back.

1. start the command, move the selection to 'bar'
2. press 'ctrl-x' to switch to RAW MODE
3. move from 'bar' to 'baz'
4. press 'ctrl-x' to switch to NORMAL MODE
5. notice you are back to 'bar'

```bash
printf "%s\n" {a..z}{a..z}{a..z}  |
  FZF_DEFAULT_OPTS_FILE= FZF_DEFAULT_OPTS= fzf \
    --query '^ba' --bind 'ctrl-x:transform:
  if [[ -n $FZF_RAW ]]; then
    echo "change-header(NORMAL MODE)+disable-raw+change-prompt(> )+pos:$FZF_PROMPT"
  else
    echo "change-header(RAW MODE)+enable-raw+change-prompt:$FZF_POS"
  fi'
```

alex-huff:
> One approach would be to save the position before switching to raw mode, for example as FZF_PROMPT (or to some other FZF env variable/file), and then restore the position when switching back.

Nice approach. You could also use `FZF_GHOST` and `change-ghost` instead if you wanted to be more discrete.

junegunn:
The current behavior is:

- Normal -> Raw
    - The cursor stays on the same matching item.
- Raw -> Normal
    - If the cursor is on a matching item: it stays on that item.
    - If the cursor is on a non-matching item: since that item is not visible in normal mode, the cursor falls back to the last highlighted matching item.

> navigation without change of selected item

I can see how this could be desirable in some cases, but I don't think always returning to the pre-raw-mode highlighted item should be the default.

Thanks @LangLangBart for the nice suggestion. Yes, we can store states in multiple fzf elements or in a temp file.

But I wonder if we can make the process easier. Would having `store-pos` / `restore-pos` help? (Maybe we can extend the idea to a stack-like approach with `push-pos` and `pop-pos` combo.) If we make the stored position local to each mode, we can probably avoid writing `transform`:

```sh
--bind 'ctrl-x:store-pos+toggle-raw+restore-pos'
```

Or we can make the `restore-pos` optional.

```sh
--bind 'ctrl-x:store-pos+toggle-raw,ctrl-o:restore-pos'
```

alex-huff:
> ```shell
> --bind 'ctrl-x:store-pos+toggle-raw+restore-pos'
> ```

I'm confused how this would work. Wouldn't the position that got stored before you entered raw mode always get overridden by the current position when you toggle back to normal mode?

junegunn:
> If we make the stored position local to each mode

junegunn:
While that might seem unintuitive, if we think about it, having a single stored postion across two modes isn't really great either, because the position would point to different items in each mode.

alex-huff:
> If we make the stored position local to each mode

My bad I must have skipped right over this.

Using `store-pos`/`restore-pos` could the user have the position automatically restored when they go back to normal mode but not the other way around? Assuming the primary use-case of switching to raw mode is to see the surrounding items of a match, it is probably undesirable to have the position jump back to wherever it was when you last left raw mode.

It seems the user cannot opt in to raw -> normal restoring without also opting in to normal -> raw restoring unless they bind `restore-pos` to a different key than `toggle-raw` and manually invoke it like you suggested.

junegunn:
> Assuming the primary use-case of switching to raw mode is to see the surrounding items of a match, it is probably undesirable to have the position jump back to wherever it was when you last left raw mode.

Good point. So this means we want different actions to be performed depending on the current mode, and it makes `transform` with a condtional branch unavoidable.

maxaykin:
While you are having the very good discussion of my question (thank you guys for that!) I would like to ask one more related question: would it be possible to introduce another mode (independent of raw mode) that would allow navigation (i.e. scrolling using keys/mouse) without changing the selection?

Then the way forward for the current use case would be something as simple as this:
`
--bind 'ctrl-x:toggle-raw+toggle-navigation'
`

Moreover, I believe that it would be helpful for some other cases when hiding not matching items is not required (e.g. you are looking at a listing of a directory, you don't know yet what to type in the query field and you like having the preview of the currently selected file open while searching for another file).

Perhaps, there could be variants. For example, in navigation mode the selection could be not changed while scrolling using mouse (the wheel and dragging the scroll bar) and using PageUp/PageDown keys but clicking on an item or using Up/Down keys could change the selection.

I know this is a bit out of scope of this feature now. :) Sorry for that.

maxaykin:
> Is CTRL-X the most appropriate key for the feature?

Just to remind: in fzf.vim ctrl-x opens a new split

junegunn:
I'm not sure if adding another mode brings much value here. It might just add confusion, since you would have to be constantly aware of how behaviors differ between modes.

> you like having the preview of the currently selected file open while searching for another file

Do you find yourself needing that often? It feels like a niche use case to me. One workaround is to multi-select the item, and use `{+}` placeholder in your preview command template.

```sh
fzf --multi 1 --preview 'cat {+}'
```

> Just to remind: in fzf.vim ctrl-x opens a new split

Oh, that's true. It didn't cross my mind.

junegunn:
Key to bind to `toggle-raw`

* `CTRL-R`
    * Intuitive and easy to remember.
    * Possible conflict with "reload" bindings.
* `CTRL-X`
    * Less mnemonic; but X stands for "expand".
    * Possible conflict with "execute" bindings.
    * Already used in fzf.vim for split open.

So what else do we have?

* `ALT-R` (strong candidate)
    * Rarely used for other purposes (not sure).
    * Easy to remember.
    * Slightly harder to press than `CTRL-R` (on my macbook)
* ...?

maxaykin:
> `CTRL-R`
> * Intuitive and easy to remember.

I guess you think it is intuitive because 'R' is the first letter of 'raw' but I am afraid that for most people the name of raw mode itself would be not very intuitive. :blush:

maxaykin:
> > you like having the preview of the currently selected file open while searching for another file
> 
> Do you find yourself needing that often? It feels like a niche use case to me.

Well, that was just an example. Here is another one: you found some old commit and you are scrolling the commits history to see what new changes were done after it; without keeping the selection it is quite easy to get lost which would require to repeat the search. This is what happens to me relatively often.

Besides specific use cases the thing is that it seems to be very common and natural nowadays that scrolling is separated from movement of cursor position. In many text editors, UI list components, in web pages with input fields you can stop typing (or selecting an item) at any moment, have a look around the whole page/document and then get back to where you stopped either by scrolling to the cursor position or simply by entering the next character or pressing a navigation key like Up/Down  which brings you back immediately.

junegunn:
> In many text editors

But not in Vim, right? I don't think the model maps well to terminal programs, because users generally expect the cursor to stay visible on screen. Anyway, I recommend you experiment with what @LangLangBart suggested above.

> I am afraid that for most people the name of raw mode itself would be not very intuitive. 😊

That's true. Any better key suggestions?

maxaykin:
> But not in Vim, right? I don't think the model maps well to terminal programs, because users generally expect the cursor to stay visible on screen.

In Vim (at least with my setup) it works partially: scrolling with mouse wheel (or dragging scrollbar in gvim) preserves the cursor until the line where it is located goes out of the window bounds. In midnight commander mouse wheel acts really the same as in fzf.

But do we really want the modern fzf tool behave as these old-school programs which grew up with much stricter limitations? Besides, we are talking about an optional mode for those who want it, and this is very similar to raw mode which is very cool and desired (not by everyone, I suppose). Moreover, even old terminal programs are evolving. For example, now in vim you have very nice resizeable popup windows with elegant slim scrollbars (which are very missing in the ordinary windows unfortunately; I hope, they will be added there too one day).

> I recommend you experiment with what @LangLangBart suggested above.

Thank you both for that but so far I don't feel that it fully match what I need. I believe it shouldn't be a workaround. Each new workaround makes fzf's args even more complex. Perhaps, I will try to implement it myself locally, if you don't change your mind.

maxaykin:
> > I am afraid that for most people the name of raw mode itself would be not very intuitive. 😊
> 
> That's true. Any better key suggestions?

I don't have any at the moment. In some of my presets I use Alt-R to toggle recursive search and I don't want to mess with that. So, I've started using Alt-Z for toggling raw mode, maybe I will change it later.

maxaykin:
By the way, some time ago I made a local change in fzf to make AltUp/AltDown jump to the first/last item. I really wanted to have some bindings like these by default (again, because this is quite common and in many other programs/scenarios you have ability to jump to the beginning and the end of something from the box).

Now I found that they are bound to up-match/down-match and I have to switch to something else. I think I will stick to AltHome/AltEnd which is inline with AltUp/AltDown logic. Would you consider implementing these bindings as defaults?

junegunn:
> a local change in fzf to make AltUp/AltDown jump to the first/last item

> Now I found that they are bound to up-match/down-match and I have to switch to something else.

A local change to the source code? Couldn't you just set the bindings in your `$FZF_DEFAULT_OPTS`? Those settings will still be respected regardless of what we do here.

maxaykin:
> A local change to the source code? Couldn't you just set the bindings in your `$FZF_DEFAULT_OPTS`? Those settings will still be respected regardless of what we do here.

Yep, I forgot about this option. Thank you! Though, I have few more changes in the code which can't be solved as simple as this... :smile:

maxaykin:
Didn't you think about adding more colors for raw mode? Perhaps it is worth adding a color for matching items and background colors for both matching and non-matching items.

I think I would prefer highlighting the matching items with some background color (like selected-bg) over dimming all non-matching items.

maxaykin:
Even though there --raw switch, starting fzf in raw mode is not so useful as I expected it to be because typing a query does not lead to jumping to any match. Is it possible to improve this?

junegunn:
> Even though there --raw switch, starting fzf in raw mode is not so useful as I expected it to be because typing a query does not lead to jumping to any match. Is it possible to improve this?

```sh
fzf --raw --bind 'result:first+transform:[[ $FZF_RAW = 0 ]] && echo up-match'
```

> [!NOTE]
> It should be `down-match` if `--reverse` layout is used.
> Maybe we should make fzf export the direction of the list so you can just write `echo $FZF_DIRECTION-match`.
> This also applies to `up-selected`.
> To address this issue in `toggle-up` and `toggle-down`, we introduced `toggle-in` and `toggle-out`, but we can take the same approach there as well.

That said, I probably won't be starting fzf in `--raw` mode by default, and I haven't thought much about the use case.

alex-huff:
> The current behavior is:
> 
> * Normal -> Raw
>   
>   * The cursor stays on the same matching item.
> * Raw -> Normal
>   
>   * If the cursor is on a matching item: it stays on that item.
>   * If the cursor is on a non-matching item: since that item is not visible in normal mode, the cursor falls back to the last highlighted matching item.

This seemed to me like a good default behavior but today I came across a version of @maxaykin's scenario where this is particularly annoying/unintuitive:

When navigating around in raw mode if your last selected match barely matches (has a low score), going back to normal mode can result in your cursor being hundreds or thousands away from the useful (high score) matches. As far as I know, by default fzf doesn't have any keys bound to `first`, so there isn't a good way for the user to decrease their position without modifying the query, spamming page down, or using the scrollbar.

Here is a somewhat contrived example:

1. With `fzf` as my query, I search through git logs and find a commit of interest: 0a8ff789 (Do not unset **FZF**\_DEFAULT\_* variables when using winpty).
2. I enter raw mode to see the context.
3. After scrolling through the immediately following commits, I determine 0a8ff789 is not the commit I was looking for so I go back to normal mode to inspect a different commit.
4. At this point, my cursor position is ~400 because I unfortunately briefly selected commit 6684771c (**F**ix CTRL-**Z** **f**or tcell renderer by using the official API) which, although very close to my commit of interest in raw mode, is very far up in the list when toggling back to normal mode since it is a very poor match.

https://github.com/user-attachments/assets/a4c96d9b-03f0-4964-81c1-1ce5ea6b94d3

I wonder if saving/restoring the normal mode position is a better default behavior.

Think about the possible cases where a user goes from raw -> normal mode.
1. `The user is selecting an unmatched item.` If the user switches to normal mode from this state they likely don't care about whatever the last selected matched item was and just want to return to where they were last time in normal mode.
2. `The user incidentally has a matched item selected.` What I mean by this is they were navigating around in raw mode to see the context and happened to end up on a matched item, or maybe they never even moved. They had no intent in selecting this item. Similar to case 1, when switching back to normal mode they likely want to return to where they were last in normal mode.
3. `The user has a matched item of interest selected.` Given the current behavior, If while in raw mode the user finds and selects an item of interest, what is the point of switching to normal mode? Switching to normal mode would allow them to see their selected item next to other matched items with a similar score. Is this useful? If a user finds an item they care about in raw mode, viewing that item in the context of normal mode is unlikely to give them any valuable information. They probably will end up accepting the item instead of switching back to normal mode.

Another way to think of it is: Why would the user want to go from raw -> normal mode? I think generally the answer to this question is one of the following:
1. The user is done evaluating the context surrounding their item of interest.
2. The user was looking for some close by unmatched item that they could not find. (If they found it they would have accepted it)

Either way, when the user returns to normal mode I think they want to be back where they were before entering raw mode.

junegunn:
> Either way, when the user returns to normal mode I think they want to be back where they were before entering raw mode.

Here's a counterexample. If you see the demo gif I posted,

fzf-git-objects-raw

you can see me navigating commit hashes in chronological order using <kbd>CTRL-G-F</kbd> binding from [fzf-git.sh](https://github.com/junegunn/fzf-git.sh). Here, sorting is disabled with `--no-sort`, and I'm using an exact-match query (`'fix`). In this setup, moving through the list in raw mode works nicely; I don't end up stranded on a "deserted island". In this particular use case, I thought the current default made most sense. I mainly turn off raw mode just to hide the extra information, but I want to keep navigating from the same position I left off in raw mode.

That said, as you pointed out, when sorting is enabled, jumping to the last highlighted match can indeed leave you lost.

So should we make the behavior differ depending on whether sorting is enabled? Hmm, no, I guess it will just make the feature confusing. Anyway, we should provide a good default that is straightforward to understand, while still allowing both behaviors to be supported.

alex-huff:
> Here, sorting is disabled with `--no-sort`

I didn't think about the `--no-sort`. Forgot it existed. With sorting disabled the current behavior is definitely much better since the ordering of matches between both modes is identical. The metaphor of raw mode being an "expansion" of normal mode is valid.

> I mainly turn off raw mode just to hide the extra information, but I want to keep navigating from the same position I left off in raw mode.

Yes this makes sense to me for `--no-sort` but I think it is quite unintuitive without.

junegunn:
> So should we make the behavior differ depending on whether sorting is enabled?

Is this the best option we have? One can always use `transform` to achieve sophisticated customization, but it's obviously not for the faintest of hearts. And as discussed above, `store-pos` / `restore-pos` combo cannot solve this without a conditional branch via `transform`.

junegunn:
Another idea is to add separate actions like `toggle-raw-track` / `disable-raw-track`, but I don't love it :(

alex-huff:
Even without `--no-sort`, if `--scheme=history` is used with an exact-match query, the sorting of matched items between both modes would be identical correct? For this use-case the existing logical is also desirable.

alex-huff:
> > So should we make the behavior differ depending on whether sorting is enabled?
> 
> Is this the best option we have?

I think altering the behavior based on some presence of options is bound to be confusing. This is really hard :(

alex-huff:
Another thing about the current behavior is if you navigate around in raw mode with [half-]page-up/[half-]page-down you will likely skip over matches and land on unmatched items. When you switch back to normal mode would it make more sense to pick the closest matched item instead of the last selected matched item?

alex-huff:
> > > So should we make the behavior differ depending on whether sorting is enabled?
> > 
> > 
> > Is this the best option we have?
> 
> I think altering the behavior based on some presence of options is bound to be confusing. This is really hard :(

But maybe the added confusion is worth it because it protects the user from very unexpected behavior. If they navigate around in raw mode without `--no-sort` treating it like an "expanded" mode they can easily and unknowingly skip over the item they are looking for when they switch back to normal mode thinking they are just "collapsing" the items. The fact that the order of matched items is not consistent between modes (without `--no-sort`) is non-obvious and makes cycling through matches using `up-match` and `down-match` while simultaneously toggling modes a bad idea. I don't know.

maxaykin:
Hello guys, it is quite difficult for me to follow your discussion and to consider all these use cases but still I would like to make few notes regarding the behavior in general.

It looks like you think of raw mode as a temporary helper for normal mode. I would rather consider it as an alternative. Most probably I will enable it at start of fzf for inspection of commits history and list of running processes (output of 'ps). Perhaps, it will be also main for searches in vim's buffer. In these cases I may or may not need to switch to normal mode. So, I support the idea "to pick the closest matched item" when switching between the modes.

I have tried the following suggestion quickly and it doesn't seem to be the best option:
> ```shell
> fzf --raw --bind 'result:first+transform:[[ $FZF_RAW = 0 ]] && echo up-match'
> ```
No score is accounted, unconditonal 'first' changes the behavior for normal mode as well and 'first' should be 'last' in case of reverse. I understand, that this binding can be more or less adapted but it would be great if in raw mode selection would be done by default (without additional scripting) and in the same way as in normal mode, if possible. I believe this is just what users will expect.

> The only thing which seems still missing for me is navigation without change of selected item or preserving the selection somehow...
> Maybe an option that would make raw mode have it own cursor or would it be overcomplicated?

When I wrote this I mainly thought about keeping the selection while scrolling as such, regardless of what mode is currently enabled. Although I provided a proposal for switching from raw to normal I still think it would be great to have (an optional) ability to scroll without loosing the selection in general. Perhaps it would be also helpful for some of the cases you discussed.

junegunn:
> It looks like you think of raw mode as a temporary helper for normal mode.

Yeah, I do see it that way. Some users have requested for navigational functionality comparable to Vim, but then I wonder then why they don't use Vim. For example, for browsing Git commit hashes, I don't use fzf, I use Vim with [a plugin](https://github.com/junegunn/gv.vim). I have a shell alias for that and I use it every day:

```sh
alias gv='vim +GV +"autocmd BufWipeout <buffer> qall"'
```

I only use fzf for quickly picking commit hashes on the command-line using [fzf-git.sh](https://github.com/junegunn/fzf-git.sh) bindings, and I don't spend much time inside fzf in that case.

So "navigational functionality comparable to Vim" has not been an important goal of fzf.

> unconditonal 'first' changes the behavior for normal mode as well

You can make it conditional if you want.

> 'first' should be 'last' in case of reverse.

No, you don't have to change it. `first` refers to the match with the highest ("1st") rank. It used to be called `top`, but I renamed it to `first` to make it less confusing, though I can see how it might still cause some. When the list is not sorted (or in raw mode), `first` will point to the first item in the input list.

And as noted above, `up-match` can be replaced to `$FZF_DIRECTION-match` now so that it's layout-agnostic.

> No score is accounted

Raw mode presents the list in the original order, unsorted. If the cursor is automatically moved to the match with the highest score, your cursor could jump erratically across the list every time the query changes, which can be confusing and inconsistent with how fzf works in "normal" mode. In "normal" mode, your cursor position is not automatically updated, it just happens to be at the "first" position by default.

> Maybe an option that would make raw mode have it own cursor or would it be overcomplicated?

I think so.

> keeping the selection while scrolling

Using `offset-up` and `offset-down` can help to some degree. Like how Vim partially supports that.

(Excerpt from https://github.com/junegunn/fzf/releases/tag/0.43.0)

```sh
# Scrolling will behave similarly to CTRL-E and CTRL-Y of vim
fzf --bind scroll-up:offset-up,scroll-down:offset-down --scroll-off 0
```

maxaykin:
> Raw mode presents the list in the original order, unsorted. If the cursor is automatically moved to the match with the highest score, your cursor could jump erratically across the list every time the query changes, which can be confusing and inconsistent with how fzf works in "normal" mode. In "normal" mode, your cursor position is not automatically updated, it just happens to be at the "first" position by default.

Ok but the current behavior in "raw" mode is also inconsistent with how fzf works in "normal" mode. I mean in "normal" mode you type the query and see the result immediately while in "raw" mode (without additional scripting) you may not see any change at all if the nearest match is out of the displayed items. Maybe jumping to it would be a compromise?

> > It looks like you think of raw mode as a temporary helper for normal mode.
> 
> Yeah, I do see it that way. Some users have requested for navigational functionality comparable to Vim, but then I wonder then why they don't use Vim. For example, for browsing Git commit hashes, I don't use fzf, I use Vim with [a plugin](https://github.com/junegunn/gv.vim).

I noticed that you tend to use additional wrappers, layers and programs/plugins. This is not my way. I like to place all the functionality I need in one place, I prefer keeping it simple and using the same approaches for as many tasks as possible. For example, now I am moving my fzf-related code from vim functions to a single bash script. This has many advantages for me and the one of them is much simpler customization of the behavior because of no need to study and to debug the fzf vim plugin as an intermediate layer. Another important benefit is the ability to use the same fzf-related functions from both vim and the shell. So, starting vim with one more plugin just to browse git commits is too much for me. This way one may suggest using gitk instead of fzf :smile: 

I understand that users have different needs and preferences but still the tool should meet some basic expectations (which probably come from users' experience with other programs) and provide the needed functionality out of the box, at least as options. When you reject a proposal because it does not comply with the original concept of fzf, please think of this "raw" mode being introduced. It is probably out of the concept as well.

Anyway, sorry for bothering you with my verbose comments. Even though not all of what I would like to have is accepted I really enjoy my experience of using your tool. Thank you once again!

junegunn:
@maxaykin From what I gather, your request isn't directly tied to the idea of raw mode. You're asking for:

- an action to move the cursor to the best match, allowing you to write `--bind result:best`
- an enhancement to `offset-up` and `offset-down` so they can move the scroll offset beyond the limit of the visible screen.

Both of these could be implemented independently of raw mode.

junegunn:
> This way one may suggest using gitk instead of fzf 😄

Why not? I think that's a fair suggestion.

maxaykin:
> * an action to move the cursor to the best match, allowing you to write `--bind result:best`

Hmm, it's interesting but doesn't it make sense only in "raw" mode? Anyway, my complaint was about a possible confusion when a user sees no changes after entering a query in "raw" mode if all the displayed items are non-matches. My thought was that a jump to any (the nearest/next?) match would be better.

There was also a proposal of new colors:
> Didn't you think about adding more colors for raw mode? Perhaps it is worth adding a color for matching items and background colors for both matching and non-matching items.

The following is really out of scope of this feature:
> * an enhancement to `offset-up` and `offset-down` so they can move the scroll offset beyond the limit of the visible screen.

By the way, I have just had a try of it and it seems the actions work incorrectly with `--layout=reverse-list`: if you change direction of scrolling when the cursor is "locked" at 4th visible item, the list of items will still be moved in the same direction as before.

junegunn:
> Anyway, my complaint was about a possible confusion when a user sees no changes after entering a query in "raw" mode if all the displayed items are non-matches. My thought was that a jump to any (the nearest/next?) match would be better.

I really don't see it that way. Quite the opposite. Raw mode can be seen as a kind of "debug" mode where you can carefully inspect the full list and see exactly which items match the current query. I can change the query and observe which entries on screen match and which don't. Having the viewpoint automatically move would make it much harder to inspect the results. If I want to move across the matches, I can just press `CTRL-N` or `CTRL-P`.

> Didn't you think about adding more colors for raw mode?

The answer is no. I don't want more options there.

> The following is really out of scope of this feature

I agree, so I hope we can stop discussing about having a separate cursor in raw mode, or adding another navigation mode here.

> I have just had a try of it and it seems the actions work incorrectly with `--layout=reverse-list`

That might be a bug. I rarely use the `reverse-list` layout so it's possible that some features haven't been thoroughly tested with it.

junegunn:
@alex-huff 

> When you switch back to normal mode would it make more sense to pick the closest matched item instead of the last selected matched item?

I've tried to address that in edd08d5c.

alex-huff:
> @alex-huff
> 
> > When you switch back to normal mode would it make more sense to pick the closest matched item instead of the last selected matched item?
> 
> I've tried to address that in [edd08d5](https://github.com/junegunn/fzf/commit/edd08d5cc65f4acddff30e331747475402c2f619).

Thanks, appears to work as expected from my testing.

junegunn:
@maxaykin

- 7288daff071e3e860d26f5375a7567a1a44ff9ef implements `best` action.
- The bug of `offset-up` and `offset-down` was fixed in f4937c1df9c1d5b1f43f87bbffb3239c1f6a5c2b. Thanks for the report.

junegunn:
I think I'm going to wrap this up at this point. We haven't concluded whether we should move the cursor back to the previous position in non-raw mode when exiting raw mode, but I think we can only decide after rolling it out and hearing what users actually think in the real world. And we do have a way to implement the alternative behavior (using `$FZF_POS`).

alex-huff:
Thanks for working on this. Great addition to fzf.

Issue #4417, opened 2025-06-09
Option to disable Ctrl+R bindings (zsh)


seflue:
With alternative populare tools for incremental history search like [atuin](https://docs.atuin.sh/), it would be nice if the CTRL-R key binding could be disabled as well as CTRL-T and ALT-C. Otherwise I would have to trade fzf for atuin, but I would like to use both. And achieving that is currently quite difficult and brittle, because I need to maintain my own version of the fzf zsh integration, which for other users is not even an option.

It's also inconsisten, that CTRL-R is the only fzf binding, which has no corresponding env variable `$FZF_CTRL_R_COMMAND`.

junegunn:
Can't you just set up atuin later in your shell configuration file to override CTRL-R from fzf?

laur89:
In general junegunn is right, this _can be_ solved by project/plugin load order. However, consider `vi` mode in zsh that offers redo via `ctrl+r` in cmd mode -- it will still collide with fzf's ctrl+r binding regardless.

seflue:
> Can't you just set up atuin later in your shell configuration file to override CTRL-R from fzf?

That is my workaround. However, I am using zinit as plugin manager and with lazy loading you have no control, when something get's loaded, for the advantage that the shell starts much faster. A lot of zsh plugins allow very fine granual configuration via env variable, but for those, which doesn't, I have break out of the pattern. This has some disadvantages: The shell config get's more complicated and having to control, when something get's loaded makes the configuration much more brittle.

It is possible, no question, but it makes tools much less composable. And fzf for me is actually a tool, which is all about composability. So having the env variable to control the Ctrl+R keybinding would really help zsh users.

@junegunn Would you be open for a PR, which implements that?

junegunn:
Sure, please go ahead. Just make sure it works consistently across bash, zsh, and fish. By the way, we can't use the same `FZF_{KEY_NAME}_COMMAND=` convention, because CTRL-R doesn't use a simple, single command to generate the list. Maybe `FZF_CTRL_R_DISABLED`?

PR #4535, opened 2025-09-29
feat: Allow disabling Ctrl-R binding in shell integration

jdevera:
Use the `FZF_CTRL_R_DISABLED` environment variable with any value to disable the binding to Ctrl-R in shell integrations.

Fixes: #4417

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ This version introduces many new features centered around the new "raw" mode.
 | Enhancement | Key     | `CTRL-N`            | `down` -> `down-match`                             |
 | Enhancement | Key     | `CTRL-P`            | `up` -> `up-match`                                 |
 | Enhancement | Shell   | `CTRL-R` binding    | Toggle raw mode with `ALT-R`                       |
+| Enhancement | Shell   | `CTRL-R` binding    | Opt-out with an empty `FZF_CTRL_R_COMMAND`         |
 
 ### Introducing "raw" mode
 
@@ -230,6 +231,17 @@ socket is exported as `$FZF_SOCK`, analogous to `$FZF_PORT` for TCP sockets.
 As described above, `$FZF_RAW` is now exported to child processes in raw mode,
 indicating whether the current item is a match (`1`) or not (`0`). It is not
 defined when not in raw mode.
+
+#### `$FZF_CTRL_R_COMMAND`
+
+You can opt-out `CTRL-R` binding from the shell integration by setting
+`FZF_CTRL_R_COMMAND` to an empty string. Setting it to any other value is not
+supported and will result in a warning.
+
+```sh
+# Disable the CTRL-R binding from the shell integration
+FZF_CTRL_R_COMMAND= eval "$(fzf --bash)"
+```
 
 ### Added key support for `--bind`
 


jdevera:
Copilot's comment do raise the fact that the interface is weird for this setting. I followed the pattern of `FZF_ALT_C_COMMAND` but the semantics are different here.

The closest I find to correct if we want to keep the pattern of setting something to emtpy to disable it is to name the variable `FZF_CTRL_R_BINDING`.

Alternatively, we could redesign this interface (while keeping backwards compatibility with the two vars that already exist).

The shell integration provides these features, AFAIK:

* C-R binding
* C-T binding
* A-C binding
* Completion

And we could have env vars that disable each before loading. The current name `FZF_ALT_C_COMMAND` suggests one could add an alternative command, yet I don't see that in the code. So how about this plan:

1. Add new variables to control shell integration:
  * `FZF_SHELL_EXCLUDE_CTRL_R`
  * `FZF_SHELL_EXCLUDE_CTRL_T`
  * `FZF_SHELL_EXCLUDE_ALT_C`
  * `FZF_SHELL_EXCLUDE_COMPLETION`
  
2. Handle backwards compatibility by checking early for the two currently available vars and setting the new ones accordingly
3. Remove the old vars from the docs
4. Add the new vars to the docs

I'm happy to work on this once we agree on the interface.

junegunn:
> The current name `FZF_ALT_C_COMMAND` suggests one could add an alternative command, yet I don't see that in the code.

It is used. You can test it.

The semantic difference was noted in my original comment: https://github.com/junegunn/fzf/issues/4417#issuecomment-2995668509

jdevera:
Indeed, and sorry for all the back and forth for a simple change.

New plan: I keep this PR about excluding C-R and/or completion from the shell integration, and leave the C-T and A-C alone, since they are a different feature (setting alternatives rather than skipping).

Still, these variables:

* `FZF_SHELL_EXCLUDE_CTRL_R`
* `FZF_SHELL_EXCLUDE_COMPLETION`

But I could not find any examples of boolean env vars in fzf other than `NO_COLOR`, so I'm using that as a basis: If the flag is set to any non-empty value, then it's true, otherwise (unset or set to empty) it is false.

jdevera:
I've made the changes for C-R only. If you like them, I'll add the completion ones. Thanks again for the quick engagement!

junegunn:
Thanks for the update, but I'm not actually happy with the direction for the following reasons:

* I prefer the name `FZF_CTRL_R_DISABLED` because it shares the same prefix with the related configs.
* I've tried to avoid adding more configuration variables whenever possible. Adding a set of `EXCLUDE` variables is not what I'm looking for. If you don't want a part of the shell integration, you can:
    1. Directly source the individual script file from your config. I've noticed some advanced users actually prefer this over using `fzf --bash` or `fzf --zsh`, because it's much faster.
    1. Use the magic comments around each section to filter out unwanted parts
         - https://github.com/junegunn/fzf/pull/3675#issuecomment-1992799268
         - ```sh
           fzf --bash | sed '/### completion/,/### end/d'
           ```

jdevera:
All fair 👍
How do you want to proceed? This was just not 100% clear to me:

a) Support disable C-R with `FZF_CTRL_R_DISABLED` *only* (no completion option)
b) Discard this PR altogether

junegunn:
> a) Support disable C-R with FZF_CTRL_R_DISABLED only (no completion option

It would be ideal if we could support `FZF_CTRL_R_COMMAND` and follow the same convention here as well, but as you may have noticed, it's not simple because that custom command, unlike the other ones, should follow a strict protocol to be compatible with the rest of the script. For example, it should generate a null-terminated list to support multi-line history entries, and each item should start with an actual index number followed by a tab character. So considering the complexity, few will actually try to define their own command. And I don't think it's worth it either to add a conditional branch to lift these requirements for custom command.

So adding `FZF_CTRL_R_DISABLED` is something unavoidable.

An alternative would be to add `FZF_CTRL_R_COMMAND` just like the others, but only use it for disabling the feature, by checking if it's defined to be an empty-string. If it's defined to be a non-empty string, print a warning message that it's not "yet" supported. What do you think?

jdevera:
> An alternative would be to add FZF_CTRL_R_COMMAND just like the others, but only use it for disabling the feature, by checking if it's defined to be an empty-string. If it's defined to be a non-empty string, print a warning message that it's not "yet" supported. What do you think?

I think this is a great alternative, it's consistent with the rest in API, and warns users where it has potential for surprise. I gave it a go:

```
❯ FZF_CTRL_R_COMMAND=cat source shell/key-bindings.bash
warning: FZF_CTRL_R_COMMAND is set to a custom command, but custom commands are not yet supported for CTRL-R
```

In this case, the binding is still applied to the default one, since I assume that if a user is setting this with some non-empty value, they **want** the binding.

shell/key-bindings.bash
```diff
@@ -132,9 +133,14 @@ if ((BASH_VERSINFO[0] < 4)); then
   fi
 
   # CTRL-R - Paste the selected command from history into the command line
-  bind -m emacs-standard '"\C-r": "\C-e \C-u\C-y\ey\C-u`__fzf_history__`\e\C-e\er"'
-  bind -m vi-command '"\C-r": "\C-z\C-r\C-z"'
-  bind -m vi-insert '"\C-r": "\C-z\C-r\C-z"'
+  if [[ ${FZF_CTRL_R_COMMAND-x} != "" ]]; then
+    if [[ -n ${FZF_CTRL_R_COMMAND+x} && -n $FZF_CTRL_R_COMMAND ]]; then
```
junegunn: Is the former condition required? Shouldn't `-n ${FZF_CTRL_R_COMMAND-}` suffice?

jdevera: I will suffice, and it's also easier to reason about. Will change.

jdevera:
I hope we end up squashing these commits 😅

shell/key-bindings.bash
jdevera: I also learned that Fish (not a shell I use) is more permissive with unset variables, and so I simplified the condition there.

junegunn: Thanks. Reference: https://github.com/fish-shell/fish-shell/issues/4163

junegunn:
> I hope we end up squashing these commits 😅

Sure!

shell/key-bindings.zsh
```diff
@@ -150,10 +151,15 @@ fzf-history-widget() {
   zle reset-prompt
   return $ret
 }
-zle     -N            fzf-history-widget
-bindkey -M emacs '^R' fzf-history-widget
-bindkey -M vicmd '^R' fzf-history-widget
-bindkey -M viins '^R' fzf-history-widget
+if [[ "${FZF_CTRL_R_COMMAND-x}" != "" ]]; then
```
junegunn:
```suggestion
if [[ ${FZF_CTRL_R_COMMAND-x} != "" ]]; then
```

README.md
```diff
@@ -218,13 +218,13 @@ Add the following line to your shell configuration file.
...
+> You can disable CTRL-T, CTRL-R, or ALT-C bindings by setting the
+> corresponding `*_COMMAND` variable to an empty string when sourcing the
+> script. For example, to disable CTRL-R and ALT-C:
 >
-> * bash: `FZF_ALT_C_COMMAND= eval "$(fzf --bash)"`
-> * zsh: `FZF_ALT_C_COMMAND= source <(fzf --zsh)`
-> * fish: `fzf --fish | FZF_ALT_C_COMMAND= source`
+> * bash: `FZF_CTRL_R_COMMAND= FZF_ALT_C_COMMAND= eval "$(fzf --bash)"`
+> * zsh: `FZF_CTRL_R_COMMAND= FZF_ALT_C_COMMAND= source <(fzf --zsh)`
+> * fish: `fzf --fish | FZF_CTRL_R_COMMAND= FZF_ALT_C_COMMAND= source`
```
junegunn, line 227: This change will be immediately reflected in the README page, so to avoid confusion, I'll merge this PR just before I release a new version. Thanks.

jdevera: Great, thanks for the nice engagement.

junegunn:
Merged, thanks!

Issue #4572, opened 2025-10-29
binding `backspace` is ignored


Jaehaks:
### Checklist

- [x] I have read through the manual page (`man fzf`)
- [x] I have searched through the existing issues
- [x] For bug reports, I have checked if the bug is reproducible in the latest version of fzf

### Output of `fzf --version`

0.66.1 (c1b259c0)

### OS

- [ ] Linux
- [ ] macOS
- [x] Windows
- [ ] Etc.

### Shell

- [ ] bash
- [ ] zsh
- [ ] fish

### Problem / Steps to reproduce

After update, default backspace behavior is changed to other, not `backward-delete-char`.
I am Windows 10 user and use cmd prompt.

I add `backspace` keymap to option like this, but it doesn't work.
`backward-delete-char` does work when I set the other key like `ctrl-p`.

```powershell
set FZF_DEFAULT_OPTS=^
--layout=reverse ^
--border ^
--color=hl:#FF0000,hl+:#00FF00 ^
--preview="bat --color=always {}" ^
--preview-window "down,50%%,hidden" ^
--bind=ctrl-j:up,^
ctrl-k:down,^
ctrl-d:half-page-down,^
ctrl-u:half-page-up,^
ctrl-l:forward-char,^
ctrl-h:backward-char,^
backspace:backward-delete-char,^
ctrl-a:beginning-of-line,^
ctrl-e:end-of-line,^
ctrl-i:toggle-preview,^
alt-k:preview-half-page-down,^
alt-j:preview-half-page-up
```

junegunn:
Did it work in 0.66.0?

Jaehaks:
backspace work as `backward-delete-char` in `0.66.0` without any settings

junegunn:
Then I think I broke it in a0cabe021d76de5b6227d23b09808ffbe114aadd.

masmu:
I think it could be as easy as reverting the changes to `src/tui/tcell.go`.

@junegunn Or are there any reasons why this would break things in regards to [issue 4556](https://github.com/junegunn/fzf/issues/4556)?

I might get my hands on a windows machine in the next days ... Lets see.

junegunn:
The change was needed to make fzf with tcell build distinguish ctrl-h and backspace on my macbook. You can test it with `go run -tags tcell main.go --bind  ..`. Maybe things are a little different in Windows and I should have tested there.

masmu:
[...]

PR #4582, opened 2025-11-10
Fix modifier detection for Backspace / Ctrl-H on Windows (#4572)

masmu:
Windows sends different key events and modifier combinations to the FullscreenRenderer than a FullscreenRenderer on Linux (`-tags tcell`). This led to Ctrl+H being misinterpreted (and therefore unbindable) on some Windows builds.

Basically reverts changes to `src/tui/tcell.go` introduced by `a0cabe0`.


junegunn:
Thanks, let me just merge this because we don't use tcell on *nix anyway.

PR #4595, opened 2025-11-18
feat(zsh): Handle multi-line history selection

LangLangBart:
Related discussion https://github.com/junegunn/fzf/discussions/4591

@drelephant

Related PR with implementation in fish: #4280

---

Add the ability to select multiple zsh history entries

### DIFF
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -170,8 +170,8 @@ fzf-history-widget() {
         fi
       done
       if (( ${#cmds[@]} )); then
-        # Join by newline after stripping trailing whitespace from each command
-        BUFFER="${(pj:\n:)${(@)cmds%%[[:space:]]#}}"
+        # Join by newline after stripping trailing newlines from each command
+        BUFFER="${(pj:\n:)${(@)cmds%%$'\n'#}}"
         CURSOR=${#BUFFER}
       fi
     else # selected is a custom query, not from history
--- a/test/test_shell_integration.rb
+++ b/test/test_shell_integration.rb
@@ -490,7 +490,7 @@ class TestZsh < TestBase
     tmux.prepare
     tmux.send_keys 'echo "bar', :Enter, 'foo"', :Enter
     tmux.prepare
-    tmux.send_keys 'echo "trailing"  ', :Enter
+    tmux.send_keys 'echo "trailing_space "', :Enter
     tmux.prepare
     tmux.send_keys 'cat <<EOF | wc -c', :Enter, 'qux thud', :Enter, 'EOF', :Enter
     tmux.prepare
@@ -525,7 +525,7 @@ class TestZsh < TestBase
     tmux.until { |lines| assert_includes lines[-2], '(3)' }
     tmux.send_keys :Enter
     tmux.until do |lines|
-      assert_equal ['cat <<EOF | wc -c', 'qux thud', 'EOF', 'echo "trailing"', 'echo "bar', 'foo"'], lines[-6..]
+      assert_equal ['cat <<EOF | wc -c', 'qux thud', 'EOF', 'echo "trailing_space "', 'echo "bar', 'foo"'], lines[-6..]
     end
   end
 


shell/key-bindings.zsh
```diff
@@ -128,27 +128,44 @@ fi
...
+    if [[ $selected == [[:blank:]]#(<->)[[:blank:]]* ]]; then
+      # Split at newlines
+      for line in ${(f)selected}; do
+        if (( extracted_with_perl )); then
+          [[ $line == (#b)(<->)(#B)$'\t'* ]] && cmds+=("${history[${match[1]}]}")
+        elif [[ $line == [[:blank:]]#(#b)(<->)(#B)[[:blank:]]* ]]; then
+          if zmodload -F zsh/parameter p:history 2>/dev/null;then
+            cmds+=("${history[${match[1]}]}")
+          else
+            cmds+=("${(g::)"$(fc -ln ${match[1]} ${match[1]})"}")
```
LangLangBart: Usually the associative `history` array is accessible; fall
back to `fc` and use `(g::)` to process escape sequences.
 
- [zsh: 14 Parameter Expansion Flags - g:opts:](https://zsh.sourceforge.io/Doc/Release/Expansion.html#Expansion:~:text=pj%3A%5Cn%3A%E2%80%99.-,g%3Aopts%3A,-Process%20escape%20sequences)
- https://www.bashsupport.com/zsh/parameter-expansion-flags/g/

example:

```bash
{ 
  seq 5
  echo
} | wc -l
```

```bash
str=$(fc -ln -1 -1)
typeset -p str
# typeset str='{ \n  seq 5\n  echo\n} | wc -l'

str=${(g::)"$(fc -ln -1 -1)"}
typeset -p str
# typeset str=$'{ \n  seq 5\n  echo\n} | wc -l'
```

shell/key-bindings.zsh
```diff
@@ -128,27 +128,44 @@ fi
...
+  local -a mbegin mend match
   if [ -n "$selected" ]; then
-    if [[ $(__fzf_exec_awk '{print $1; exit}' <<< "$selected") =~ ^[1-9][0-9]* ]]; then
-      zle vi-fetch-history -n $MATCH
+    # Heuristic to check if the selected value is from history or a custom query
+    if [[ $selected == [[:blank:]]#(<->)[[:blank:]]* ]]; then
+      # Split at newlines
+      for line in ${(f)selected}; do
+        if (( extracted_with_perl )); then
+          [[ $line == (#b)(<->)(#B)$'\t'* ]] && cmds+=("${history[${match[1]}]}")
```
LangLangBart: the glob is intended to match the `printf ... | perl ...` pattern
- `(#b)` Starts backreferences, `(#B)` deactivate backreferences
- `(<->)` will be accessible as `${match[1]}` if the glob matches

- [zsh: 14 Globbing Flags - backreferences](https://zsh.sourceforge.io/Doc/Release/Expansion.html#Expansion:~:text=Activate%20backreferences%20for%20parenthesised)

shell/key-bindings.zsh
```diff
@@ -128,25 +128,42 @@ fi
...
+  local selected extracted_with_perl=0
+  setopt localoptions noglobsubst noposixbuiltins pipefail no_aliases noglob nobash_rematch extendedglob 2> /dev/null
   # Ensure the module is loaded if not already, and the required features, such
   # as the associative 'history' array, which maps event numbers to full history
   # lines, are set. Also, make sure Perl is installed for multi-line output.
   if zmodload -F zsh/parameter p:{commands,history} 2>/dev/null && (( ${+commands[perl]} )); then
     selected="$(printf '%s\t%s\000' "${(kv)history[@]}" |
       perl -0 -ne 'if (!$seen{(/^\s*[0-9]+\**\t(.*)/s, $1)}++) { s/\n/\n\t/g; print; }' |
-      FZF_DEFAULT_OPTS=$(__fzf_defaults "" "-n2..,.. --scheme=history --bind=ctrl-r:toggle-sort,alt-r:toggle-raw --wrap-sign '\t↳ ' --highlight-line ${FZF_CTRL_R_OPTS-} --query=${(qqq)LBUFFER} +m --read0") \
+      FZF_DEFAULT_OPTS=$(__fzf_defaults "" "-n2..,.. --scheme=history --bind=ctrl-r:toggle-sort,alt-r:toggle-raw --wrap-sign '\t↳ ' --highlight-line -m ${FZF_CTRL_R_OPTS-} --query=${(qqq)LBUFFER} --read0") \
```
LangLangBart: Place the `-m` flag before `FZF_CTRL_R_OPTS` to allow users to optionally disable multi-selection

```bash
export FZF_CTRL_R_OPTS='--no-multi'
```

shell/key-bindings.zsh
```diff
@@ -128,25 +128,42 @@ fi
...
+          [[ $line == (#b)(<->)(#B)$'\t'* ]] && cmds+=("${history[${match[1]}]}")
+        elif [[ $line == [[:blank:]]#(#b)(<->)(#B)[[:blank:]]* ]]; then
+          if zmodload -F zsh/parameter p:history 2>/dev/null;then
+            cmds+=("${history[${match[1]}]}")
+          else
+            cmds+=("${(g::)"$(fc -ln ${match[1]} ${match[1]})"}")
+          fi
+        fi
+      done
+      LBUFFER="${(pj:\n:)${(@)cmds%%$'\n'##}}"
```
LangLangBart: `(pj:\n:)` Join  with a newline; it is equivalent to `(F)`, but I prefer having `j:...:` for mnemonic reasons
  - I decided to not use semicolons to join the commands, as it can lead to issues with heredoc commands

`${(@)cmds%%$'\n'##}`: remove from each array element all trailing newline chars

- [zsh: 14 Parameter Expansion Flags - pj:\n:](https://zsh.sourceforge.io/Doc/Release/Expansion.html#Expansion:~:text=Join%20the%20words%20of%20arrays%20together%20using%20newline%20as%20a%20separator)
- https://www.bashsupport.com/zsh/parameter-expansion-flags/f_uppercase/
- https://www.bashsupport.com/zsh/parameter-expansion-flags/_at_/

shell/key-bindings.zsh
```diff
@@ -128,25 +128,42 @@ fi
...
+  local -a cmds
+  # Avoid leaking auto assigned values when using backreferences '(#b)'
+  local -a mbegin mend match
   if [ -n "$selected" ]; then
-    if [[ $(__fzf_exec_awk '{print $1; exit}' <<< "$selected") =~ ^[1-9][0-9]* ]]; then
-      zle vi-fetch-history -n $MATCH
+    # Heuristic to check if the selected value is from history or a custom query
+    if [[ $selected == [[:blank:]]#<->[[:blank:]]* ]]; then
+      # Split at newlines
+      for line in ${(ps:\n:)selected}; do
```
LangLangBart, line 156: `(ps:\n:)` Split at newline; it is equivalent to `(f)`, but I prefer having `s:...:` for mnemonic reasons

- [zsh: 14 Parameter Expansion Flags - ps:\n:](https://zsh.sourceforge.io/Doc/Release/Expansion.html#Expansion:~:text=Split%20the%20result%20of%20the%20expansion%20at%20newlines)
- https://www.bashsupport.com/zsh/parameter-expansion-flags/f/

shell/key-bindings.zsh
```diff
@@ -128,25 +128,42 @@ fi
...
   fi
   local ret=$?
+  local -a cmds
+  # Avoid leaking auto assigned values when using backreferences '(#b)'
+  local -a mbegin mend match
   if [ -n "$selected" ]; then
-    if [[ $(__fzf_exec_awk '{print $1; exit}' <<< "$selected") =~ ^[1-9][0-9]* ]]; then
-      zle vi-fetch-history -n $MATCH
+    # Heuristic to check if the selected value is from history or a custom query
+    if [[ $selected == [[:blank:]]#<->[[:blank:]]* ]]; then
```
LangLangBart: A glob to check whether the selection might be a history entry or from the query, e.g., when the user used `accept-or-print-query`.

It’s also possible to keep the existing `$(__fzf_exec_awk '{pri...` line, but the following lines all make use of the enabled `extendedglob` option, so I’ve altered this line to be consistent with the rest and its shorter.

- `[[:blank:]]#`: matches leading whitespace
- `<->`: matches any number

shell/key-bindings.zsh
```diff
@@ -128,25 +128,42 @@ fi
...
+      # Split at newlines
+      for line in ${(ps:\n:)selected}; do
+        if (( extracted_with_perl )); then
+          [[ $line == (#b)(<->)(#B)$'\t'* ]] && cmds+=("${history[${match[1]}]}")
+        elif [[ $line == [[:blank:]]#(#b)(<->)(#B)[[:blank:]]* ]]; then
+          if zmodload -F zsh/parameter p:history 2>/dev/null && (( ${+history[${match[1]}]} )); then
+            cmds+=("${history[${match[1]}]}")
```
LangLangBart: ❌❌❌ This is the part I’m concerned about: the history number here was created
with the `fc` builtin.

In the past, we’ve stumbled over foreign commands (the ones with a `*`) not
appearing in the `history` array:
- Ref: https://github.com/junegunn/fzf/issues/4061
- Ref: https://zsh.org/mla/users/2024/msg00692.html

What the user selects via fzf and what is actually delivered to them may differ;
this was partially alleviated by using `(( ${+history[${match[1]}]} ))`, but it
can still lead to issues for users. Even though I can't provide a good example,
at the moment.

Another problem with `"${(g::)"$(fc -ln ${match[1]} ${match[1]})"}"` is that it
treats every escape sequence the same, whereas retrieving from `history`
preserves the command exactly as it was entered.

For example, consider this demo command:

```bash
{
  seq 5
  echo '\n'
} | wc -l
```

✅ Using the `"${history[${match[1]}]}"` array to retrieve the command yields the
same command the user entered:

```bash
{
  seq 5
  echo '\n'
} | wc -l
```

❌ Using `"${(g::)"$(fc -ln ${match[1]} ${match[1]})"}"` causes `'\n'` to be
expanded:

```bash
{
  seq 5
  echo '
'
} | wc -l
```

---

SUMMARY: As long as the user has Perl and uses the `history` array, there
shouldn’t be an issue. The `fc` builtin and `history` array crossover is a
potenial issue.

LangLangBart: This issue has been fixed: for retrieving 'fc' history event numbers, rely on the previously used 'zle vi-fetch-history,' then read the command from the BUFFER.

junegunn:
@drelephant Can you test this and see if it works as expected? You can clone this repo, `make install`, and run `source <(bin/fzf --zsh)`.

LangLangBart:
rebased onto master and triggered the CI again; tests now pass, whereas previously it failed:
- https://github.com/junegunn/fzf/actions/runs/19523769699/job/55892204472

---

~~@junegunn Is there an easy way to temporarily disable or hide Perl when running a certain test, so I can checkZsh shell integration on OS versions that don’t have Perl installed by default?~~

**EDIT1**  `unset 'commands[perl]'` does the trick
- Ref: https://zsh.sourceforge.io/Doc/Release/Shell-Builtin-Commands.html#index-unset

test/test_shell_integration.rb
```diff
@@ -462,6 +462,70 @@ def test_complete_quoted_command
...
+  end
+
+  def disable_perl
+    tmux.send_keys "unset 'commands[perl]'", :Enter
+    tmux.prepare
+    # Verify perl is actually unset (0 = not found)
+    tmux.send_keys 'echo ${+commands[perl]}', :Enter
+    tmux.until { |lines| assert_equal '0', lines[-1] }
+    tmux.prepare
+  end
```
LangLangBart: I didn't want to duplicate the logic from `TestZsh` for running with `unset 'commands[perl]'`.

As an unwanted side effect, this now triggers not just the `test_ctrl_r_*` tests, but also unrelated tests.

```txt
TestZsh#test_alt_c = 0.28 s = .
TestZshNoPerl#test_ctrl_r_multi_selection = 0.56 s = .
TestZshNoPerl#test_ctrl_r_abort = 0.62 s = .
TestZshNoPerl#test_unset_completion = 0.59 s = .
TestZshNoPerl#test_ssh_completion = 0.47 s = .
TestZshNoPerl#test_file_completion_unicode = 0.38 s = .
TestZshNoPerl#test_ctrl_r = 0.84 s = .
TestZshNoPerl#test_dir_completion = 0.37 s = .
TestZshNoPerl#test_ctrl_r_multiline = 0.37 s = .
TestZshNoPerl#test_process_completion = 0.41 s = .
TestZshNoPerl#test_file_completion_root = 0.28 s = .
TestZshNoPerl#test_ctrl_t = 0.49 s = .
TestZshNoPerl#test_ctrl_r_accept_or_print_query = 0.69 s = .
TestZshNoPerl#test_file_completion = 0.70 s = .
TestZshNoPerl#test_alt_c_command = 0.62 s = .
TestZshNoPerl#test_ctrl_r_no_multi_selection = 0.70 s = .
TestZshNoPerl#test_ctrl_t_unicode = 0.52 s = .
TestZshNoPerl#test_custom_completion_api = 0.66 s = .
TestZshNoPerl#test_custom_completion = 0.36 s = .
TestZshNoPerl#test_complete_quoted_command = 0.78 s = .
TestZshNoPerl#test_completion_in_command_sequence = 1.55 s = .
TestZshNoPerl#test_alt_c = 0.42 s = .
TestRaw#test_raw_mode = 0.17 s =
```

LangLangBart: handled with a `test_perl_and_awk` helper function

junegunn:
Sorry, I haven't had the bandwidth to review this yet. I just requested Copilot's review, let's see if it can provide meaningful insights. But feel free to dismiss any comments from it that you consider irrelevant or inaccurate.

test/test_shell_integration.rb
```diff
@@ -462,6 +462,84 @@ def test_complete_quoted_command
       tmux.send_keys 'C-c'
     end
   end
+
+  # Helper function to run test with Perl and again with Awk
+  def self.test_perl_and_awk(name, &block)
```
junegunn, line 467: Nice idea!

shell/key-bindings.zsh
```diff
@@ -128,25 +128,51 @@ fi
...
+          zle .push-line
+          zle vi-fetch-history -n ${match[1]}
+          (( "${#BUFFER}" )) && cmds+=("${BUFFER}")
+          BUFFER=""
+          zle .get-line
+        fi
+      done
+      if (( ${#cmds[@]} )); then
+        # Join by newline after stripping trailing whitespace from each command
+        BUFFER="${(pj:\n:)${(@)cmds%%[[:space:]]#}}"
```
junegunn: Just wondering. If I'm not mistaken, in the previous version, this code would only strip trailing new lines, but now it strips `[[:space:]]`. Could you elaborate on the change?

https://github.com/junegunn/fzf/pull/4595#discussion_r2536577262

LangLangBart: During testing, a command had a trailing newline & space character, and since the
original command only removed trailing newlines, it failed to remove anything. For example:

```bash
zsh -o extendedglob -fc $'
ORIGINAL="1 foo\nbar\n "
O_WSPACE="${ORIGINAL%%[[:space:]]#}"
O_NEWLIN="${ORIGINAL%%\n#}"
typeset -p ORIGINAL O_WSPACE O_NEWLIN'
```

```txt
typeset ORIGINAL=$'1 foo\nbar\n '
typeset O_WSPACE=$'1 foo\nbar'
typeset O_NEWLIN=$'1 foo\nbar\n '
```

My thinking was nobody wants trailing whitespace on a command, right 😬 ?

(Though there is a chance a user may complain, as it affects their workflow – https://www.explainxkcd.com/wiki/index.php/1172:_Workflow.)

LangLangBart:
> Is there anything you want to address?

There is the comment by Copilot.

When entering `1 foobar€` into the query input and having no results, and with
`export FZF_CTRL_R_OPTS='--bind enter:accept-or-print-query'` set, **Perl**
would not surface anything to the user, while **Awk** would see `1` as the
history event number and fetch the corresponding command.

Status quo with the latest release (0.67.0), both **Perl** and **Awk**
incorrectly retrieve the history value for `1` instead of printing `1 foobar€`
to the command-line buffer.

Do you think the I should make the heuristic stricter? At least for **Perl** it
would then correctly print `1 foobar€` to the command line buffer, but **Awk** would
still be "fooled".

```diff
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -151,5 +151,6 @@ fzf-history-widget() {
   if [ -n "$selected" ]; then
     # Heuristic to check if the selected value is from history or a custom query
-    if [[ $selected == [[:blank:]]#<->\*#[[:blank:]]* ]]; then
+    if (( extracted_with_perl )) && [[ $selected == <->$'\t'* ]] || \
+      (( ! extracted_with_perl )) && [[ $selected == [[:blank:]]#<->\*#[[:blank:]]* ]]; then
       # Split at newlines
       for line in ${(ps:\n:)selected}; do
```

Of course I could then also add a proper test, which only **Perl** would actually
pass, that is the reason I have simply dismissed this case?

shell/key-bindings.zsh
junegunn: > My thinking was nobody wants trailing whitespace on a command, right 😬 ?

Yeah. One concern is that when:

- User accidentally types in a command with extra spaces.
- User wants to repeat the command, so they use CTRL-R to select the previous command, and press enter.
- Now when they press CTRL-R again, they see two duplicate commands in the list, only differ in the trailing spaces.

```
100   whoami<invisible spaces>
101   whoami

File: shell/key-bindings.zsh [hunk 11]
```diff
@@ -128,25 +128,51 @@ fi
 
 # CTRL-R - Paste the selected command from history into the command line
 fzf-history-widget() {
-  local selected
-  setopt localoptions noglobsubst noposixbuiltins pipefail no_aliases noglob nobash_rematch 2> /dev/null
+  local selected extracted_with_perl=0
+  setopt localoptions noglobsubst noposixbuiltins pipefail no_aliases no_glob no_ksharrays extendedglob 2> /dev/null
```
LangLangBart, line 132: Option changes remarks:
- `no_ksharrays` is needed. By default, array elements are numbered from one in
  zsh; when a user sets `ksharrays`, array elements are numbered from zero
  - I believe 3a6af27586c6 was the right approach, and reverting it was a
    mistake, but the original author deleted their commit post so I can't see what the
    actual issue was. `emulate -L zsh` would therefore automatically set the
    default array behavior, and there would be no need for setting `no_ksharrays`
- `no_glob` and `extendedglob` do not conflict with one another and serve
  different purposes (man zshoptions)
- `extendedglob` provides additional pattern‑matching syntax (such as `#` or
  flags like `(#b)`)
- `nobash_rematch` was only useful for `=~` usage

junegunn:
> Status quo with the latest release (0.67.0), both **Perl** and **Awk** incorrectly retrieve the history value for `1` instead of printing `1 foobar€` to the command-line buffer.

That is unfortunate.

> Do you think the I should make the heuristic stricter? At least for **Perl** it
> would then correctly print `1 foobar€` to the command line buffer, but **Awk** would
> still be "fooled".

I think it's understandable we can't "perfect" this. fzf can be configured in numerous ways and it's unrealistic for us to make these scripts compatible with every possible setup.

That said, if the added complexity is not too significant, we can try to cover more cases. I'll leave the decision up to you.

shell/key-bindings.zsh
LangLangBart: Right, I was able to reproduce what you described. I am using
'romkatv/zsh4humans', and it has a mechanism[^1] built in by default that
automatically removes trailing spaces from a command when it is executed, that
is why it took me a bit to reproduce your issue and find out why its not
behaving the same way as you described and only did in 'zsh -f' mode.

---

Anyway, are you suggesting it’s better to keep the original trailing‑newline
removal?

Because even when reverting back, your original issue still remains: a user sees
two commands.

[^1]: https://github.com/romkatv/zsh4humans/blob/cd6c4770c802c3a17b4c43e5587adabb9a370a75/fn/-z4h-zle-line-finish#L15-L19

LangLangBart:
>  I'll leave the decision up to you.

Ok, I think we can make the heuristic a bit stricter for **Perl/$history**, and technically for **Awk/fc** as well, since by default the history event number (using **fc**) and the actual command are separated by two spaces **or**  a `*` for foreign commands and one space, and this has been the case for a long time[^1].

I will work on that and let you know later today, so that we can also cover the `1 foobar€` case and have a test with it.

[^1]: https://github.com/zsh-users/zsh/blame/5539bc3fd59a2577bf1f951430c20e2b1e7b4dce/Src/builtin.c#L1829

shell/key-bindings.zsh
```diff
@@ -128,25 +128,51 @@ fi
...
   fi
   local ret=$?
+  local -a cmds
+  # Avoid leaking auto assigned values when using backreferences '(#b)'
+  local -a mbegin mend match
   if [ -n "$selected" ]; then
-    if [[ $(__fzf_exec_awk '{print $1; exit}' <<< "$selected") =~ ^[1-9][0-9]* ]]; then
-      zle vi-fetch-history -n $MATCH
+    # Heuristic to check if the selected value is from history or a custom query
+    if [[ $selected == [[:blank:]]#<->\*#[[:blank:]]* ]]; then
```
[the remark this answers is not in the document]
LangLangBart: see 6bf96a9542a2

drelephant:
> @drelephant Can you test this and see if it works as expected? You can clone this repo, `make install`, and run `source <(bin/fzf --zsh)`.

I did that (using `./install` instead of `make install` which came up with errors) and now I can't multi-select using `tab` after pressing `ctrl-r`.

I was previously using the code from [here](https://github.com/junegunn/fzf/discussions/4591#discussioncomment-14989724) in my `.zshrc` but I took that out. Should I still need that in my `.zshrc`?

Also, before I took that code out of the `.zshrc`, I noticed that if I do this:

1. type something onto the command line, eg "vi"
2. press `ctrl-r`
3. go up to select a previous usage of vi and hit enter

Then the command line will now have "vivi file" (ie it kept what was on the line before entering fzf.

I don't think it did that before updating to the latest commit?

Anyway, I'm on the latest commit now without anything extra in my `.zshrc` except
```
export FZF_CTRL_R_OPTS="
  --preview 'echo {}' --preview-window down:3:hidden:wrap
  --bind 'ctrl-p:toggle-preview'
  --multi
  --no-height"
```
And I can't multi-select using tab key.

shell/key-bindings.zsh
junegunn: > Anyway, are you suggesting it’s better to keep the original trailing‑newline removal?

Hmm, yeah, I think it's preferable that we don't modify the original command.

> Because even when reverting back, your original issue still remains: a user sees two commands.

I didn't test your previous version. I might be missing something, but on zsh, when I repeat the same command multiple times, there's no duplication in the list. Are we on the same page?

```sh
$ foo<space><space>
$ foo<space><space>
$ foo<space><space>

# CTRL-R shows only one "foo<space><space>"

$ foo<space>
$ foo

# CTRL-R shows "foo<space><space>", "foo<space>", and "foo"
```

junegunn:
@drelephant 

> I did that (using `./install` instead of `make install` which came up with errors) and now I can't multi-select using `tab` after pressing `ctrl-r`.

`./install` will download the latest pre-built binary, it does not build a fzf binary from source. We use `source <(fzf --zsh)` these days so you're using the the shell integration scripts embedded in that binary.

If you can't build the binary from source, you can just source the script files from your zsh.

shell/key-bindings.zsh
LangLangBart: > Are we on the same page?

Yes, we are. When viewing these commands, they appear to be the same, with the trailing space being the only difference.



> Hmm, yeah, I think it's preferable that we don't modify the original command.

ok, will adjust that, and adopt test `BUFFER="${(pj:\n:)${(@)cmds%%$'\n'#}}"`

LangLangBart:
> Then the command line will now have "vivi file"

should be fixed in this version; the one from the discussion used the LBUFFER assignment instead of BUFFER

> I did that (using ./install instead of make install which came up with errors) 

thanks @junegunn 

@drelephant If you still have trouble testing, one way would be:

1. download the file: https://github.com/LangLangBart/fzf/blob/zsh-hist-multi-select/shell/key-bindings.zsh  
2. run `source <path to your downloaded file>`  
3. press `ctrl‑r`

---

I think I am done.

updated the test, checked older zsh versions with the command below: 

| Version | Release Date        | AWK | Perl |
| ------- | ------------------- | --- | ---- |
| 4.3.15  | 2011-12-17          | ✅   | ✅    |
| 5.0.2   | 2012-12-21          | ✅   | ✅    |
| 5.4   | 2017-08-07           | ✅   | ✅    |
| 5.8     | 2020-02-15          | ✅   | ✅    |
| 5.9     | 2022-05-14 (latest) | ✅   | ✅    |

```bash
# Requires fzf repo directory and 'zsh-hist-multi-select' branch checked out
# https://hub.docker.com/r/ohmyzsh/zsh/tags
podman run -qit --rm -e ARCH=$(uname -m) -v $PWD:/repo -w /repo ohmyzsh/zsh:5.9 sh -uec '
apt update >/dev/null 2>&1
apt install -y curl jq >/dev/null 2>&1
mkdir /fzf-bin
download_url=$(curl -s https://api.github.com/repos/junegunn/fzf/releases/latest  | \
  jq --arg arch_string "linux_$ARCH" \
    -r '\''.assets[].browser_download_url | select(contains($arch_string))'\'')
curl -sfLo /tmp/fzf.tar.gz "$download_url"
tar -xf /tmp/fzf.tar.gz -C /fzf-bin
{
  echo "export PATH=\"/fzf-bin:\$PATH\""
  echo "source /repo/shell/key-bindings.zsh"
  echo "PS1=\"FZF \$(fzf --version) > \""
  echo "RPS1=\"ZSH \$(zsh --version)\""
  echo "print -s -- \"ls -la\""
  echo "print -s -- \"cat /etc/os-release\""
} > ~/.zshrc
zsh'
```

junegunn:
Merged, thanks a ton for the great work!

drelephant:
> `./install` will download the latest pre-built binary, it does not build a fzf binary from source. We use `source <(fzf --zsh)` these days so you're using the the shell integration scripts embedded in that binary.
> 
> If you can't build the binary from source, you can just source the script files from your zsh.

I just did a `git pull`
```
remote: Enumerating objects: 16, done.
remote: Counting objects: 100% (15/15), done.
remote: Compressing objects: 100% (6/6), done.
remote: Total 6 (delta 5), reused 0 (delta 0), pack-reused 0 (from 0)
Unpacking objects: 100% (6/6), 2.59 KiB | 1.29 MiB/s, done.
From https://github.com/junegunn/fzf
   af2a81d..cfc37ca  master     -> origin/master
Updating af2a81d..cfc37ca
Fast-forward
 shell/key-bindings.zsh         | 39 +++++++++++++++++----
 test/test_shell_integration.rb | 78 ++++++++++++++++++++++++++++++++++++++++++
 2 files changed, 111 insertions(+), 6 deletions(-)
```
So I'm on the latest now.

Then `source <(fzf --zsh)`, `source ~/.zshrc ; exec zsh`

And I couldn't multi-select, but then I did `source ~/.fzf/shell/key-bindings.zsh` and then it worked.

I can add that line to my `~/.zshrc`, but is it odd that `source <(fzf --zsh)` didn't work properly?

junegunn:
`fzf --zsh` prints the script embedded in that binary, so updating the external script file doesn't affect its. That's why we told you to build the binary from the source.

PR #4596, opened 2025-11-18
fix(terminal): correct display width calculation with maxWidth

### LINKED ISSUE #4593: Multi-character ellipsis leaves right-side artifacts with freeze-right option enabled
### Checklist

### Output of `fzf --version`

0.67.0 (2ab923f3)

### OS

- [ ] Linux
- [x] macOS
- [ ] Windows
- [ ] Etc.

### Shell

- [ ] bash
- [x] zsh
- [ ] fish

Long ellipsis causes character artifact 

issue started with: 3f499f055ee04fa9a3f0691b00c1088da2bab475

### to reproduce

1. Run the command below; ellipsis should be more than 2 chars
2. Keep pressing the up arrow key 
3. Notice the chars on the right side remain visible 

```bash
{
  seq 30
  printf '%.0sg' {1..1000}
  echo
  seq 32 100
} |
  FZF_DEFAULT_OPTS='' fzf \
    --ellipsis='...' \
    --freeze-right 1
```

Image

---

workaround: trim the ellipsis to 2 dots only

LangLangBart:
fix #4593

---

reproduction described in the ref ticket

~~is a test needed ?~~

### DIFF
--- a/test/test_core.rb
+++ b/test/test_core.rb
@@ -1235,6 +1235,16 @@ class TestCore < TestInteractive
     end
   end
 
+  def test_freeze_right_with_ellipsis_and_scrolling
+    tmux.send_keys "{ seq 6; ruby -e 'print \"g\"*1000, \"\\n\"'; seq 8 100; } | #{FZF} --ellipsis='777' --freeze-right 1 --scroll-off 0 --bind a:offset-up", :Enter
+    tmux.until { |lines| assert_equal '  100/100', lines[-2] }
+    tmux.send_keys(*Array.new(6) { :a })
+    tmux.until do |lines|
+      assert_match(/> 777g+$/, lines[-3])
+      assert_equal 1, lines.count { |l| l.end_with?('g') }
+    end
+  end
+
   def test_backward_eof
     tmux.send_keys "echo foo | #{FZF} --bind 'backward-eof:reload(seq 100)'", :Enter
     tmux.until { |lines| lines.item_count == 1 && lines.match_count == 1 }


junegunn:
> is a test needed ?

Preferably, yes, to avoid future regressions. I know writing an integration test can be not so straightforward, so I can add it for you if that's easier.

LangLangBart:
> Preferably, yes, to avoid future regressions.

An attempt was made... 😬

test/test_core.rb
```diff
@@ -1235,6 +1235,21 @@ def test_freeze_right_exceed_range_with_freeze_left
...
 
+  def test_freeze_right_with_ellipsis_and_scrolling
+    tmux.send_keys "{ seq 30; ruby -e 'print \"g\"*1000, \"\\n\"'; seq 32 100; } | #{FZF} --ellipsis='...' --freeze-right 1", :Enter
+    tmux.until { |lines| assert_equal '  100/100', lines[-2] }
+    tmux.send_keys(*Array.new(30) { :Up })
+    tmux.until do |lines|
+      line = lines.find { |l| l.start_with?('>') }
+      assert_match(/> \.\.\.g+$/, line)
+    end
+    tmux.send_keys :Up
```
junegunn: Thanks, but I think this doesn't necessarily cause scrolling.

How about triggering `offset-up` action?

```sh
{ seq 6; ruby -e 'print "g"*1000, "\n"'; seq 8 100; } |
  fzf --ellipsis 777 --freeze-right 1 --scroll-off 0 --bind a:offset-up
````

- Press `a` 6 times, and see if there's a line starting with `> 777ggg`
- And there's only one line that ends with `g`

LangLangBart: Adjusted the test as you described.

junegunn (approved):
LGTM!

junegunn:
Merged, thanks!

Issue #4599, opened 2025-11-19
[Feature] Add alt-gutter color option as an alternative for alt-bg


hedgieinsocks:
There is this option:

```
alt-bg                Alternate background color to create striped lines
```

But the result often looks too jarring for the eye. This being said, I would like to see the following option to be implemented and exposed:

```
alt-gutter          Alternate color for the gutter on the left to create striped pattern
```

I believe it could be a nice and elegant minimalistic way to improve readability for mixed one-line/multi-line lists without using `alt-bg` or `gap/gap-line`.

hedgieinsocks:
The idea is to achieve something subtle like this: 

<img width="682" height="448" alt="Image" src="https://github.com/user-attachments/assets/d0b1362f-c2c5-43e0-81d4-71c8c2698293" />

PR #4602, opened 2025-11-20
feat: add alt-gutter color option

hedgieinsocks:
Closes https://github.com/junegunn/fzf/issues/4599

Note: I don't quite understand that jumping mode code, so perhaps there could be a more elegant and proper way to handle `altGutter` bool assignment.

### DIFF
--- a/src/terminal.go
+++ b/src/terminal.go
@@ -3294,8 +3294,6 @@ func (t *Terminal) printItem(result Result, line int, maxLine int, index int, cu
 	extraWidth := 0
 	alt := false
 	altBg := t.theme.AltBg
-	altGutter := false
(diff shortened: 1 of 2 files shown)


junegunn:
Thanks. I was able to simplify the code in 740cd39 by making `AltGutter` inherit from `Gutter`. This also fixes the issue where non-color attributes were not applied.

```sh
fzf --gutter x --color alt-gutter:bold
```

hedgieinsocks:
Thank you for accepting the idea. I'm perfectly happy with the changes. Please feel free to shape it further if need be.

junegunn:
Merged, thanks!

Issue #4609, opened 2025-11-23
min-height not respected with auto height


georgeharker:
### Checklist

- [x] I have read through the manual page (`man fzf`)
- [x] I have searched through the existing issues
- [x] For bug reports, I have checked if the bug is reproducible in the latest version of fzf

### Output of `fzf --version`

0.67.0 (Homebrew)

### OS

- [ ] Linux
- [x] macOS
- [ ] Windows
- [ ] Etc.

### Shell

- [ ] bash
- [x] zsh
- [ ] fish

### Problem / Steps to reproduce

When using `--min-height` combined with '--height=~` the window is often too small .

Eg

```
export FZF_DEFAULT_OPTS="${FZF_DEFAULT_OPTS:-} \
  --margin=0,5%,0,5% \
  --height='~95%' \
  --min-height=16 \
  --highlight-line \
  --info=inline-right \
  --ansi \
  --border=rounded \
"
```

in a completion with 4 items only shows 2 lines.  This seems to be an interplay between the border requirements (2 lines) and the computation of height,  My understanding of `min-height` is that it supposed to dictate the min number of items to show.  It seems to have no effect with auto-computed height, and auto-computed height seems to not correctly account for the border (padding / margin?) needs.

junegunn:
This is actually the intended behavior. `--min-height` was added to address the limitation of `--height=X%` where `X%` can be too short in short windows. `--height=~X%` was added later to address the need to minimize the size of the fzf and remove unnecessary blank lines in it, and it overrules `--min-height`.

> in a completion with 4 items only shows 2 lines

```sh
seq 4 | FZF_DEFAULT_OPTS=" \
  --margin=0,5%,0,5% \
  --height='~95%' \
  --min-height=16 \
  --highlight-line \
  --info=inline-right \
  --ansi \
  --border=rounded \
" fzf
```

<img width="1920" height="330" alt="Image" src="https://github.com/user-attachments/assets/4030c896-349a-4741-913b-d74aa3382f6d" />

I see all 4 items. Do you have any other options in your `$FZF_DEFAULT_OPTS`?

georgeharker:
This turned out to be an issue with `fzf-tab` overriding height even when it was set to use the FZF_DEFAULT_OPTS flags.  My appologies!

PR #4617, opened 2025-11-25
Fix issues with --min-height and auto height combined # 4609 on fzf gh

georgeharker:
See #4609

Auto computed height did not appear to properly account for min-height.


georgeharker:
turns out to be an issue with fzf-tab

PR #4622, opened 2025-12-04
zsh foreign test

LangLangBart:
### test(zsh): add test for C-r with foreign commands

Adds a test to retrieve foreign commands for user w/ SHARE_HISTORY set
- Ref: #4595

### ci: make docker command configurable via variable

allows me to run the test with podman

```bash
DOCKER=podman make docker-test
```

### DIFF
--- a/test/test_shell_integration.rb
+++ b/test/test_shell_integration.rb
@@ -552,22 +552,26 @@ class TestZsh < TestBase
     # SHARE_HISTORY picks up foreign commands; marked with * in fc
     tmux.send_keys 'setopt SHARE_HISTORY', :Enter
     tmux.prepare
-    tmux.send_keys 'local_cmd_abc', :Enter
+    tmux.send_keys 'fzf_cmd_local', :Enter
     tmux.prepare
     # Mock foreign command (for testing only; don't edit your HISTFILE this way)
-    tmux.send_keys "echo ': 0:0;foreign_cmd_xyz' >>$HISTFILE", :Enter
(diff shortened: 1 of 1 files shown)


test/test_shell_integration.rb
```diff
@@ -540,6 +540,41 @@ def prepare_ctrl_r_test
       assert_equal ['cat <<EOF | wc -c', 'qux thud', 'EOF'], lines[-3..]
     end
   end
+
+  # NOTE: 'Perl/$history' won't see foreign cmds immediately, unlike 'awk/fc'.
+  # Perl passes only because another cmd runs between mocking and triggering C-r
+  # https://github.com/junegunn/fzf/issues/4061
+  # https://zsh.org/mla/users/2024/msg00692.html
```
LangLangBart, line 547: Just as a reminder

We can't use `Perl/fc` because, unlike bash, zsh escapes actual newlines as `\n`.

### bash: fc preserves actual newlines

```bash
printf 'foo\n
bar\n'
# foo

# bar

fc -rl -1
# 1        printf 'foo\n
# bar\n'
```

### zsh: fc escapes newlines to \n

```bash
printf 'foo\n
bar\n'
# foo

# bar

fc -rl -1
    # 1  printf 'foo\n\nbar\n'

printf '%s\t%s\n' "${(kv)history[@]}"
# 1       printf 'foo\n
# bar\n'
```

---

...and trying to patch the issue by running `fc -RI` before calling the
`$history` array caused problems the last time we tried.

PR #4605, opened 2025-11-21
Add fish completion support

lalvarezt:
Adds fuzzy completion for fish shell, triggered by `**<TAB>`, matching the functionality of bash/zsh completions.

PS: I didn't test the tmux part since I don't use it, but I tried to keep it faithful to the original code

### Configuration Examples

```fish
# Custom command lists
set -gx FZF_COMPLETION_DIR_COMMANDS "cd pushd rmdir yazi"
set -gx FZF_COMPLETION_FILE_COMMANDS "bat cat head less tail"

# Custom walkers (e.g., include hidden entries for all)
set -gx FZF_COMPLETION_DIR_WALKER "dir,follow,hidden"
set -gx FZF_COMPLETION_FILE_WALKER "file,follow,hidden"
set -gx FZF_COMPLETION_PATH_WALKER "file,dir,follow,hidden"

# Custom options with exclusions
set -l skip ".git,.cache,node_modules,target"
set -gx FZF_COMPLETION_PATH_OPTS "--walker-skip '$skip'"
set -gx FZF_COMPLETION_DIR_OPTS "--walker-skip '$skip'"
set -gx FZF_COMPLETION_FILE_OPTS "--walker-skip '$skip'"
```

### Usage

```fish
vim **<TAB>      # Complete files and directories
cd **<TAB>       # Complete directories only
cat **<TAB>      # Complete files only
kill **<TAB>     # Complete process IDs
```

PR #4630, opened 2025-12-09
shell: nushell integration scripts

### LINKED ISSUE #4122: [Feature Request] `nushell` integration
### Checklist

### Output of `fzf --version`

0.56.3 (v0.56.3)

### OS

- [ ] Linux
- [ ] macOS
- [ ] Windows
- [ ] Etc.

### Shell

- [ ] bash
- [ ] zsh
- [ ] fish

Hi, I am currently switching to `nushell` as my main shell. I noticed that `fzf` does not have a nushell integration.
I would like to propose to add a `nushell` integration, like it is done for `zsh`, `bash` and `fish`.

I really miss the <kbd>CTRL</kbd>+<kbd>t</kbd> keybind in nushell, as that was my goto way of selecting files in my terminal.

sim590:
* Keybindings `alt_c`, `ctrl_r` and `ctrl_t`
* Completion functions:
  * base implementation for `**` completion
  * extra completion functionalities for programs such as:
    * [pass](https://www.passwordstore.org/)
    * ssh, scp, sftp, telnet
    * kill
    * cd, pushd, rmdir

Closes: #4122


sim590:
@imsys: just letting you know that I disabled some of the completion cases (in 5742835) as they don't work for me and produce bad output:

image

So, I decided it was best to just comment them out. I left the implementations untouched if anybody wants to fix them at some point. I frankly was not sure how to use those either, so I didn't bother fixing them.

sim590:
I'm going to put this in draft because I just realized something else is broken.

sim590:
I just fixed a major bug in 55de1bf. Then, 4f10ad7 fixes a minor thing. And finally, I added support for completing Archlinux pacman packages.

I think this is ready. I could indeed add more programs for which to provide completion, but I'm going to stop there.

jochumdev:
You guys might be interested into that:

https://git.sr.ht/~jochumdev/gists/tree/main/item/nu/show/README.md

shell/key-bindings.nu
```diff
@@ -0,0 +1,101 @@
...
+# Dependencies: `fd`, `bat, `tree`.
+
+# Code provided by @igor-ramazanov
+# Source: https://github.com/junegunn/fzf/issues/4122#issuecomment-2607368316
+
+
+export-env {
+  $env.FZF_TMUX_OPTS       = $env.FZF_TMUX_OPTS?       | default "--height 40%"
+  $env.FZF_CTRL_T_COMMAND  = $env.FZF_CTRL_T_COMMAND?  | default "fd --type file --hidden"
+  $env.FZF_CTRL_T_OPTS     = $env.FZF_CTRL_T_OPTS?     | default "--preview 'bat --color=always --style=full --line-range=:500 {}' "
```
sim590: @jochumdev are you suggesting to integrate this as the preview program for the FZF view? Like, for the `CTRL-T` binding, here?

sim590: Actually, now that I'm thinking about it, I don't know why we'd use `bat` or any other external program as default since there's a builtin command in nushell that does a good job just like `bat`: the command `open`... I think I'm going to change `bat` to `open` here.

As for the script you linked, I'm not sure about using it as a default for the same reason as `bat`. I think that best is built-in first for compatibility sake.

Please, tell me if I'm missing the point of the suggestion.

sim590: It is true though that `open` doesn't actually give the same output. I thought for a second that it did something similar, but it doesn't (excuse my noobness).

I'm not sure anymore I want `open` as a default. I like the argument of "built-in is best", but at the same time, this is an opt-in script... And `bat` does a way better job.

So, may be we could consider your script after all!

sim590: So, based on your [comment here](https://git.sr.ht/~jochumdev/gists/tree/main/item/nu/show/show#L7), I guess that you're suggesting to merge `show.nu` as `fzf-preview.nu` may be under `bin/` directory? Anyway, I'll stop trying to figure out, and just wait on your answer. :smile:

jochumdev: Hey :)

First: I integrated your work into my nushell, you did a wonderful job.
Second: I refactored my "show" script. For now I think the best is to link it somewhere as an alternative to the bat usage, I don't know how you wanna let people install your "nu" integration, yet.

Third: 

I did some work based on yours, I added support for "sudo" , having no trigger at all and preview of pacman packages, are you interested in a follow up PR on your repo?

jochumdev: New findings :)

Use `open` with https://git.sr.ht/~jochumdev/gists/blob/main/nu/show/hook.nu

junegunn:
Hi, this is interesting. I don't use nushell and have no experience with it, so it will take me some time to get started on reviewing this.

Also, to incorporate this to the project,  we need to address a few more things:

- Documentation
- Alignment with other implementations (bash, zsh, and fish) in terms of functionality and customizability
    - If this feels too limiting, making this a separate third-party project might be a better option. e.g., https://github.com/Aloxaf/fzf-tab
- Tests
- Updates to the install and uninstall scripts

There is ongoing work to add fish fuzzy completion in https://github.com/junegunn/fzf/pull/4605. You might want to take a look at the discussion there.

jochumdev:
Hey @junegunn 

Many, many thanks for fzf, I use it a lot!

For nushell there's a plugin <https://github.com/idanarye/nu_plugin_skim> which adds deep integration into the shell (structured data in -> structured data out).

I'll switch to that for my use with nushell.

Thank you again!

shell/key-bindings.nu
sim590: >I did some work based on yours, I added support for "sudo" , having no trigger at all and preview of pacman packages, are you interested in a follow up PR on your repo?

That's awesome. I also noticed today that it didn't work correctly with `sudo`. Please, make a PR and I'll pull your commits on this one!



> New findings :)
> 
> Use `open` with https://git.sr.ht/~jochumdev/gists/blob/main/nu/show/hook.nu

The link is broken. :confused:

jochumdev: > That's awesome. I also noticed today that it didn't work correctly with sudo. Please, make a PR and I'll pull your commits on this one!

Coming.

> The link is broken. 😕

Removed hook.nu it messes with `open --raw` ,if you still want it:

https://git.sr.ht/~jochumdev/gists/tree/71d7cdad2d137b36637e54069879ca6eac2c06a2/item/nu/show/hook.nu

Please also revisit `show` it got major new features:

- Support for "--jump-to-line" 
- Integrates a variant of https://junegunn.github.io/fzf/tips/ripgrep-integration/#wrap-up into it.

sim590:
Just want to mention that this is still on my todo list. I'll follow-up on this soon.

sim590:
Hi @junegunn,

I've done a significant overhaul of the nushell integration to address all the points you raised. Here's a summary of the changes:

### Alignment with bash/zsh/fish

**Key bindings (`key-bindings.nu`):**
- CTRL-T, CTRL-R, ALT-C now use the same fzf options as bash/zsh (e.g. `--scheme=history`, `--bind=ctrl-r:toggle-sort`, `--wrap-sign`, `--highlight-line`, `+m` for CTRL-R)
- Added `__fzf_defaults` to properly merge `FZF_DEFAULT_OPTS` and `FZF_DEFAULT_OPTS_FILE`
- Added `__fzfcmd` for `FZF_TMUX` / fzf-tmux support
- Use fzf's built-in walker instead of `fd` for CTRL-T and ALT-C (matching bash/zsh default behavior)
- Key bindings can be individually disabled by setting `FZF_CTRL_T_COMMAND`, `FZF_CTRL_R_COMMAND`, or `FZF_ALT_C_COMMAND` to an empty string
- CTRL-R replaces the command line content instead of inserting (matching bash/zsh behavior)

**Completion (`completion.nu`):**
- `FZF_COMPLETION_DIR_COMMANDS` defaults to `['cd', 'pushd', 'rmdir']` (matching bash/zsh)
- Dedicated handlers for: `kill`, `ssh`/`scp`/`sftp`/`telnet`, `pacman`, `pass`
- Removed dead code for `export`/`unset`/`unalias` — these commands don't exist in nushell (their equivalents are `$env.VAR = ...`, `hide-env`, and nushell has no `unalias`), so fzf completion handlers for them are unreachable
- Extensibility is documented for users who want to add custom completions

### Installation approach

Unlike bash/zsh where the install script appends a `source` line to the shell rc file, nushell's `source` is a parser keyword — not a command — meaning the path must be a string literal or a `const`, and it cannot be used dynamically. The install script instead generates a `fzf.nu` file in the nushell `autoload/` directory (`$XDG_CONFIG_HOME/nushell/autoload/`), which nushell loads automatically at startup. This is similar to how fish integration works with its `conf.d/` directory.

### Infrastructure

- Added `--nushell` flag to fzf to emit shell integration scripts (like `--bash`, `--zsh`, `--fish`)
- Updated `install` and `uninstall` scripts with nushell support
- Added nushell to `README.md`

### Tests

- Added `TestNushell` test class with 7 passing tests: `test_ctrl_t`, `test_ctrl_t_unicode`, `test_alt_c`, `test_alt_c_command`, `test_ctrl_r`, `test_ctrl_r_abort`, and `test_ctrl_r_multiline` (omitted — nushell doesn't support multiline command recall the same way)
- All 18 bash tests still pass, unaffected by the changes

### Notable nushell-specific decisions

- CTRL-T and ALT-C pipe custom commands through `sh -c` rather than `nu -c`, because a nushell subprocess intercepts TTY input and prevents fzf keybindings from working
- History deduplication in CTRL-R uses `reverse | uniq` on plaintext history
- The `pacman` and `pass` completions are nushell-specific additions that don't exist in bash/zsh — happy to remove them if you'd prefer strict parity

### Question: git completion?

While reviewing coverage, I noticed that none of the shell implementations (bash, zsh, fish) include dedicated fzf completion for git (branches, tags, remotes, etc.). This could be a useful addition for nushell (and potentially other shells), but I don't want to exceed the expected scope of this PR. Would this be something you'd welcome as a follow-up, or is it better left to third-party tools?

junegunn:
Hi, please rebase work onto the latest master. In the meanwhile, fish got its fuzzy completion.

shell/key-bindings.nu
sim590: Hi @jochumdev, just a follow-up: I've integrated your sudo support, paru alias, and pacman preview (`pacman -Si {}`) into the current codebase in commit sim590/fzf@10787b85 with a `Co-authored-by` credit.

The codebase has been significantly reworked since December (alignment with bash/zsh, bug fixes, tests, etc.), so I cherry-picked the relevant features manually rather than merging your branch directly. Your PR on sim590/fzf has been closed with a summary.

Thanks again for the contributions!

sim590:
Done!

todo.txt
```diff
@@ -0,0 +1,8 @@
+# Quicktask v1.4
+
+TÂCHES COURANTES:
+  - pass: peut pas passer `-m`. Il faut retirer.
+    @ Added [dim 2025-12-07]
+TÂCHES COMPLÉTÉES:
+
+# vim:ft=quicktask
```
junegunn: 👋

sim590: oops. :smile:

README.md
```diff
@@ -635,6 +643,23 @@ unalias **<TAB>
...
+  built-in commands (e.g. `cd`, `ls`). Fuzzy completion with `**<TAB>` only
+  works for external commands.
+- Custom completion extensibility (e.g. `_fzf_complete_COMMAND` in bash/zsh)
+  is not available. Custom completions are defined via a `match` statement
+  in `completion.nu`.
+- The following environment variables are supported:
+  `FZF_COMPLETION_TRIGGER`, `FZF_COMPLETION_OPTS`,
+  `FZF_COMPLETION_PATH_OPTS`, `FZF_COMPLETION_DIR_OPTS`,
+  `FZF_COMPLETION_DIR_COMMANDS`.
+
```
junegunn: It might be better to move this to a dedicated section, similar to 'Fuzzy completion for fish'.

sim590: Done! :smiley_cat:

shell/completion.nu
```diff
@@ -0,0 +1,506 @@
...
+#  / __/ / /_/ __/
+# /_/   /___/_/ completion.nu
+
+
+# An implementation of completion.nu
+# This loads FZF as an Nushell External Completer
+# https://www.nushell.sh/cookbook/external_completers.html
+
+# It's the most stable implementation.
+# The drawback is that it does't work for completing some commands, like 'cd' and 'ls' on Nushell >= 0.103.0
```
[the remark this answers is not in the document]
sim590: Fixed in 8d405a7c.

shell/key-bindings.nu
```diff
@@ -0,0 +1,158 @@
...
+          let fzf_args = ($fzfcmd | skip 1);
+          let alt_c_cmd = ($env.FZF_ALT_C_COMMAND? | default null);
+          let result = if ($alt_c_cmd == null) or ($alt_c_cmd | is-empty) {
+            with-env { FZF_DEFAULT_OPTS: $fzf_opts, FZF_DEFAULT_OPTS_FILE: '' } { ^($fzfcmd | first) ...$fzf_args }
+          } else {
+            let fzf_cmd_str = ($fzfcmd | str join ' ');
+            let sh_cmd = [$alt_c_cmd '|' $fzf_cmd_str] | str join ' ';
+            with-env { FZF_DEFAULT_OPTS: $fzf_opts, FZF_DEFAULT_OPTS_FILE: '' } { ^sh -c $sh_cmd }
+          };
+          cd $result;
```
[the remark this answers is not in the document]
sim590: Fixed in 8d405a7c — added `if ($result | is-not-empty) { cd $result }`.

install
```diff
@@ -251,7 +256,7 @@ for shell in $shells; do
   fzf_completion="source \"$fzf_base/shell/completion.${shell}\""
   fzf_key_bindings="source \"$fzf_base/shell/key-bindings.${shell}\""
   [[ $shell == fish ]] && continue
-  src=${prefix_expand}.${shell}
+  [[ $shell == nushell ]] && continue
```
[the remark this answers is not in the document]
sim590: Good catch — this was a real bug introduced during merge conflict resolution. The `src=${prefix_expand}.${shell}` assignment was accidentally dropped. Fixed in 17b194e1.

test/test_shell_integration.rb
```diff
@@ -1103,3 +1103,130 @@ def test_ctrl_r_multi
     end
   end
 end
+
+class TestNushell < TestBase
+  include TestShell
+
+  def shell
+    :nushell
+  end
```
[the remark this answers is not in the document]
sim590: Fixed — added `test_file_completion` in 8963f385. It exercises single-selection `**<TAB>` completion since Nushell's completer replaces the current token rather than appending (multi-selection works differently than bash/zsh).

test/lib/common.rb
```diff
@@ -78,19 +78,63 @@ def fish
...
+      send_keys 'function fish_prompt; end; clear', :Enter
+      self.until(&:empty?)
+    elsif shell == :nushell
+      # Clear history from previous tests to avoid contamination
+      FileUtils.rm_f('/tmp/fzf-nushell-xdg/nushell/history.txt')
+      # Wait for nushell to be ready: send a marker and wait for it
+      sleep 2
+      send_keys 'clear', :Enter
+      self.until(&:empty?)
+      send_keys '"ready"', :Enter
```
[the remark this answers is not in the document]
sim590: This is actually correct as-is. With plaintext history format (configured in `env.nu`), Nushell stores `history.txt` under `XDG_CONFIG_HOME/nushell/`, not under `XDG_DATA_HOME`. The deletion path `/tmp/fzf-nushell-xdg/nushell/history.txt` matches the actual file location, so tests are deterministic.

shell/completion.nu
```diff
@@ -0,0 +1,506 @@
...
+  }
+
+  let walker_type = if ($suffix == '/') {
+      "dir,follow"
+  } else {
+      "file,dir,follow,hidden"
+  }
+  # Use the 'walker_root' calculated at the beginning
+  let $fzf_all_opts = ["--scheme=path", "--walker", $walker_type, "--walker-root", $walker_root] | append $fzf_opts_arg
+                                                                                                 | append $completion_type_opts
```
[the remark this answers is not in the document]
sim590: Fixed in 17b194e1 — changed to `let fzf_all_opts`. Nushell does accept the `let $var` form, but the idiomatic syntax omits the `$`.

shell/completion.nu
```diff
@@ -0,0 +1,506 @@
...
+
+# Specific path completion wrapper
+def _fzf_path_completion_nu [prefix: string] {
+  # Zsh args: base, lbuf, _fzf_compgen_path, "-m", "", " "
+  # Nu: prefix, empty command name (use find), ["-m"], "", " "
+  __fzf_generic_path_completion_nu $prefix "" ["-m"] ""
+}
+
+# General completion helper for commands that feed a list to fzf
+# This is called by ssh, export, unalias, kill. everything exept path and dir
```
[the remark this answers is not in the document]
sim590: Fixed in 8d405a7c.

shell/completion.nu
```diff
@@ -0,0 +1,506 @@
...
+# Options for fzf completion in general. e.g. '--border'
+$env.FZF_COMPLETION_OPTS = $env.FZF_COMPLETION_OPTS? | default ''
+
+# Options specific to path completion. e.g. '--extended'
+$env.FZF_COMPLETION_PATH_OPTS = $env.FZF_COMPLETION_PATH_OPTS? | default ''
+# Options specific to directory completion. e.g. '--extended'
+$env.FZF_COMPLETION_DIR_OPTS = $env.FZF_COMPLETION_DIR_OPTS? | default ''
+
+$env.FZF_COMPLETION_DIR_COMMANDS = $env.FZF_COMPLETION_DIR_COMMANDS? | default ['cd', 'pushd', 'rmdir']
+$env.FZF_COMPLETION_VAR_COMMANDS = $env.FZF_COMPLETION_VAR_COMMANDS? | default ['export', 'unset', 'printenv']
```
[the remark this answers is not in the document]
sim590: Fixed in 8d405a7c — removed the orphaned variable.

sim590:
I added nushell installation to the CI workflow (`.github/workflows/linux.yml`) using the official Gemfury apt repository (as documented on [nushell.sh](https://www.nushell.sh/book/installation.html#package-managers)), so that the nushell integration tests can run in the pipeline.

I also fixed the typos spell checker false positive on `pacman -Slq` and a couple of rubocop warnings in the test file.

README.md
```diff
@@ -210,9 +211,17 @@ Add the following line to your shell configuration file.
   # Set up fzf key bindings
   fzf --fish | source
   ```
+* Nushell — Nushell does not support piping into `source`, so the install
+  script generates a file in the autoload directory. If you didn't use the
+  install script, you can manually set it up:
+  ```nu
+  # Generate the integration script
+  # (run this in your regular shell, not in Nushell)
```
junegunn: Care to explain what this means? What are the regular shells, anything but Nushell? Can they interpret this command?

sim590: You're right, that was poorly worded. I've rewritten it to use pure Nushell syntax:

```nu
fzf --nushell | save -f ($nu.default-config-dir | path join "autoload" "fzf.nu")
```

This can be run directly in Nushell — no need for another shell.

shell/completion.nu
```diff
@@ -0,0 +1,505 @@
...
+  let passwordstore_files_gen_closure = {||
+    ls ~/.password-store/**/*.gpg | get name | each {$in | str replace -r '^.*?\.password-store/(.*).gpg' '${1}' }
+  }
+  _fzf_complete_nu $prefix $passwordstore_files_gen_closure ["+m"]
+}
+
+# Kill completion post-processor (extracts PID)
+def _fzf_complete_kill_post_get_pid [selected_line: string] {
+  # Assuming standard ps output where PID is the second column
+  $selected_line | lines | each { $in | from ssv --noheaders | get 0.column1 } | to text
```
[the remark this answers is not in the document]
sim590: This is a false positive. `from ssv --noheaders` assigns generic column names: `column0` = USER, `column1` = PID, `column2` = PPID, etc. So `get 0.column1` correctly extracts the PID. The comment on the line above documents this.

README.md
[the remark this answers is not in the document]
sim590: Already fixed in 691dfa3e — the instructions now use pure Nushell syntax with `save -f`.

README.md
```diff
@@ -210,9 +211,17 @@ Add the following line to your shell configuration file.
...
+  fzf --nushell > ($nu.default-config-dir | path join "autoload" "fzf.nu")
+  ```
 
 > [!NOTE]
-> `--bash`, `--zsh`, and `--fish` options are only available in fzf 0.48.0 or
+> `--bash`, `--zsh`, `--fish`, and `--nushell` options are only available in fzf 0.48.0 or
```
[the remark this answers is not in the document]
sim590: Fixed in 305a6508 — I removed `--nushell` from this note since it only applies to 0.48.0 which introduced `--bash`, `--zsh`, and `--fish`. The note should be updated to include `--nushell` once the version number for this release is known.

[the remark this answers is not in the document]
junegunn: We could just write `only available in recent versions of fzf`.

README.md
```diff
@@ -210,6 +211,13 @@ Add the following line to your shell configuration file.
   # Set up fzf key bindings
   fzf --fish | source
   ```
+* Nushell — Nushell does not support piping into `source`, so the install
+  script generates a file in the autoload directory. If you didn't use the
+  install script, you can manually set it up:
+  ```nu
+  # Generate the integration script
+  fzf --nushell | save -f ($nu.default-config-dir | path join "autoload" "fzf.nu")
```
junegunn: First time trying nushell, and this gives me an error:

```
~/github/fzf> fzf --nushell | save -f ($nu.default-config-dir | path join "autoload" "fzf.nu")
Error: nu::shell::io::directory_not_found

  × Directory not found
   ╭─[repl_entry #2:1:25]
 1 │ fzf --nushell | save -f ($nu.default-config-dir | path join "autoload" "fzf.nu")
   ·                         ────────────────────────────┬───────────────────────────
   ·                                                     ╰── Directory not found
   ╰────
  help: '/Users/***/Library/Application Support/nushell/autoload' does not exist
```

Can we update the instruction to handle this case as well?

sim590: It's possible that the `autoload` subdirectory doesn't exist. We could add a line that creates it if it's not there.

README.md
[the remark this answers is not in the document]
sim590: Good idea — updated in adbd8e70.

README.md
sim590: Fixed in adbd8e70 by adding a `mkdir` step before `save`:

```nu
mkdir ($nu.default-config-dir | path join "autoload")
fzf --nushell | save -f ($nu.default-config-dir | path join "autoload" "fzf.nu")
```

Nushell's `mkdir` is idempotent (no error if the directory already exists).

test/test_shell_integration.rb
```diff
@@ -1103,3 +1103,148 @@ def test_ctrl_r_multi
...
+
+    tmux.prepare
+    tmux.send_keys 'C-t'
+    tmux.until { |lines| assert_equal 100, lines.match_count }
+    tmux.send_keys :Tab, :Tab, :Tab
+    tmux.until { |lines| assert lines.any_include?(' (3)') }
+    tmux.send_keys :Enter
+    tmux.until { |lines| assert lines.any_include?('1 2 3') }
+    tmux.send_keys 'C-c'
+  end
```
junegunn: This looks identical to the one in `TestShell`.

test/test_shell_integration.rb
```diff
@@ -1103,3 +1103,148 @@ def test_ctrl_r_multi
...
+
+    tmux.prepare
+    tmux.send_keys :Escape, :c
+    tmux.until { |lines| assert_equal 1, lines.match_count }
+    tmux.send_keys :Enter
+
+    tmux.prepare
+    tmux.send_keys 'pwd', :Enter
+    tmux.until { |lines| assert_equal '/tmp', lines[-1] }
+  end
```
junegunn: Same here. Except for `:pwd` and `'pwd'` which should make no difference.

Please review these test cases if redefinitions are necessary.

junegunn (changes requested):

junegunn:
Let me just randomly post issues I noticed in the comments. 

```
~/github/fzf> pass **Error executing data_gen closure. Closure code: closure_635. Actual error: {msg: No matches found for Expand("~/.password-store/**/*.gpg"), debug: Generic(GenericError { code: "nu::shell::error", error: "No matches found for Expand(\"~/.password-store/**/*.gpg\")", msg: "Pattern, file or folder not found", site: Span(Span[165878..165904]), help: Some("no matches found"), inner: [], source: None }), raw: Generic(GenericError { code: "nu::shell::error", error: "No matches found for Expand(\"~/.password-store/**/*.gpg\")", msg: "Pattern, file or folder not found", site: Span(Span[165878..165904]), help: Some("no matches found"), inner: [], source: None }), rendered: Error: nu::shell::error

                                              × No matches found for Expand("~/.password-store/**/*.gpg")
                                                                                                              ╭─[/Users/***/Library/Application Support/nushell/autoload/fzf.nu:528:8]
                                                                              527 │   let passwordstore_files_gen_closure = {||
                528 │     ls ~/.password-store/**/*.gpg | get name | each {$in | str replace -r '^.*?\.password-store/(.*).gpg' '${1}' }
                             ·        ─────────────┬────────────
                                                                     ·                     ╰── Pattern, file or folder not found
                 529 │   }
                               ╰────
                                      help: no matches found
                                                            , json: {"msg":"No matches found for Expand(\"~/.password-store/**/*.gpg\")","labels":[{"text":"Pattern, file or folder not found","span":{"start":165878,"end":165904}}],"code":"nu::shell::error","url":null,"help":"no matches found","inner":[]}}
```

junegunn:
Fuzzy completion does not respect `--tmux` option in my `$FZF_DEFAULT_OPTS`. CTRL-T does.

```sh
export FZF_DEFAULT_OPTS='--tmux 90%,70%'
```

README.md
```diff
@@ -210,10 +211,18 @@ Add the following line to your shell configuration file.
   # Set up fzf key bindings
   fzf --fish | source
   ```
+* Nushell — Nushell does not support piping into `source`, so the install
```
junegunn:
```suggestion
* Nushell -- Nushell does not support piping into `source`, so the install
```

junegunn:
Multi-selection (tab) in fuzzy completion doesn't work properly.

```
~/github/fzf> vim .editorconfig
::: .gitignore
::: .goreleaser.yml
```

It works for `^kill`.

sim590:
Thanks for testing! Here's a summary of the fixes:

- **`pass` crash**: Fixed in 8057be90 — added `try/catch` so `pass **<TAB>` no longer errors when `~/.password-store` doesn't exist.
- **Multi-selection**: Fixed in e127c4ef — selected paths are now joined with `lines | str join ' '` instead of being returned as separate entries.
- **`--tmux` in `FZF_DEFAULT_OPTS`**: Fixed in 0025d228 — `__fzf_defaults_nu` now builds `FZF_DEFAULT_OPTS` as a string (like bash's `__fzf_defaults`), so `--tmux` takes precedence over `--height`. The root cause was twofold: (1) `is-string` doesn't exist in Nushell, so a filter was silently dropping all `FZF_DEFAULT_OPTS` entries, and (2) `--height` was passed as a CLI argument (higher index) instead of through `FZF_DEFAULT_OPTS` (where `--tmux` could override it).

README.md
sim590: Applied, thanks!

test/test_shell_integration.rb
sim590: You're right, removed. `TestNushell` includes `TestShell`, so these are inherited automatically.

test/test_shell_integration.rb
sim590: Same — removed the redundant redefinition.

README.md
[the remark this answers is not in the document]
sim590: Applied in a previous commit, thanks!

README.md
sim590: Already handled — `mkdir` was added before `save -f` in a previous commit.

test/test_shell_integration.rb
junegunn: Please review the other tests as well to see if we can minimize code duplication.

sim590: Done — also removed `test_ctrl_r` (identical to `TestShell`). The remaining overrides now have comments explaining why they differ:

- `test_ctrl_t_unicode`: uses `^echo` (external command) since Nushell's builtin `echo` outputs structured data
- `test_file_completion`: Nushell-specific, tests single-selection only (the external completer replaces the token rather than appending)
- `test_ctrl_r_abort`: only tests with `foo` — single/double quotes cause issues in Nushell's line editor

test/test_shell_integration.rb
```diff
@@ -1175,37 +1179,11 @@ def test_file_completion
...
-    tmux.send_keys :Enter
-    tmux.until { |lines| assert_equal 'echo 3rd', lines[-1] }
-    tmux.send_keys :Enter
-    tmux.until { |lines| assert_equal '3rd', lines[-1] }
-  end
-
   # Nushell does not support multiline command recall the same way
   # as bash/zsh/fish, so test_ctrl_r_multiline is omitted.
 
+  # Override: only test with 'foo' — single and double quotes cause
```
junegunn: Coding agents love em-dashes, but I don't want to see non-ascii chars in the source code.

test/test_shell_integration.rb
junegunn: > * Nushell-specific, tests single-selection only (the external completer replaces the token rather than appending)

Can you elaborate? We missed the multi-selection bug because of the skipping, so I'd like to keep the test if possible.

junegunn:
Coding agents can miss subtle issues that only surface in hands-on use, which matters a lot for an interactive tool like fzf. Could you share a list of what you manually tested, ideally with screenshots?

sim590:
> Coding agents can miss subtle issues that only surface in hands-on use, which matters a lot for an interactive tool like fzf.

Yes. I agree.

> Could you share a list of what you manually tested, ideally with screenshots?

I tested:

* all keybindings
* `vim **<TAB>`
  * with no prefix.
  * with absolute path prefix
  * with relative path prefix
  * with `~`
  * with `$env.HOME/**`, but that doesn't work. I think that this is a limitation of Nushell's completion dipatcher
* `^cd **<TAB>`
* `^kill **<TAB>`
* `pass **<TAB>`
* `pacman -S **<TAB>`
* `pacman -Q **<TAB>`

All of these work as expected. Here are some screenshots...

image

image

image

image

image

image

vim and `ctrl+t`:

image

image

test/test_shell_integration.rb
sim590: Fixed in 6f1ed5a.

test/test_shell_integration.rb
sim590: Good point. The test now covers multi-selection (Tab+Tab with 2 files), single selection, and hidden files. The previous override only tested single selection, which is how the multi-selection bug slipped through. Fixed in 4ec77b94.

README.md
```diff
@@ -823,6 +834,23 @@ function _fzf_post_complete_foo
...
+Fuzzy completion in Nushell works via the
+[external completer](https://www.nushell.sh/cookbook/external_completers.html)
+mechanism. There are some differences compared to bash and zsh:
+
+- On Nushell >= 0.103.0, the external completer is no longer called for
+  built-in commands (e.g. `cd`, `ls`). Fuzzy completion with `**<TAB>` only
+  works for external commands.
+- Custom completion extensibility (e.g. `_fzf_complete_COMMAND` in bash/zsh)
+  is not available. Custom completions are defined via a `match` statement
+  in `completion.nu`.
```
junegunn: Would it be possible to implement this scheme? A hard-coded match statement feels quite limiting. It could easily turn into a "kitchen sink", with users continually requesting more commands to be added.

For what it's worth, I'm on macOS and don't use tools like pacman or pass, so having them in the default list feels a bit off. I imagine this concern will only grow as more items get added over time.

sim590: You're right, the hard-coded `match` is not ideal. I investigated a dynamic dispatch approach that works within Nushell's constraints.

Users could define an `$env.FZF_COMPLETERS` record in their `config.nu` with closures keyed by command name. Each closure receives the prefix and returns a list of candidates, which are then fed to fzf:

```nu
# In config.nu
$env.FZF_COMPLETERS = {
    git: {|prefix| ["branch-main", "branch-dev", "branch-feature"]}
    pacman: {|prefix| ^pacman -Slq | lines}
}
```

In `completion.nu`, we'd check `$env.FZF_COMPLETERS` before falling through to the built-in defaults:

```nu
let user_completers = ($env.FZF_COMPLETERS? | default {})
if ($cmd_word in $user_completers) {
    # call _fzf_complete_nu with the user's closure
} else {
    match $cmd_word {
        "ssh" | ... => { ... }
        "kill"      => { ... }
        ...
    }
}
```

I tested this end-to-end: defining a completer in `config.nu`, triggering `git **<TAB>`, selecting from fzf, and confirming the result is inserted on the command line. It works.

With this in place, we could remove `pacman`, `paru`, and `pass` from the default `match` and keep only the universal ones (`ssh`, `kill`). The question is where to provide the removed implementations as examples for users -- perhaps in the README or in a separate example file. What do you think?

**Limitation:** The closures must be defined in `config.nu`, not in a separate autoload file, because Nushell doesn't propagate `$env` mutations from autoload files to later-loaded files. This is consistent with how users configure other Nushell settings.

junegunn: > With this in place, we could remove `pacman`, `paru`, and `pass` from the default `match` and keep only the universal ones (`ssh`, `kill`). The question is where to provide the removed implementations as examples for users -- perhaps in the README or in a separate example file. What do you think?

Sounds good to me. How about putting them under `shell/completion-examples.nu` (any better name suggestion?) and add a link to the file in the README file?

sim590:
I just fixed a corner case. Expanding `~/**` was bringing FZF up, but it was not substituting the result on the command line. It's fixed.

README.md
sim590: Done. I implemented the `$env.FZF_COMPLETERS` mechanism and moved `pacman`, `paru`, and `pass` to `shell/completion-examples.nu`.

Each closure receives `(prefix, spans)` and can return either a simple list of candidates, or a record `{ candidates: [...], opts: [...] }` for custom fzf options (e.g. `--preview`). The README now documents the approach and links to the examples file.

junegunn: The "Combine example" feels redundant. We already have a single-record example for pacman and paru and I think it should be obvious to readers. Maybe we can restructure the code like so:

```nu
$env.FZF_COMPLETERS = {}

# Example 1
$env.FZF_COMPLETERS.pacman = {

}
```

(Does this make sense? I'm not a nushell user)

---

This file will not be sourced by default. How about uncommenting the code so that it's easier for the users to read (with syntax highlighting) and copy when they want it?

---

Does this support the "post" hook like the other shells? For reference, zsh's kill completion dog-foods its "post" mechanism via `_fzf_complete_kill_post`.

shell/completion.nu
```diff
@@ -0,0 +1,498 @@
...
+    # Each closure receives the prefix (text before the trigger) and the full
+    # command spans (e.g. ["pacman", "-S", "vim**"]), and should return either:
+    #   - a list of candidate strings, or
+    #   - a record { candidates: [...], opts: [...] } to pass custom fzf options.
+    # See shell/completion-examples.nu for examples.
+    let user_completers = ($env.FZF_COMPLETERS? | default {})
+    if ($cmd_word in $user_completers) {
+      let user_gen = ($user_completers | get $cmd_word)
+      let user_result = (do $user_gen $prefix $cmd_spans)
+      "\n" | save -a /tmp/fzf-debug.log
```
junegunn: Please remove this debug log.

shell/key-bindings.nu
```diff
@@ -0,0 +1,158 @@
...
+# When not defined (null), the binding is enabled (using fzf's built-in walker).
+def __fzf_binding_enabled [var_name: string]: nothing -> bool {
+  let val = ($env | get -o $var_name)
+  # null = not defined = enabled; "" = explicitly disabled
+  $val == null or ($val | into string | is-not-empty)
+}
+
+# Update the $env.config
+export-env {
+  let already_loaded = ($env.config.keybindings | any { |kb| $kb.name == 'fzf_files' })
```
junegunn: Is checking only `fzf_files` correct? A user could source the file with `FZF_CTRL_T_COMMAND=""` and re-source without it.

shell/completion.nu
```diff
@@ -0,0 +1,498 @@
...
+# This ensures external completions are enabled and sets our wrapper.
+$env.config = $env.config | upsert completions {
+  external: {
+    enable: true
+    completer: $fzf_wrapper_completer
+  }
+}
+
+#  vim: set sts=2 ts=2 sw=2 tw=120 et :
+
```
junegunn: Unnecessary empty line.

shell/key-bindings.nu
```diff
@@ -0,0 +1,158 @@
...
+    {
+      send: executehostcommand
+      cmd: "commandline edit --replace (
+        let fzf_opts = (__fzf_defaults '' $'--scheme=history --bind=ctrl-r:toggle-sort --wrap-sign \"\t↳ \" --highlight-line ($env.FZF_CTRL_R_OPTS) +m --read0');
+        let fzfcmd = (__fzfcmd);
+        let fzf_args = ($fzfcmd | skip 1);
+        history
+          | get command
+          | reverse
+          | uniq
```
junegunn, line 107: Will this block until the full list is ready? Could that become a problem with very large histories? (i.e. no streaming)

junegunn: It seems so:

```
seq 1 100000000 | reverse | uniq 
```

junegunn (changes requested):

shell/key-bindings.nu
sim590: Fixed — now checks all three binding names (`fzf_files`, `fzf_dirs`, `history_menu`) instead of just `fzf_files`.

shell/key-bindings.nu
sim590: Nushell's `history` command works differently from bash's `fc` — it loads the entire history as an in-memory table, bounded by `$env.config.history.max_size` (default: 100,000). So `reverse` and `uniq` operate on an already-materialized list and are effectively free compared to the initial load.

This means `reverse | uniq` can't block in the way `seq 1 100000000 | reverse | uniq` does — the data is already fully materialized by `history` before `reverse` even runs. Even with an aggressive `max_size`, no one realistically has 100M history entries.

The alternative would be to skip `reverse` and pass `--tac` to fzf, but nushell's `uniq` keeps the *first* occurrence (oldest), so deduplication would show commands at their oldest position rather than most recent — wrong behavior.

I've added a comment in the code explaining why `reverse | uniq` is safe here.

README.md
sim590: Done — restructured with per-key assignment and removed the combined example. See 4d81f66d.

shell/completion.nu
sim590: Removed in 450f0b5.

shell/completion.nu
sim590: Removed in ba6dea59.

sim590: Correction: ba6dea59 removed a different empty line (between the dispatch block and the return section). The trailing empty line after the vim modeline (line 498 in the original diff) is now removed in 48372860.

shell/key-bindings.nu
junegunn: > So `reverse` and `uniq` operate on an already-materialized list and are effectively free compared to the initial load.

You mean `history | reverse | head -1` takes the same amount of time as `history | head -1` even when the list is huge?

shell/key-bindings.nu
junegunn: Question: what happens when we blindly re-run the code here? In bash and zsh, it will just overwrite the existing bindings, so an existence check is not necessary.

shell/key-bindings.nu
sim590: You're right to push back on "effectively free." I ran some tests:

| Command | Time |
|---|---|
| `seq 1 100000000 \| lines \| last 1` | ~6s |
| `seq 1 100000000 \| lines \| reverse \| first 1` | ~10s |
| `seq 1 100000000 \| reverse \| uniq` | >40s (interrupted) |

So `reverse | uniq` is indeed expensive on large datasets. However, this comparison uses 100M items from an unbounded stream. Nushell's `history` is bounded by `$env.config.history.max_size` (default: 100,000). At that scale, `reverse | uniq` should be instant — could you try `seq 1 100000 | reverse | uniq | length` to confirm?

The challenge is that nushell's `uniq` only keeps the *first* occurrence, so without `reverse`, duplicates would appear at their oldest position (wrong behavior for history). There's no `--last` flag on `uniq`, and no streaming deduplication primitive in nushell.

If you'd prefer, I can update the in-code comment in `key-bindings.nu` to be more honest about the tradeoff — something like: "This is O(n) but acceptable for typical history sizes (bounded by max_size, default 100K). For very large histories, this may add noticeable latency."

junegunn: I wasn't asking about `seq 1 1000000...`, but about `history`. You mentioned that it's somehow different from bash and zsh, so I wanted to clarify if nushell has a mechanism to reverse the list without any overhead.

So can I conclude that nushell is no different from bash or zsh in that reverse has to consume the entire list before it can output the first element?

junegunn: In bash and zsh implementations, we use `fc` with the `-r` flag to generate the list in reverse order, so fzf can see the latest commands immediately no matter how large the entire list is. No initial delay at all for the user.

So from what I gather, nushell has no such option, but the performance shouldn't be an issue in realistic scenarios?

README.md
junegunn: Thanks. Any thoughts on the 'post' hook part?

shell/key-bindings.nu
sim590: In nushell, `$env.config.keybindings` is a plain list of records, and `append` simply adds to the end without checking for duplicates. For example:

```nu
$env.config.keybindings | append {name: "fzf_files", ...} | append {name: "fzf_files", ...}
```

This would create two entries with the same name. Nushell doesn't deduplicate by name automatically, unlike bash/zsh where `bind`/`bindkey` overwrites the existing binding for the same key. That's why the existence check is needed.

shell/key-bindings.nu
sim590: Yes, that's correct. Nushell's `history` command has no `-r` flag — it always returns entries in chronological order. `reverse` has to consume the full list. But with the default `max_size` of 100,000, this should be imperceptible in practice.

README.md
sim590: Sorry for the lack of clarity — the design evolved as we went. The `post` hook is actually already implemented. The `$env.FZF_COMPLETERS` closure can now return a record with a `post` key containing a closure that processes the fzf selection before inserting it on the command line:

```nu
$env.FZF_COMPLETERS.mycommand = {|prefix, spans|
    {
        candidates: ["Alice 100", "Bob 200"]
        opts: ["+m"]
        post: {|sel| $sel | split row ' ' | last}
    }
}
```

This mirrors the `_fzf_complete_*_post` pattern from bash/zsh. The built-in `kill` completer already uses this mechanism to extract PIDs from `ps` output. See commit 673f979b.

shell/key-bindings.nu
sim590: Actually, there's a better approach. Instead of checking for existing bindings and skipping, we can filter out any previous fzf bindings and re-add the enabled ones:

```nu
export-env {
  let fzf_names = ['fzf_files', 'fzf_dirs', 'history_menu']
  # Remove any existing fzf bindings, then add the enabled ones
  mut bindings = ($env.config.keybindings | where { |kb| $kb.name not-in $fzf_names })
  if (__fzf_binding_enabled 'FZF_ALT_C_COMMAND') { $bindings = ($bindings | append $alt_c) }
  if (__fzf_binding_enabled 'FZF_CTRL_R_COMMAND') { $bindings = ($bindings | append $ctrl_r) }
  if (__fzf_binding_enabled 'FZF_CTRL_T_COMMAND') { $bindings = ($bindings | append $ctrl_t) }
  $env.config.keybindings = $bindings
}
```

This way, re-sourcing always replaces the fzf bindings instead of skipping — closer to how bash/zsh `bind`/`bindkey` behaves (overwrite). No duplicates possible, and if the user changes `FZF_CTRL_T_COMMAND` between sources, the bindings update correctly.

Would you prefer this approach?

junegunn: I like this better. What's your call? By the way, shouldn't `history_menu` have the same `fzf_` prefix? Is there a reason you chose that name?

shell/key-bindings.nu
junegunn: Okay, let's see if someone complains.

README.md
junegunn: Got it.

shell/completion.nu
```diff
@@ -0,0 +1,501 @@
...
+# This function is registered with Nushell's external completion system.
+# It gets called when Tab is pressed.
+let fzf_external_completer = {|spans|
+  let trigger: string = $env.FZF_COMPLETION_TRIGGER? | default '**'
+
+  if ($trigger | is-empty)     { return null } # Cannot work with empty trigger
+  if (($spans | length ) == 0) { return null } # Nothing to complete
+
+  let last_span = $spans | last
+  let line_before_cursor = $spans | str join ' ' # Reconstruct line for context
```
junegunn: Unused?

shell/completion.nu
```diff
@@ -0,0 +1,501 @@
...
+  ]
+  | flatten # Combine all lists into a single stream
+  | where {|it| not ($it | is-empty) } # Remove empty entries
+  | sort | uniq # Sort and remove duplicates
+}
+
+
+# Base function for path/directory completion
+def __fzf_generic_path_completion_nu [ prefix:           string       # The text before the trigger
+                                     , compgen_cmd_name: string       # not used
```
junegunn: Any reason to keep this unused argument?

shell/completion.nu
```diff
@@ -0,0 +1,501 @@
...
+                                     , fzf_opts_arg:     list<string> # Extra options for fzf
+                                     , suffix:           string       # Suffix to add to selection (e.g. , "/")
+                                     ] {
+  # --- Determine walker root and initial query from the raw prefix ---
+  let raw_prefix = $prefix # Use the original prefix before any expansion
```
junegunn: Is this alias necessary? I don't see `prefix` or `raw_prefix` being modified somewhere.

shell/completion.nu
```diff
@@ -0,0 +1,501 @@
...
+#   / /_/_  / / /_
+#  / __/ / /_/ __/
+# /_/   /___/_/ completion.nu
+
+
+# An implementation of completion.nu
+# This loads FZF as a Nushell External Completer
+# https://www.nushell.sh/cookbook/external_completers.html
+
+# It's the most stable implementation.
```
junegunn: I don't think we need this line.

shell/completion.nu
```diff
@@ -0,0 +1,501 @@
...
+  #   3. $FZF_DEFAULT_OPTS, $append
+  $"--height ($height_opt) --min-height 20+ --bind=ctrl-z:ignore ($prepend) ($file_opts) ($default_opts) ($append)" | str trim
+}
+
+# Wrapper for running fzf or fzf-tmux
+def __fzf_comprun_nu [ context_name: string       # e.g., "fzf-completion" , "fzf-helper" - mainly for potential debugging
+                     , query:        string       # The initial query string for fzf
+                     , fzf_opts_arg: list<string> # Remaining options for fzf/fzf-tmux
+                     ] {
+  # in which case $stdin_content will be null.
```
junegunn: Not sure what this tries to deliver.

shell/completion.nu
```diff
@@ -0,0 +1,501 @@
...
+
+# Specific path completion wrapper
+def _fzf_path_completion_nu [prefix: string] {
+  # Zsh args: base, lbuf, _fzf_compgen_path, "-m", "", " "
+  # Nu: prefix, empty command name (use find), ["-m"], "", " "
+  __fzf_generic_path_completion_nu $prefix "" ["-m"] ""
+}
+
+# General completion helper for commands that feed a list to fzf
+# This is called by ssh, export, unalias, kill. everything except path and dir
```
junegunn: Seems outdated.

shell/key-bindings.nu
sim590: Done — renamed `history_menu` to `fzf_history` for consistency, and implemented the filter-and-append approach. Re-sourcing now replaces fzf bindings instead of skipping.

shell/completion.nu
sim590: Removed.

shell/completion.nu
sim590: Removed.

shell/completion.nu
sim590: Removed — it was a leftover from an earlier design that mirrored the bash/zsh `_fzf_compgen_path` argument.

shell/completion.nu
sim590: Removed — `$prefix` is used directly now.

shell/completion.nu
sim590: Removed.

shell/completion.nu
sim590: Updated to "ssh, kill, and user-defined completers."

shell/completion.nu
```diff
@@ -0,0 +1,489 @@
...
+# --- Helper Functions ---
+
+# Helper to build default fzf options list
+def __fzf_defaults_nu [prepend: string, append: string] {
+  let default_opts      = $env.FZF_DEFAULT_OPTS? | default ''
+  let default_opts_file = $env.FZF_DEFAULT_OPTS_FILE? | default ''
+  let height_opt        = $env.FZF_TMUX_HEIGHT? | default '40%'
+
+  let file_opts = try {
+     open $default_opts_file | str trim
```
junegunn: Should we use `--raw` here? I noticed it's a bit different from the equivalent in key-bindings.nu.

shell/completion.nu
```diff
@@ -0,0 +1,489 @@
...
+
+  # Determine if fzf should generate its own candidates via walker
+  let has_walker = ($fzf_prefinal_opt | find '--walker' | is-not-empty)
+
+  # Check for custom comprun function (Nu equivalent)
+  if (which _fzf_comprun_nu | is-not-empty) {
+    # Note: Nushell doesn't have a direct equivalent to Zsh/Bash `type -t _fzf_comprun`.
+    # This check assumes a user might define a custom command named `_fzf_comprun_nu`.
+    _fzf_comprun_nu $context_name $query ...$fzf_prefinal_opt # Pass args correctly to custom function
+  } else if ($env.TMUX_PANE? | is-not-empty) and (($env.FZF_TMUX? | default 0) != 0 or ($env.FZF_TMUX_OPTS? | is-not-empty)) {
```
junegunn: In key-bindings.nu:

```nushell
let in_tmux = ($env.TMUX_PANE? | default '' | into string | is-not-empty)
```

Which one is correct?

shell/completion.nu
```diff
@@ -0,0 +1,489 @@
...
+$env.FZF_COMPLETION_PATH_OPTS = $env.FZF_COMPLETION_PATH_OPTS? | default ''
+# Options specific to directory completion. e.g. '--extended'
+$env.FZF_COMPLETION_DIR_OPTS = $env.FZF_COMPLETION_DIR_OPTS? | default ''
+
+$env.FZF_COMPLETION_DIR_COMMANDS = $env.FZF_COMPLETION_DIR_COMMANDS? | default ['cd', 'pushd', 'rmdir']
+
+# --- Helper Functions ---
+
+# Helper to build default fzf options list
+def __fzf_defaults_nu [prepend: string, append: string] {
```
junegunn: Most (all?) functions in this file have `_nu` suffix. Could you explain why you decided to put it?

test/lib/common.rb
```diff
@@ -78,19 +78,63 @@ def fish
...
-    send_keys 'function fish_prompt; end; clear', :Enter
-    self.until(&:empty?)
+    if shell == :fish
+      send_keys 'function fish_prompt; end; clear', :Enter
+      self.until(&:empty?)
+    elsif shell == :nushell
+      # Clear history from previous tests to avoid contamination
+      FileUtils.rm_f('/tmp/fzf-nushell-xdg/nushell/history.txt')
+      # Wait for nushell to be ready: send a marker and wait for it
+      sleep 2
```
junegunn: Can there be a better way to ensure that nushell is ready than sleeping for an arbitrary period of time?

shell/key-bindings.nu
```diff
@@ -0,0 +1,164 @@
...
+      send: executehostcommand
+      cmd: "commandline edit --replace (
+        let fzf_opts = (__fzf_defaults '' $'--scheme=history --bind=ctrl-r:toggle-sort --wrap-sign \"\t↳ \" --highlight-line ($env.FZF_CTRL_R_OPTS) +m --read0');
+        let fzfcmd = (__fzfcmd);
+        let fzf_args = ($fzfcmd | skip 1);
+        # reverse | uniq: show most recent first, deduplicate keeping the latest.
+        # Nushell's `history` loads the full history as an in-memory table
+        # (bounded by $env.config.history.max_size, default 100,000), so
+        # reverse and uniq operate on an already-materialized list and are
+        # effectively free.
```
junegunn:
```suggestion
        # Nushell's `history` loads the full history as an in-memory table
        # (bounded by $env.config.history.max_size, default 100,000), so
        # reverse and uniq run on an already-materialized list. This is O(n)
        # but acceptable for typical history sizes; unlike bash/zsh `fc -r`,
        # there is no streaming primitive that would let fzf show the latest
        # entries before the full list is consumed.
```

shell/completion.nu
```diff
@@ -0,0 +1,489 @@
...
+  } else {
+    # 3. FZF didn't handle it, so call the previous completer (if it exists).
+    if $previous_external_completer != null {
+      do $previous_external_completer $spans
+    } else {
+      # 4. No previous completer, and FZF didn't handle it. Return null.
+      null
+    }
+  }
+}
```
junegunn: Is it safe to re-source this file? It looks like each load captures the current completer as `$previous_external_completer` and wraps it, so re-sourcing would nest a new wrapper around the previous wrapper, growing the chain on every reload. Could you confirm?

shell/completion.nu
sim590: Good catch — fixed in 7b71302c. Without `--raw`, Nushell tries to parse the file based on its extension, which breaks for plain option files. Both files now use `open --raw`.

shell/completion.nu
sim590: Harmonized in b640f91d — completion.nu now uses the same form as key-bindings.nu: `$env.TMUX_PANE? | default '' | into string | is-not-empty`. The extra `into string` is needed because `TMUX_PANE` can be an integer depending on context.

shell/completion.nu
sim590: Removed in b4d33a65 — the `_nu` suffix was unnecessary. The only function that keeps it is `__fzf_defaults_nu` because `__fzf_defaults` (without suffix) is already defined in key-bindings.nu with a different signature, and both files can be sourced in the same session.

shell/completion.nu
sim590: Fixed in 298e0e2d — added a guard: `$env.__fzf_completer_registered` is set to `true` after the first registration, and subsequent loads skip the wrapper creation entirely. This prevents nesting on re-source.

test/lib/common.rb
sim590: Replaced in a60cdb23 + becb14f6 — we now poll by sending `print "fzf-ready"` and waiting for a line that is exactly `fzf-ready` in the terminal output, with up to 5 retries. No more arbitrary sleep.

shell/completion.nu
junegunn: Do they have to differ? In bash and zsh, we just put the identical definition in both files.

shell/completion.nu
junegunn: Not a big issue, but the commit hashes in your latest comments are broken. Looks like the agent hallucinated the hashes in those replies?

sim590: Yeah that's weird. I don't know where it got those hashes from. It didn't rebase... Sometimes it does that if it rebased  earlier. The right hash should be 0b1d82e37f6601cccd8b070b8901ce15b6b8bfb3. I'm guessing it might have rebased the branch without me asking him to :smile:

shell/completion.nu
sim590: No, they don't have to differ. Done in 6485735c — `__fzf_defaults_nu` is renamed to `__fzf_defaults` with the identical implementation as key-bindings.nu. Nushell silently accepts the redefinition when both files are sourced in the same session (e.g. via autoload), which matches the bash/zsh behavior.

shell/completion.nu
```diff
@@ -0,0 +1,489 @@
...
+    # Return null to let Nushell fall back to other completers (e.g., default file completion).
+    null
+  }
+}
+
+# --- WRAPPER AND REGISTRATION ---
+
+# Guard against re-sourcing: wrapping the completer multiple times would
+# nest wrappers and grow the call chain on every reload.
+if ($env.__fzf_completer_registered? | default false) != true {
```
junegunn, line 452: This means we can no longer re-source the file to apply new changes in the file? Would it be possible to allow that?

sim590: Good point — fixed in 7bf47c75. Instead of blocking re-sourcing entirely, we now save the pre-fzf completer in `$env.__fzf_previous_completer` on first load. On subsequent sources, that saved reference is used as the base rather than the current completer (which is already the fzf wrapper), so re-sourcing always produces a single wrapper with no nesting.

junegunn:
I'm not a nushell user so I might be missing something, but that the fuzzy completion is not always available is quite confusing to me. `vim **<tab>` works, `ls **<tab>` doesn't. `kill **<tab>` doesn't work, unless I change it to `^kill **<tab>`. This is confusing, also a bit frustrating to see that fuzzy completion doesn't pop up only after I typed in `**`. Is there a way to make the situation better? Or is this a fundamental limitation of nushell?

shell/completion.nu
```diff
@@ -0,0 +1,489 @@
...
+
+  if ($last_span | str ends-with $trigger) {
+    # --- Trigger Found ---
+
+    # Skip sudo to determine the actual command
+    let cmd_spans = if ($spans | first) == "sudo" { $spans | skip 1 } else { $spans }
+    let cmd_word = ($cmd_spans | first | default "")
+
+    # Calculate the prefix (part before the trigger in the last span)
+    let prefix = $last_span | str substring 0..(-1 * ($trigger | str length) - 1)
```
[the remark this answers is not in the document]
sim590: Not a bug. In Nushell, `str substring 0..-3` on a 2-character string (`"**"`) returns an empty string `""`, which is the intended prefix for the bare-trigger case. Verified:

```nushell
'**' | str substring 0..-3 | str length
# => 0
```

Nushell clamps out-of-range indices rather than throwing an error.

install
```diff
@@ -436,6 +443,15 @@ if [[ $shells =~ fish ]]; then
   fi
 fi
 
+if [[ "$shells" =~ nushell ]]; then
+  echo "Setting up Nushell integration ..."
+  mkdir -p "$nushell_autoload_dir"
+  echo -n "  Generate fzf.nu ... "
+  "$fzf_base"/bin/fzf --nushell > "$nushell_autoload_dir/fzf.nu"
+  echo "OK"
```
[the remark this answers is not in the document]
sim590: This is intentional and consistent with fish (`fzf --fish | source` also loads both keybindings and completion unconditionally). The `--[no-]key-bindings` / `--[no-]completion` flags are documented as bash & zsh only (line 25 of the install script). A finer-grained approach for nushell and fish could be considered in a follow-up.

sim590:
These are fundamental limitations of Nushell's completion architecture:

- **`ls **<TAB>` / `kill **<TAB>` don't work** because `ls` and `kill` are Nushell built-in commands. Since [Nushell 0.103.0](https://www.nushell.sh/blog/2025-03-18-nushell_0_103_0.html#external-completers-are-no-longer-used-for-internal-commands-toc) ([nushell/nushell#15086](https://github.com/nushell/nushell/pull/15086)), the external completer is no longer invoked for internal commands. Using `^kill **<TAB>` forces the external binary, which goes through our completer — that's why it works.

- **The `**` trigger requirement** exists because Nushell's external completer is only called ["if no Nushell completions were found"](https://www.nushell.sh/book/custom_completions.html#external-completions). For external commands (like `vim`, `^kill`), Nushell has no built-in completions, so our completer gets called and can check for the trigger. The trigger is a deliberate design choice (same as bash/zsh) to avoid replacing the user's existing completions for every command unconditionally.

AFAIK, there's no way around these limitations without changes on the Nushell side.

install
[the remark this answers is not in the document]
junegunn: Not true. Fish supports those parameters:

https://github.com/sim590/fzf/blob/bae407bf0f6b1ea88083c2136b0d30c97644c639/install#L387-L402

I think nushell can easily implement the same here:

https://github.com/sim590/fzf/blob/bae407bf0f6b1ea88083c2136b0d30c97644c639/install#L449-L450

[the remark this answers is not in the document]
sim590: You're right, my mistake — fixed in 9bf1b8ac. The nushell block now follows the same pattern as fish: when both are enabled, `fzf --nushell` is used as before; when only one is enabled, the relevant section is extracted from `fzf --nushell` output via `awk` and embedded directly in `fzf.nu` (no hardcoded `source` paths).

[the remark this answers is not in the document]
sim590: One question though: for the partial case (only key-bindings or only completion), the other shells generate `source "$fzf_base/shell/..."` lines with an absolute path to the repo. Since nushell's `fzf.nu` lives in the autoload directory (a copy, not a symlink), I instead extract the relevant section from `fzf --nushell` output via `awk` and embed the content directly — no dependency on the repo path.

Are you okay with this approach, or would you prefer we align with the other shells and use `source` with an absolute path?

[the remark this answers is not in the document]
junegunn: I'm okay with the approach, but the `awk` dance seems unnecessary since we have access to the original files and can just `cp` the right one to the destination, no?

Another concern: should we pick a name other than `fzf.nu` to avoid possibly overwriting a user's custom script that happens to share the name? Maybe a name less likely to conflict?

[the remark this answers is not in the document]
sim590: Good point about the name conflict. I think we can address this in two ways:

1. **Rename the file** to something less likely to conflict with a user's own scripts — e.g. `_fzf_integration.nu`. The leading underscore and longer name make it unlikely anyone would pick that name naturally.

2. **Add an overwrite check** before writing — if the file already exists, prompt the user before replacing it (similar to how the fish block checks for an existing `fish_user_key_bindings.fish`).

As for the `awk` approach — you're right, we can simply `cp` the source file directly for the partial case.

Would you like both safeguards, or is the rename alone sufficient?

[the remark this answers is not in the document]
junegunn: Naming it `_fzf_integration.nu` should suffice.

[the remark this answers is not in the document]
sim590: Alright. I did it in a6fb1026edefd91a384d7cbf0f7dbea1ed46980f.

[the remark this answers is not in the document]
junegunn: Thanks, I think we also need to update `uninstall`.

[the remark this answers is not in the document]
sim590: Done in 37d83c1b.

junegunn (dismissed):
Thanks for the persistence; it's good to go.

But I'll merge it right before I release a new version so the README doesn't mention unreleased features and confuse users.

junegunn:
Ah, we also need to update https://github.com/junegunn/fzf/blob/263eb4732fc6268f9fb35cffb634903ea8e2a26b/man/man1/fzf.1#L1352-L1369

sim590:
> Ah, we also need to update
> 
> https://github.com/junegunn/fzf/blob/263eb4732fc6268f9fb35cffb634903ea8e2a26b/man/man1/fzf.1#L1352-L1369

Did it in 52a23e3

junegunn (approved):

junegunn:
Merged to the devel branch. Will be included in the next version that I'm going to release shortly. Thanks!

junegunn:
Included in https://github.com/junegunn/fzf/releases/tag/v0.73.0

sim590:
Nice !!!! :tada: :tada: :tada:

fdncred:
nice work! I'd love to have conversations about how to better support fzf in nushell.

PR #4636, opened 2025-12-16
Fix --accept-nth being ignored in filter mode

### LINKED ISSUE #4615: --accept-nth ignored with --filter
`--accept-nth` has no effect when used in combination with `--filter`

the `--query` option respects `--accept-nth`

```
% echo 'one two three' | fzf -d ' ' --with-nth 1 --accept-nth -1 -q one
three
```

but `--filter` does not

```
% echo 'one two three' | fzf -d ' ' --with-nth 1 --accept-nth -1 -f one
one two three
```

charemma:
The `--accept-nth` option wasn't working with `--filter` mode. It would just ignore the field selection and print the whole line instead of the specified fields.

I ran into this when trying to use filter mode in a script where I wanted to filter by one field but output another (like showing names but returning IDs). Turns out filter mode just wasn't applying the field transformer at all - it was only working in interactive mode.

Fixed it by adding a `buildItemTransformer()` helper function that's now used consistently across filter mode (both streaming and non-streaming paths) and the select1/exit0 modes.

Tested with various delimiter and field configs, all the go tests still pass.

Examples that now work:
```bash
$ echo 'one two three' | fzf -d ' ' --with-nth 1 --accept-nth -1 -f one
three

$ echo 'ID001:John:Developer' | fzf -d ':' --with-nth 2 --accept-nth 1 -f John
ID001
```

Fixes #4615

### DIFF
--- a/test/test_filter.rb
+++ b/test/test_filter.rb
@@ -312,4 +312,18 @@ class TestFilter < TestBase
       assert_equal expected, result
     end
   end
+
+  def test_accept_nth
+    # Single field selection
+    assert_equal 'three', `echo 'one two three' | #{FZF} -d' ' --with-nth 1 --accept-nth -1 -f one`.chomp
+
+    # Multiple field selection
+    writelines(['ID001:John:Developer', 'ID002:Jane:Manager', 'ID003:Bob:Designer'])
+    assert_equal 'ID001', `#{FZF} -d: --with-nth 2 --accept-nth 1 -f John < #{tempname}`.chomp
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks, could you add a test case for this in https://github.com/junegunn/fzf/blob/master/test/test_filter.rb?

charemma:
Added a test case in test_filter.rb that covers single field selection, multiple fields, and different delimiters.

junegunn:
Merged, thanks!

PR #4641, opened 2025-12-19
Use builtin min, max

mrclmr:


mrclmr:
Sorry for not executing locally
```
FZF_VERSION=test make test
```
Build should pass now. If you are open for more PRs I would solve the deprecation warnings in `.goreleaser.yml`.

src/algo/algo.go
```diff
@@ -589,7 +589,7 @@ func FuzzyMatchV2(caseSensitive bool, normalize bool, forward bool, input *util.
 					if b >= bonusBoundary && b > fb {
 						consecutive = 1
 					} else {
-						b = util.Max16(b, util.Max16(bonusConsecutive, fb))
+						b = max(b, max(bonusConsecutive, fb))
```
junegunn: If I'm not mistaken, this builtin max can take multiple arguments, right? So we can simplify some of these calls.

mrclmr: Nested max calls are replaced with single max call.

mrclmr:
Sorry for the timeline confusion and the triggered GitHub Action workflows. I accidentally updated the PR branch and not master in my fork. So I had a merge commit, removed it and finally force pushed a rebased PR with up-to-date master.

junegunn:
Merged, thanks!

PR #4650, opened 2026-01-01
Parse envvar in FZF_DEFAULT_OPTS_FILE

phanen:
To move most of the fzf flags to a single file

### DIFF
--- a/src/options.go
+++ b/src/options.go
@@ -3709,10 +3709,10 @@ func postProcessOptions(opts *Options) error {
 	return nil
 }
 
-func parseShellWords(str string) ([]string, error) {
+func parseShellWords(str string, parseEnv bool) ([]string, error) {
 	parser := shellwords.NewParser()
 	parser.ParseComment = true
-	parser.ParseEnv = true
+	parser.ParseEnv = parseEnv
 	return parser.Parse(str)
 }
 
@@ -3729,7 +3729,7 @@ func ParseOptions(useDefaults bool, args []string) (*Options, error) {
 				return nil, errors.New("$FZF_DEFAULT_OPTS_FILE: " + err.Error())
 			}
 
-			words, parseErr := parseShellWords(string(bytes))
+			words, parseErr := parseShellWords(string(bytes), true)
 			if parseErr != nil {
 				return nil, errors.New(path + ": " + parseErr.Error())
 			}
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks, I understand the motivation. However, I'm not sure about the capabilities and limitations of the `ParseEnv` feature of the library and there are questions need be addressed:

- Does it support alternative forms like `${FOO}`?
- Does it support more sophisticated expressions like `${FOO:-default-foo}`?
- Does it require quoting when a value contains multiple words?
    - e.g., `--border-label "$FZF_LABEL"` vs. `--border-label $FZF_LABEL`
    - The behavior differs between Bash and Zsh, so a shell-specific interpretation could be confusing for users.
- Is there any chance it can break the surrounding expressions when the value contains double or single quotes?

So overall, I'm not confident this is the best way.

phanen:
> * Does it support alternative forms like `${FOO}`?
> 
> * Does it support more sophisticated expressions like `${FOO:-default-foo}`?

it did expanding `${FOO}` but not support `${FOO:-default-foo}`.
But each shell can have different syntax, if possible I want at least `$FOO` is supported, which is convention 90% of the time.

> Does it require quoting when a value contains multiple words?
> Is there any chance it can break the surrounding expressions when the value contains double or single quotes?

Did consider a lot about it. Have to agree this introduce a new syntax different from shell. Just tested, even in single quote the `$VARNAME` is expanded, so this can cause confusing. If user use some shell cmd in `--bind`, some environment var may be expanded before it should be in the subshell.

phanen:
Maybe add some document about this behavior. If a user want to put config in `$FZF_DEFAULT_OPTS_FILE`, he should realize this is different from shell.

junegunn:
> Maybe add some document about this behavior. If a user want to put config in `$FZF_DEFAULT_OPTS_FILE`, he should realize this is different from shell.

Hmm, I'm a bit reluctant to introduce a feature that could feel half-baked or potentially confusing.

On second thought, having environmental variables in `$FZF_DEFAULT_OPTS_FILE` means that it's dependent on the user's shell configuration anyway. If that's the case, I don't see a strong benefit here. Users could just as well implement a small script that dynamically generates `$FZF_DEFAULT_OPTS` and `source ~/fzf-env.sh`.

phanen:
Yeah, have to agree no significant benefit. At first I just found changing this file can apply config immediately without resource/restart the shell, so I try to move all my flags into file. but found only one flag remaining in the shellrc: `--history="$XDG_DATA_HOME/fzfhistory"` so I create a pr here. I have no problem to leave one more flag in the init shell since FZF_DEFAULT_OPTS_FILE should always be set too. If this may potentially cause more confusing than convient for some user.

PR #4651, opened 2026-01-01
fix: sync preview hidden state across alternate layouts

### LINKED ISSUE #4100: [Bug] alternate preview layout maintains separate hidden state
### Checklist

### Output of `fzf --version`

0.56.2

### OS

- [X] Linux
- [ ] macOS
- [ ] Windows
- [ ] Etc.

### Shell

- [ ] bash
- [X] zsh
- [ ] fish

Consider the below command:
```sh
echo "line1\nline2" | fzf --preview 'echo $FZF_PREVIEW_COLUMNS' --preview-window 'right,60%,<80(down:45%)' --preview-window 'nohidden'
```

Run the command in any terminal width and use a keybind to hide the preview:
- If the preview was `down` extend the terminal width until preview is shown on the right
- If the preview was `right` decrease the terminal width until preview is shown at the bottom

It's very easy to do if you have a tmux split into two panes and then use `<tmux-prefix>-z` to "zoom" the pane which extends it to full terminal width.

It seems that the hidden state is saved per layout (horizontal/veritcal), when extending the terminal width the preview will lose its hidden state and will be un-hidden, when decreasing the terminal width again the preview will be hidden again.

majiayu000:
Fixes #4100

## Changes
- Add togglePreview() helper to sync hidden state to both main and alternative preview layouts
- Call togglePreview() after toggle/show/hide/close actions to ensure state consistency

When using alternate preview layouts (e.g., `right,60%,<80(down:45%)`), hiding the preview in one layout now persists when the terminal is resized and the alternate layout becomes active.

### DIFF
--- a/src/terminal.go
+++ b/src/terminal.go
@@ -4866,6 +4866,13 @@ func (t *Terminal) hasPreviewWindowOnRight() bool {
 	return t.hasPreviewWindow() && t.activePreviewOpts.position == posRight
 }
 
+func (t *Terminal) togglePreview(hidden bool) {
+	t.previewOpts.hidden = hidden
+	if t.previewOpts.alternative != nil {
+		t.previewOpts.alternative.hidden = hidden
+	}
+}
+
 func (t *Terminal) currentItem() *Item {
 	cnt := t.merger.Length()
 	if t.cy >= 0 && cnt > 0 && cnt > t.cy {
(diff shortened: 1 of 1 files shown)


majiayu000:
Closing this PR due to extended inactivity. Happy to reopen if there's still interest — feel free to comment.

PR #4652, opened 2026-01-01
Add --no-clear-on-start option

phanen:
This can be used in shell integration, to avoid clear the cmdline
content

before:
image

after:
image

### DIFF
--- a/src/options.go
+++ b/src/options.go
@@ -650,6 +650,7 @@ type Options struct {
 	ListenAddr        *listenAddress
 	Unsafe            bool
 	ClearOnExit       bool
+	KeepScreen        bool
 	WalkerOpts        walkerOpts
 	WalkerRoot        []string
 	WalkerSkip        []string
@@ -3277,6 +3278,8 @@ func parseOptions(index *int, opts *Options, allArgs []string) error {
 			opts.ClearOnExit = true
 		case "--no-clear":
 			opts.ClearOnExit = false
+		case "--no-clear-on-start":
+			opts.KeepScreen = true
 		case "--walker":
 			str, err := nextString("walker options required [file][,dir][,follow][,hidden]")
 			if err != nil {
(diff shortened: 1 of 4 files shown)


junegunn:
Can you explain more about the screenshots? Which feature are you testing, and on which shell?

phanen:
This is fish, I also tested in zsh.
Cursor is on `|`: `nvim | filename1 filename2`.
I trigger fzf shell integration to help me insert a new filename. Without this pr, filename1/filename2 is hidden. (clear by `csi J`)
This pr just avoid clear it when we enter.

junegunn:
I think we can just reorganize the code so that it doesn't delete the current line if the cursor is not at the front. This should be acceptable in almost all cases. What do you think?

```diff
diff --git a/src/tui/light.go b/src/tui/light.go
index ed537322..97d1effe 100644
--- a/src/tui/light.go
+++ b/src/tui/light.go
@@ -207,11 +207,6 @@ func (r *LightRenderer) Init() error {
 	if r.fullscreen {
 		r.smcup()
 	} else {
-		// We assume that --no-clear is used for repetitive relaunching of fzf.
-		// So we do not clear the lower bottom of the screen.
-		if r.clearOnExit {
-			r.csi("J")
-		}
 		y, x := r.findOffset()
 		r.mouse = r.mouse && y >= 0
 		// When --no-clear is used for repetitive relaunching, there is a small
@@ -222,6 +217,11 @@ func (r *LightRenderer) Init() error {
 			r.upOneLine = true
 			r.makeSpace()
 		}
+		// We assume that --no-clear is used for repetitive relaunching of fzf.
+		// So we do not clear the lower bottom of the screen.
+		if r.clearOnExit {
+			r.csi("J")
+		}
 		for i := 1; i < r.MaxY(); i++ {
 			r.makeSpace()
 		}
```

phanen:
yes, this seems a better solution.

PR #4663, opened 2026-01-20
Fix page-up/page-down not executing subsequent actions

### LINKED ISSUE #4649: page-down+track-current ignores track-current action
### Checklist

### Output of `fzf --version`

0.67.0 (2ab923f3)

### OS

- [x] Linux
- [ ] macOS
- [ ] Windows
- [ ] Etc.

### Shell

- [ ] bash
- [x] zsh
- [ ] fish

With --bind 'page-up:page-up+track-current' The +t does not activate.  

If I have track-current bound to another key I can toggle it on/off after paging up/down.  It seems to be an issue of combining it with a position move.

namest504:
Fixes #4649
 
When using bindings like `page-up:page-up+track-current`, the `track-current` action wasn't being executed. This affected all page movement actions combined with `+`.
 
The issue was caused by a `break` statement at line 6420 that exited the switch case before subsequent actions could run. Additionally, there was logic that would immediately disable track-current when it detected cursor movement.
 
This PR restructures the code to use if-else and moves `req(reqList)` outside so both single-line and multi-line modes can execute it properly. The auto-disable logic is also commented out to prevent track-current from being turned off right after it's set.
 
Tested with various bindings and confirmed they now work as expected.

```bash
seq 1 100 | ./fzf --bind 'page-up:page-up+track-current'
seq 1 100 | ./fzf --bind 'page-down:page-down+track-current'
seq 1 100 | ./fzf --bind 'up:up+track-current'
seq 1 100 | ./fzf --bind 'down:down+track-current'
```

### DIFF
--- a/test/test_core.rb
+++ b/test/test_core.rb
@@ -1629,6 +1629,77 @@ class TestCore < TestInteractive
     end
   end
 
+  def test_page_up_action_chain
+    tmux.send_keys "seq 100 | #{FZF} --bind 'page-up:page-up+track'", :Enter
+    tmux.until do |lines|
+      assert_equal 100, lines.match_count
+    end
+    tmux.send_keys :PageUp
+    tmux.until do |lines|
+      assert_includes lines[-2], '+t'
+    end
+  end
+
+  def test_page_down_action_chain
+    tmux.send_keys "seq 100 | #{FZF} --height 100% --bind 'page-down:page-down+track'", :Enter
+    tmux.until do |lines|
+      assert_equal 100, lines.match_count
+    end
+    tmux.send_keys :PageDown
+    tmux.until do |lines|
+      assert_includes lines[-2], '+t'
+    end
+  end
+
+  def test_up_action_chain
+    tmux.send_keys "seq 100 | #{FZF} --bind 'up:up+track'", :Enter
+    tmux.until do |lines|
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks for your interest in the project. But I'm afraid your approach here isn't accurate.

> there was logic that would immediately disable track-current when it detected cursor movement

This is the intended behavior. Excerpt from man page:

```
track-current                (track the current item; automatically disabled if focus changes)
```

We automatically cancel the tracking when we detect a manual focus change. Global tracking, which is not automatically cancelled, can be enabled using the `--track` option or `toggle-track` action, and fzf will display `+T` instead of `+t` in the info line to distinguish the two modes.

The actual problem here is that we incorrectly disables the tracking when we shouldn't, and it happens when `track-current` is combined with any action that changes the focus.

- Renderer
    - Renders the screen
    - Remembers the current item index
- Event handler: Processes actions bound to the key
    - `up` / `page-up`, etc: Changes the focus
    - `track-current`: Enables the tracking of the "current" item
- Renderer
    - Notices that the current item has changed due to the movement
    - Incorrectly disables the tracking

To fix this, we have to store the current item index at the moment when the `track-current` action is triggered, and use the information for decide whether to cancel the tracking.

PR #4668, opened 2026-02-01
fix(terminal): handle SIGHUP signal

LangLangBart:
### to reproduce:

1. start fzf with sleep 12345

```bash
fzf --bind 'start:reload:true' --preview 'sleep 12345'
```

2. kill fzf with SIGHUP

```bash
kill -SIGHUP $(pgrep -x fzf)
```

3. check for the sleep command still running

```bash
ps -o pid,ppid,pcpu,state,command -p $(pgrep -x sleep)
```

### additional notes

Occurs most often for me when closing a window in Apple’s default Terminal while fzf is still running

sig

**EDIT1**: add image

### DIFF
--- a/src/terminal.go
+++ b/src/terminal.go
@@ -5209,7 +5209,7 @@ func (t *Terminal) Loop() error {
 
 	{ // Late initialization
 		intChan := make(chan os.Signal, 1)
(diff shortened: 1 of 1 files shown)


junegunn:
Merged, thanks!

PR #4670, opened 2026-02-04
Fix bug in advanced guide for using fzf as a ripgrep launcher

skbolton:
With the previous version if you passed in an initial search term that resulted in 0 results you would get rg error output in the prompt. It looks like down the change path the `|| true` clause was added but it also needs to be on the initial query.


junegunn:
Thanks!

Issue #4258, opened 2025-02-19
``--preview-window wrap`` broken preview scrolling in some condition.


fsc0:
### Checklist

- [x] I have read through the manual page (`man fzf`)
- [x] I have searched through the existing issues
- [x] For bug reports, I have checked if the bug is reproducible in the latest version of fzf

### Output of `fzf --version`

0.60.0 (6fa8295a)

### OS

- [x] Linux
- [ ] macOS
- [ ] Windows
- [ ] Etc.

### Shell

- [x] bash
- [ ] zsh
- [ ] fish

### Problem / Steps to reproduce

I have a text file named ``preview-scroll-test.txt``
```
Line 1
Line 2
Line 3
Line 4
Line 5
Line 6
Line 7
Line 8
Line 9
Line 10
Line 11
Line 12
Line 13
Line 14
Line 15
Line 16
Line 17
Line 18
Line 19
Line 20
Line 21
Line 22
Line 23 ######################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
Line 24
Line 25
```

And I saved below script like ``fzf-preview-scroll-test.sh``
```bash
#!/usr/bin/env bash
: | FZF_DEFAULT_OPTS= fzf \
	--preview-window wrap \
	--preview 'cat /tmp/preview-scroll-test.txt' \
	--bind 'up:preview-page-up,down:preview-page-down'
```

When I execute the script.
``Line 24~`` is hidden cause of very precise wrapping.
And ``up`` and ``down`` bindings not working.

![Image](https://github.com/user-attachments/assets/bae0eadc-5f77-4545-821a-fb59555131b2)

Let's remove those X phrase on ``Line 23``.

```
Line 1
Line 2
Line 3
Line 4
Line 5
Line 6
Line 7
Line 8
Line 9
Line 10
Line 11
Line 12
Line 13
Line 14
Line 15
Line 16
Line 17
Line 18
Line 19
Line 20
Line 21
Line 22
Line 23 ######################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################
Line 24
Line 25
```

Then execute the script again.
Line 24 shows up but scrolling still not working.

![Image](https://github.com/user-attachments/assets/e4a474d7-4418-42e3-a64e-c71af5533b0b)

[...]

Issue #3243, opened 2023-04-07
Preview window doesn't scroll to the end when combining 'follow' and 'wrap'


rakoo:
- [x] I have read through the manual page (`man fzf`)
- [x] I have the latest version of fzf
- [x] I have searched through the existing issues
=> Using tip from `master`

## Info

- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [ ] bash
    - [ ] zsh
    - [ ] fish
    - [x] sh

## Problem / Steps to reproduce

I want to use the `follow` option to scroll to the last line and follow the output of my preview script. Also, the preview script has long line, so I want them to be wrapped; I use the `wrap` option.

Problem: it seems like `fzf` calculates the scroll first and goes there, *then* wraps; what I see first is not the last line but some line in the middle.

Here's a reproducer, expecting your screen to be less  than 200 chars wide and 200 lines tall:

```
SHELL=sh fzf --preview-window follow,wrap --preview 'for i in $(seq 200); do echo -n $i; for j in $(seq 200); do echo -n '-'; done; echo $i; done'
```

Each line is `$i---------$i`, where `i` is iterated from 1 to 200, and the `---` is 200 chars wide.

What I expect: see the last line wrapped, which means something like

```
[more lines on top]
199-------------------------
-------------------------199
200-------------------------
-------------------------200
[nothing below, we're at the bottom]
```

What I see: 

```
[more lines on top]
171-------------------------
-------------------------171
172-------------------------
-------------------------172
[more lines below]
```

rakoo:
Oh and btw: thanks for fzf, it's amazing, I'm slowly replacing EVERETHING with fzf-based scripts

junegunn:
> Problem: it seems like `fzf` calculates the scroll first and goes there, _then_ wraps

Correct. This is a known issue. The current implementation calculates the offset first assuming there's no wrap before actually processing the preview contents (extracting and interpreting ANSI color codes, wrapping lines, and rendering the text on the preview window). Unfortunately, the whole process is pretty complicated and I think it's not going to be trivial to get it right.

rakoo:
Junegunn Choi ***@***.***> wrote:
[...]

PR #4672, opened 2026-02-05
fish: Command history improvements

### LINKED ISSUE #4661: key-bindings.fish: much faster ctrl-r in Awk for those without Perl
### Checklist

### Output of `fzf --version`

0.67.0 (Homebrew)

### OS

- [x] Linux
- [x] macOS
- [ ] Windows
- [ ] Etc.

### Shell

- [ ] bash
- [ ] zsh
- [x] fish

I don't have perl installed (it's not there by default on Fedora) so when I press CTRL-R I'm using @bitraid's clever Fish script (#4171) to preprocess the history output. Unfortunately, despite the optimizations it's still around 20x slower than Perl on my history file with 40k entries.

I didn't want to install Perl so I rewrote it as an AWK one-liner:

```
builtin history -z --reverse | tr '\0' '\036' | awk 'BEGIN{RS="\036"; ORS="\036"} {gsub(/\n/, "\n\t"); print NR "\t" $0}' | tr '\036' '\0'
```

tested on macOS and Linux to have identical output to the Perl command and perform within 1-2 ms.

It's entirely POSIX-compliant and could replace both the other methods of pre-processing in the script. If you want it I'll open a PR.

The only downside is that it'll break if you have any RECORD SEPERATOR (https://codepoints.net/U+001E?lang=en) in your Fish history

## Benchmarks

```
Benchmark 1: ./awk.fish
  Time (mean ± σ):      65.7 ms ±   2.7 ms    [User: 59.4 ms, System: 29.6 ms]
  Range (min … max):    61.7 ms …  70.8 ms    43 runs
 
Benchmark 2: ./perl.fish
  Time (mean ± σ):      63.1 ms ±   2.3 ms    [User: 61.6 ms, System: 23.0 ms]
  Range (min … max):    60.0 ms …  69.1 ms    49 runs
 
Benchmark 3: ./pure_fish.fish
  Time (mean ± σ):      1.263 s ±  0.011 s    [User: 0.917 s, System: 0.678 s]
  Range (min … max):    1.250 s …  1.279 s    10 runs
 
Summary
  ./perl.fish ran
    1.04 ± 0.06 times faster than ./awk.fish
   20.01 ± 0.76 times faster than ./pure_fish.fish
```

bitraid:
*- Update -*

This change provides the following improvements:

- Changes the view of the command history list, so that the script no longer depends on perl for performance.
- Enables syntax color highlighting on fish v4.3.3 and newer.
- Provides a preview window with the selected commands, and expanded view of the highlighted command if available.
- Improves the delete functionality, by successfully handling very large numbers of selected commands.
- Inserts commands in their formatted form with `<Alt-Enter>`.

---
*- Original -*

This change fixes the load delay of large history files when perl is missing, by using an alternative view that displays the history item date (in `yymmdd` format) instead of its count number. I think that this more useful so I dropped the use of perl completely (If you disagree, we can bring it back).

It has also a couple more changes/improvements:

- There is generally less overhead because there is no preprocessing of the history list, which is now also loaded with the most recent items first. Only the delete command processes the whole selected items because trailing newlines are stripped when using `{+2..}`.
- There is no line break on multi-line commands (can no longer add a tab after newlines) and newlines are represented by `␊` (like native fish history), but I don't think that this is a drawback because multi-line commands wasn't much easier to read before, since there was no indentation, while now there should be more space in these cases, so users can better fit a preview with a command like `"string collect -- {2..} | fish_indent --ansi"` that does indentation and syntax highlighting. Also, wrapping is enabled by default, because I think it is more useful to be able to view the whole commands.

---

Fixes #4661

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,15 @@ CHANGELOG
 - Added `--preview-wrap-sign` to set a different wrap indicator for the preview
   window
 - Added `alt-gutter` color option (#4602) (@hedgieinsocks)
+- Added `$FZF_WRAP` environment variable to child processes (`char` or `word`
+  when wrapping is enabled) (#4672) (@bitraid)
+- fish: Improved command history (CTRL-R) (#4672) (@bitraid)
+    - Enabled syntax highlighting in the list on fish 4.3.3+
+    - Added syntax-highlighted preview window that auto-shows for long or
+      multi-line commands
+    - Added `ALT-ENTER` to reformat and insert selected commands
+    - Improved handling of bulk deletion of selected history entries
+      (`SHIFT-DELETE`)
 - Added fish completion support (#4605) (@lalvarezt)
 - zsh: Handle multi-line history selection (#4595) (@LangLangBart)
 - Bug fixes


bitraid:
The failing tests is something we experienced before. The cause is the change in the way the history command is started, from a pipe to a subshell, in order to make the loading of the history items non-blocking. Even though the list loads instantly, I hope we'll be able to use the optimal method, by figuring out why in this case the history records seem to be unavailable for the tests, and fixing the issue.

bitraid:
I reverted to the original way of running history, after I remembered and confirmed that the command is non-blocking after all, but the tests still fail. I can't think of what else could be causing the issue...

lalvarezt:
hi @bitraid it's because of `--no-multi-line`, this flattens the new lines to ␊ and the test is not expecting that

lalvarezt:
also doing this on your local copy will let you run the tests. I didn't commit it in my PR because it was working online, but there's some flakiness with the timing

```diff
diff --git a/test/test_shell_integration.rb b/test/test_shell_integration.rb
index f51d7195..ea958835 100644
--- a/test/test_shell_integration.rb
+++ b/test/test_shell_integration.rb
@@ -114,6 +114,7 @@ module TestShell
     end
     tmux.send_keys 'echo 4th', :Enter
     tmux.prepare
+    sleep(1)
     tmux.send_keys 'C-r'
     tmux.until { |lines| assert_operator lines.match_count, :>, 0 }
     tmux.send_keys 'e3d'
@@ -135,6 +136,7 @@ module TestShell
     tmux.send_keys 'echo "foo', :Enter, 'bar"', :Enter
     tmux.until { |lines| assert_equal %w[foo bar], lines[-2..] }
     tmux.prepare
+    sleep(1)
     tmux.send_keys 'C-r'
     tmux.until { |lines| assert_equal '>', lines[-1] }
     tmux.send_keys 'foo bar'
@@ -1019,6 +1021,7 @@ class TestFish < TestBase
     tmux.prepare
     tmux.send_keys 'echo "bar', :Enter, 'foo"', :Enter
     tmux.prepare
+    sleep(1)
     tmux.send_keys 'C-l', 'C-r'
     block = <<~BLOCK
       echo "foo
```

bitraid:
> also doing this on your local copy will let you run the tests. I didn't commit it in my PR because it was working online, but there's some flakiness with the timing

Yes, I had similar experiences with some tests in the past, where they randomly failed locally, or cases where they passed locally but failed online. I was expecting to have some issues again, since I undid a workaround that was for that exact reason, but I wasn't expecting the rest of the changes to cause issues.

bitraid:
I added a workaround that doesn't disable multi-line display when running tests, by checking the value of `$XDG_CONFIG_HOME`. I don't think multi-line looks good without at least a leading tab. Thanks a lot for your help @lalvarezt , you saved me a lot of trouble.

junegunn:
> * I don't think that this is a drawback because multi-line commands wasn't much easier to read before, since there was no indentation

There is indentation in the current implementation. Am I missing something?

image

bitraid:
> There is indentation in the current implementation. Am I missing something?

~The current version, yes. But this is an attempt for an implementation that doesn't require preprocessing the history list.~

EDIT: I misunderstood the question - see my next reply.

bitraid:
> There is indentation in the current implementation. Am I missing something?

I meant proper indentation of typed commands:
``` 
for i in *
    if test -d "$i"
        echo $i is a dir
    end
end
```

EDIT: The commands must be typed not pasted.[](url)

junegunn:
So fish automatically shows the lines indented while typing that its history entry ends up without actual indentation. I can't help but feel removing multi-line display is a step backward, but I see your point.

I'm not sure about the `YYYYMMDD` format though. Because without the index numbers, it's hard to see the relative ordering of matches from the same day, which can be sometimes important. Adding `HH:MM:SS` helps, but it eats a lot of horizontal space and gets annoying in narrow split panes.

One idea is to put the time on a separate line like so:

```sh
history -Rz --show-time="%F %R:%S%n" | fzf --scheme=history --multi --read0 --print0 --tac --delimiter="\n" --nth=2.. --accept-nth=2.. --color=fg:dim,nth:regular --wrap --gap --highlight-line
```

But then it takes too many lines, so it's hard to say if it's an objective improvement.

Also, in some rare cases where there are a bunch of very similar matches, I want the option to narrow things down by typing the history index directly.

Anyway, that's my initial impression.

bitraid:
@junegunn Thanks for the input. I made the following modifications:

- Changed the date format to Unix time (seconds since Epoch), so the relative ordering of matches from the same day can be distinguished. The field is also now searchable, so it can be used to narrow down many similar matches. It's no longer very useful as date directly, but it can be converted and displayed in a preview or label transform by user configuration.
- Enabled ANSI colors for syntax highlighting. This is another advantage of using the output of `history` directly, and I think that it's a very nice improvement and something that fzf history lacked compared to native fish history search. It seems to have broken the tests again though, but I'm sure it can be fixed.

junegunn:
Thanks, syntax highlighting is really nice. Two things:

1.

I was using fish 4.1 (not a fish user) and I ran into this message:

```
history: --color=always: unknown option
```

Might be better to enable the option and `--ansi` with a version check.

2.

I think it's not ideal to have "Workaround for passing the tests" in the code. Why don't we just skip the otherwise failing test on fish? I don't see much point because fish by default doesn't support multi-line display now.

```diff
diff --git a/test/test_shell_integration.rb b/test/test_shell_integration.rb
index f51d7195..3687026f 100644
--- a/test/test_shell_integration.rb
+++ b/test/test_shell_integration.rb
@@ -129,6 +129,8 @@ module TestShell
   end
 
   def test_ctrl_r_multiline
+    skip('multi-line command history disabled in fish') if shell == :fish
+
     # NOTE: Current bash implementation shows an extra new line if there's
     # only entry in the history
     tmux.send_keys ':', :Enter
```

junegunn:
Or we could sprinkle some conditional branches to the tests. For example:

```diff
diff --git a/test/test_shell_integration.rb b/test/test_shell_integration.rb
index f51d7195..411731d5 100644
--- a/test/test_shell_integration.rb
+++ b/test/test_shell_integration.rb
@@ -1020,15 +1020,23 @@ class TestFish < TestBase
     tmux.send_keys 'echo "bar', :Enter, 'foo"', :Enter
     tmux.prepare
     tmux.send_keys 'C-l', 'C-r'
+    offset = -6
     block = <<~BLOCK
       echo "foo
       bar"
       echo "bar
       foo"
     BLOCK
+    if shell == :fish
+      offset = -4
+      block = <<~FISH
+        echo "foo␊bar"
+        echo "bar␊foo"
+      FISH
+    end
     tmux.until do |lines|
       block.lines.each_with_index do |line, idx|
-        assert_includes lines[-6 + idx], line.chomp
+        assert_includes lines[idx + offset], line.chomp
       end
     end
     tmux.send_keys :BTab, :BTab
```

bitraid:
There is now a version check before applying the options for syntax color highlighting. Also, I made a small improvement to the code of the initial version check.

To make up for the lack of line breaks in list view, a preview window is now used and is only displayed when the highlighted command can be expanded, or when commands are selected (more info in commit message). I understand that a setting like this is usually better left for users, but I think that it's a good default which users will find useful. There are no style modifications and will only be applied if a preview command isn't already set by the user. There is also added the `<alt-enter>` key bind, which inserts the selected commands after first being formatted by `fish_indent` (as shown in preview). These changes are in separate commits, so if you don't like this feature, it can be easily removed.

While writing this feature, I noticed that the `follow` parameter of `--preview-window` option doesn't work correctly when `wrap` is also set and the preview window has wrapped lines (seems like the wrapped lines are not being accounted for scrolling). Can you confirm the bug?

```fish
for i in (seq 0 9); string pad -w 200 -c $i $i; end | fzf --preview="echo {+}" --preview-window=10%,follow,wrap -m --reverse
```

shell/key-bindings.fish
```diff
@@ -182,36 +182,58 @@ function fzf_key_bindings
...
+          set -- date_cmd '(date -d @{1} \\"+%F %a %T\\")'
+        else if date -r 0 '+%s' 2>/dev/null | string match -q 0
+          # BSD date
+          set -- date_cmd '(date -r {1} \\"+%F %a %T\\")'
+        end
+      end
+
+      # Prepend the options to allow user customizations
+      set -p -- FZF_DEFAULT_OPTS \
+        '--bind="focus,resize:transform:if test \\"$FZF_COLUMNS\\" -gt 100 -a \\\\( \\"$FZF_SELECT_COUNT\\" -gt 0 -o (string length -- {}) -gt (math $FZF_COLUMNS - 4) -o (string collect -- {2..} | fish_indent | count) -gt 1 \\\\); echo show-preview; else echo hide-preview; end"' \
```
junegunn: This makes scrolling through the list noticeably sluggish. We should use [`bg-transform`](https://junegunn.github.io/fzf/releases/0.63.0/#asynchronous-transformation) instead.

---

I think it's nice to show the list of selected commands in the preview window, but I'm not sure about showing it for a single long command. Any reason you decided to not use `--wrap`?`

FWIW, There is a default binding for `toggle-wrap` (`ctrl-/`), so users can press it if they want to see the full command.

junegunn:
> Can you confirm the bug?

Yeah, it's a known issue.

- https://github.com/junegunn/fzf/issues/3243
- https://github.com/junegunn/fzf/issues/4258

Given the complexity involved, I don't have plans to fix it for now.

shell/key-bindings.fish
bitraid: > This makes scrolling through the list noticeably sluggish. We should use [`bg-transform`](https://junegunn.github.io/fzf/releases/0.63.0/#asynchronous-transformation) instead.

Right.

> I think it's nice to show the list of selected commands in the preview window, but I'm not sure about showing it for a single long command. Any reason you decided to not use `--wrap`?`

No big reason really. Just because it was not enabled before this PR, and without it the preview window won't cause extra wrapping and reduce the number of visible list items. I now realize though, that displaying the preview for a line that needs wrapping is not a good choice because it is going to be displayed even when line wrapping is enabled. I better change that.

bitraid:
> Yeah, it's a known issue.

That's OK, I'll make some changes to the script.

shell/key-bindings.fish
junegunn: Thanks. I like the direction overall.

Here's what I'm suggesting:

- Do not enfore `--wrap` or `--no-wrap`, leave it up to the users. Let them use `CTRL-/` binding for toggling it.
- Preview for multiple selections. This can be achieved by binding `bg-transform` to `multi` event. i.e. `--bind multi:bg-transform:...`

What do you think? I'm open to suggestions.

bitraid: > * Do not enfore `--wrap` or `--no-wrap`, leave it up to the users. Let them use `CTRL-/` binding for toggling it.

That makes sense. Setting default wrap was still possible by `$FZF_CTRL_R_OPTS`, but lets remove it completely so it can also be set by `$FZF_DEFAULT_OPTS`.

> Preview for multiple selections. This can be achieved by binding bg-transform to multi event. i.e. --bind multi:bg-transform:...

So, no automatic (formatted) preview of any sort of highlighted command (like `for i in a b c␊echo $i␊end` or `for i in a b c; echo $i; end`), correct? This would remove the need for setting the `follow` option of the preview window (which doesn't work well with `wrap`) and would simplify the code.

Should the column limit and the bind to `resize` be kept?

junegunn: > any sort of highlighted command

Oh, I see your point.

So the preview window is not just for long commands, but also for multi-line commands. When I tested this, there were only long, wrapped commands in my command history, so I didn't see much benefit of the preview window compared to `--wrap`. I mean, those commands are wrapped because the screen is not wide enough, so, opening a preview window on the right to make the screen narrower didn't seem ideal. But for multi-line commands, I see the appeal.

Another option would be to:
- Enable `--wrap` by default.
- Automatically enable preview **only** for multi-line commands.

bitraid: > So the preview window is not just for long commands, but also for multi-line commands.

(I have the details of when the preview is shown in the commit message, so you might have missed it.). Yes, the main purpose of the highlighted command preview, is to better view multi-line commands (properly indented this time), but also single-line chained commands that can be broken down by the formatter (the formatted form can be inserted in the command line with `ALT-ENTER` for easier editing). The logic behind also showing single-line commands that just don't fit in the available columns, is that it can be used to fully view the highlighted command without having to enable wrapping for all list items (could be useful for example, when scrolling a large list of very long commands). But for working correctly, it should only trigger when line wrapping is disabled. So, if you think is worth for fzf to export `$FZF_WRAP`, we could make it work, otherwise we can just remove this preview.

junegunn: > also single-line chained commands that can be broken down by the formatter

I see, I didn't have such a command in my history, so I couldn't see the merit.

With that in mind, I can agree with the current implementation. Changing `transform` to `bg-transform` is all we need.

It's a little unfortunate that `--wrap-sign` is shared between the list section and the preview window. I can add a separate `--preview-wrap-sign` option.

test/test_shell_integration.rb
```diff
@@ -114,6 +114,7 @@ def test_ctrl_r
     end
     tmux.send_keys 'echo 4th', :Enter
     tmux.prepare
+    sleep(1)
```
junegunn: Hmm, can there be a better way? This indicates that the "prepare" implementation is not good enough.

shell/key-bindings.fish
bitraid: These are three distinct cases, so we can have the preview for multi-lines and lines that can be broken down (no matter if they fit or not), while not for lines that only need wrapping. But if we also keep the preview for single lines that need wrapping, it should only be shown when line wrapping in the list is disabled. So, fzf would need to provide an environment variable with the state of `wrap`, something like:

```diff
diff --git a/src/terminal.go b/src/terminal.go
index 04e202c6..06ebc32d 100644
--- a/src/terminal.go
+++ b/src/terminal.go
@@ -1352,6 +1352,7 @@ func (t *Terminal) environImpl(forPreview bool) []string {
 	env = append(env, fmt.Sprintf("FZF_LINES=%d", t.areaLines))
 	env = append(env, fmt.Sprintf("FZF_COLUMNS=%d", t.areaColumns))
 	env = append(env, fmt.Sprintf("FZF_POS=%d", min(t.merger.Length(), t.cy+1)))
+	env = append(env, fmt.Sprintf("FZF_WRAP=%t", t.wrap))
 	env = append(env, fmt.Sprintf("FZF_CLICK_HEADER_LINE=%d", t.clickHeaderLine))
 	env = append(env, fmt.Sprintf("FZF_CLICK_HEADER_COLUMN=%d", t.clickHeaderColumn))
 	env = append(env, fmt.Sprintf("FZF_CLICK_FOOTER_LINE=%d", t.clickFooterLine))
```

test/test_shell_integration.rb
bitraid: Maybe there is a better way. I haven't spent much time on the tests - I applied this change from https://github.com/junegunn/fzf/pull/4672#issuecomment-3854987754. I will try to find some time and experiment.

shell/key-bindings.fish
junegunn: > something like

Looks good to me. Feel free to add it to this PR branch. If you do, please also mention that in https://github.com/junegunn/fzf/blob/b56d614ba2f901c64eed454f181badec0b1cedff/man/man1/fzf.1#L1366.

bitraid:
I've made the discussed changes. I wasn't able to find a better fix for the tests, other than adding a small delay to `tmux.prepare`.

> It's a little unfortunate that `--wrap-sign` is shared between the list section and the preview window. I can add a separate `--preview-wrap-sign` option.

That would be great.

There is also the issue of using `follow` together with `wrap` for the preview window. I think the simplest solution would be to show the highlighted before the selected commands, and disable `follow`.

junegunn:
Thanks.

> There is also the issue of using follow together with wrap for the preview window. I think the simplest solution would be to show the highlighted before the selected commands, and disable follow.

You mean the issue of not showing the wrapped lines at the bottom of the window? I think we can leave it as an open issue for now.


Another issue I noticed while testing the latest patch is that it add a small delay until the list is ready. I believe it's because we changed `eval $FZF_DEFAULT_COMMAND \| (__fzfcmd)` to `eval (__fzfcmd)`, leaving fzf to start the default command. But on my system `fish -c "some-command"` has some noticeable delay:

```
time fish -c :

real    0m0.247s
user    0m0.022s
sys     0m0.033s
```

and I see that when I hit CTRL-R.

junegunn:
The `follow` issue is fixed in c338df02 (and 74e98cac5cb7cc0ce08228ac45e2894bc690c4e8). Also, 69e9abda implemented word wrapping in the list section, so we probably should assign different values to `$FZF_WRAP`

- FZF_WRAP = char (`wrap && !wrapWord`)
- FZF_WRAP = word (`wrap && wrapWord`)
- or undefined (`!wrap`)

bitraid:
> The `follow` issue is fixed in [c338df0](https://github.com/junegunn/fzf/commit/c338df02c4b1e111d1fdd6a20edef15ad40bbe0e) (and [74e98ca](https://github.com/junegunn/fzf/commit/74e98cac5cb7cc0ce08228ac45e2894bc690c4e8)).

Nice, I tested it and it works great.

---

> Also, [69e9abd](https://github.com/junegunn/fzf/commit/69e9abdab4e7aea98437f40dbaeab56c7431be18) implemented word wrapping in the list section, so we probably should assign different values to `$FZF_WRAP`
>
> * FZF_WRAP = char (`wrap && !wrapWord`)
> * FZF_WRAP = word (`wrap && wrapWord`)
> * or undefined (`!wrap`)

I updated the values. I want to mention here an issue with the wrap symbol of the preview, where in some cases has the wrong color or doesn't display at all. I don't know if this is a known problem or not (searching the issues gave no results), so here is a brief example. Run (fish):

```fish
echo test | fzf --preview='echo (set_color red)(string repeat -n 5 1234567890)(set_color normal)word' --preview-window=wrap
```

- Modify the terminal size so that the numbers wrap to the second line: The wrap sign color is red.
- Modify the terminal size so that `word` is wrapped at the beginning of the second line: The wrap sign is not shown.

---

> But on my system `fish -c "some-command"` has some noticeable delay:

I reverted the change, but I wonder about the cause of the delay (commands in configuration files?) and how common this issue is. On my systems, fish starts instantly - I can't even notice the difference between using `transform` and `bg-transform`.

In this last push, I changed the way the delete command works, so it can handle a very large number of selected items. For not exceeding the argument length limit of `history` command, each item is now individually passed to `history` by iterating `{+f}`. Because I haven't noticed a delay compared to the previous method of passing all the items at once, the new method is used for any number/length of selected commands. It would be interesting if you could also try its performance, although it shouldn't make much difference because even though `history` is called multiple times, fish is still only started once. On my tests, deleting 10K records is done without noticeable delay, while at 50K there is a delay of a couple of seconds. You can run the command below to generate a large temporary history file for tests (the first variable sets the number in thousands of records):

```fish
set k 10; set t (math (date +%s) - $k x 1000); for j in (seq $k); set h; for i in (seq 1000); set -a -- h "- cmd: echo "(string repeat -n 10 $i$j$t(random) | base64 -w0)\n"  when: $(math $t + $i)"; end; string collect -- $h >>~/.local/share/fish/fzf_history; echo {$j}000; end
```

To switch to the generated history file run `set -x fish_history fzf`, and `set -e fish_history` to switch back.

---

Finaly, I happen to notice a typo I made when I suggested the changes to `install` script in #4605. The information in the line bellow should display `fish_user_key_bindings` instead of `fzf_user_key_bindings`

https://github.com/junegunn/fzf/blob/4522868fc0d3d9a41a58250810e22949879f092f/install#L442

Let me know if you want me to fix it by adding a commit here.

junegunn:
> I want to mention here an issue with the wrap symbol of the preview

Thanks for the report, I'll look into it.

> I wonder about the cause of the delay

I just reinstalled fish and removed all files under `~/.config/fish`, but I'm still noticing the delay.

```sh
▌~/.config/fish > ls
▌~/.config/fish > fish --version
fish, version 4.5.0
▌~/.config/fish > time fish -c :

real    0m0.251s
user    0m0.028s
sys     0m0.034s
```

I don't know why, but it's safer to assume this kind of delay can happen and prepare for it accordingly.

> On my tests, deleting 10K records is done without noticeable delay

There was a delay, but it's less than 0.5 seconds so it should never be an issue. BTW, I totally forgot about the `shift-delete` binding, I think we have to mention it somewhere.

Overall, everything looks great. I'll finish the review later in the day. Thanks!

shell/key-bindings.fish
```diff
@@ -182,36 +182,58 @@ function fzf_key_bindings
     set -l -- total_lines (count $command_line)
     set -l -- fzf_query (string escape -- $command_line[$current_line])
 
-    set -lx FZF_DEFAULT_OPTS (__fzf_defaults '' \
-      '--nth=2..,.. --scheme=history --multi --wrap-sign="\t↳ "' \
-      '--bind=\'shift-delete:execute-silent(eval history delete --exact --case-sensitive -- (string escape -n -- {+} | string replace -r -a "^\d*\\\\\\t|(?<=\\\\\\n)\\\\\\t" ""))+reload(eval $FZF_DEFAULT_COMMAND)\'' \
+    set -lx -- FZF_DEFAULT_OPTS (__fzf_defaults '' \
+      '--nth=2..,.. --scheme=history --multi --no-multi-line --no-wrap --wrap-sign="\t\t\t↳ " --preview-wrap-sign="↳ "' \
```
junegunn, line 186: I've rebased your branch onto the latest master and applied the new `--preview-wrap-sign` option.

I noticed you included a dot in the wrap sign (`↳·`), but I removed it for consistency. Let me know if there's a specific reason for keeping the dot that I might be missing.

shell/key-bindings.fish
```diff
@@ -182,36 +182,58 @@ function fzf_key_bindings
...
+          # BSD date
+          set -- date_cmd '(date -r {1} \\"+%F %a %T\\")'
+        end
+      end
+
+      # Prepend the options to allow user customizations
+      set -p -- FZF_DEFAULT_OPTS \
+        '--bind="focus,resize:bg-transform:if test \\"$FZF_COLUMNS\\" -gt 100 -a \\\\( \\"$FZF_SELECT_COUNT\\" -gt 0 -o \\\\( -z \\"$FZF_WRAP\\" -a (string length -- {}) -gt (math $FZF_COLUMNS - 4) \\\\) -o (string collect -- {2..} | fish_indent | count) -gt 1 \\\\); echo show-preview; else echo hide-preview; end"' \
+        '--preview="string collect -- (test \\"$FZF_SELECT_COUNT\\" -gt 0; and string collect -- {+2..}) \\"\\n# \\"'$date_cmd' {2..} | fish_indent --ansi"' \
+        '--preview-window="right,50%,wrap-word,follow,info,hidden"'
```
junegunn, line 210: I also changed `wrap` to `wrap-word` as I find it more useful.

junegunn:
> I don't know if this is a known problem or not (searching the issues gave no results), so here is a brief example.

I believe this is fixed now in 0ecbf3f475d6a69f7c38e92727cf28531420d70a.

shell/key-bindings.fish
bitraid: Thanks.

The reason for the middle dot is not very important, it is just for being distinct from an actual space. If the first character after the sign (which ends with a space) is also a space, it can be a little hard to recognize if a word that follows is separate or wrapped from the previous line. Now that word wrapping is available, it is even less important.

shell/key-bindings.fish
bitraid: Yes, it's probably more useful.

bitraid:
> > I don't know if this is a known problem or not (searching the issues gave no results), so here is a brief example.
> 
> I believe this is fixed now in [0ecbf3f](https://github.com/junegunn/fzf/commit/0ecbf3f475d6a69f7c38e92727cf28531420d70a).

Yes, the missing sign is now fixed. Is it expected for the sign to take the color of the wrapped word, though? This doesn't happen with the wrapping sign of the list.

junegunn:
> Is it expected for the sign to take the color of the wrapped word, though?

It wasn't like originally intended, but I kind of liked how it looked (especially in the preview window) and decided to keep it that way.

Anyway, I think this is now ready for merge. Thanks again for working on this!

PR #4676, opened 2026-02-08
Make symlinks to directories to return as directories, not files (#4675)

### LINKED ISSUE #4675: Symbolic links to directories show as files, not directories
### Checklist

### Output of `fzf --version`

0.67 (devel)

### OS

- [x] Linux
- [ ] macOS
- [ ] Windows
- [ ] Etc.

### Shell

- [ ] bash
- [ ] zsh
- [ ] fish

When fzf is filtering the filesystem for directories, even when fzf is following symbolic links, the directory a symlink points to never appears. Child directories do appear, just not the directory that is the symlink.

But when filtering for files, directory symlinks do appear.

## Reproduction

```
mkdir /tmp/asdf
cd /tmp/asdf

mkdir foo
ln -s foo bar

find -L * -type d
# bar
# foo

fzf --walker=dir,follow
# foo

fzf --walker=file
# bar
```

skk64:
Fixes #4675

Changes the way directories and symbolic links are detected so that symlinks to directories are handled identically to directories when opts.follow is enabled.

### DIFF
--- a/src/reader.go
+++ b/src/reader.go
@@ -303,8 +303,12 @@ func (r *Reader) readFiles(roots []string, opts walkerOpts, ignores []string) bo
 		}
 		path = trimPath(path)
 		if path != "." {
-			isDir := de.IsDir()
-			if isDir || opts.follow && isSymlinkToDir(path, de) {
+			isDirSymlink := isSymlinkToDir(path, de)
+			if isDirSymlink && !opts.follow {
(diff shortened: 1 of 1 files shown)


junegunn:
The updated behavior now matches that of the other tools (https://github.com/junegunn/fzf/issues/4675#issuecomment-3866064637), thanks!

PR #4684, opened 2026-02-21
Replace printf with builtin printf to by pass local indirections

DarrenBishop:
On macos, having run `brew install coreutils`, which installed GNU version of `printf`, this script/completion would sometimes complain about the (now missing) `-v` option usage.

This change set ensures the `-v` option is available where needed.

Note: there is a precedent of qualifying which tool to run e.g. `command find ...`, `command dirname ...`, etc, so hopefully `builtin printf ...` should not cause any offense.

### DIFF
--- a/shell/completion.bash
+++ b/shell/completion.bash
@@ -38,9 +38,9 @@ if [[ $- =~ i ]]; then
 # the changes. See code comments in "common.sh" for the implementation details.
 
 __fzf_defaults() {
-  printf '%s\n' "--height ${FZF_TMUX_HEIGHT:-40%} --min-height 20+ --bind=ctrl-z:ignore $1"
+  builtin printf '%s\n' "--height ${FZF_TMUX_HEIGHT:-40%} --min-height 20+ --bind=ctrl-z:ignore $1"
   command cat "${FZF_DEFAULT_OPTS_FILE-}" 2> /dev/null
-  printf '%s\n' "${FZF_DEFAULT_OPTS-} $2"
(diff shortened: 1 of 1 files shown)


junegunn:
> On macos, having run `brew install coreutils`, which installed GNU version of `printf`

Isn't it named `gprintf`? You have an alias?

junegunn:
Whatever the case, this is the right thing to do. Merged, thanks.

DarrenBishop:
> > On macos, having run `brew install coreutils`, which installed GNU version of `printf`
> 
> Isn't it named `gprintf`? You have an alias?

Hey so sorry, I started a reply earlier; yes, reminded by your comment, I realise that both:
1. All/most GNU replacements are installed with 'g'-prefixed name, and that
1. I routinely apply the `PATH` update suggested in the `brew install...` output, effectively putting non-prefixed aliases earlier on the `PATH`
    ```
    brew install coreutils
    ==> Fetching downloads for: coreutils                                                                                                                                                                                                                   
    
    ...
    
    ==> Caveats
    Commands also provided by macOS and the commands dir, dircolors, vdir have been installed with the prefix "g".
    If you need to use these commands with their normal names, you can add a "gnubin" directory to your PATH with:
      PATH="/opt/homebrew/opt/coreutils/libexec/gnubin:$PATH"
    ```

DarrenBishop:
> Whatever the case, this is the right thing to do. Merged, thanks.

Awesome. And FZF is awesome

Issue #4659, opened 2026-01-11
[Feature request] add `change-header-lines` action


ibhagwan:
Use case: when resuming fzf-lua history/oldfiles picker (same as fzf.vim's `:History`) we need to set the first line as a header but **only if the first line is the current buffer**, to do so we run a `transform:reload` on the resume event.

To set the header line we need to run: `change-header-lines(1)+reload:...`.

ibhagwan:
While we’re at it can we also add a `change-with-nth` action?

junegunn:
Both features are non-trivial to implement because they don't just change the visual style, they also affect the search and indexing logic.

I think I can add `change-header-lines` relatively easily but I'm not sure about `change-with-nth` since it would require reindexing the entire list.

junegunn:
Need to fix test failures.

junegunn:
@ibhagwan 
`change-with-nth` is being implemented in #4691.

The issue is that changing `--with-nth` requires re-transforming the entire list, which means we need to retain the original input stream. fzf currently avoids holding the original bytes in memory unless `--with-nth` is specified, in order to keep the memory footprint low.

Given that, I'm considering enabling `change-with-nth` only when fzf is initially started with `--with-nth` (i.e., at least `--with-nth ..`). Does that sound reasonable?

ibhagwan:
> [@ibhagwan](https://github.com/ibhagwan) `change-with-nth` is being implemented in [#4691](https://github.com/junegunn/fzf/pull/4691).
> 
> The issue is that changing `--with-nth` requires re-transforming the entire list, which means we need to retain the original input stream. fzf currently avoids holding the original bytes in memory unless `--with-nth` is specified, in order to keep the memory footprint low.
> 
> Given that, I'm considering enabling `change-with-nth` only when fzf is initially started with `--with-nth` (i.e., at least `--with-nth ..`). Does that sound reasonable?

Very reasonable, ty @junegunn !

Issue #2743, opened 2022-02-26
Feature Request add `change-nth` and `change-with-nth` actions


salmankhilji:
- [x] I have read through the manual page (`man fzf`)
- [x] I have the latest version of fzf
- [x] I have searched through the existing issues

## Info

- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Etc.
- Shell
    - [x] bash
    - [ ] zsh
    - [ ] fish

## Problem / Steps to reproduce

Using a customized version of `rigrep` integration as explained within the [Advanced examples](https://github.com/junegunn/fzf/blob/master/ADVANCED.md), after I `reload` my data, I'd like to be able to change the column that `fzf` searches in.  At the moment, this option is only set using the CLI options and it would be nice to be able to change that at runtime.

salmankhilji:
I have just realized that for my specific scenario, I can use `--nth=-1` as all of my data needs to limit the search scope to the very last column.  (I had not discovered negative numbers when I filed this feature request.)

Hence the priority to implement this feature may be dropped.

LangLangBart:
- change-with-nth: **0.70.0**[^1]
- change-nth: **0.58.0**[^2]

[^1]: https://github.com/junegunn/fzf/blob/master/CHANGELOG.md#0700
[^2]: https://github.com/junegunn/fzf/blob/master/CHANGELOG.md#0580

PR #4691, opened 2026-02-24
Add change-with-nth action to dynamically change --with-nth

junegunn:
https://github.com/user-attachments/assets/40abe7e6-8c98-4a13-85b7-17e69372ef9e

### DIFF
--- a/src/terminal.go
+++ b/src/terminal.go
@@ -340,6 +340,7 @@ type Terminal struct {
 	nthAttr              tui.Attr
 	nth                  []Range
 	nthCurrent           []Range
+	withNthDefault       string
 	withNthExpr          string
 	withNthEnabled       bool
 	acceptNth            func([]Token, int32) string
@@ -1093,6 +1094,7 @@ func NewTerminal(opts *Options, eventBox *util.EventBox, executor *util.Executor
 		nthAttr:            opts.Theme.Nth.Attr,
 		nth:                opts.Nth,
 		nthCurrent:         opts.Nth,
+		withNthDefault:     opts.WithNthExpr,
 		withNthExpr:        opts.WithNthExpr,
 		withNthEnabled:     opts.WithNth != nil,
 		tabstop:            opts.Tabstop,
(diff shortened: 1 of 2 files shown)


LangLangBart:
Tried it; it worked without any issues. Thanks.

- Ref to open issue: #2743

junegunn:
@LangLangBart Thanks for testing it! This is a non-trivial change, so I'd like to take a closer look before merging.

rootInfluence:
The `--with-nth` option currently requires specifying field indices at startup, but a dynamic action to change this during preview/selection would be powerful for exploratory workflows. You could implement this similar to how `change-preview` works—by adding a new action that accepts a field specification and updates the preview/display without restarting the search.

junegunn:
@ibhagwan If you have a chance, please test this.

ibhagwan:
> @ibhagwan If you have a chance, please test this.

Absolutely @junegunn, reopened https://github.com/ibhagwan/fzf-lua/issues/2515, will keep you posted.

ibhagwan:
@junegunn, found a small issue not sure if by design, the `change-nth-action` won't work unless the initial command had a `--with-nth` specified.

This works:
```sh
echo "a\tb\tc" | ./target/fzf-linux_amd64 --with-nth=".." --delimiter="[\t]" --bind="ctrl-y:transform:echo 'change-header(2..)+change-with-nth(2..)'"
```
But this doesn't (due to not starting with `--with-nth`):
```sh
echo "a\tb\tc" | ./target/fzf-linux_amd64 --delimiter="[\t]" --bind="ctrl-y:transform:echo 'change-header(2..)+change-with-nth(2..)'"
```
image

junegunn:
Yeah, this is by design, I mentioned it here: https://github.com/junegunn/fzf/issues/4659#issuecomment-3951410397. And the limitation is documented in the `--with-nth` section of the man page.

That said, we could potentially improve usability by inspecting the bound actions and enabling it when a `{change|(bg-)transform}-with-nth` action is present. However, that's not always feasible due to `--listen` and `transform`, so I think the current way is acceptable.

ibhagwan:
> Yeah, this is by design, I mentioned it here: [#4659 (comment)](https://github.com/junegunn/fzf/issues/4659#issuecomment-3951410397). And the limitation is documented in the `--with-nth` section of the man page.
> 
> That said, we could potentially improve usability by inspecting the bound actions and enabling it when a `{change|(bg-)transform}-with-nth` action is present. However, that's not always feasible due to `--listen` and `transform`, so I think the current way is acceptable.

Ty @junegunn, it's no big deal to add `--with-nth=".."` on my end.

PR #4693, opened 2026-02-26
Support --nth independent of --with-nth

### LINKED ISSUE #4257: [feature request] make --nth independent of --with-nth
### Checklist

### Output of `fzf --version`

0.59.0 (bbe1721a)

### OS

- [x] Linux
- [ ] macOS
- [ ] Windows
- [ ] Etc.

### Shell

- [ ] bash
- [x] zsh
- [ ] fish

```
cat <<EOF | fzf -d: --with-nth=2 --nth=1..2
col1:col2
EOF
```

Search for `col1`, which returns no result

It would be great to be able to search "hidden" fields as well.

I've seen https://github.com/junegunn/fzf/issues/2432, but that trailed off into another direction and was "closed" by a fix for ellipsis.

prashantv:
Currently, `--nth` only works if the displayed text contains the expected field, and uses the same delimiter.

Use the original text with `--nth` so it works independently of `--with-nth`.

Fixes #4257

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,10 @@ CHANGELOG
     - bash: CTRL-R now supports multi-select and `shift-delete` to delete history entries (#4715)
     - fish: Improved command history (CTRL-R) (#4703) (@bitraid)
 - `GET /` HTTP endpoint now includes `positions` field in each match entry, providing the indices of matched characters for external highlighting (#4726)
+- `--nth` now always matches against the original input line, even when
+  `--with-nth` is set. Previously `--nth` operated on the `--with-nth`
+  transformed text, coupling display and search. They are now independent:
+  `--with-nth` controls display, `--nth` controls search scope.
 - Bug fixes
     - `--walker=follow` no longer follows symlinks whose target is an ancestor of the walker root, avoiding severe resource exhaustion when a symlink points outside the tree (e.g. Wine's `z:` → `/`) (#4710)
     - Fixed AWK tokenizer not treating a new line character as whitespace
--- a/man/man1/fzf.1
+++ b/man/man1/fzf.1
@@ -113,11 +113,12 @@ Fuzzy matching algorithm (default: v2)
 .TP
 .BI "\-n, \-\-nth=" "N[,..]"
 Comma-separated list of field index expressions for limiting search scope.
-See \fBFIELD INDEX EXPRESSION\fR for the details. When you use this option with
-\fB\-\-with\-nth\fR, the field index expressions are calculated against the
-transformed lines (unlike in \fB\-\-preview\fR where fields are extracted from
-the original lines) because fzf doesn't allow searching against the hidden
-fields.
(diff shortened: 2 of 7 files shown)


prashantv:
@junegunn Are you open to a change that allows `--nth` to work even the output delimiter doesn't match the input delimiter, and allows matching against hidden fields?

junegunn:
Sorry but I don't plan to, and I can't accept patches that break backward compatibility. The whole program was built on the assumption that matching occurs only on visible parts. `with-nth` is now dynamic via `change-nth` or `transform-nth`, so one can unhide hidden parts when they want to match against them.

prashantv:
I totally understand keeping backwards compatibility, and open to other solutions to solve my problem. My specific issue is documented in https://github.com/junegunn/fzf/issues/4257#issuecomment-3960295169 -- the input delimiter doesn't match the output delimiter. Is there a way to match with a specific field when delimiters mismatch?

junegunn:
I see. My suggestion would be to pre-process the input using a separate program instead of soley relying on `--with-nth`.

```sh
echo -e "1,a\n2,b" | tr , $'\t' | fzf --with-nth "{1}   # {2}" --nth 1
```

A single `--delimiter` is used throughout the program (e.g. `--nth`, `--with-nth`, `--accept-nth`, `--preview 'echo {1}'` / `transform-border-label:echo {2}`, ...). If we start allowing different delimiters for different options, it can quickly get out of hands.

prashantv:
Thanks for the suggestion, but I did consider that but it doesn't work as the delimiter is intentionally a more complex pattern that is much less likely to occur in the input, while the output prioritizes simplicity for display (e.g., using space delimiter + colours to show `{2}`).

My real delimiter is `\t:::\t`, as I'm dealing with shell commands which can contain tabs or commas so the delimiter can't be tab or comma, and when displaying, I intentionally use a space + escape codes to change the colour.

It seems like there's interest in a purely-display version of the string that isn't used for matching (allowing hidden fields, and solving for more complex delimiters in input vs output). Rather than breaking backwards compatibility, would you be open to a different flag to support display formatting without impacting the matching / delimiters used to parse?

junegunn:
I've been using zero-width spaces in such cases [in my scripts](https://github.com/junegunn/everything.fzf/blob/6e2b5488f3434755f634a191e73ed91ed10c078f/jira.fzf#L44).

```sh
ZWS=$'\u200B'

echo -e "1,a\n2,b" | tr , $ZWS | fzf --delimiter $ZWS --with-nth "{1}$ZWS   # {2}" --nth 1
```

> would you be open to a different flag to support display formatting without impacting the matching / delimiters used to parse?

Probably not. I've rejected similar requests several times in the past. The primary reason is that it can be very confusing to understand why an item is matched by a fuzzy-matching query when you can't visually confirm which parts are being matched. It becomes even harder to reason about how matches are scored and sorted.

fzf is essentially an interactive grep program, and I can't help but feel that such use cases don't align well with its core design.

prashantv:
Thanks for the suggestion, using a 0-width space works great for my use-case!

PR #4699, opened 2026-03-02
Remove tmppath pledge on OpenBSD

lcheylus:
"tmppath" pledge is no longer supported.
See commit https://github.com/openbsd/src/commit/c883e836f48a8ca9ee3be9de6ebf4b36751f2196

### DIFF
--- a/typos.toml
+++ b/typos.toml
@@ -7,4 +7,4 @@ tabe = "tabe"
 Iterm = "Iterm"
 
 [files]
-extend-exclude = ["README.md"]
+extend-exclude = ["README.md", "*.s"]


src/protector/protector_openbsd.go
```diff
@@ -6,5 +6,5 @@ import "golang.org/x/sys/unix"
 
 // Protect calls OS specific protections like pledge on OpenBSD
 func Protect() {
-	unix.PledgePromises("stdio dpath wpath rpath tty proc exec inet tmppath")
+	unix.PledgePromises("stdio cpath dpath wpath rpath tty proc exec inet")
```
junegunn, line 9: So `cpath` instead of `tmppath`?

Is there anything I should know about before merging this?

junegunn:
Merged, thanks!

PR #4701, opened 2026-03-03
Proposal: Add `--track-nth` option for field-based tracking across reloads

110y:
## What

- Add a new option `--track-nth` option for the nth field-based tracking across reloads

## Why

I'm often using fzf with `--track` and `reload` to filtering dynamic contents. While it's useful, it sometimes leads me to an undesired entry when I use fzf for some kind of dynamic contents since its tracking is based on the index. For example, if I use fzf for filtering Kubernetes Pods (`kubectl get pods`) with the preview for the pod spec, I would like to make fzf track the entry based on the name of Pod via its nth field value, not the index.

### DIFF
--- a/shell/completion.bash
+++ b/shell/completion.bash
@@ -211,6 +211,7 @@ _fzf_opts_completion() {
     --tiebreak
     --tmux
     --track
+    --track-nth
     --version
     --walker
     --walker-root
--- a/src/options.go
+++ b/src/options.go
@@ -101,6 +101,8 @@ Usage: fzf [options]
     --no-multi-line          Disable multi-line display of items when using --read0
     --raw                    Enable raw mode (show non-matching items)
     --track                  Track the current selection when the result is updated
+    --track-nth=N[,..]       Track the current selection by the value of specific fields
+                             when the result is updated
     --tac                    Reverse the order of the input
     --gap[=N]                Render empty lines between each item
     --gap-line[=STR]         Draw horizontal line on each gap using the string
@@ -594,6 +596,7 @@ type Options struct {
 	Sort              int
 	Raw               bool
 	Track             trackOption
+	TrackNth          []Range
 	Tac               bool
 	Tail              int
 	Criteria          []criterion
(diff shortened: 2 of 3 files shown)


110y:
@junegunn 
Hi, thank you for your great work for this project.
Is this type of proposal acceptable? I appreciate if you could take a look this PR when you have time, thanks :pray:

junegunn:
Hi, as you might have noticed `--track` (global tracking) or `track` action (one-off tracking of the current item) were not designed with the `reload` action in mind. Because `reload` runs an arbitrary command and there is no guarantee that the resulting list is related to the previous one. It could be a completely different set of things and we do not know if it's desirable or even possible to keep tracking.

Another fundamental difficulty in preserving the tracking across `reload`s is that fzf reloads the list asynchronously, and the related item from the new list may not immediately be available.

```sh
seq 1000 | sort -R | fzf --track --bind 'space:reload:for i in $(seq 1000); do echo $i; sleep 0.1; done'
```

So while I understand the desire for this functionality, we need to think more carefully about how such a mechanism should be implemented.

That said, one could attempt to approximate the behavior using a more sophisticated `transform` action like so:

```sh
sort -R /usr/share/dict/words | fzf \
    --bind 'start:execute-silent:rm -f /tmp/prev-word' \
    --bind 'ctrl-r:reload:echo {1} > /tmp/prev-word; sort -R /usr/share/dict/words' \
    --bind 'load:transform:
      if [[ -r /tmp/prev-word ]]; then
        word=$(cat /tmp/prev-word)
        line=$(grep -n "^$word$" {*f} | head -1 | sed "s/:.*//")
        if [[ $line ]]; then
          echo "pos($line)+offset-middle"
        fi
      fi
    '
```

But this isn't quite nice because of the inherent asynchronous nature of fzf.

So overall, I'm open to the idea, but I'd like to spend more time thinking about how it could be implemented properly.

110y:
@junegunn 

Thank you for your explanation and introducing new `--id-nth` 🙏 
Now I confirmed that I can do what I want 🎉 

Again, thank you for your great work all the time!
`FZF` is literally an essential tool for my daily life ❤️

PR #4716, opened 2026-03-12
Update FZF_DEFAULT_COMMAND to show full timestamp

0rtz:
Make ctrl-r in fish shell more human readable


0rtz:
So i wanted to get rid of those ugly unix timestamps before commands in ctrl-r completion, but just noticed its actually a bit more tricky
Will look into this when i have more time

junegunn:
This will address your needs in the next release: https://github.com/junegunn/fzf/pull/4703

PR #4719, opened 2026-03-16
Add "unix" and "fattr" promises

dgl:
Without this fzf --listen=/tmp/foo.sock fails on OpenBSD:

```
fzf[66741]: pledge "unix", syscall 97
zsh: abort (core dumped)  fzf --listen=/tmp/fzf.sock
```

Further adjust the way pledges are used, so if --listen isn't specified it doesn't ask for either inet or unix.

### DIFF
--- a/src/protector/protector_openbsd.go
+++ b/src/protector/protector_openbsd.go
@@ -6,5 +6,5 @@ import "golang.org/x/sys/unix"
 
 // Protect calls OS specific protections like pledge on OpenBSD
 func Protect() {
-	unix.PledgePromises("stdio cpath dpath wpath rpath tty proc exec inet")
+	unix.PledgePromises("stdio cpath dpath wpath rpath inet fattr unix tty proc exec")
 }


src/protector/protector_openbsd.go
```diff
@@ -2,9 +2,17 @@
...
-import "golang.org/x/sys/unix"
+import (
+	"strings"
+
+	"golang.org/x/sys/unix"
+)
 
 // Protect calls OS specific protections like pledge on OpenBSD
-func Protect() {
-	unix.PledgePromises("stdio cpath dpath wpath rpath tty proc exec inet")
```
junegunn, line 9: Thanks.

I'd like to keep it simple and just hard-code the entries here instead.

A problem with having the conditional branch in `main()` is that fzf [can be used as a library](https://junegunn.github.io/fzf/tips/using-fzf-in-your-program/) and each library user has to do the same in their code. We could move the code to the `fzf.Run()`, but I just don't want to worry about it :)

dgl:
How about something like this? It keeps the reduction of access as well as pledging unix, but doesn't put the actual promise names in main.go.

junegunn:
Simpler, but it still can be considered a breaking change for library users.

I'm not really familiar with the whole concept of "pledge". What's the worst thing that can happen from putting `unix` in the list by default? This makes me wonder, should we conditionally add the other ones as well according the enabled features? For example, we may not need `wpath` and `dpath` if we don't use the `--tmux` option.

- https://github.com/junegunn/fzf/commit/295b89631b5baf1ca88d1b143805c73c7bff6a89

junegunn:
My point is if it doesn't do much harm, I'd much prefer to have a single, static list that specifies what fzf might need regardless of the current options.

dgl:
Separating out "can access the filesystem" from "can talk to the network" is valuable. Ideally pledges are as granular as possible. In this case I don't see much value in splitting out the tmux related ones into their own set, but splitting out the network ones means the binary goes from "can access filesystem and network" to just "can access filesystem", meaning any security issue should be more contained.

However given this already had inet, adding `unix fattr` isn't going to change much.

junegunn:
Understood, thanks for the explanation.

> However given this already had inet, adding unix fattr isn't going to change much.

Right, that's how I feel. Could you update your PR to simply add those to the list?

junegunn:
Merged, thanks!

Issue #3834, opened 2024-06-02
Examine Profile-guided optimization (PGO)


FilipAndersson245:
### Checklist

- [X] I have read through the manual page (`man fzf`)
- [X] I have searched through the existing issues
- [ ] For bug reports, I have checked if the bug is reproducible in the latest version of fzf

### Output of `fzf --version`

0.52.1 (6432f00)

### OS

- [X] Linux
- [X] macOS
- [X] Windows
- [ ] Etc.

### Shell

- [X] bash
- [X] zsh
- [X] fish

### Problem / Steps to reproduce

Profile-guided optimization (PGO), also known as feedback-directed optimization (FDO), is a compiler optimization technique that feeds information (a profile) from representative runs of the application back into to the compiler for the next build of the application, which uses that information to make more informed optimization decisions. For example, the compiler may decide to more aggressively inline functions which the profile indicates are called frequently.

https://go.dev/doc/pgo

Enabling PGO usually result in somewhere between `2-14%` better performance, I think it would be interesting to enable this for fzf.

junegunn:
Thanks, I will take a look.

mvanhorn:
Submitted a PR to add PGO support: #4725. It adds a `default.pgo` CPU profile generated from the existing benchmark suite, which Go 1.21+ automatically picks up during build. Also includes a `make profile` target for regenerating the profile as code changes.

PR #4725, opened 2026-03-20
Add profile-guided optimization (PGO) support

mvanhorn:
## Summary

- Add `default.pgo` CPU profile generated from the existing benchmark suite (`src/algo` and `src/` packages)
- Go 1.21+ automatically detects this file during build, enabling PGO without any build flag changes
- Add `make profile` Makefile target to regenerate the profile when code changes significantly

## How it works

Go's PGO (available since Go 1.21, fzf uses Go 1.23) automatically reads `default.pgo` from the main package directory during `go build`. The compiler uses the CPU profile to optimize hot paths - typically yielding 2-7% performance improvement for CPU-bound workloads like fuzzy matching.

The `make profile` target:
1. Runs benchmarks from `src/algo` and `src/` with `-cpuprofile`
2. Merges the two profiles with `go tool pprof -proto`
3. Outputs the merged profile as `default.pgo`

Both `go build` and `goreleaser` builds benefit automatically since they run from the repo root where `default.pgo` lives.

## Test plan

- [x] `make test` passes with `default.pgo` present
- [x] `make profile` successfully generates the profile
- [x] Go compiler auto-detects the profile during build (verified with `go build -v`)

Ref #3834

This contribution was developed with AI assistance (Claude Code).


FilipAndersson245:
What is the performance difference with PGO?

mvanhorn:
Measured it on this branch (Apple M-series, Go 1.26.4, hyperfine, 8 runs each). End-to-end filtering of a 2M-line synthetic path corpus:

- narrow query (`srcmainhandler`): PGO 81.1ms +/- 0.9 vs non-PGO 82.0ms +/- 2.2 (1.01x, within noise)
- broad query (`srclib`, heavy scoring): 143ms both, no measurable difference

src/algo microbenchmarks (`-benchtime=2s -count=3`) are mixed: LastIndexByteTwo_100/loop improves ~12% (24.4 -> 21.5 ns/op) while _1000/loop regresses ~8% (206 -> 222 ns/op); the asm paths are unchanged, as expected.

Honest summary: with the bundled default.pgo on this hardware, PGO is a wash end-to-end (<=1%). The wins Go's PGO docs cite (2-7%) may show up on the exact profiled workload or other architectures, but I can't demonstrate a user-visible gain from these numbers. If that doesn't clear the bar for carrying a default.pgo in-tree, I'm fine closing this.

mvanhorn:
Measured it properly rather than guessing. Same commit (749e63c), Go 1.26.5, Apple Silicon: `-pgo=auto` (picking up default.pgo) vs `-pgo=off`, byte-identical output on all test patterns.

End to end, `fzf --filter` over 2M-line and 800k-line inputs showed no measurable difference: every delta stayed inside the ±5% run-to-run noise and several flipped sign across interleaved repetitions. The only consistent wins were in the src/algo microbenchmarks for the pure-Go loop fallbacks (LastIndexByteTwo/loop about 15-30% faster), but amd64/arm64 use the assembly implementations at runtime, so that does not reach user-visible filtering speed on mainstream platforms.

So on this hardware the honest answer is: no observable end-to-end gain. If PGO earns its keep here, it would be on platforms without the asm fast paths.

PR #4731, opened 2026-03-21
fish: Completion script rewrite (SHIFT-TAB)

bitraid:
## Contribution Policy

---

## Acknowledgement

---

The current completion script suffers from various issues:
- It crashes on open quotes, single or trailing unescaped `\` or `$`, and commands like `env -i cmd`. Also, due to an oversight in `complete`, a single escaped backslash (`\\`) in command line crashes the shell.
- Completion of tokens with escaped/unescaped `$` work in the opposite way: Unescaped `$` does file completion while escaped `$` does variable name completion (instead of the other way around).
- It does not handle names containing newlines.
- Descriptions text is not searchable and not wrapped by default. If wrapping is enabled, the wrapped text is not aligned.
- I think it is not correct for the current token to automatically become the fzf query on startup. The generated list is already based upon the token, and having the token also as the query modifies the natural order (as listed by `tab`) of the results and gives the wrong impression that by deleting the query the whole list is going to be displayed.

This implementation is designed from the beginning as a replacement for the native `<shift-tab>` functionality and is cleaner and more robust, while providing the additional ability of recognizing tokens that are wildcard expansion patterns, and perform search on the expanded lists. It consists of a single function (`fzf_complete`) of about 100 LOC, which can also be used by custom completion functions. The function is defined as a wrapper of fzf, so its command line options (same as fzf) can be auto completed. Since the script makes use of fish completion functions, it should be advised for users to create/update regular completion functions (which will work for both `tab` and `shift-tab`), and create custom fzf completion functions only when needing to modify fzf options or want different results for `shift-tab` (for example, have no limit on the number of results for a command like `git checkout`).

Minimal custom function example:
```fish
function _fzf_complete_foo
  ls -la | fzf_complete --header-lines 3
end

function _fzf_post_complete_foo
  # Or use an external command instead of the while loop: awk '{print $NF}'
  while read -l result
    string split -r -m 1 -f 2 -- ' ' $result
  end
end
```

Null terminated items from custom functions (that don't have custom post-functions) are supported, even though it doesn't make much sense for completion items to insert newlines (or other unescaped whitespace characters). For that reason, custom post completion functions are always expected to return line terminated items:

```fish
function _fzf_complete_foo
  ls --zero | fzf_complete --read0 --print0
end

function _fzf_post_complete_foo
  while read -lz result
    string escape -- $result
  end
end
```

A more useful example that demonstrates the different ways the function can be used:

```fish
# Customize git completion
function _fzf_complete_git
    # Show header text with active branch for all git completions
    set -lx -- FZF_COMPLETION_OPTS --header=\'(git branch --color=always 2>/dev/null)\'

    # No other changes when less than 3 arguments, or when completing options
    if not set -q argv[3]; or string match -q -- '-*' $argv[-1]
        fzf_complete
        return
    end

    # Check subcommand
    switch $argv[2]
        case checkout diff log show
            # Set preview and display all branches and commits for subcommands: checkout, diff, log, show
            begin
                git branch --all --format='%(refname:short)'
                git log --all --oneline --color=always
            end | fzf_complete --no-multi --ansi --accept-nth=1 --query=$argv[-1] --preview='git show --color=always {1}'

        case add rm mv
            # Only set preview for subcommands: add, rm, mv
            fzf_complete --preview="git diff --color=always {r1}"

        case '*'
            # No changes for other subcommands
            fzf_complete
    end
end
```

The README isn't updated yet. There are some things than must be decided:

- The script uses the `--visible` (`-V`) option of `string length`, and `--escape` option of `complete -C`, both of which were introduced in fish v3.4.0. This makes the version requirement different than that of `key-bindings.fish`, which is v3.1b1. There are three different ways we can handle this:
  - Make the script compatible with v3.1b1 by adding version checks and alternative commands for versions older than 3.4.0. But maybe supporting such old versions doesn't worth the added complexity, because probably there are hardly any users that would benefit from it. The current completion script also uses `string length -V`, and there were no complaints yet.
  - Make v3.4.0 the minimum required version for both scripts and simplify the code of `key-bindings.fish`. A `key-bindings-legacy.fish` could be left in the repository, so that users with old fish versions can still have a way to use the key bindings.
  - Have different version requirements for the two scripts.

- The file names in completion lists (but not in glob lists) are escaped. This can be changed, but:
  - This is the most direct and reliable way of escaping only when needed. If the escaping was to be done by the script, it would first need to check if it is a variable name completion (by checking if the list contains the string `\tVariable: `, which is not very elegant or 100% reliable) and not apply escaping in this case. It should also check after escaping if the token starts with unescaped `~`, and unescape the leading `~` from the result.
  - Because completion items can have descriptions separated by the tab character, this would be the only way of correctly handling filenames with tabs. Unfortunately, `complete` seems to split the item before escaping, so names that contain tabs are still not handled correctly. This is something that the native fish completion also suffers from and will probably be fixed in the future.

- The environment variable `$FZF_EXPANSION_OPTS` is used for customizing the options of glob expansion search, in addition to the usual `$FZF_COMPLETION_OPTS` for setting completion options. The list in glob expansions is certain to be files or directories, so it makes sense to have separate options, but it is also necessary for accessing the selections because different templates must be used (`{+}` for expansions, `{+r1}` for completions). It would be possible to use a single variable though, if the names in both completion and expansion were escaped or unescaped, and also have `\x00` as delimiter for expansion (so the same template can be used).

- The custom post completion functions are named `_fzf_post_complete_COMMAND` instead of `_fzf_compete_COMMAND_post` that is used for other shells. This is done for being able to handle binaries named `foo` and `foo_post`. If having the same function name as other shells is preferred, we can change it.

- The script tries to detect the command to be completed, by stripping any leading variable definitions and launcher commands (`builtin`, `command`, `doas`, `env`, `sudo`) from the command line. If a custom function is defined for that command, the function is called and the (stripped) command line tokens are passed as arguments (`$argv[1]` is the command name, `$argv[-1]` is the current token). Any options of the launcher commands are removed, but in order to also support options that accept parameters and remove the parameters too, it would require much code (for something rarely used), whereas now a single regex is used. Let me know if this is OK.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,10 @@ CHANGELOG
         - With the reduced per-entry cost, the cache now has broader coverage.
 - Shell integration improvements
     - bash: CTRL-R now supports multi-select and `shift-delete` to delete history entries (#4715)
-    - fish: Improved command history (CTRL-R) (#4703) (@bitraid)
+    - fish:
+        - Improved command history (CTRL-R) (#4703) (@bitraid)
+        - Rewrite completion script (SHIFT-TAB) (#4731) (@bitraid)
+        - Increase minimum fish version requirement to 3.4.0 (#4731) (@bitraid)
 - `GET /` HTTP endpoint now includes `positions` field in each match entry, providing the indices of matched characters for external highlighting (#4726)
 - Bug fixes
     - `--walker=follow` no longer follows symlinks whose target is an ancestor of the walker root, avoiding severe resource exhaustion when a symlink points outside the tree (e.g. Wine's `z:` → `/`) (#4710)


junegunn:
I've rebased devel onto the latest master, so the CI error should resolve.

EDIT: Hmm, it doesnt. Let me rebase your branch.

shell/completion.fish
```diff
@@ -4,238 +4,116 @@
...
-    # Calculate tabstop based on longest completion item (sample first 500 for performance)
-    set -l -- tabstop 20
-    set -l -- sample_size (math "min(500, "(count $completions)")")
-    for c in $completions[1..$sample_size]
-      set -l -- len (string length -V -- (string split -- \t $c))
-      test -n "$len[2]" -a "$len[1]" -gt "$tabstop"
-      and set -- tabstop $len[1]
+      set -- result (string collect -- $list |
+        $fzf_cmd --delimiter="\t" --tabstop=$tabstop --wrap-sign=\t"↳ " --accept-nth=1)
     end
```
lalvarezt, line 150: I'd suggest you reconsider this change. If you don't bind tabstop calculation this tanks performance badly if the completion list is big
You don't have to do it this way but consider using the n first values

lalvarezt: would it be worth it to implement some heuristic to not do tabstop calculation when doing path completion (there the description field doesn't exist)?. Perhaps sample the first n entries of the list, if there's a tab then get the first m and calc tabstop on those. wdyt?

junegunn:
Thanks. Some questions:

It looks like this change breaks backward compatibility for the custom completion API, right? Could you explain the rationale behind it? Also, what happens to users who rely on the previous API for custom completions? e.g.

```fish
function _fzf_complete_foo
  function _fzf_complete_foo_post
    awk '{print $NF}'
  end
  _fzf_complete --multi --reverse --header-lines=3 -- $argv < (ls -al | psub)

  functions -e _fzf_complete_foo_post
end
```

Could the new code detect old-style functions and still call them, at least with a warning? Maybe keep `_fzf_complete` (with `_` prefix) as a compatibility shim?

> * The custom post completion functions are named `_fzf_post_complete_COMMAND` instead of `_fzf_compete_COMMAND_post` that is used for other shells. This is done for being able to handle binaries named `foo` and `foo_post`. If having the same function name as other shells is preferred, we can change it.

I see the point, but I'd lean toward consistency and handle the `_post` edge case only if it becomes a real issue. The earlier work aimed to keep the API consistent with other shell implementations, which I thought was a strong design choice.

junegunn:
Please take a look at the Copilot comments, but feel free to disregard irrelevant ones. Copilot often speaks nonsense.

bitraid:
> The earlier work aimed to keep the API consistent with other shell implementations, which I thought was a strong design choice.

> It looks like this change breaks backward compatibility for the custom completion API, right? Could you explain the rationale behind it? Also, what happens to users who rely on the previous API for custom completions?

I just went with the most optimal implementation this time, since completion already works different than the other shells and it's a recent addition to fish. It is easy to maintain compatibility though: Rename the main function and custom post functions, and remove `--` and any following arguments that may be present in the command line (these are no longer used by the script and all arguments are passed to fzf, which will cause issues if they contain `$argv`). It doesn't matter if the input comes from a pipe or a process substitution redirection, but there is no `__fzf_complete_native` function (instead, the single function is called with no redirected input). We can also support this though, by creating a copy of the function (`functions -c _fzf_complete __fzf_complete_native`) when the script is sourced. Let me know if I should make those changes (or some of them).

shell/completion.fish
bitraid: Even though the results that come from completion functions are usually already limited, and it would need tens of thousands of files for a noticeable delay, having a limit is not a bad idea. I will add it.

shell/completion.fish
```diff
@@ -4,238 +4,116 @@
...
+
+  set -l result
+
+  # Get the completion list from stdin when it's not a tty
+  if not isatty stdin
+    set -l -- custom_post_func _fzf_post_complete_$cmd[1]
+    if functions -q $custom_post_func
+      $fzf_cmd | $custom_post_func $cmd | while read -l r; set -a -- result $r; end
+    else if string match -q -- '*--print0*' "$FZF_DEFAULT_OPTS"
+      $fzf_cmd | while read -lz r; set -a -- result $r; end
```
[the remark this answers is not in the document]
bitraid: The same can also be said for when `--print0` is not used: a selection that contains spaces will be split. But what if this is the desired behavior (insert multiple word arguments from single selection) or the selection contains variable names or paths starting with `~`? As I mentioned in the description, the escaping (along with any other processing of the results) should be done by post completion functions. We could reverse that, so that the results are always escaped when there is no post completion function, and require post completion functions when the results shouldn't be escaped. But this seems a little backwards, and it's not how it's done in other shells either.

shell/completion.fish
```diff
@@ -4,238 +4,116 @@
...
-        end
+    # Use complete builtin (fish completion functions)
+    set -l -- list (complete -C --escape -- (commandline -pc))
+    if test -n "$list"
+      # Determine the tabstop length for description alignment
+      set -l -- tabstop 4
+      for i in $list
+        set -l -- len (string length -V -- (string split -- \t $i))
+        test -n "$len[2]" -a "$len[1]" -gt "$tabstop" -a "$len[1]" -lt (math $COLUMNS - 40)
+        and set -- tabstop (math $len[1] + 4)
```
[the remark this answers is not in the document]
bitraid: I'm waiting for feedback from the maintainer on how to handle the version requirement inconsistency between the two scripts.

shell/completion.fish
```diff
@@ -4,238 +4,116 @@
...
+  set -l -- glob_pattern (string join -- '' (string match -r -- $r_glob $token;
+    and string match -r -- $r_quote $token))
+
+  # Set command line tokes without any leading variable definitions or launcher
+  # commands (including their options, but not any option arguments).
+  set -l -- r_cmd '^(?:(?:builtin|command|doas|env|sudo|\\w+=\\S*|-\\S+)\\s+)*\\K.+'
+  set -l -- cmd (commandline -pc | string match -r $r_cmd | string split ' ')
+
+  # Set fzf options
+  test -z $FZF_TMUX_HEIGHT
```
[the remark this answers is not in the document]
bitraid: This is incorrect, `test -z` without any other argument returns true (for the same reason that `test -n` also returns true, unless the `test-require-arg` feature is set - which will be the default in the feature - and then `test -n` will return false, `test foo` will be an error, but `test -z` will still return true). The use of `set -q` is not a good choice because we might end up using a variable defined with an empty value. I'll add the quotes anyway.

shell/completion.fish
```diff
@@ -4,238 +4,116 @@
...
+  set -l -- r_quote '^(?>(?:\\\\.|"(?:[^"\\\\]|\\\\.)*"|\'(?:[^\'\\\\]|\\\\.)*\'|[^\'"\\\\]+)*)\\K[\'"]'
+
+  # The expansion pattern is the token with any open quote closed, or is empty.
+  set -l -- glob_pattern (string join -- '' (string match -r -- $r_glob $token;
+    and string match -r -- $r_quote $token))
+
+  # Set command line tokes without any leading variable definitions or launcher
+  # commands (including their options, but not any option arguments).
+  set -l -- r_cmd '^(?:(?:builtin|command|doas|env|sudo|\\w+=\\S*|-\\S+)\\s+)*\\K.+'
+  set -l -- cmd (commandline -pc | string match -r $r_cmd | string split ' ')
```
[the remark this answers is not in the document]
bitraid: Even though the custom functions are mainly interested in subcommands and options (which are words), we better address this. The suggestion is wrong though, because the regex must be used on the whole command line, not on each token. But I think the fix is simple enough.

shell/completion.fish
```diff
@@ -4,238 +4,116 @@
...
-      echo "fzf-tmux $FZF_TMUX_OPTS -- "
-    else if test "$FZF_TMUX" = "1"
-      echo "fzf-tmux -d$FZF_TMUX_HEIGHT -- "
+# - $FZF_COMPLETION_OPTS
+# - $FZF_EXPANSION_OPTS
+
+function fzf_complete -w fzf -d 'fzf command completion and wildcard expansion search'
+  # Remove any trailing unescaped backslash from token
+  set -l -- token (commandline -t | string replace -r '(?<!\\\\)(?:\\\\\\\\)*\\K\\\\$' '')
+  and commandline -rt -- $token
```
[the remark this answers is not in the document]
bitraid: Right. This somehow slipped my mind, even though I have done the same thing in `key-bindings.fish`.

shell/completion.fish
```diff
@@ -4,238 +4,116 @@
...
-    else
-      set -- tokens (commandline -opc)
-    end
+  # Wildcard expansion
+  else if test -n "$glob_pattern"
+    # Set the command to be run by fzf, so there is a visual indicator and an
+    # easy way to abort on long recursive searches.
+    set -lx -- FZF_DEFAULT_COMMAND "for i in $glob_pattern;" \
+      'test -d "$i"; and string match -qv -- "*/" $i; and set -- i $i/;' \
+      'string join0 -- $i; end'
```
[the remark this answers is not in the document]
bitraid: This is incorrect. Variables in fish are lists, with its elements staying as they are even if they have spaces, tabs, or newlines. Unlike other shells, fish does not split the variable when it is used in an expansion. The only reason for quoting, is to ensure that the variable is passed as one element, even if it might be unset or have multiple elements.

junegunn:
## Option 1

- `fzf_complete` - The new, simpler API designed for ease of use
- `_fzf_complete` - A compatibility shim aligned with the bash/zsh style for backward compatibility (i.e. fzf args + `--` + mandatory `$argv`)

## Option 2

Make `_fzf_complete` print deprecation warning (and do nothing).

---

What do you think? Although the fish completion is new, we neglected to mark it as experimental, so we should at least let the users know.

junegunn:
> Make v3.4.0 the minimum required version for both scripts and simplify the code of key-bindings.fish. 

Let's go with this. I can't imagine someone installing the latest version of fzf while still being stuck on a 4-year-old version of fish. We should mention it on the CHANGELOG though.

shell/completion.fish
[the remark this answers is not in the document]
junegunn: https://github.com/junegunn/fzf/pull/4731#issuecomment-4107588869

shell/completion.fish
lalvarezt: I did some quick testing and got the following. 

Median timing over 3 runs per sample size:

- 1000: 90.788 ms
- 2000: 173.345 ms
- 3000: 274.771 ms
- 4000: 364.846 ms
- 5000: 431.007 ms
- 6000: 497.994 ms
- 7000: 608.992 ms
- 8000: 744.913 ms
- 9000: 773.352 ms
- 10000: 909.278 ms

Even a 1000 items starts to be noticeable

lalvarezt: something like this should work. The idea is to skip tabstop calculation altogether if there aren't tabs at all (like most `ls /usr/bin/<shift-tab>` type of listing)

```diff
diff --git a/shell/completion.fish b/shell/completion.fish
index a40c458e..1381f3b1 100644
--- a/shell/completion.fish
+++ b/shell/completion.fish
@@ -84,10 +84,17 @@ function fzf_complete -w fzf -d 'fzf command completion and wildcard expansion s
     if test -n "$list"
       # Determine the tabstop length for description alignment
       set -l -- tabstop 4
-      for i in $list
-        set -l -- len (string length -V -- (string split -- \t $i))
-        test -n "$len[2]" -a "$len[1]" -gt "$tabstop" -a "$len[1]" -lt (math $COLUMNS - 40)
-        and set -- tabstop (math $len[1] + 4)
+      set -l -- tab_detection_limit 10
+      set -l -- tabstop_sample_limit 500
+      set -l -- head_size (math "min($tab_detection_limit, "(count $list)")")
+      set -l -- tab (printf '\t')
+      if string match -q -- "*$tab*" $list[1..$head_size]
+        set -l -- sample $list[1..(math "min($tabstop_sample_limit, "(count $list)")")]
+        for i in $sample
+          set -l -- len (string length -V -- (string split -- $tab $i))
+          test -n "$len[2]" -a "$len[1]" -gt "$tabstop" -a "$len[1]" -lt (math $COLUMNS - 40)
+          and set -- tabstop (math $len[1] + 4)
+        end
       end
 
       set -- result (string collect -- $list |
```

bitraid:
> ## Option 1
> * `fzf_complete` - The new, simpler API designed for ease of use
> * `_fzf_complete` - A compatibility shim aligned with the bash/zsh style for backward compatibility (i.e. fzf args + `--` + mandatory `$argv`)

Good idea, lets do that.

> > Make v3.4.0 the minimum required version for both scripts and simplify the code of key-bindings.fish.
> 
> Let's go with this. I can't imagine someone installing the latest version of fzf while still being stuck on a 4-year-old version of fish. We should mention it on the CHANGELOG though.

Great, now there are more commands and more command options available, as well as the command substitution syntax `$()` which can be used inside double quotes.


I might delay the changes a little because one of my dogs thew my laptop off the desk and now is dead (the laptop not the dog!).

shell/completion.fish
bitraid: I'm curious about the hardware you used for the measurements, because the results are similar to an old AMD Ryzen 5 3400G I use for testing. On a 5-year-old ThinkPad (i7-1165G7) it takes less than half of that time. It doesn't really matter though, because we are going to apply a limit anyway. But I don't think that is correct to assume that if the first 10 items don't have descriptions, there won't be any afterwards - the list can be a mix of paths and arguments, and any of them may or may not have descriptions. I suggest to just process the first 500 items only, which should be fast enough even on old hardware. It is not needed to calculate the actual length for the loop, a simple `for i in $list[1.500]` will do. I will also apply the following optimization, which improves the performance of the loop, especially when there are no tabs (more than 2 times faster):

```diff
diff --git a/completion.fish b/completion.fish
index a40c458..8572f4f 100644
--- a/completion.fish
+++ b/completion.fish
@@ -84,10 +84,12 @@ function fzf_complete -w fzf -d 'fzf command completion and wildcard expansion s
     if test -n "$list"
       # Determine the tabstop length for description alignment
       set -l -- tabstop 4
-      for i in $list
-        set -l -- len (string length -V -- (string split -- \t $i))
-        test -n "$len[2]" -a "$len[1]" -gt "$tabstop" -a "$len[1]" -lt (math $COLUMNS - 40)
-        and set -- tabstop (math $len[1] + 4)
+      set -l -- max_columns (math $COLUMNS - 40)
+      for i in $list[1..500]
+        set -l -- item (string split -f 1 -- \t $i)
+        and set -l -- len (string length -V -- $i)
+        and test "$len" -gt "$tabstop" -a "$len" -lt $max_columns
+        and set -- tabstop (math $len + 4)
       end

       set -- result (string collect -- $list |
```

lalvarezt: Windows + WSL and other stuff running 😅.
The heuristic for 10 entries it was assuming we're only doing either arguments or paths, but as you say, it is possible to have mixed lists.

lalvarezt: uhm, there might be a problem. for entries with tabs, `item` is only the first column, but `len` is the length of the entire original entry, including everything after the tab.

try `git <shitf-tab>` and `docker <shift-tab>`

image

image

doing `$item` instead of `$i` and a bigger `tabstop` should do it (I saw some cases where leaving it in 4 causes the columns to not be aligned correctly)

`        and set -l -- len (string length -V -- $item)`
 
EDIT: imagine a case where adding 4 wouldn't be sufficient?

bitraid: Yes, `$i` was a typo, it should be `$item`. Maybe the alignment issue has to do with the value returned by `string length`. If you have such a case, can you try without the `-V` flag?

lalvarezt: this this `git <shift-tab>` (with and without `-v`)

image

notice the behavior when we're just one char before a tabstop multiple (in this case 15)

bitraid: Thanks. This should fix it:

```diff
diff --git a/shell/completion.fish b/shell/completion.fish
index a40c458e..5e221be7 100644
--- a/shell/completion.fish
+++ b/shell/completion.fish
@@ -84,11 +84,14 @@ function fzf_complete -w fzf -d 'fzf command completion and wildcard expansion s
     if test -n "$list"
       # Determine the tabstop length for description alignment
       set -l -- tabstop 4
-      for i in $list
-        set -l -- len (string length -V -- (string split -- \t $i))
-        test -n "$len[2]" -a "$len[1]" -gt "$tabstop" -a "$len[1]" -lt (math $COLUMNS - 40)
-        and set -- tabstop (math $len[1] + 4)
+      set -l -- max_columns (math $COLUMNS - 40)
+      for i in $list[1..500]
+        set -l -- item (string split -f 1 -- \t $i)
+        and set -l -- len (string length -V -- $item)
+        and test "$len" -gt "$tabstop" -a "$len" -lt $max_columns
+        and set -- tabstop $len
       end
+      set -- tabstop (math $tabstop + 4)
 
       set -- result (string collect -- $list |
         $fzf_cmd --delimiter="\t" --tabstop=$tabstop --wrap-sign=\t"↳ " --accept-nth=1)
```

lalvarezt: what about

```diff
diff --git a/shell/completion.fish b/shell/completion.fish
index a40c458e..de03a894 100644
--- a/shell/completion.fish
+++ b/shell/completion.fish
@@ -84,11 +84,14 @@ function fzf_complete -w fzf -d 'fzf command completion and wildcard expansion s
     if test -n "$list"
       # Determine the tabstop length for description alignment
       set -l -- tabstop 4
-      for i in $list
-        set -l -- len (string length -V -- (string split -- \t $i))
-        test -n "$len[2]" -a "$len[1]" -gt "$tabstop" -a "$len[1]" -lt (math $COLUMNS - 40)
-        and set -- tabstop (math $len[1] + 4)
-      end
+      set -l -- max_columns (math $COLUMNS - 40)
+      for i in $list[1..500]
+        set -l -- item (string split -f 1 -- \t $i)
+        and set -l -- len (string length -V -- $item)
+        and set -l -- new_tabstop (math "$len + 4")
+        and test "$new_tabstop" -gt "$tabstop" -a "$new_tabstop" -le "$max_columns"
+        and set -- tabstop "$new_tabstop"
+       end
 
       set -- result (string collect -- $list |
         $fzf_cmd --delimiter="\t" --tabstop=$tabstop --wrap-sign=\t"↳ " --accept-nth=1)
```

EDIT: I didn't see your previous message before sending this one

lalvarezt: it's basically the same thing 😅, yes this works

bitraid: Almost the same - adding once outside the loop is a little faster. It also makes the unchanged value of tabstop `8`, which is probably a better default.

lalvarezt: what about setting the initial value of the tabstop to a bigger value (say 40), to prevent a cramped UI if the options are short?

bitraid: I'm not sure about that. On small terminal sizes it might waste a lot of space. I also find it easier to make out which description belong to which option (without having to highlight) when they are closer.

lalvarezt: your call

bitraid: What do you think about not hard coding the initial value of tabstop, but extracting it from `$FZF_DEFAULT_OPTS` if it exists?

lalvarezt: That could work. We just need to clarify that this is used as base and will increase (because the user might expect that fixed value)

junegunn: I imagine it's extremely rare to see `--tabstop` in someone's `$FZF_DEFAULT_OPTS`.

bitraid: > I imagine it's extremely rare to see `--tabstop` in someone's `$FZF_DEFAULT_OPTS`.

I mean the final `$FZF_DEFAULT_OPTS`, which will include `$FZF_COMPLETION_OPTS` and any arguments passed to the function, where tabstop could be set to define the minimum starting position of the descriptions.

Another option would be to have the initial value of tabstop be dynamically set according to the available columns, although I think that is more useful for the extra space to be used for the descriptions text instead.

junegunn:
No rush, just wondering if this is close to landing so I can decide whether to put out a quick patch for #4737 in the meantime and make a new fzf release without this.

bitraid:
I found some issues with `commandline` and `complete` when the current token is split by a line break or has newlines inside quotes. I've been trying some workarounds and I think I have a solution that works good enough. I will probably push it later today (I need to test it a little more).

In case that this doesn't make it to the new release, the patch below fixes #4737 for the current script:

```diff
diff --git a/shell/completion.fish b/shell/completion.fish
index a1281c31..d9c74b79 100644
--- a/shell/completion.fish
+++ b/shell/completion.fish
@@ -214,6 +214,12 @@ function fzf_completion_setup
 
   # Main completion function
   function fzf-completion
+    # Restore the default shift-tab behavior on tab completions
+    if commandline --paging-mode
+      commandline -f complete-and-search
+      return
+    end
+
     set -l -- tokens (__fzf_cmd_tokens)
     set -l -- current_token (commandline -t)
     set -l -- cmd_name $tokens[1]
```

junegunn:
I see, thanks! If this can be completed within a week, I'd like to wait and include it in the next release.

bitraid:
OK, good. I'll push the changes tomorrow (Monday) then - I'll try to fix an edge case where there is still an issue.

bitraid:
I pushed the changes - please test.

junegunn:
Thanks! I ran some basic tests and everything worked as expected.

- The shift-tab bug has been fixed.
- The examples in the current README all work as before. Great!
- The description part is no longer dimmed. Is this intended? To be clear, I don't have a strong opinion on the style.

And could you elaborate a bit more on`$FZF_EXPANSION_OPTS`? I read your explanation in the PR description, but I'm having trouble understanding it. How do you expect users to make use of it in practice? Could you share a few concrete examples?

bitraid:
> The description part is no longer dimmed. Is this intended?

The intention is for the descriptions to be searchable (and unless there is an option that I'm not aware of, it is not possible to dim the field), which I think is more useful, don't you agree?


> And could you elaborate a bit more on `$FZF_EXPANSION_OPTS?`

The `$FZF_EXPANSION_OPTS` sets the options for when searching expansion lists (the command line token is a wildcard pattern). The main reason is that the filenames in completion lists are escaped (I explain the reason in the PR description) and they belong to the first field (filenames can also have descriptions - for example in `git add` it displays if a file is modified or untracked). So, to set a preview for example, the option should be `--preview="test -f {r1}; and bat -- {r1}"`, while for expansion lists would be `--preview="test -f {}; and bat -- {}`. We could have only `$FZF_COMPLETION_OPTS`, if the files in expansion lists were also escaped (or the files in completion lists were unescaped, but that would have the drawbacks I mention in the PR description), and also have `--delimiter=\x00` to the default expansion list options, so that in both cases `{r1}` would work. Let me know if you prefer that, or if you have a better idea.

junegunn:
I see. I wasn't familiar with the term "expansion list", but I suppose fish users would know it. Sounds fine to me. By the way, this doesn't seem to work. Could you take a look?

```sh
touch 'foo " bar'

git add <shift-tab>
  # git add \"foo\ \\\"\ bar\"
  # fatal: pathspec '"foo \" bar"' did not match any files
```

bitraid:
> I see. I wasn't familiar with the term "expansion list", but I suppose fish users would know it. Sounds fine to me. By the way, this doesn't seem to work. Could you take a look?
> 
> ```shell
> touch 'foo " bar'
> 
> git add <shift-tab>
>   # git add \"foo\ \\\"\ bar\"
>   # fatal: pathspec '"foo \" bar"' did not match any files
> ```

This is an issue with the fish completion function of git. You will notice you get the same result with `<tab>`.

junegunn:
> You will notice you get the same result with `<tab>`.

Ha, you're right!

Anything else to decide? If not, we should update the README and CHANGELOG.

bitraid:
> Anything else to decide?

I can't think of anything else. I just forgot to mention a couple of changes with the last commit (maybe worth mentioning in README):
- The `tabstop` value can be set by `$FZF_COMPLETION_OPTS` or passed by custom functions to set the minimum starting position of descriptions.
- The custom post completion function names can be either `_fzf_complete_COMMAND_post` or `_fzf_post_complete_COMMAND`, with the latter having precedence (so that the post completion function of `foo` won't collide with the completion function of `foo_post`).


> If not, we should update the README and CHANGELOG.

Should I update the README in this PR, or do you prefer to make the changes yourself in `master`?

junegunn:
> * The custom post completion function names can be either `_fzf_complete_COMMAND_post` or `_fzf_post_complete_COMMAND`

Yeah, that's a good strategy. We could adopt it in bash and zsh as well in the future, though I wouldn't worry too much about it in practice.

> Should I update the README in this PR, or do you prefer to make the changes yourself in `master`?

Could you update here? I'll make the final edit when I merge your changes from devel to master. Thanks.

bitraid:
Sorry for taking so long - there was an issue with the arguments passed to custom functions (it is now fixed).
Please let me know if the information in the README is missing something or needs any changes.

README.md
```diff
@@ -732,60 +732,77 @@ _fzf_complete_foo_post() {
...
+fzf options for a specific command or want different results for `Shift-TAB`:
 
 ```fish
-function _fzf_complete_foo
-  function _fzf_complete_foo_post
-    awk '{print $NF}'
+# Customize git completion
+function _fzf_complete_git
+  # Show header text with active branch for all git completions
+  set -lx -- FZF_COMPLETION_OPTS --header="'"(git branch --color=always 2>/dev/null)"'"
```
junegunn: Could you check this again? It looks a little strange to me. It's showing a single inactive branch in the header, which seems to be on the last line in the output of `git branch --color=always 2>/dev/null`.

junegunn:
Thanks! Would you like to add an entry to the CHANGELOG.md as well? I think we should at least mention that the new version requirement of 3.4.0.

README.md
bitraid: Oops... that was supposed to have `--show-current` instead of `--color=always`.

bitraid:
> Thanks! Would you like to add an entry to the CHANGELOG.md as well? I think we should at least mention that the new version requirement of 3.4.0.

Done.

README.md
```diff
@@ -732,60 +732,77 @@ _fzf_complete_foo_post() {
...
+    case checkout diff log show
+      # Set preview and display all branches and commits for subcommands: checkout, diff, log, show
+      begin
+        git branch --all --format='%(refname:short)'
+        git log --all --oneline --color=always
+      end | fzf_complete --no-multi --ansi --accept-nth=1 --query=$argv[-1] --preview='git show --color=always {1}'
+
+    case add rm mv
+      # Only set preview for subcommands: add, rm, mv
+      fzf_complete --preview="git diff --color=always {r1}"
```
junegunn, line 785: I think this is a good place to explain when to use the `r` flag. To my understanding, `r` flag is needed when we populate the input list using Fish's native completion (not expansion list, not stdin list), correct?

bitraid: Yes. Do you think that something like this is good enough?

```diff
diff --git a/README.md b/README.md
index 48d367e8..c3f69268 100644
--- a/README.md
+++ b/README.md
@@ -781,6 +781,8 @@ function _fzf_complete_git

     case add rm mv
       # Only set preview for subcommands: add, rm, mv
+      # Special characters in fish completion lists are escaped, so the r flag must be used.
       fzf_complete --preview="git diff --color=always {r1}"

     case '*'
```

junegunn: LGTM.

junegunn:
Thanks a lot! Merged to devel branch. I'll release a new version with your work this weekend.

llakala:
> I think it is not correct for the current token to automatically become the fzf query on startup. The generated list is already based upon the token, and having the token also as the query modifies the natural order (as listed by tab) of the results and gives the wrong impression that by deleting the query the whole list is going to be displayed.

I've been getting much worse completions as a result of this change. When I type `git l` and complete:
image
Then only things starting with `l` show up, but because `l` isn't autofilled into my search results, when I type `o`, `log` doesn't appear first anymroe.
image
While I understand the intentions of this change, I think it's going to result in a degraded experience for most users.

bitraid:
This is not the reason that `log` is no longer the first item after you enter `o`, but because now it is possible to also search the descriptions. To restore the previous behavior of applying the search on options only, set:

```fish
set -gx FZF_COMPLETION_OPTS '--nth=1'
```

bitraid:
Alternatively, if you want to keep the descriptions search and mimic the native `shift-tab` behavior, set the `--no-sort` option.

PR #4734, opened 2026-03-22
Fix nthTransformer parts slice preallocation

cuiweixie:
Use make([]NthParts, 0, len(indexes)) so the slice starts empty with reserved capacity. The previous length-len(indexes) allocation left leading zero NthParts entries before appended elements.

## Contribution Policy

---

## Acknowledgement


junegunn:
Thanks!

PR #4736, opened 2026-03-23
fix: avoid heap allocation of server variable in startHttpServer

### LINKED ISSUE #4735: `server` can be moved into go func to avoid heap allocation
### Checklist

### Output of `fzf --version`

0.65.2 (416aff86) and commit 14e3995

### OS

- [x] Linux
- [x] macOS
- [x] Windows
- [ ] Etc.

### Shell

- [x] bash
- [x] zsh
- [x] fish

Use `go build -gcflags="-m" ./... &> build.log`, you can see the following line, which means `server` is allocated in **heap** by escape analysis in Go compiler.

```bash
src/server.go:101:2: moved to heap: server
```

And in assembly it's allocated in heap. The executable is generated by `go build -a -ldflags "-X main.version=0.65.2 -X main.revision=416aff86" -tags ""` and the assembly is generated by `go tool objdump -S fzf > fzf.plan9 `

```
	server := httpServer{
  0x59afc1		488d05b8820700		LEAQ 0x782b8(IP), AX		
  0x59afc8		e8b3b7e7ff		CALL runtime.newobject(SB)	# here: newobject for httpServer
  0x59afcd		48898424e8000000	MOVQ AX, 0xe8(SP)		
		apiKey:        []byte(apiKey),
  0x59afd5		31c0			XORL AX, AX				
  0x59afd7		488b5c2470		MOVQ 0x70(SP), BX			
  0x59afdc		488b4c2440		MOVQ 0x40(SP), CX			
  0x59afe1		e89ad6ebff		CALL runtime.stringtoslicebyte(SB)	
  0x59afe6		4889842490000000	MOVQ AX, 0x90(SP)			
  0x59afee		48899c2498000000	MOVQ BX, 0x98(SP)			
  0x59aff6		48898c24a0000000	MOVQ CX, 0xa0(SP)			
		actionChannel: actionChannel,
  0x59affe		488b8c2418010000	MOVQ 0x118(SP), CX	
  0x59b006		48898c24a8000000	MOVQ CX, 0xa8(SP)	
		getHandler:    getHandler,
  0x59b00e		488b8c2420010000	MOVQ 0x120(SP), CX	
  0x59b016		48898c24b0000000	MOVQ CX, 0xb0(SP)	
```

https://github.com/junegunn/fzf/blob/14e3995a06e6a92a48d982c45f48595be1199002/src/server.go#L125-L146

But in the code above, actually `server` is only used in `go func()`. So actually it can be allocated in **stack** in `go func()` by moving the definition of `server` into `go func()`. Sure, the size of `server` is small, with only 40 bytes and `server` seems to be a single instance. So moving `server` to stack does not reduce memory consumption and only reduces only one dynamic allocation.

After moving `server` definition to `go func()`, `server` is allocated in stack.

```go
	go func() {
		server := httpServer{
			apiKey:        []byte(apiKey),
			actionChannel: actionChannel,
			getHandler:    getHandler,
		}
		for {
			conn, err := listener.Accept()
			if err != nil {
				if errors.Is(err, net.ErrClosed) {
					return
				}
				continue
			}
			conn.Write([]byte(server.handleHttpRequest(conn)))
			conn.Close()
		}
	}()
```

> It's really a small problem. It's detected by CodeQL with Go escape analysis extension(doc [here](https://github.com/Lslightly/QLStat/blob/main/doc/adapters/escape_analysis.md) and linter [here](https://github.com/Lslightly/QLStat/blob/main/qlsrc/escape_ext/heapvar_should_move.ql)). And fzf has only 1 issue detected. (Yes, the static analysis is quite limited now.)

mango766:
This PR addresses issue #4735 by moving the `server` variable declaration inside the goroutine in `startHttpServer()`. 

Previously, the `server` variable was allocated on the heap because Go's escape analysis detected that it was captured by a goroutine. By moving the declaration inside the goroutine itself, the variable can now be allocated on the stack instead, eliminating an unnecessary dynamic allocation of ~40 bytes.

The fix is minimal - just moving the struct declaration from outside to inside the `go func()` block. This doesn't change any behavior, only improves performance slightly by avoiding heap allocation.

---

Fixes #4735

### DIFF
--- a/src/server.go
+++ b/src/server.go
@@ -122,13 +122,12 @@ func startHttpServer(address listenAddress, actionChannel chan []*action, getHan
 		}
 	}
 
-	server := httpServer{
-		apiKey:        []byte(apiKey),
-		actionChannel: actionChannel,
-		getHandler:    getHandler,
-	}
-
 	go func() {
+		server := httpServer{
+			apiKey:        []byte(apiKey),
+			actionChannel: actionChannel,
+			getHandler:    getHandler,
+		}
 		for {
 			conn, err := listener.Accept()
 			if err != nil {


Lslightly:
Copy, paste and steal by AI and become a contributor as a human? I don't think this is a real contribution but just makes my effort a case of PR fixes of your agent which just copies the solution in issue.

> I deleted the comment before since I know there are also needs for agent development for making minimal issue to be fixed automatically but in fact this PR just copy the solution in issue. I'm here to claim the contribution. You only get the credit of copying solution.
> Yeah, again this comment becomes the data of AI.

junegunn:
@Lslightly If you open a pull request, I'll review it over this.

I'm noticing that more and more contributors are acting as middlemen, relying entirely on AI, which has sadly made me question the value of pull requests.

Issue #4816, opened 2026-05-29
[ZSH] ALT-C binding unexpectedly resolves symbolic links in $PWD


silverneko:
### Checklist

- [x] I have read through the manual page (`man fzf`)
- [x] I have searched through the existing issues
- [x] For bug reports, I have checked if the bug is reproducible in the latest version of fzf

### Output of `fzf --version`

0.73.1 (ce4bef75)

### OS

- [x] Linux
- [ ] macOS
- [ ] Windows
- [ ] Etc.

### Shell

- [ ] bash
- [x] zsh
- [ ] fish

### Problem / Steps to reproduce

`zsh 5.9 (x86_64-ubuntu-linux-gnu)`

When the $PWD parts contains symbolic link, ALT-C would expand the $PWD to its realpath, breaking the directory traversal path. It's probably simpler to understand by demonstration:

```zsh
$ cd ~
$ ln -s /usr usr
$ cd usr
# PWD is now `$HOME/usr`

$ <ALT-C> & select "bin/"
# I expect the result to be `cd bin/` and I expect to end up with PWD = "$HOME/usr/bin"

$ builtin cd -- /usr/bin
# PWD is now `/usr/bin`
# fzf-cd-widget inserts the "builting cd -- /usr/bin" commandline instead.
# In other words, the `$HOME/usr` part of the PWD got resolved to its realpath, so the end result became `/usr/bin`

$ cd ..
# PWD is "/usr", instead of "~/usr"
```

The root cause is that in `fzf-cd-widget()` shell/key-bindings.zsh, the target directory is calculated by:
```zsh
local dir=(result of fzf)
BUFFER="builtin cd -- ${(q)dir:a}"
```

And the `:a` modifier expands the value of `$dir` with respect to the real working directory `getcwd()` (/usr) instead of the _logical_ working directory `$PWD` (~/usr).

---

To remedy this, we can explicitly append the `$PWD` instead of relying one `:a`:
```
BUFFER="builtin cd -- ${(q):-"${PWD%/}/${dir%/}"}"
```

Or use the same trick as shell/key-bindings.bash:
```
BUFFER="builtin cd -- ${(q)$(builtin cd -- "${dir}" && builtin pwd)}"
```

silverneko:
I've uploaded a fix f803731 in my fork, PTAL!

I prefer `BUFFER="builtin cd -- ${(q):-"${PWD%/}/${dir%/}"}"` because it feels less hacky and more direct.

[...]

PR #4817, opened 2026-05-30
[shell][zsh] Don't resolve symlinks in ALT-c (#4816)

junegunn:
This way ALT-c behaves more aligned with `cd`.

Imagine a setup like:
```
/foo -> foo_real
/foo_real/bar
```

Right now if we first `cd foo` (a symlink to `foo_real`), and then use ALT-c to goto `bar`, then we would end up executing `cd /foo_real/bar` instead of `cd /foo/bar`. `$PWD = /foo_real/bar`.

For comparison, if we first `cd foo` and then `cd bar`, we end up with `$PWD = /foo/bar`.

This commit changes the internal logic of `fzf-cd-widget` to first run `cd <result of FZF_ALT_C_COMMAND>` in a subshell to simulate the behavior of `cd`, and then insert the target PWD into the shell history. This way we get behavior consistent with the builtin `cd` command, while also recording reusable shell history.

## Contribution Policy

---

## Acknowledgement

### DIFF
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -110,8 +110,14 @@ fzf-cd-widget() {
     zle redisplay
     return 0
   fi
+  # Use subshell expansion to get the absolute PWD of the target dir.
+  # This allows the recorded shell history to be reused even from a different
+  # working directory.
+  # If failed, fallback to the unexpanded path to surface the error to the user.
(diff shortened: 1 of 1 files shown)


LangLangBart:
asked zsh devs for confirmation if the CWD shall be resolved in case of a
symlink, hoping it may be a bug, but most agreed it was intended, thx for fixing
it; a test would have been nice

- <https://zsh.org/mla/workers/2026/msg00485.html>

PR #4821, opened 2026-06-05
extra tests for alt-c

LangLangBart:
- **test: install nushell in Dockerfile**
  - setup copied from nushell.sh
    - <https://www.nushell.sh/book/installation.html>
  - Used here-documents for readability (can be inlined when preferred)
    - <https://docs.docker.com/reference/dockerfile/#here-documents>
- **test: ALT-C regression tests**
  - test both issues discussed in #4816

## Contribution Policy

---

## Acknowledgement

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,15 @@
 FROM rubylang/ruby:3.4.1-noble
-RUN apt-get update -y && apt install -y git make golang zsh fish tmux
+RUN apt-get update && apt-get install -y git make golang zsh fish tmux
+
+# https://www.nushell.sh/book/installation.html
+RUN <<EOF
+    set -ex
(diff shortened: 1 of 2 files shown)


LangLangBart:
> _test: install nushell in Dockerfile_

Dockerfile change was needed to run the tests locally.

```bash
podman build -qt fzf-ubuntu . &&
  podman run --rm -qit fzf-ubuntu bash -ic 'tmux new -d && ruby /fzf/test/runner.rb --name "/test_alt_c/"'
```

Otherwise, you’ll get errors.

```txt
.....................no such window: 1
can't find pane: 1
can't find pane: 1
can't find pane: 1
can't find pane: 1
...
```

junegunn (approved):
Thanks!

PR #4863, opened 2026-07-19
fish: Don't use exported variable for history command

### LINKED ISSUE #4862: Setting FZF_DEFAULT_COMMAND breaks the Fish history widget after Shift+Delete
### Checklist

### Output of `fzf --version`

0.74.0 (6765f464)

### OS

- [x] Linux
- [ ] macOS
- [ ] Windows
- [ ] Etc.

### Shell

- [ ] bash
- [ ] zsh
- [x] fish

fish: 4.8.1
terminal: kitty

Steps to reproduce
1. Set a global FZF_DEFAULT_COMMAND, for example:
2. set -gx FZF_DEFAULT_COMMAND 'fd --type file'
3. Start a new Fish session.
4. Press Ctrl+R to open the history widget.
5. Press Shift+Delete on any history entry.

Expected behavior

The selected history entry is deleted and the history widget reloads, continuing to display the updated history.

Actual behavior

After the reload, the history widget is broken and no longer shows the expected history entries.

Configuration files

~/.config/fish/conf.d/env.fish
```
~/.config/fish/conf.d/env.fish
set -gx FZF_DEFAULT_OPTS "\
--color=bg+:#313244,bg:#1E1E2E,spinner:#F5E0DC,hl:#F38BA8 \
--color=fg:#CDD6F4,header:#F38BA8,info:#CBA6F7,pointer:#F5E0DC \
--color=marker:#B4BEFE,fg+:#CDD6F4,prompt:#CBA6F7,hl+:#F38BA8 \
--color=selected-bg:#45475A \
--color=border:#6C7086,label:#CDD6F4"

set -gx FZF_DEFAULT_COMMAND 'fd --type file'
set -gx FZF_CTRL_T_COMMAND "fd --hidden"
set -gx FZF_ALT_C_COMMAND "fd --hidden --type directory"
set -gx FZF_CTRL_R_OPTS "--style full --height 20 --with-nth=1,3.."
```
~/.config/fish/config.fish
```
function fish_greeting
    if test "$KITTY_WINDOW_ID" = 1
        fastfetch --config examples/13
    end
end

function fish_user_key_bindings
    fish_vi_key_bindings
    bind -M insert ctrl-z '__fish_echo fg 2>/dev/null'
end

if status is-interactive
    fzf --fish | source
    starship init fish | source
    zoxide init --cmd cd fish | source
end

fish_config theme choose default
```

https://github.com/user-attachments/assets/ae9d4759-aa5c-4d23-8b37-d6fdfd058051

bitraid:
The history command is no longer loaded by an exported variable because its value might be overwritten by user initialization scripts of global/non-interactive shell session types.

The command is also now improved in the following ways:
- The timestamp/date prefix has always the $fish_color_comment color, even on fish versions where the syntax highlighting colors are not supported.
- The $fish_color_comment value is taken from the current shell session (there are cases where the variable might be undefined in a non-interactive subshell).

Fix #4862

## Contribution Policy

---

## Acknowledgement

### DIFF
--- a/shell/key-bindings.fish
+++ b/shell/key-bindings.fish
@@ -127,14 +127,21 @@ function fzf_key_bindings
     set -l -- total_lines (count $command_line)
     set -l -- fzf_query (string escape -- $command_line[$current_line])
 
+    set -lx -- FZF_DEFAULT_COMMAND "builtin history -z --show-time=(set_color $fish_color_comment 2>/dev/null; or set_color normal)\"%F %a %T%t%s%t\"(set_color normal)"
+
+    # Enable syntax highlighting colors on fish v4.3.3 and newer
+    if string match -qr -- '^\\d\\d+|^4\\.[4-9]|^4\\.3\\.[3-9]' $version
+      set -a -- FZF_DEFAULT_COMMAND "--color=always"
+    end
+
     set -lx -- FZF_DEFAULT_OPTS (__fzf_defaults '' \
       '--with-nth=2.. --nth=2..,.. --scheme=history --multi --no-multi-line' \
       '--no-wrap --wrap-sign="\t\t\t↳ " --preview-wrap-sign="↳ " --freeze-left=1' \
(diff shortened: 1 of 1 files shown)


junegunn (approved):
LGTM, thanks!

PR #4872, opened 2026-07-27
test(zsh): check capture for fzf exit status in widget

### LINKED ISSUE #4871: `fzf-history-widget` ignores `fzf` exit status when Perl is available
### Checklist

### Output of `fzf --version`

0.74 (devel)

### OS

- [x] Linux
- [ ] macOS
- [ ] Windows
- [ ] Etc.

### Shell

- [ ] bash
- [x] zsh
- [ ] fish

## Description

In `shell/key-bindings.zsh`, when `fzf-history-widget` executes the `if` branch (when `perl` is installed), `local ret=$?` captures the exit status of `extracted_with_perl=1` instead of the command substitution containing `fzf`.

Because `extracted_with_perl=1` succeeds and exits with status `0`, `$?` is reset to `0`. As a result, `fzf-history-widget` always returns `0` when `perl` is available, even if the user cancels `fzf` (which exits with status `130`).

## Steps to Reproduce

I have used a docker container to reproduce: `docker run --rm -it debian:trixie-20260223`

```shell
apt-get update -y
apt-get install -y git curl golang zsh perl perl-base
git clone https://github.com/junegunn/fzf /opt/fzf
cd /opt/fzf
go build
mv ./fzf /usr/bin/fzf
zsh
```

Now we are in zsh. Test 1 (Bugged behavior)

```shell
which perl
# Must have path
source <(fzf --zsh)
FZF_CTRL_R_OPTS="--bind 'ctrl-x:become(exit 55)'" fzf-history-widget
# now press Ctrl-x in the Widget
echo $?
# You will receive 0 (expected -- 55) (!!)
```

Exit from zsh and run it again:
```shell
exit
# Now we are in bash
mv $(which perl) $(which perl).disable
zsh
```

Test 2  (Correct behavior)

```shell
which perl
# perl must be not found
source <(fzf --zsh)
FZF_CTRL_R_OPTS="--bind 'ctrl-x:become(exit 55)'" fzf-history-widget
# now press Ctrl-x in the Widget
echo $?
# You will receive 55
```

## Suggested Fix

```patch
index bd2e4378..85a3821b 100644
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -139,11 +139,11 @@ fzf-history-widget() {
   # as the associative 'history' array, which maps event numbers to full history
   # lines, are set. Also, make sure Perl is installed for multi-line output.
   if zmodload -F zsh/parameter p:{commands,history} 2>/dev/null && (( ${+commands[perl]} )); then
+    extracted_with_perl=1
     selected="$(printf '%s\t%s\000' "${(kv)history[@]}" |
       FZF_DEFAULT_OPTS=$(__fzf_defaults "" "-n2..,.. --scheme=history --bind=ctrl-r:toggle-sort,alt-r:toggle-raw --wrap-sign '\t↳ ' --highlight-line --multi ${FZF_CTRL_R_OPTS-} --query=${(qqq)LBUFFER} --read0") \
       FZF_DEFAULT_OPTS_FILE='' $(__fzfcmd))"
-      extracted_with_perl=1
   else
     selected="$(fc -rl 1 | __fzf_exec_awk '{ cmd=$0; sub(/^[ \t]*[0-9]+\**[ \t]+/, "", cmd); if (!seen[cmd]++) print $0 }' |
       FZF_DEFAULT_OPTS=$(__fzf_defaults "" "-n2..,.. --scheme=history --bind=ctrl-r:toggle-sort,alt-r:toggle-raw --wrap-sign '\t↳ ' --highlight-line --multi ${FZF_CTRL_R_OPTS-} --query=${(qqq)LBUFFER}") \

```

LangLangBart:
fix #4871

- first add a regression test => let it run to see only perl fails
- commit the suggested fix

## Contribution Policy

---

## Acknowledgement

### DIFF
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -139,11 +139,11 @@ fzf-history-widget() {
   # as the associative 'history' array, which maps event numbers to full history
   # lines, are set. Also, make sure Perl is installed for multi-line output.
   if zmodload -F zsh/parameter p:{commands,history} 2>/dev/null && (( ${+commands[perl]} )); then
+    extracted_with_perl=1
     selected="$(printf '%s\t%s\000' "${(kv)history[@]}" |
       perl -0 -ne 'if (!$seen{(/^\s*[0-9]+\**\t(.*)/s, $1)}++) { s/\n/\n\t/g; print; }' |
       FZF_DEFAULT_OPTS=$(__fzf_defaults "" "-n2..,.. --scheme=history --bind=ctrl-r:toggle-sort,alt-r:toggle-raw --wrap-sign '\t↳ ' --highlight-line --multi ${FZF_CTRL_R_OPTS-} --query=${(qqq)LBUFFER} --read0") \
       FZF_DEFAULT_OPTS_FILE='' $(__fzfcmd))"
-      extracted_with_perl=1
   else
     selected="$(fc -rl 1 | __fzf_exec_awk '{ cmd=$0; sub(/^[ \t]*[0-9]+\**[ \t]+/, "", cmd); if (!seen[cmd]++) print $0 }' |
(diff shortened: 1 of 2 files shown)


junegunn (approved):
Thanks for handling this. Confirmed the test fails without the patch but passes with it.

PR #4880, opened 2026-07-30
test(zsh): verify chpwd fires once in alt-c widget

### LINKED ISSUE #4879: zsh cd widget calls zsh's chpwd hook functions twice
### Checklist

### Output of `fzf --version`

0.74 (devel) # git descript = v0.74.1-22-gb6615382

### OS

- [x] Linux
- [ ] macOS
- [ ] Windows
- [ ] Etc.

### Shell

- [ ] bash
- [x] zsh
- [ ] fish

# context 

In zsh the fzf-cd-widget (bound to alt-c by default) executes cd twice.  Once in a subshell to find the correct logical dir to change to and then once it is pushed onto the editor stack and executed by the line editor.  This "double call to cd" was introduced in 7d647c70c286720c225ebb8ab75896e5dbee6af6.

# problem 

I have a function registered with zsh that runs on every directory change.  The feature in zsh is called "hook function" and the relevant function one has to define is `chpwd` (or list several function names in the array `$chpwd_functions`).

This function is now executed twice when I use the widget/key binding.  Depending on what the hook function does this might be bad.

# repoduce

1. load the fzf bindings: `source <(fzf --zsh)`
2. define a chpwd function: `chpwd() { echo hook is running; }`
3. use the key binding alt-c
4. select entry
5. observe the output "hook is running" twice, once before the shell prompt  and "builtin cd ..." command line that the fzf-cd-widget injects (to set the `dir` variabel in the widget) and once after it (for the actual directory change in the shell).

# proposed solution 

The zsh [cd builtin has a `-q` option](https://zsh.sourceforge.io/Doc/Release/Shell-Builtin-Commands.html) that stops the shell from executing the hooks for this particular directory change. So I propose this change:

~~~diff
index 85a3821b..21c2ea12 100644
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -115,7 +115,7 @@ fzf-cd-widget() {
   # working directory.
   # If failed, fallback to the unexpanded path to surface the error to the user.
   # NOTE: Don't use the `:a` modifier as it resolves symlinks like `pwd -P`.
-  dir=$(builtin cd >/dev/null -- "${dir}" && echo "${PWD}" || echo "${dir}")
+  dir=$(builtin cd -q >/dev/null -- "${dir}" && echo "${PWD}" || echo "${dir}")
   zle push-line # Clear buffer. Auto-restored on next prompt.
   BUFFER="builtin cd -- ${(q)dir}"
   zle accept-line
~~~

I would also suggest a test case that executes the steps I described in the reproduction and then checks that the output from the hook is only seen once.  Sadly I do not know enough ruby to write that down just now.

LangLangBart:
fix #4879

- regression test => fails
- commit the suggested fix

## Contribution Policy

---

## Acknowledgement

### DIFF
--- a/shell/key-bindings.zsh
+++ b/shell/key-bindings.zsh
@@ -115,7 +115,7 @@ fzf-cd-widget() {
   # working directory.
   # If failed, fallback to the unexpanded path to surface the error to the user.
   # NOTE: Don't use the `:a` modifier as it resolves symlinks like `pwd -P`.
-  dir=$(builtin cd >/dev/null -- "${dir}" && echo "${PWD}" || echo "${dir}")
+  dir=$(builtin cd -q >/dev/null -- "${dir}" && echo "${PWD}" || echo "${dir}")
   zle push-line # Clear buffer. Auto-restored on next prompt.
   BUFFER="builtin cd -- ${(q)dir}"
   zle accept-line
--- a/test/test_shell_integration.rb
+++ b/test/test_shell_integration.rb
@@ -965,6 +965,18 @@ class TestZsh < TestBase
     end
   end
 
+  # Duplicate 'chpwd' calls overcount visits => skews rank tracking tools (e.g. 'zoxide')
+  def test_alt_c_chpwd_hook_once
+    tmux.send_keys "chpwd() { echo 'chpwd hook fired' >&2 }", :Enter
+    tmux.prepare
+    tmux.send_keys :Escape, :c
+    tmux.until { |lines| assert_operator lines.match_count, :>, 0 }
+    tmux.send_keys :Enter
+    tmux.until do |lines|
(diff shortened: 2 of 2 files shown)


LangLangBart:
`-q` flag for the `cd` builtin has been around since 4.3.6 (2008-04-02)

Refs:

- <https://github.com/zsh-users/zsh/commit/50d9cdeae4a6>
- <https://zsh.sourceforge.io/releases.html#:~:text=Major%20changes%20between%20versions%204.3.5%20and%204.3.6>

junegunn (approved):

junegunn:
@LangLangBart Thanks for the pull request.
@lucc Thanks for reporting the issue and providing the fix.

lucc:
Thanks for your quick reaction guys :)

PR #4909, opened 2026-09-06
fish: Improve custom CTRL-T command execution

### LINKED ISSUE #4498: CTRL-T binding not using last token with custom FZF_CTRL_T_COMMAND
### Checklist

### Output of `fzf --version`

0.65.1 (Fedora)

### OS

- [x] Linux
- [ ] macOS
- [ ] Windows
- [ ] Etc.

### Shell

- [ ] bash
- [ ] zsh
- [x] fish

fzf version: `0.65.1`
fish version: `4.0.2`

I've been exploring the shell integration features of fzf with fish shell, and I came across the CTRL-T binding and ways to customize it. From what I can see in the readme, referencing $dir in FZF_CTRL_T_COMMAND should allow me to use the current last token on the command line as the root directory. I've tried many different ways to format the command, with both find and fd, but I'm finding that any value in that variable causes the root directory to be the cwd, and not the last token.

The correct behavior is working if FZF_CTRL_T_COMMAND is unset. Following the example in the readme, hitting CTRL-T with the following text in the command line:

`ls /var/`

Does list files and directories relative to /var/

The problem occurs if you set FZF_CTRL_T_COMMAND to either:

- As suggested in the readme with find:
    - `set -g FZF_CTRL_T_COMMAND "command find -L \$dir -type f 2> /dev/null | sed '1d; s#^\./##'"`
- As suggested in #4251 with fd:
    - `set -g FZF_CTRL_T_COMMAND "command fd -L --type f  --hidden --follow --exclude .git . \$dir"`

And then try to use the CTRL-T binding from the same starting text as above, it will incorrectly list files relative to your current working directory.

I've tried this with my own fish config, and from a blank config started with `fish -N` and the behavior is the same. I haven't been able to find any value for FZF_CTRL_T_COMMAND that shows the list relative to the directory.

Based on previous issues I've found, like the one referenced above this seems to have been working before. I know there's been some bigger fish version changes lately, so it's possible the current way of setting that token isn't working with the newest version of fish.

bitraid:
The custom CTRL-T command is now executed by the script in the current shell session, instead of being called by fzf through a subshell. This ensures that the value of $dir can't be altered by user initialization scripts, and is also a little faster.

Fix #4498

## Contribution Policy

---

## Acknowledgement

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ CHANGELOG
     - fzf now opens in a tmux or Zellij floating pane by default, so the window it was started from stays visible and can be used while fzf is running
         - Requires tmux 3.7+ or Zellij 0.44+
         - Set `g:fzf_layout` to pick a different layout
+- fish: Fixed custom CTRL-T command not using the prefixed target directory in some cases (#4498)
 
 0.74.3
 ------
--- a/shell/key-bindings.fish
+++ b/shell/key-bindings.fish
@@ -104,7 +104,7 @@ function fzf_key_bindings
   # Store current token in $dir as root for the 'find' command
   function fzf-file-widget -d "List files and folders"
     set -l commandline (__fzf_parse_commandline)
-    set -lx dir $commandline[1]
+    set -l dir $commandline[1]
     set -l fzf_query $commandline[2]
     set -l prefix $commandline[3]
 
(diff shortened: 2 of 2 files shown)


junegunn (approved):

junegunn:
Thanks!

PR #4912, opened 2026-09-07
fish: Optimize description alignment of completions

bitraid:
## Contribution Policy

---

## Acknowledgement


lalvarezt:
@bitraid what do you think about
```fish
# Determine the tabstop length for description alignment
set -l -- max_columns (math $COLUMNS - 40)
for len in (string length -V -- (string replace -r -- \t'.*$' '' $list[1..500]))
    test "$len" -gt "$tabstop" -a "$len" -lt "$max_columns"
    and set -- tabstop $len
end
```

shell/completion.fish
```diff
@@ -124,10 +124,8 @@ function fzf_complete -w fzf -d 'fzf command completion and wildcard expansion s
 
       # Determine the tabstop length for description alignment
       set -l -- max_columns (math $COLUMNS - 40)
-      for i in $list[1..500]
-        set -l -- item (string split -f 1 -- \t $i)
-        and set -l -- len (string length -V -- $item)
-        and test "$len" -gt "$tabstop" -a "$len" -lt "$max_columns"
+      for len in (string match -r -- '^[^\\t]*(?=\\t)' $list[1..500] | string length -V)
```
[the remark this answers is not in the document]
bitraid: - In fish, inside single quotes, the backslash escapes single quotes and backslashes. So, `\\t` is consumed as `\t`. Using `\t` is the same, but I find easier to read and edit regexes when backslashes are always escaped.

- We are intentionally ignoring strings without tabs (descriptions) - so, the `string replace` suggestion is no good.

bitraid:
> @bitraid what do you think about
> 
> ```fish
> # Determine the tabstop length for description alignment
> set -l -- max_columns (math $COLUMNS - 40)
> for len in (string length -V -- (string replace -r -- \t'.*$' '' $list[1..500]))
>     test "$len" -gt "$tabstop" -a "$len" -lt "$max_columns"
>     and set -- tabstop $len
> end
> ```

This is basically what copilot suggested, which is wrong, because we don't want to update the tabstop when there is no description.

CHANGELOG.md
```diff
@@ -21,7 +21,9 @@ CHANGELOG
     - fzf now opens in a tmux or Zellij floating pane by default, so the window it was started from stays visible and can be used while fzf is running
         - Requires tmux 3.7+ or Zellij 0.44+
         - Set `g:fzf_layout` to pick a different layout
-- fish: Fixed custom CTRL-T command not using the prefixed target directory in some cases (#4498)
+- fish:
+    - Fixed custom CTRL-T command not using the prefixed target directory in some cases (#4498)
+    - Optimized description alignment of completion items (#4910)
```
junegunn, line 26:
```suggestion
    - Fixed custom CTRL-T command not using the prefixed target directory in some cases (#4498) (@bitraid)
    - Optimized description alignment of completion items (#4910) (@bitraid)
```

junegunn (approved):

junegunn:
@bitraid Thanks for the quality work!

@lalvarezt Thank you for the report and suggestion!

lalvarezt:
But that's my point, why is it wrong. Example

```
foo<TAB>description
very_long_completion_without_description
bar<TAB>description
```
Should be 
```
foo                 description
very-long-option
bar                 description
```

CHANGELOG.md
junegunn: Oh, this was not included. I'll update this later.

bitraid:
> But that's my point, why is it wrong. 

We are treating options without descriptions as filenames, which allows the code to be more efficient when doing the standard file completion, and in cases that there is a very long option without description, it also provides more space for the descriptions (allowing the display of more items in some cases), while still being aligned. I think that this is better for most cases, but it is a trade-off. If the maintainer thinks that it would be more popular the other way, I have no objection changing it.

Example:

- Current:
```
>
  83/83 (0) ─────────────────────────────────────────────────────────────────────────
  ▌ -X                            Sort by extension
  ▌ -x                            Multi-column output, horizontally listed
  ▌ -Z                            Display security context so it fits on most displays
  ▌ --all                         Show hidden
  ▌ --almost-all                  Show hidden except . and ..
  ▌ --author                      Print author
  ▌ --block-size                  Set block size
  ▌ --classify                    Append filetype indicator (*/=>@|)
  ▌ --color                       Use colors
  ▌ --color=                      Use colors                                           │
  ▌ --context                     Display security context so it fits on most displays │
  ▌ --dereference                 Follow symlinks                                      │
  ▌ --dereference-command-line    Follow symlinks                                      │
  ▌ --dereference-command-line-symlink-to-dir
  ▌ --directory                   List directories, not their content
  ▌ --dired                       Generate dired output
  ▌ --escape                      Octal escapes for non-graphic characters
  ▌ --file-type                   Append filetype indicator
  ▌ --format                      List format
  ▌ --full-time                   Long format, full-iso time
```

- Suggested:
```
>
  83/83 (0) ─────────────────────────────────────────────────────────────────────────
  ▌ -X                                           Sort by extension
  ▌ -x                                           Multi-column output, horizontally
  ▌                                              ↳ listed
  ▌ -Z                                           Display security context so it fits
  ▌                                              ↳ on most displays
  ▌ --all                                        Show hidden
  ▌ --almost-all                                 Show hidden except . and ..
  ▌ --author                                     Print author
  ▌ --block-size                                 Set block size
  ▌ --classify                                   Append filetype indicator (*/=>@|)    │
  ▌ --color                                      Use colors                            │
  ▌ --color=                                     Use colors                            │
  ▌ --context                                    Display security context so it fits   │
  ▌                                              ↳ on most displays
  ▌ --dereference                                Follow symlinks
  ▌ --dereference-command-line                   Follow symlinks
  ▌ --dereference-command-line-symlink-to-dir
  ▌ --directory                                  List directories, not their content
  ▌ --dired                                      Generate dired output
  ▌ --escape                                     Octal escapes for non-graphic
```

junegunn:
I prefer the current implementation for exactly the reason @bitraid mentioned. The suggested model risks shifting everything too far away because of a single exceptionally long entry.

